PR #40, opened 2018-12-10
Check the return value for orm.RegisterDataBase()

MeteorKL:
I compile kubeedge with the command below:
```bash
env GOOS=linux GOARCH=arm GOARM=5 go build cmd/edge_core.go
```
But it warns that `must have one register DataBase alias named Default`, which is pretty misleading.
After checking the return value for orm.RegisterDataBase(), I get the root case below:
```
Binary was compiled with 'CGO_ENABLED=0', go-sqlite3 requires cgo to work.
```

### DIFF
--- a/pkg/common/dbm/db.go
+++ b/pkg/common/dbm/db.go
@@ -42,8 +42,12 @@ func init() {
 		dataSource = "edge.db"
 	}
 
-	orm.RegisterDriver(driverName, orm.DRSqlite)
-	orm.RegisterDataBase(dbName, driverName, dataSource)
+	if err := orm.RegisterDriver(driverName, orm.DRSqlite); err != nil {
+		log.LOGGER.Fatalf("Failed to register driver: %v", err)
+	}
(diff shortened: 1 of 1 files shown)


m1093782566:
@MeteorKL 

Do you think we should add some information on enabling CGO when compile the kubeedge binary?

cc @Baoqiang-Zhang @qizha

m1093782566:
@MeteorKL Please fix the gofmt error, then LGTM.

MeteorKL:
@m1093782566  
I think so.

BTY, I finally use `xgo` to compile kubeedge with the command below. With xgo, I don't need to care about cgo and the cross-compiler. Maybe you can investigate whether xgo is easy and good to use in different environments.
```
~/go/bin/xgo --targets=linux/arm-7 github.com/kubeedge/kubeedge/cmd
```

PS: I use xgo to compile the binary of Raspberry Pi 3b+ in CentOS.

m1093782566:
COOL!

Issue #46, opened 2018-12-17
Data race in edged syncPod()


rohitsardesai83:
Build with race flag and start edge_core. Deploy an application from IEF to the edge node. The logs show a data race condition in the syncPod() and StartGarbageCollection() methods in edged.
```
==================
WARNING: DATA RACE
Write at 0x00c420422008 by goroutine 52:
  github.com/kubeedge/kubeedge/pkg/edged.(*edged).syncPod()
      /home/root1/go/src/github.com/kubeedge/kubeedge/pkg/edged/edged.go:603 +0xe2d
  github.com/kubeedge/kubeedge/pkg/edged.(*edged).Start()
      /home/root1/go/src/github.com/kubeedge/kubeedge/pkg/edged/edged.go:198 +0x1845

Previous read at 0x00c420422008 by goroutine 90:
  github.com/kubeedge/kubeedge/pkg/edged.(*edged).StartGarbageCollection.func2()
      /home/root1/go/src/github.com/kubeedge/kubeedge/pkg/edged/edged.go:330 +0x4e
  github.com/kubeedge/kubeedge/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1()
      /home/root1/go/src/github.com/kubeedge/kubeedge/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133 +0x6f
  github.com/kubeedge/kubeedge/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil()
      /home/root1/go/src/github.com/kubeedge/kubeedge/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134 +0xcd
  github.com/kubeedge/kubeedge/vendor/k8s.io/apimachinery/pkg/util/wait.Until()
      /home/root1/go/src/github.com/kubeedge/kubeedge/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88 +0x5a

Goroutine 52 (running) created at:
  github.com/kubeedge/kubeedge/beehive/pkg/core.StartModules()
      /home/root1/go/src/github.com/kubeedge/kubeedge/beehive/pkg/core/core.go:22 +0x196
  github.com/kubeedge/kubeedge/beehive/pkg/core.Run()
      /home/root1/go/src/github.com/kubeedge/kubeedge/beehive/pkg/core/core.go:46 +0x2f
  main.main()
[...]

PR #49, opened 2018-12-19
Issue #46: Add lock for podInitReady flag to prevent data race

rohitsardesai83:
Try to fix issue: #46

### DIFF
--- a/pkg/edged/edged.go
+++ b/pkg/edged/edged.go
@@ -3,6 +3,7 @@ package edged
 import (
 	"encoding/json"
 	"fmt"
+	"sync"
 	"time"
 
 	"k8s.io/api/core/v1"
@@ -79,10 +80,17 @@ const (
 	EdgeController                   = "controller"
 )
 
+// podReady holds the initPodReady flag and its lock
+type podReady struct {
+	// initPodReady is flag to check Pod ready status
+	initPodReady bool
+	// podReadyLock is used to guard initPodReady flag
+	podReadyLock sync.RWMutex
+}
+
 //Define edged
 type edged struct {
 	context                   *context.Context
-	initPodReady              bool
 	hostname                  string
 	namespace                 string
 	nodeName                  string
@@ -113,6 +121,8 @@ type edged struct {
 	rootDirectory             string
 	gpuPluginEnabled          bool
 	version                   string
+	// podReady is structure with initPodReady flag and its lock
+	podReady
 	// cache for secret
 	secretStore    cache.Store
 	configMapStore cache.Store
@@ -204,6 +214,20 @@ func (e *edged) Cleanup() {
 	}
 	e.context.Cleanup(e.Name())
 	log.LOGGER.Info("edged exit!")
+}
+
+// isInitPodReady is used to safely return initPodReady flag
+func (e *edged) isInitPodReady() bool {
+	e.podReadyLock.RLock()
+	defer e.podReadyLock.RUnlock()
+	return e.initPodReady
+}
+
+// setInitPodReady is used to safely set initPodReady flag
+func (e *edged) setInitPodReady(readyStatus bool) {
+	e.podReadyLock.Lock()
+	defer e.podReadyLock.Unlock()
+	e.initPodReady = readyStatus
 }
 
 func getConfig() *Config {
@@ -327,7 +351,7 @@ func (e *edged) StartGarbageCollection() {
 	}, ImageGCPeriod, utilwait.NeverStop)
 
 	go utilwait.Until(func() {
-		if e.initPodReady {
+		if e.isInitPodReady() {
 			err := e.containerGCManager.GarbageCollect()
 			if err != nil {
 				log.LOGGER.Errorf("Container garbage collection failed: %v", err)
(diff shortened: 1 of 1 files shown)


pkg/edged/edged.go
```diff
@@ -57,6 +57,7 @@ import (
 	utilpod "github.com/kubeedge/kubeedge/pkg/edged/util/pod"
 	"github.com/kubeedge/kubeedge/pkg/edged/util/record"
 	"github.com/kubeedge/kubeedge/pkg/metamanager"
+	"sync"
```
m1093782566: Please move the "sync" around the 

"fmt"
"time"

pkg/edged/edged.go
```diff
@@ -113,6 +114,7 @@ type edged struct {
 	rootDirectory             string
 	gpuPluginEnabled          bool
 	version                   string
+	podReadyLock              sync.Mutex
```
m1093782566: comment please.

m1093782566:
@rohitsardesai83 Could you please provide the related issue link?

rohitsardesai83:
@m1093782566 , here is the github issue : https://github.com/kubeedge/kubeedge/issues/46

pkg/edged/edged.go
m1093782566: If `podReadyLock` is only used for protecting `initPodReady` flag, why not put them into a another struct?

pkg/edged/edged.go
rohitsardesai83: done.

pkg/edged/edged.go
rohitsardesai83: moved the flag and the associated lock to a separate struct. embedded this struct as a member variable of edged struct.

pkg/edged/edged.go
```diff
@@ -327,6 +329,8 @@ func (e *edged) StartGarbageCollection() {
 	}, ImageGCPeriod, utilwait.NeverStop)
 
 	go utilwait.Until(func() {
+		e.podReadyLock.Lock()
```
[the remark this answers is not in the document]
rohitsardesai83: It is possible , but not much effective I feel since the only job this function would do is to return the e.podInitReady variable's value. Since the callers are only limited to 2 in this file , I feel it's ok for each function to acquire the lock , do the required stuff and release it . Abstracting locking / unlocking into a utility function seems somewhat less readable to me. Thoughts ?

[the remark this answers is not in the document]
m1093782566: I can not see any readability issue if we abstract a function and do lock/unlock inside the function - callers will no need to worry about the data race stuff.

The function can help callers to avoid misusing the locks. For example, 

```
e.podReadyLock.RLock()
defer e.podReadyLock.RUnlock()
if e.initPodReady {	 		
  err := e.containerGCManager.GarbageCollect()
  ...
}
```

If we try to acquire the lock in `containerGCManager.GarbageCollect()` as well, there would be a deadlock problem. Right?

VS.

```
func (e*) IsInitPodReady() bool {
  e.lock()
  defer e.unlock
  return e.initPodReady
}

if e.IsInitPodReady() {
  err := e.containerGCManager.GarbageCollect()
}
```

pkg/edged/edged.go
```diff
@@ -327,6 +336,8 @@ func (e *edged) StartGarbageCollection() {
 	}, ImageGCPeriod, utilwait.NeverStop)
 
 	go utilwait.Until(func() {
+		e.podReadyLock.RLock()
+		defer e.podReadyLock.RUnlock()
```
m1093782566: If we try to acquire the lock in containerGCManager.GarbageCollect() as well, there would be a deadlock problem. Right?

pkg/edged/edged.go
```diff
@@ -600,14 +616,18 @@ func (e *edged) syncPod() {
 						log.LOGGER.Errorf("handle podList failed: %v", err)
 						continue
 					}
+					e.podReadyLock.Lock()
```
m1093782566: Can we use `e.isInitPodReady()` here?

pkg/edged/edged.go
```diff
@@ -600,14 +616,18 @@ func (e *edged) syncPod() {
...
 					e.initPodReady = true
+					e.podReadyLock.Unlock()
 				} else if op == model.ResponseOperation && resID == "" && request.GetSource() == EdgeController {
 					err := e.handlePodListFromEdgeController(content)
 					if err != nil {
 						log.LOGGER.Errorf("handle controllerPodList failed: %v", err)
 						continue
 					}
+					e.podReadyLock.Lock()
```
m1093782566: Same as above.

pkg/edged/edged.go
```diff
@@ -113,6 +121,7 @@ type edged struct {
 	rootDirectory             string
 	gpuPluginEnabled          bool
 	version                   string
+	podReady
```
m1093782566: comment?

m1093782566:
LGTM overall, except I want to see some comments for the new field `podReady` of edged.

m1093782566:
Assign @Baoqiang-Zhang for another approval.

cc @qizha

PR #55, opened 2018-12-26
Added UT for metaManager, corresponding Mocks and FeedbackError in msg processor.

rohitsardesai83:


pkg/common/util/config.go
```diff
@@ -0,0 +1,36 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/ServiceComb/go-archaius/sources/file-source"
+
+	"github.com/kubeedge/kubeedge/beehive/pkg/common/config"
+)
+
+// LoadConfig is function to Load Configarations
+func LoadConfig() error {
```
m1093782566: Why do we create this function? Only test codes need it?

pkg/metamanager/dao/meta_test.go
```diff
@@ -0,0 +1,196 @@
+package dao_test
```
m1093782566: I think the package name should be `dao` instead of `dao_test`?

m1093782566: And please add the LICENSE HEAD.

pkg/metamanager/module_test.go
```diff
@@ -0,0 +1,91 @@
+package metamanager
```
m1093782566: LICENSE HEAD please.

pkg/metamanager/msg_processor_test.go
```diff
@@ -0,0 +1,719 @@
+package metamanager
```
m1093782566: LICENSE HEAD please.

m1093782566:
cc @Baoqiang-Zhang @qizha 

We are implement UTs for this project!

pkg/metamanager/msg_processor.go
```diff
@@ -260,6 +260,8 @@ func (m *metaManager) processRemoteQuery(message model.Message) {
 		content, err := json.Marshal(resp.GetContent())
 		if err != nil {
 			log.LOGGER.Errorf("marshal remote query response content failed, %s", msgDebugInfo(&resp))
+			feedbackError(err, "Error to marshal message content", message, m.context)
```
m1093782566: We should be a bit careful about `feedbackError` here. Any reason why should do this here? I think this PR is adding some UTs, does test codes rely on the error messages transported in the message channel?

pkg/common/util/config.go
rohitsardesai83: yes.

sids-b: yes, this will be common and used for all UT's

pkg/metamanager/msg_processor.go
sids-b: processRemoteQuery queries about a resource from cloud through websocket(EdgeHub). Hence if there is a marshall error in the content, we need to feedback to cloud. There is a Database error case below this. Since database save error is a local error, we are not sending feedback to cloud.

pkg/common/util/config.go
m1093782566: Is there any way that adding this function without increase the size of kubeedge binary?

pkg/metamanager/msg_processor.go
sids-b: Some of the UT cases rely on this change

m1093782566: Is it possible to get rid of this rely?

rohitsardesai83: I feel it is OK to feedback the error in case of an unmarshal error. There are other instances in the code which do the same.

pkg/common/util/config.go
sids-b: Done.

m1093782566:
I think we probably need a document to describe why/how to generate mock packages for other contributors.

mocks/beehive/fake_module.go
```diff
@@ -0,0 +1,78 @@
+// Code generated by MockGen. DO NOT EDIT.
```
m1093782566: I am unable to find any place where reference this package? I can't find any reason to generate this mock package, am I missing something?

sids-b:
> I think we probably need a document to describe why/how to generate mock packages for other contributors.

Yeah, we can create an issue for the same and will create a document soon.

mocks/beehive/fake_module.go
sids-b: Please see msg_processor_test.go Line31 where we have imported this package and used for creating fake modules.

m1093782566: Okay, it's useful :)

sids-b:
Yes @m1093782566 , We are using lot of mocks to avoid dependencies on other modules which makes it difficult to use table driven tests.  Right now we are focusing on testing all functionalities of a module and improving coverage. We have kept each test-case in each function separate with comment on top of each test case specifying which functionality it tests. Further wherever possible, we will use table driven tests.

m1093782566:
LGTM overall except a developing document for test codes.

sids-b:
@m1093782566 I have raised an issue for this Issue #57 . We will come up with a document soon.

m1093782566:
I am going to merge this PR but the developing document needs to be added in follow-ups.

PR #56, opened 2018-12-26
GOPATH can be empty, it will cause unclear error like "directory not …

tianxiaoliang:
…found", so give explicit error


m1093782566:
Thanks for catching this @tianxiaoliang

Issue #54, opened 2018-12-26
boolean validation error.


sids-b:
https://github.com/kubeedge/kubeedge/blob/3d599fea0b7e0823839043854beb7fdd1495d342/pkg/devicetwin/dtcommon/util.go#L33

the condition should be 
if strings.Compare(value, "true") != 0 && strings.Compare(value, "false") != 0 {

m1093782566:
@sids-b I am shocked about that.

This kind bug is one of the reason that we should add more unit test codes for this project...

m1093782566:
cc @Baoqiang-Zhang

sids-b:
> @sids-b I am shocked about that.
> 
> This kind bug is one of the reasons that we should add more unit test codes for this project...

@m1093782566 Found this issue while writing UT for this package

sids-b:
Issue addressed in pull request #59

PR #59, opened 2018-12-27
Fix for Issue #54

sids-b:


m1093782566:
THANKS for catching this! 

BTW, what can we do for the tests?

sids-b:
> THANKS for catching this!
> 
> BTW, what can we do for the tests?

@m1093782566 I have started with UT of deviceTwin. Will start raising PR's soon

m1093782566:
Thanks @sids-b! I would expect to see some documents before accepting two many codes as we discussed before?

Issue #57, opened 2018-12-27
Document on Mocks and Test Code Development


sids-b:
Document which explains when/why/how to generate and use mocks while writing test codes to help contributors.

PR #65, opened 2018-12-30
docs: guide for unit test. 

sids-b:
this adds a guide for writing unit tests.

Fixes Issue #57

### DIFF
--- a/docs/guides/unit_test_guide.md
+++ b/docs/guides/unit_test_guide.md
@@ -1,0 +1,74 @@
+# Unit Test Guide
+The purpose of this document is to give introduction about unit tests and to help contributors in writing unit tests.
+ 
+ ## Unit Test
+ Read this [article](http://softwaretestingfundamentals.com/unit-testing/) for a simple introduction about unit tests and benefits of unit testing.
+ 
+ Go has its own built-in package called testing and command called ```go test```.
+ 
+ For more detailed information on golang's builtin testing package read this [document](https://golang.org/pkg/testing/]).
+ 
+ ### Mocks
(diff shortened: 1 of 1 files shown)


sids-b:
@rohitsardesai83 @m1093782566 Please review. Thanks

rohitsardesai83:
@m1093782566 , what is the intention of having documentation  for how to write unit tests using gomock in kubeedge project?

m1093782566:
> @m1093782566, what is the intention of having documentation on how to write unit tests using gomock in kubeedge project?

I think gomock it's not straightforward to users and a document should help other contributors.

docs/guides/unit_test_guide.md
```diff
@@ -0,0 +1,70 @@
...
+```
+6) If we want Insert function of ormer interface which has return types as (int64,err) to return (1 nil), it can be done in 1 line in your test file using gomock.
+```shell
+ormerMock.EXPECT().Insert(gomock.Any()).Return(int64(1), nil).Times(1)
+```
+Expect() : is to tell that a function of ormermock will be called.
+Insert(gomock.Any()) : expect Insert to be called with any parameter.
+Return(int64(1), nil) : return 1 and error nil
+Times(1): expect insert to be called once and return 1 and nil only once.
+7) So whenever insert is called, it will return 1 and nil , thus removing the dependency on external implementation.
```
m1093782566: nit: please check the preview of this line:

```
7) So whenever insert is called, it will return 1 and nil , thus removing the dependency on external implementation.
```

The format is inconsistent with other lines.

LGTM overall except one nit. We can merge this PR once you fix it.

PR #71, opened 2019-01-05
Extend eventBus to support MQTT connection

kexun:
feature list:
1. eventBus support both internal mqtt broker and external broker (like mosquitto broker.)
2. make eventBus configurable with 3 mode. (0: internal mqtt broker enable only. 1: internal and external mqtt broker enable. 2: external mqtt broker enable only.)
3. eventBus can receive and send message by internal mqtt broker without sub or pub topic by mqtt client.
4. mqtt protocol V3.1.1 support.
5. qos0 support.
6. no persistence store, in memory only.


sids-b:
@kexun can you please run dep ensure so that I can clone and test?

m1093782566:
@kexun 

It would be easy for review if you can separate the commits, e,g. 1st for vendor, 2nd for configuration and 3rd for logic changes...

kexun:
> @kexun
> 
> It would be easy to review if you can separate the commits, e,g. 1st for vendor, 2nd for configuration and 3rd for logic changes...

i have separated the commits, please review.

m1093782566:
A quick comment: I just noticed that there are even some go test files in the vendor, please note that we are using dep to manage kubeedge 3rd-party dependency.

m1093782566:
@kexun 

You may check out https://github.com/kubeedge/kubeedge/pull/66 for reference?

conf/edge.yaml
```diff
@@ -1,5 +1,7 @@
 mqtt:
-    server: tcp://127.0.0.1:1883
+    internal-mqtt: tcp://127.0.0.1:1884
```
m1093782566: We are changing the APIs here, @Baoqiang-Zhang @qizha 

Any comments?

pkg/eventbus/event_bus.go
```diff
@@ -3,23 +3,34 @@ package eventbus
 import (
 	"encoding/json"
 	"fmt"
-
```
sids-b: @kexun , can we keep go, kubeedge and vendor packages seperated by an empty line ?

pkg/eventbus/event_bus.go
```diff
@@ -35,19 +46,39 @@ func (eb *eventbus) Start(c *context.Context) {
 	// no need to call TopicInit now, we have fixed topic
 	eb.context = c
 
-	mqttURL := config.CONFIG.GetConfigurationByKey("mqtt.server")
+	mqttURL := config.CONFIG.GetConfigurationByKey("mqtt.internal-mqtt")
 	nodeID := config.CONFIG.GetConfigurationByKey("edgehub.controller.node-id")
+
+	thirdpartyMqttUrl := config.CONFIG.GetConfigurationByKey("mqtt.thirdparty-mqtt")
 	if mqttURL == nil || nodeID == nil {
```
sids-b: @kexun , if mode is 2, i.e external mqtt mode, we may not give mqtt.internal-mqtt in config. Is this panic required in that case ??

conf/edge.yaml
Baoqiang-Zhang: edge.yaml is a config file for kubeedge, no APIs changing.  @m1093782566

pkg/eventbus/event_bus.go
Baoqiang-Zhang: +1

conf/edge.yaml
m1093782566: Configuration parameters is also API - it's API of binary file. We need to keep backward compatibility as well for component configuration and there should be a mechanism to deprecate old configuration. 

I am shocked to see the opinionated behavior change.

m1093782566: A general comment: kubeedge is open sourced now and someone(I don't know the exact people though) probably is running it in their system, any change to user behavior should be cautious.

conf/edge.yaml
```diff
@@ -1,5 +1,7 @@
...
+    thirdparty-mqtt: tcp://127.0.0.1:1883
```
rohitsardesai83: maybe we could rename it to external-mqtt ?

pkg/eventbus/event_bus.go
```diff
@@ -3,23 +3,34 @@ package eventbus
...
 	"github.com/kubeedge/kubeedge/beehive/pkg/core"
 	"github.com/kubeedge/kubeedge/beehive/pkg/core/context"
-
 	"github.com/kubeedge/kubeedge/pkg/eventbus/common/util"
 	mqttBus "github.com/kubeedge/kubeedge/pkg/eventbus/mqtt"
 )
 
+var mqttServer *mqttBus.MqttServer
+
+const (
```
rohitsardesai83: we can use "iota" instead of manually defining the integer values.

pkg/eventbus/event_bus.go
```diff
@@ -35,19 +47,42 @@ func (eb *eventbus) Start(c *context.Context) {
...
+		hub.InitPubClient()
+	}
+
+	if eb.mqttMode <= doubleMqttMode {
+		internalMqttURL := config.CONFIG.GetConfigurationByKey("mqtt.internal-mqtt")
+		if internalMqttURL == nil {
+			panic("mqtt url is not configured")
+		}
+		// launch an internal mqtt server only
+		// launch a mqtt server
```
sids-b: both comments look similar

m1093782566:
> @m1093782566 We need to add LICENSE head in files added by us in eventBus correct ?

Yes. Any new file should need a LICENSE head.

pkg/eventbus/event_bus.go
```diff
@@ -8,18 +8,30 @@ import (
...
 	"github.com/kubeedge/kubeedge/beehive/pkg/core"
 	"github.com/kubeedge/kubeedge/beehive/pkg/core/context"
-
 	"github.com/kubeedge/kubeedge/pkg/eventbus/common/util"
 	mqttBus "github.com/kubeedge/kubeedge/pkg/eventbus/mqtt"
 )
 
+var mqttServer *mqttBus.MqttServer
+
+const (
```
diaojuan: maybe it will look better to change the const name of doubleMqttMode to  bothMqttMode?

pkg/eventbus/event_bus.go
```diff
@@ -35,19 +47,42 @@ func (eb *eventbus) Start(c *context.Context) {
 	// no need to call TopicInit now, we have fixed topic
 	eb.context = c
 
-	mqttURL := config.CONFIG.GetConfigurationByKey("mqtt.server")
 	nodeID := config.CONFIG.GetConfigurationByKey("edgehub.controller.node-id")
-	if mqttURL == nil || nodeID == nil {
+	if nodeID == nil {
 		panic("mqtt url or node id not configured")
```
diaojuan: panic message should only keep the node id , no mqtt url.

kexun: tks, i will fix it.

pkg/eventbus/event_bus.go
```diff
@@ -35,19 +47,42 @@ func (eb *eventbus) Start(c *context.Context) {
...
+		}
+		mqttBus.MQTTHub = hub
+		hub.InitSubClient()
+		hub.InitPubClient()
+	}
+
+	if eb.mqttMode <= doubleMqttMode {
+		internalMqttURL := config.CONFIG.GetConfigurationByKey("mqtt.internal-mqtt")
+		if internalMqttURL == nil {
+			panic("mqtt url is not configured")
```
diaojuan: pannic info : "internal mqtt url is not configured "will be more helpful when to identify problem.

PR #84, opened 2019-01-12
documents: add README_zh.md

jahentao:
documents: add README_zh.md

add README_zh.md file for Chinese developers. Since the HuaweiCloud IEF is only available in China now, the project is mainly open to Chinese developers at present.


jahentao:
> I want to say what a good translation it is!

Thank you. This project is interesting and has great potential, more developers are needed to make it better.

PR #86, opened 2019-01-14
scripts:Unit Tests for pkg/devicetwin/devicetwin.go

lidiyag:


pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+	//TestModule is name of test.
+	TestModule = "test"
+	//GroupResource is name of resource group resource.
+	GroupResource = "resource"
+)
+
+// ormerMock is mocked Ormer implementation.
+var ormerMock *beego.MockOrmer
+
+// querySeterMock is mocked QurySeter implementation.
```
sids-b: nit: QurySeter-> QuerySeter

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+	initMocks(t)
+	registerFakeModule()
+	core.Register(&DeviceTwin{})
+	dt := DeviceTwin{}
+	mainContext.AddModule(dt.Name())
+	mainContext.AddModuleGroup(dt.Name(), dt.Group())
+	dt.context = mainContext
+	controller, err := InitDTController(mainContext)
+	if err != nil {
+		log.LOGGER.Errorf("Start device twin failed, due to %v", err)
```
sids-b: Can we use t.Errorf instead of log ? no need to continue the test if InitDTController fails

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
+package devicetwin
+
+import (
+	"github.com/kubeedge/kubeedge/beehive/pkg/common/config"
+	"github.com/kubeedge/kubeedge/pkg/devicetwin/dtcontext"
+	"github.com/kubeedge/kubeedge/pkg/devicetwin/dtmodule"
+	"testing"
+
```
sids-b: please seperate golang, kubeedge and vendor packages with a blank line

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+	dbm.DBAccess = ormerMock
+}
+
+// registerFakeModule is fuction to register all mock modules used.
+func registerFakeModule() {
+	fakeDeviceTwin.EXPECT().Name().Return(TestModule).Times(3)
+	core.Register(fakeDeviceTwin)
+	mainContext = context.GetContext(context.MsgCtxTypeChannel)
+	mainContext.AddModule(TestModule)
+
```
sids-b: remove extra blank line

sids-b:
@lidiyag please remove edge.db from your commit. We should not push it.

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+		})
+	}
+}
+
+// TestDeviceTwin_Start is function to test Start().
+func TestDeviceTwin_Start(t *testing.T) {
+	initMocks(t)
+	registerFakeModule()
+	core.Register(&DeviceTwin{})
```
rohitsardesai83: the device twin is registered in the registerFakeModule func right ? why do we again register it ?

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+	mainContext.AddModuleGroup(dt.Name(), dt.Group())
+	dt.context = mainContext
+	controller, err := InitDTController(mainContext)
+	if err != nil {
+		log.LOGGER.Errorf("Start device twin failed, due to %v", err)
+	}
+	dt.dtcontroller = controller
+	ormerMock.EXPECT().QueryTable(gomock.Any()).Return(querySeterMock).Times(200)
+	querySeterMock.EXPECT().All(gomock.Any()).Return(int64(1), nil).Times(200)
+	go dt.Start(mainContext)
```
rohitsardesai83: can we add some test code to see whether the dt.Start method worked fine ?
when the controller starts , it registers 4 different modules , memmodule, twinmodule, devicemodule, commmodule.
can we validate if these modules got registered successfully ?

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// TestDeviceTwin_Group is function to test Group().
+func TestDeviceTwin_Group(t *testing.T) {
+	tests := []struct {
+		name string
+		want string
+	}{
+		{
+			name: "test1",
```
rohitsardesai83: please name the test as per what it is supposed to test.

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// TestDeviceTwin_Name is function to test Name().
+func TestDeviceTwin_Name(t *testing.T) {
+	tests := []struct {
+		name string
+		want string
+	}{
+		{
+			name: "edgehub",
```
rohitsardesai83: why this test is named "edgehub" ?

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+	fakeDeviceTwin = beehive.NewMockModule(mockCtrl)
+	dbm.DBAccess = ormerMock
+}
+
+// registerFakeModule is fuction to register all mock modules used.
+func registerFakeModule() {
+	fakeDeviceTwin.EXPECT().Name().Return(TestModule).Times(3)
+	core.Register(fakeDeviceTwin)
+	mainContext = context.GetContext(context.MsgCtxTypeChannel)
+	mainContext.AddModule(TestModule)
```
rohitsardesai83: what is the purpose of this TestModule ?

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+// ormerMock is mocked Ormer implementation.
+var ormerMock *beego.MockOrmer
+
+// querySeterMock is mocked QurySeter implementation.
+var querySeterMock *beego.MockQuerySeter
+
+// fakeDeviceTwin is mocked implementation of devicetwin module.
+var fakeDeviceTwin *beehive.MockModule
```
sids-b: can you rename to fakeModule ? fakeDeviceTwin creates confusion.

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,148 @@
...
+	controller, err := InitDTController(mainContext)
+	if err != nil {
+		log.LOGGER.Errorf("Start device twin failed, due to %v", err)
+	}
+	dt.dtcontroller = controller
+	ormerMock.EXPECT().QueryTable(gomock.Any()).Return(querySeterMock).Times(200)
+	querySeterMock.EXPECT().All(gomock.Any()).Return(int64(1), nil).Times(200)
+	go dt.Start(mainContext)
+	msg := model.NewMessage("").BuildRouter(TestModule, GroupResource, model.ResourceTypePodStatus, model.InsertOperation)
+	dt.context.Send(TestModule, *msg)
```
rohitsardesai83: what are we testing over here ? why are we sending a pod status insert operation message on a test module ? how is it related to device twin test ? only the context is taken from dt.

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,193 @@
...
+			t.Errorf("Error while receiving message: %v", err)
+			return
+		}
+	})
+	//Checking health of MemModule
+	dtmName := dtcommon.MemModule
+	t.Run("MemModuleHealthCheck", func(t *testing.T) {
+		health, ok := controller.DTContexts.ModulesHealth.Load(dtmName)
+		if ok {
+			now := time.Now().Unix()
```
sids-b: Can we convert them to table driven tests ??

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,193 @@
...
+	initMocks(t)
+	registerFakeModule()
+	core.Register(&DeviceTwin{})
+	dt := DeviceTwin{}
+	mainContext.AddModule(dt.Name())
+	mainContext.AddModuleGroup(dt.Name(), dt.Group())
+	dt.context = mainContext
+	controller, err := InitDTController(mainContext)
+	if err != nil {
+		t.Errorf("Start device twin failed, due to %v", err)
```
sids-b: add return after this

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,193 @@
...
+			NodeID:      nodeid,
+			CommChan:    make(map[string]chan interface{}),
+			ConfirmChan: make(chan interface{}, 1000),
+		},
+		DTModules: make(map[string]dtmodule.DTModule),
+		Stop:      make(chan bool, 1),
+	}
+	t.Run("cleanuptest", func(t *testing.T) {
+		dt := &deviceTwin
+		dt.Cleanup()
```
sids-b: I don't see any checks in this function. Please check https://github.com/kubeedge/kubeedge/blob/master/pkg/metamanager/module_test.go to see how cleanup is tested.

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,193 @@
...
+func TestCleanup(t *testing.T) {
+	var deviceTwin DeviceTwin
+	deviceTwin.context = context.GetContext(context.MsgCtxTypeChannel)
+	nodeid, _ := config.CONFIG.GetValue("edgehub.controller.node-id").ToString()
+	deviceTwin.dtcontroller = &DTController{
+		HeartBeatToModule: make(map[string]chan interface{}),
+		DTContexts: &dtcontext.DTContext{
+			GroupID:     "",
+			NodeID:      nodeid,
+			CommChan:    make(map[string]chan interface{}),
```
sids-b: We only require stop channel in cleanup function.
```go
func (dt *DeviceTwin) Cleanup() {
	dt.dtcontroller.Stop <- true
	dt.context.Cleanup(dt.Name())
}
```
Why are we filling other fields here ?

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// TestName is function to test Name().
+func TestName(t *testing.T) {
+	tests := []struct {
+		name string
+		want string
+	}{
+		{
+			name: "devicetwinnametest",
```
sids-b: Use camel case

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// TestGroup is function to test Group().
+func TestGroup(t *testing.T) {
+	tests := []struct {
+		name string
+		want string
+	}{
+		{
+			name: "devicetwingrouptest",
```
sids-b: Use camel case

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,193 @@
...
+	dt := DeviceTwin{}
+	mainContext.AddModule(dt.Name())
+	mainContext.AddModuleGroup(dt.Name(), dt.Group())
+	dt.context = mainContext
+	controller, err := InitDTController(mainContext)
+	if err != nil {
+		t.Errorf("Start device twin failed, due to %v", err)
+	}
+	dt.dtcontroller = controller
+	ormerMock.EXPECT().QueryTable(gomock.Any()).Return(querySeterMock).Times(200)
```
sids-b: Please give exact times this will be used. I think 200 is a random number correct?

pkg/devicetwin/devicetwin_test.go
sids-b: +1

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,182 @@
...
+			name:       "CommModulehealthCheck",
+			moduleName: dtcommon.MemModule,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			health, ok := controller.DTContexts.ModulesHealth.Load(test.moduleName)
+			if ok {
+				now := time.Now().Unix()
+				if now-health.(int64) > 60*2 {
```
rohitsardesai83: in the code of dtcontroller start , if the modules dont ping / send heartbeat after more than 60*2 seconds , then  it is assumed that that module is unhealthy and it will attempt to start it again. so we dont need to check that case here .
tests could be like : 

1. after dtcontroller starts , it registers these modules first , so we can check if the context does have all these 4 modules.
2. get the heartbeat from the module health for each of these modules and see if ok. if not ok , just log a debug statement , saying that the heartbeat may have expired for this module and will be handled by dtcontroller to start it again.

pkg/devicetwin/devicetwin_test.go
lidiyag: There was a naming confusion involved...A test beehive module is created for sending a message from devicetwin so as to check the context is initialized properly.

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,194 @@
...
+		{
+			name:       "CommModuleHealthCheck",
+			moduleName: dtcommon.CommModule,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			moduleCheck := false
+			for _, module := range dt.dtcontroller.DTModules {
+				if test.moduleName == module.Name {
```
sids-b: There should be a ```go moduleCheck = false``` over here as we should overwrite the true value of previous iteration

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,194 @@
...
+	initMocks(t)
+	registerFakeModule()
+	core.Register(&DeviceTwin{})
+	dt := DeviceTwin{}
+	mainContext.AddModule(dt.Name())
+	mainContext.AddModuleGroup(dt.Name(), dt.Group())
+	dt.context = mainContext
+	controller, err := InitDTController(mainContext)
+	if err != nil {
+		t.Errorf("Start device twin failed, due to %v", err)
```
sids-b: the message should be InitDTController failed ...

pkg/devicetwin/devicetwin_test.go
```diff
@@ -0,0 +1,193 @@
...
+// TestStart is function to test Start().
+func TestStart(t *testing.T) {
+	initMocks(t)
+	registerFakeModule()
+	core.Register(&DeviceTwin{})
+	dt := DeviceTwin{}
+	mainContext.AddModule(dt.Name())
+	mainContext.AddModuleGroup(dt.Name(), dt.Group())
+	dt.context = mainContext
+	controller, err := InitDTController(mainContext)
```
sids-b: Even if we call InitDTController and assign it to dt.dtcontroller , it will be overwritten when we call start function. Why are we doing this activity then ? Am i missing something ?

pkg/devicetwin/devicetwin_test.go
lidiyag: This is to try sending a message from devicetwin to a module so as to check context is initialized properly.

rohitsardesai83:
LGTM.

PR #87, opened 2019-01-15
scripts: changes in Make file to run UT

### LINKED ISSUE #85: Unit test failing in CI
- The init() present in : kubeedge/pkg/edgehub/config/config.go, will panic if "kubeedge/pkg/edgehub/config" is imported. 

- This is due to the fact that the binaries are built & run at a temporary location where the config files are not present.

- Thus, it will cause the unit tests to fail before it even has a chance to execute

srivatsav123:
Made changes in make file to add env variable to fix UT panic

Fixes #85

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ else
 endif
 
 test:
+	export GOARCHAIUS_CONFIG_PATH=$(CURDIR)
 	find ${TEST_DIR} -name "*_test.go"|xargs -i dirname {}|uniq|xargs -i go test ${T} {}
 
 # verify


sids-b:
beehive config/config.go has below function
```go
// Get the configuration file path
func GetConfigDirectory() string {
	if config, err := CONFIG.GetValue(PARAMETER_CONFIG_PATH).ToString(); err == nil {
		return config
	}

	if config, err := CONFIG.GetValue(ENVIRONMENTAL_CONFIG_PATH).ToString(); err == nil {
		return config
	}

	return util.GetCurrentDirectory()
}
```
and the constants are 
```go
PARAMETER_CONFIG_PATH     = "config-path"
ENVIRONMENTAL_CONFIG_PATH = "GOARCHAIUS_CONFIG_PATH"
```
We are setting the environmental variable GOARCHAIUS_CONFIG_PATH so that we get the correct config directly and we load the config files correctly and not panic in init() functions of source code before running the Unit Tests

PR #98, opened 2019-01-18
metamanager: Fixed logical issue in UT

naveensriram:
Logically require any of the condition to met to fail the test


sids-b:
Nice catch @naveensriram . Thanks

PR #113, opened 2019-01-25
kubeedge: fixed test build ,added test in README

sids-b:
Fixed test build fail issue in kubeedge and added Run test block in
README


m1093782566:
Thanks!

PR #123, opened 2019-01-29
Removing the Unused data structure from config

kannanvr:
Removing the Unused data structure from config
DisClientconfig structure is unused .Hence , removing the unused code


m1093782566:
Okay, let't remove the unused codes for kubeedge though IEF uses them :)

m1093782566:
@sids-b 

I leave this PR to you for merging :)

PR #144, opened 2019-02-05
script to clean the unwanted log files and folders

### LINKED ISSUE #115: "make clean" not found to remove the binaries
Implement "make clean" in build steps

**Why is this needed**:
To remove older binaries. It helps developers to create fresh binaries. Also it can be used in automation script so that the setup always has freshly created binaries, in-case the tests need to be re-executed in current context.

srivatsav123:
this script will remove unwanted logs,process and folders

**What type of PR is this?**
 cleanup

to clean up unwanted logs files,process and folders
#115 
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
_If PR is about `failing-tests or flakes`, please post the related issues/tests in a comment and do not use `Fixes`_*
-->
Fixes #115 

### DIFF
--- a/hack/clean.sh
+++ b/hack/clean.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# **clean.sh**
+
+cd $GOPATH/src/github.com/kubeedge/kubeedge
+
+#removes bin folder which is created after running make verify
+remove_folder() {
+if [ -d bin ]; then
+rm -rf bin
+fi
+}
+
+#terminate edge core if running
+stop_edge_core () {
+if pgrep edge_core >/dev/null 2>&1 ; then
+     pkill -9 edge_core
+fi
+}
+
+#delete logs,covergage and database related files
+cleanup_files () {
+if [ -f edge_core ]; then
+rm -f edge_core
+fi
+
+find . -type f -name "*db" -exec rm -f {} \;
+find . -type f -name "*log" -exec rm -f {} \;
+find . -type f -name "*out" -exec rm -f {} \;
+}
+
+remove_folder
+stop_edge_core
+cleanup_files
--- a/Makefile
+++ b/Makefile
@@ -25,3 +25,6 @@ verify:
 integration_test:
 	bash test/integration/scripts/execute.sh
 
+.PHONY: clean
+clean:
+	bash -x hack/clean.sh


hack/clean.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# **clean.sh**
+
+cd $GOPATH/src/github.com/kubeedge/kubeedge
+
+removefolder() {
```
rohitsardesai83: removefolder -> remove_folders

hack/clean.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ -d bin ]; then
+rm -rf bin
+fi
+
+if [ -d .idea ]; then
```
rohitsardesai83: we should not delete any artifacts specific to any IDE. please remove this.

hack/clean.sh
```diff
@@ -0,0 +1,47 @@
...
+rm -rf .idea
+fi
+}
+
+#terminate mosquitto if running
+killmosquitto() {
```
rohitsardesai83: we can rename to stop_mosquitto.

hack/clean.sh
```diff
@@ -0,0 +1,47 @@
...
+if  pgrep mosquitto  >/dev/null 2>&1  ; then
+     pkill -9 mosquitto
+fi
+
+}
+
+#terminate edge core if running
+killedge () {
```
rohitsardesai83: we can rename to stop_edge_core

hack/clean.sh
```diff
@@ -0,0 +1,47 @@
...
+if pgrep edge_core >/dev/null 2>&1 ; then
+     pkill -9 edge_core
+fi
+
+}
+
+#delete all logs
```
rohitsardesai83: comment says only logs , but the function does a lot more. please check and improve the comment.

hack/clean.sh
```diff
@@ -0,0 +1,47 @@
...
+cleanfiles () {
```
rohitsardesai83: rename to cleanup_files()

hack/clean.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [ -f edge_core ]; then
+rm -rf edge_core
```
sameersarkar1979: edge_core is a binary. Hence no need to use "-r" flag in rm command

hack/clean.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+find . -type f -name "*db" -exec rm -f {} \;
+find . -type f -name "*log" -exec rm -f {} \;
+find . -type f -name "*out" -exec rm -f {} \;
+}
+
+removefolder
+killmosquitto
```
sameersarkar1979: You don't have to kill MQTT. There is no need as there is no development activity with in MQTT or any configurations which we are changing during our development, which requires it to be restarted everytime.

hack/clean.sh
sameersarkar1979: Expectation is we take care of only kubeedge related files, folders and binaries. This repo should not take care of IDE related issues. IDEA is developers own development env. Hence not related here. Remove these lines

hack/clean.sh
```diff
@@ -0,0 +1,47 @@
...
+if  pgrep mosquitto  >/dev/null 2>&1  ; then
+     pkill -9 mosquitto
+fi
+
+}
+
+#terminate edge core if running
+killedge () {
+if pgrep edge_core >/dev/null 2>&1 ; then
+     pkill -9 edge_core
```
sameersarkar1979: How about the edge_core.log file which was generated earlier. After stopping edge_core, move the edge_core.log file to /tmp(possibly) directory and append it with incremental numbers, each time. So that while moving there is not overwriting issue.

hack/clean.sh
srivatsav123: @rohitsardesai83 -is that ok not to kill MQTT broker , while after calling make clean?

hack/clean.sh
srivatsav123: ok

hack/clean.sh
srivatsav123: since we are unaware of the count of the user who does make clean..we cant move it to the other location and incremant the numbers..since the logs get overwritten since we move them.
The alternate way to solve this problem is to start edge core in the following way..
nohup ./edge_core >> edge_core.log 2>&1 & and move the edge_core.log when user does make clean(if its ok with this approach .we can change the command as above in the Readme.md..)

sameersarkar1979: Make clean will be executed by a developer when she/he is implementing the code and testing in her/his developmental environment. So don't bother about "since we are unaware of the count of the user who does make clean". Just check edge_core.log.1 exists in /tmp path or not, if so then re-direct the log to edge_core.log.2. 
It is a possibility to append in existing file, but after sometime it will become very huge and not maintainable.

srivatsav123: @sameersarkar1979 first time it works fine u will check for edge_core.log.1if it present, u will redirect it to  edge_core.log.2.But second time when user does make clean again also it checks edge_core.log1.Since its their it will overwrite to edge_core.log.2, again same problem persists to over come this better to use logrotate as we did in IEF...??
@rohitsardesai83 -Please give inputs on these.

sameersarkar1979: The second time edge_core.log.2 should be checked and not edge_core.log.1 again. Then third time edge_core.log.3 should be checked. and so on.

rohitsardesai83: log rotation is required , maybe we can introduce this later.

srivatsav123: @sameersarkar1979 -I have done as you suggested please review

image

hack/clean.sh
rohitsardesai83: i feel its fine not to kill MQTT broker.

hack/clean.sh
```diff
@@ -0,0 +1,62 @@
...
+fi
+}
+
+#delete logs,covergage and database related files
+cleanup_files () {
+if [ -f edge_core ]; then
+rm -f edge_core
+fi
+
+if [[ -f edge_core.log  &&  $count -eq 0 ]] ; then #if edgecore_log.tar file not exists and count is zero then its tar for the first time
```
rohitsardesai83: I feel we should handle this using logrotate utility. let's not introduce log rotation in this cleanup activity. it better we have a separate issue logged for log rotation.

hack/clean.sh
srivatsav123: ok

hack/clean.sh
srivatsav123: ok

PR #146, opened 2019-02-08
edged: added correct return for configmaps

sids-b:
**What type of PR is this?**
 /kind bug

Changed hardcoded nil to correct return for config maps. It is needed to create pods having config-maps


skdwriting:
LGTM

PR #165, opened 2019-02-15
kubeedge: added cross build job for edge

sids-b:
**What type of PR is this?**
/kind feature

Added cross build job for edge / To see compatibility with different H/W architectures

### DIFF
--- a/edge/Makefile
+++ b/edge/Makefile
@@ -28,3 +28,12 @@ integration_test:
 .PHONY: clean
 clean:
 	bash -x hack/clean.sh
+
+.PHONY: cross_build
+cross_build:
+	export GOARCH=arm
+	export GOOS="linux"
+	export GOARM=6
+	export CGO_ENABLED=1
+	export CC=arm-linux-gnueabi-gcc
+	go build cmd/edge_core.go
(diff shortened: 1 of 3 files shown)


edge/Makefile
```diff
@@ -28,3 +28,12 @@ integration_test:
 .PHONY: clean
 clean:
 	bash -x hack/clean.sh
+
+.PHONY: cross_build
+cross_build:
+	export GOARCH=arm
```
m1093782566: So the default arch is `amd` right? What about
```
for arch in ("amd", "arm"); then
do
 export  GOARCH=$arch
 go build
done
```

sids-b: How about we keep this for loop for all other h/w platforms we support and amd we keep seperate as a basic one ??

m1093782566: OK, no problem. Either way it ok to me :)

m1093782566: Just for clarification :)

rohitsardesai83: See this : http://www.codershaven.com/multi-platform-makefile-for-go/

can we try to do something similar ?

PR #174, opened 2019-02-25
docs: Mapper Design Proposal

sids-b:
**What type of PR is this?**
  /kind design

Design document for mapper architecture. mapper is used to connect with devices of. A common device mapper design can help have consistency in mappers supported by KubeEdge


docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,117 @@
...
+  * [User cases](#user-cases)
+* [Proposal](#proposal)
+
+## Introduction
+Mapper is an application that that is used to connect and and control devices. Following are the responsibilities of mapper:
+1) Scan and connect to the device.
+2) Report the actual state of twin-attributes of device.
+3) Map the expected state of device-twin to actual state of device-twin.
+4) Collect telemetry data from device.
+5) Convert readings from device to format accepted by KubeEdge(beehive message).
```
rohitsardesai83: not sure if it has to be a beehive message. needs more discussion on the telemetry data flow design overall. how event readings would be consumed by apps interacting with the platform , whether they interact directly with a database ( say influx db ? ) where our platform saves all the data ?

sids-b: My bad, It should have been message understood by platform. For each operation like memUpdate, stateUpdate, there are different structures defied by kubeedge.

sids-b: For current implementation , there is no special handling of telemetry data. It is considered in twin I guess.

sids-b: Done, Thanks

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,117 @@
...
+**1) Event-Manager**: A device can be controlled by setting a specific value in physical register(s) of a device and readings can be acquired by getting the value from specific register(s).
+We can define an event as a group of read/write operations on a device. A device may support multiple such events. Each of this events should be supplied through config-file to event manager.
+An Operation can be defined using below defined structure
+```go
+// Operation is structure to define device operation
+type Operation struct {
+	// Action can be one of read/write corresponding to get/set respectively
+	Action string
+	// Register is physical register where operation needs to be performed
+	Register string
```
rohitsardesai83: I am not sure if "Register" has to a first-class attribute of operation. Instead of a specific property like "Register", can we have a more generic structure like a map or something which can hold such kind of info ? The map can hold any number of key-value pairs, and then say for CC2650, one such kv entry could be a register ?

rohitsardesai83: in future with device crd's , maybe this info may come from attribute_visitors in the device spec. just fyi.

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,117 @@
...
+
+a) To scan devices(wireless)/wait for device to turn on(wired) and connect to the correct device once it is Online/In-Range. It can use MAC address or any unique address provided by devices. In case of wired devices, GPIO can be an option.
+
+b) Keep a watch on the expected state of the twin-attributes of the device and perform the event(s) to make actual state equal to expected.
+
+c) To report the actual state of twin attributes.
+
+**4) Data-Converter**: Data received from the devices can be in complex formats. eg: HexDecimal with bytes shuffled. This data cannot be directly understood by KubeEdge.
+The responsibility of data-converter is the convert the readings into a format understood by KubeEdge.
+Many protocols have a standard defined for the reading returned by the devicce. Hence a common/configurable logic can be used.
```
rohitsardesai83: devicce  -> device

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,117 @@
...
+c) To report the actual state of twin attributes.
+
+**4) Data-Converter**: Data received from the devices can be in complex formats. eg: HexDecimal with bytes shuffled. This data cannot be directly understood by KubeEdge.
+The responsibility of data-converter is the convert the readings into a format understood by KubeEdge.
+Many protocols have a standard defined for the reading returned by the devicce. Hence a common/configurable logic can be used. 
+
+**5) Health-Checker**: Health-Checker can be used to periodically report the state of the device to KubeEdge.
+This can be an optional component as not all devices support health-checking. In-future can be extended to report battery-state, malfunctioning when kubeedge supports these attributes.
+
+**6) Controller**: Controller should expose API's for CRUD operations for managing Events, Schedules, Watchers, Data-Converters, Health-Checkers.
```
rohitsardesai83: we need to be more specific on the APIs. REST/gRPC/?. The apis will need documentation describing the input/ouputs etc. maybe we can add more details . thoughts ?

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,117 @@
+---
```
rohitsardesai83: overall design looks good. needs more refinement. we can keep on improving the details.
@qizha , please review.

docs/proposals/mapper-design.md
sids-b: done, thanks

docs/proposals/mapper-design.md
sids-b: Nice Catch ! Operation is at a granularity level of just 1 read/write operation. So i have removed register and added AttributeName and AttributeValue in the doc. Thanks.

docs/proposals/mapper-design.md
sids-b: I initially plan to have mqtt based API's like the one's in device-twin. Maybe I will come-up with a API-Documentation.

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,120 @@
...
+```go
+// Operation is structure to define device operation
+type Operation struct {
+	// Action can be one of read/write corresponding to get/set respectively
+	Action string
+	// AttributeName is the name of the attribute where action is to be performed.
+	// for eg uuid in BLE, holding register in modbus
+	AttributeName string
+	// AttributeValue is the value of the Attribute.
+	AttributeValue string
```
rohitsardesai83: value type is string or interface{} ?

sids-b: I think string should work fine. Inteface{} can be useful if multiple/array of values could have been possible. But since we are granularity of read/write in single attribute, string will solve the purpose.

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,120 @@
...
+1) Manage expected/actual state of a device.
+2) Collect telemetry data from devices.
+3) Schedule events and check health of the device.
+
+## Proposal
+<img src="../images/proposals/mapper.png">
+
+Please find below the detailed explanation of each component of the mapper.
+
+**1) Event-Manager**: A device can be controlled by setting a specific value in physical register(s) of a device and readings can be acquired by getting the value from specific register(s).
```
qizha: An event should be generated from the device. Any operations to a device should be defined as "action"

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,120 @@
...
+	// AttributeValue is the value of the Attribute.
+	AttributeValue string
+	// Value is the value to be written in case of write action
+	Value string
+}
+```
+An Event is a set of operation(s) and can be defined using below structure
+```go
+// Event is structure to define a device event
+type Event struct {
```
qizha: please refer to the event definiton in WOT(https://iot.mozilla.org/wot/#event-object)

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,120 @@
...
+<img src="../images/proposals/mapper.png">
+
+Please find below the detailed explanation of each component of the mapper.
+
+**1) Event-Manager**: A device can be controlled by setting a specific value in physical register(s) of a device and readings can be acquired by getting the value from specific register(s).
+We can define an event as a group of read/write operations on a device. A device may support multiple such events. Each of this events should be supplied through config-file to event manager.
+An Operation can be defined using below defined structure
+```go
+// Operation is structure to define device operation
+type Operation struct {
```
qizha: please refer to the action definition in WOT: https://iot.mozilla.org/wot/#actions-member

docs/proposals/mapper-design.md
```diff
@@ -0,0 +1,120 @@
+---
```
qizha: So what is this mapper exactly? Is it an application can connect to every devices or framework/SDK that the user need to do some development to connect to differents devices? What role is expected to use this mapper? How this mapper interact with KubeEdge device twin?

docs/proposals/mapper-design.md
zhenghaibo1008: Here are two mode to call API. async/sync mode. In sync mode, we  use service bus to pass messages. and in async mode we use eventbus to pass messages.

zhenghaibo1008: How can we access the controller from cloud？

sids-b: @zhenghaibo1008 , currently KubeEdge doesn't have a cloud part, so in v0.1 we only plan to allow access to the controller from the edge via MQTT

docs/proposals/mapper-design.md
sids-b: Thanks @qizha , i have renamed to action :)

docs/proposals/mapper-design.md
sids-b: Sure, maybe in next verson onwards we start support multiple such standards like WOT, OCF,etc

docs/proposals/mapper-design.md
sids-b: Sure :)

docs/proposals/mapper-design.md
sids-b: Mapper is an application that can connect to the devices. We will have multiple mappers like mapper-bluetooth, mapper-zigbee as the definitions and standards used by devices supporting such protocols is standard. For.eg (GATT in bluetooth). There can be custom mappers for devices supporting custom protocols. This is just a generic design for what all components a mapper should have. We plan to provide some sample and generic mappers like for BLE, Zigbee,etc in KubeEdge. User/Developer can have reference these mappers and write their own if their devices are not supported by these samples. We do not plan to impose this design on user/developer but our samples will follow these designs. There maybe some specific things in each individual mapper which are specific to protocol/device are not included here. In general, mapper watches on deviceTwin via the twin topics, if there is any change in the expected state, it is mappers job to perform operations to make actual state of device attribute equal to expected. Mapper will also collect telemetry data from devices (eg: temperature sensor readings) and convert to a format understood by KubeEdge and publish on relevant MQTT topic.

docs/proposals/mapper-design.md
zhenghaibo1008: How do you wan't to deploy mapper? modules or docker images?

sids-b: Currently we want to deploy mappers as a process, After cloud-part in there we will support both docker and process.

zhenghaibo1008: If we deploy mapper as docker, we can use configmap

docs/proposals/mapper-design.md
qizha: OK, so we need to use one concrete protocol as a sample in this design.

chiru091096: @sids-b 
What happens to the telemetrydata after it is published to a MQTT topic?
I'm following where the data is going and I'm stuck at scheduler module in bluetooth mapper where it publishes the data to a mqtt topic. I'm not able to find any subscribers to that topic.
Does the telemetry data sent  the cloud ?

PR #181, opened 2019-02-27
design: Device CRD design proposal

### LINKED ISSUE #178: Device management from cloud is required
There should be a way to manage devices from the cloud and push these updates between cloud and edge nodes. Basic CRUD operations for describing device metadata, status updates etc are needed.

**Why is this needed**:
Device management is a basic feature of an edge computing platform.

rohitsardesai83:
**What type of PR is this?**
 /kind design

Design for Device CRD. This is needed for device management from the cloud.

Partially Fixes #178 

This the design PR. It will be followed by code PR once design is approved.


rohitsardesai83:
@CindyXing , @qizha , @Baoqiang-Zhang , @kevin-wangzefeng , please review

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+  - "@m1093782566"
+creation-date: 2019-02-27
+last-updated: 2019-02-27
+status: In Progress
+---
+
+## Motivation
+
+Currently kubeedge has an edge part which manages the lifecycle of applications
+deployed on the edge node, but there is no way to manage devices from the cloud.
```
CindyXing: I wouldn't put this two lines in the design doc. Device management was in our original roadmap.

rohitsardesai83: removed the two lines. thanks.

qizha:
we can consider to combine with wot(https://iot.mozilla.org/wot/) in our device model. Then we can generate microservice interfaces to each device directly in edge. Then cowork with the EdgeMesh feature to build a full microservices system cross cloud, edge and devices.

rohitsardesai83:
> we can consider to combine with wot(https://iot.mozilla.org/wot/) in our device model. Then we can generate microservice interfaces to each device directly in edge. Then cowork with the EdgeMesh feature to build a full microservices system cross cloud, edge and devices.

@qizha . I have added actions and events in order to make it like WOT. Kindly review and provide your thoughts.

rohitsardesai83:
@CindyXing , @qizha , @Baoqiang-Zhang , I have updated the device crd yaml after discussing with haibo.
Kindly review.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+Device management is a key feature required for an edge computing platform.
```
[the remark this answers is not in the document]
qizha: I agree. Edge Computing is not only for IOT. But the device management is something needed only in IOT scenarios.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+| Protocol              | The protocol supported by the device to communicate (e.g BLE, Modbus, OPC UA). <br />The configuration can contain more details about the connection mode , server, port details etc.|
+
+The `Status` contains the following fields
+
+| Field                 | Description                                                                      |
+|-----------------------|----------------------------------------------------------------------------------|
+|Twin                   | A list of device twin attributes.Each device twin attribute is like a control attribute which can be managed from <br />the cloud or the edge. Currently, the cloud configures the `Expected`state of a device and this configuration is pushed to the edge node. The mapper sends a command to the device to change as per the expected state . It receives the Actual state of the attribute once the previous operation is complete and sends the actual state to the cloud. Offline device interaction in the edge is possible via twin attributes for control/command operations.|
+|State                  | indicates the current device state ( unknown, running, offline  etc.)|
```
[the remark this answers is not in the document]
rohitsardesai83: Yes , the initial state of the device registered in the cloud will be `unknown`. Once the device is provisioned on the edge , the actual state of the device will be reported back to the cloud.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+### Device Custom Resource Overview
+The device custom resource layout is shown below. It contains standard fields ( Name, Namespace, Labels , Annotations etc. ) which help the Kubernetes ecosystem to manage resources. In addition, it contains custom fields in the Spec and Status detailed further below.
+<img src="../images/device-crd/device-crd.png">
+
+The `Spec` contains the following fields
+
+| Field                 | Description                                                                      |
+|-----------------------|----------------------------------------------------------------------------------|
+| Attributes            | A list of device model attributes. Each device attribute describes the data type ,<br />permissible values, defaults etc.|
+| Attribute Visitors    | The attribute visitors describe how to access the attribute values. <br />Each attribute has a visitor. |
```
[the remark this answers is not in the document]
rohitsardesai83: visitor describes the internal details of how to access a particular attribute for a given device.
for example , for a [CC2650 Sensor Tag](http://processors.wiki.ti.com/index.php/CC2650_SensorTag_User%27s_Guide),  if we want to read the temperature value, we need to access a specific register with a uuid. In addition, many more details could be needed. The primary consumer of this would be the mapper which would connect to the Sensor Tag and read the data from a BLE service and charactersitic. For other protocols , the way to access the data could vary.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+| Node                  | The edge node details to which this device belongs to.|
+| Protocol              | The protocol supported by the device to communicate (e.g BLE, Modbus, OPC UA). <br />The configuration can contain more details about the connection mode , server, port details etc.|
+
+The `Status` contains the following fields
+
+| Field                 | Description                                                                      |
+|-----------------------|----------------------------------------------------------------------------------|
+|Twin                   | A list of device twin attributes.Each device twin attribute is like a control attribute which can be managed from <br />the cloud or the edge. Currently, the cloud configures the `Expected`state of a device and this configuration is pushed to the edge node. The mapper sends a command to the device to change as per the expected state . It receives the Actual state of the attribute once the previous operation is complete and sends the actual state to the cloud. Offline device interaction in the edge is possible via twin attributes for control/command operations.|
```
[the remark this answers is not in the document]
rohitsardesai83: There are some device properties which can be controlled ( accessMode : rw ). It would make sense to have expected and actual device twin attributes for such properties . If a device property is read-only like temperture from a sensor ( accessMode: r) , then it won't be logical to have an "expected" state for it. 

The device twin attributes in the DeviceStatus just report the current expected and actual values for a control device property. The property name in the spec should match the device twin name in the status.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+* provide APIs for managing devices from the cloud.
+* synchronize the device updates between cloud and edge nodes.
+
+### Proposal
+We propose using Kubernetes Custom Resource Definitions (CRDs) to describe device metadata/status and a custom controller to sync device metadata/status updates
+between the edge node and the cloud.
+
+### Syncing Device Updates From Cloud To Edge
+<img src="../images/device-crd/device-updates-cloud-edge.png">
+The device controller watches device updates in the cloud and relays them to the edge node. These updates are stored locally by the device twin . The mapper gets these updates via the broker and operates on the device based on the updates.
```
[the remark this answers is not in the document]
rohitsardesai83: How do we associate nodes with devices? Currently, we assume that when a user creates a device he can specify which edge node needs to manage this device. This information will be transmitted by the device controller downstream to the edge node as a node membership update event. The twin currently handles node membership events and stores this info locally in the db.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+// DeviceList contains a list of Device
+type DeviceList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Device `json:"items"`
+}
+```
+
+### Open Points
+- Do devices need to be namespace scoped resources or cluster wide resources ? As we are binding devices to nodes in the device spec , the namespace becomes irrelevant ?
```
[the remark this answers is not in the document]
rohitsardesai83: ok thanks.

kevin-wangzefeng:
Device models can be common, and reused among devices that have the same model, I'd suggest make device and device model different CRDs: `Device` and `DeviceModel`, and under `devices.kubeedge.io` api group.

Example of revised device and device model definition look like the following:

### Device CRD

```yaml
apiVersion: devices.kubeedge.io/v1alpha1
kind: Device
metadata:
  name: sensor-tag01
  labels:
    description: 'TI Simplelink SensorTag 2.0 with Bluetooth 4.0'
    manufacturer: 'Texas Instruments'
    model: CC2650
spec:
  macAddress: 'B0:B4:48:BD:0F:83'
  deviceModel: 
    apiVersion: devices.kubeedge.io/v1alpha1
    kind: DeviceModel
    namespace: foo
    name: sensor-tag-model
#  currently all devices are indirect, maybe we can add it later.
#  connectionType: indirect/direct
  protocol:
    name: modbus
    config:
      register: InputRegisters
      index: 0
      offset: 2
      scale : 0.1 
      is_swap: true
      is_registerswap: true
  nodeName: node-1
    
    
status:
  state: online
  twins:
    - name: TemperatureEnable
      actual:
        metadata:
          timestamp: '1550049403598'
          type: string
        value: OFF
      expected:
        metadata:
          timestamp: '1550049403598'
          type: string
        value: OFF
```

### DeviceModel CRD

```yaml
apiVersion: devices.kubeedge.io/v1alpha1
kind: DeviceModel
metadata:
  name: sensor-tag-model
  labels:
    description: 'SensorTag device attributes model'
    manufacturer: 'Texas Instruments'
spec:
  properties:
    - name: temperature
      description: 'Temperature from Temperature Sensor in Degrees Fahrenheit'
      metadata:
        dataType: float
        accessMode: r
        unit: 'degree fahrenheit'
    - name: temperature-enable
      description: 'Enable Temperature Sensor'
      metadata:
        dataType: string
        accessMode: rw

  propertyVisitors:
    - protocolName: opc-ua
      visitors:
      - name: temperature
        propertyName: Temperature
      - name: pressure
        propertyName: Pressure
      - name: humidity
        propertyName: Humidity
        
    - protocolName: modebus
      visitors:
      - name: humidity
        propertyName: Humidity
```

docs/proposals/device-crd.md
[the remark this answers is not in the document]
rohitsardesai83: agreed. updated.

docs/proposals/device-crd.md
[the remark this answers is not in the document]
rohitsardesai83: Please let me know if I am missing something.

docs/proposals/device-crd.md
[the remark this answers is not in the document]
rohitsardesai83: Kevin suggested to remove the connection type and assume that we have indirect devices by default for now.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+- Device twin attributes are not having any attribute visitors defined. Should we add visitors for them as well ?
+- How are we going to address the scalability considering huge scale of devices on top of large scale of edge nodes ?
```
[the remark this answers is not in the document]
rohitsardesai83: I have detailed the design of the device controller . Please take a look.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+- What is the customer e2e scenario ? Does customer call kubectl to register a device for a specific node ? If so, what is the process for authN/authZ of the device talking to the edge node or we assume no such need at this point ?
```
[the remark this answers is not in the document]
rohitsardesai83: I have added the relevant sections.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+    - "@rohitsardesai83"
+approvers:
+  - "@qizha"
+  - "@CindyXing"
+  - "@kevin-wangzefeng"
+  - "@Baoqiang-Zhang"
+  - "@m1093782566"
+creation-date: 2019-02-27
+last-updated: 2019-03-05
+status: In Progress
```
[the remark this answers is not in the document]
rohitsardesai83: changed to `alpha`

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+
+Device management is a key feature required for an edge computing platform.
+This proposal addresses how can we manage devices from the cloud and synchronize
+the device updates between edge nodes and cloud.
+
+### Goals
+
+Device management must:
+* provide APIs for managing devices from the cloud.
+* synchronize the device updates between cloud and edge nodes.
```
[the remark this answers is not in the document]
rohitsardesai83: added non-goals.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Proposal
+We propose using Kubernetes Custom Resource Definitions (CRDs) to describe device metadata/status and a custom controller to sync device metadata/status updates
+between the edge node and the cloud.
+
+### Syncing Device Updates From Cloud To Edge
+<img src="../images/device-crd/device-updates-cloud-edge.png">
+The device controller watches device updates in the cloud and relays them to the edge node. These updates are stored locally by the device twin . The mapper gets these updates via the broker and operates on the device based on the updates.
```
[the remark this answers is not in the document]
rohitsardesai83: added a separate section for offline scenarios with not much details. consistency issues to handle is kept as an open question.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+### Goals
+
+Device management must:
+* provide APIs for managing devices from the cloud.
+* synchronize the device updates between cloud and edge nodes.
+
+### Proposal
+We propose using Kubernetes Custom Resource Definitions (CRDs) to describe device metadata/status and a custom controller to sync device metadata/status updates
+between the edge node and the cloud.
+
```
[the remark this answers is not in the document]
rohitsardesai83: done. added user stories. moved all rest to design details

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+|-----------------------|----------------------------------------------------------------------------------|
+| DeviceInstances       | A list of device instances. Each instance contains details like device model, protocol etc.|
+| Protocols             | A list of protocols used to communicate with devices. (e.g BLE, Modbus, OPC UA). <br />The configuration can contain more details about the connection mode , server, port details etc.|
+| DeviceModels          | A list of device models which contains the device properties and propertyVisitors. <br/> Device properties can be readable/writable. If a property is writable , the visitor for that property describes details of how to read that property. <br/> For e.g temperature sensor values can be read from a register with a specific uuid.|
+| Connection Type       | The connection type field indicates whether the device is directly connected <br /> to the cloud ( active devices ),or the device is connected via the edge platform(i.e indirect connection for passive devices). |
+| Node                  | The edge node details to which this device belongs to.|
```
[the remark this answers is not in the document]
rohitsardesai83: done

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+|Twins                   | A list of device twins for the properties which can be controlled ( the writable properties defined in the device model ). Each device twin attribute is like a control attribute which can be managed from the cloud or the edge. <br />Currently, the cloud configures the `Expected`state of a device property and this configuration update is pushed to the edge node. The mapper sends a command to the device to change this property value as per the expected state . It receives the `Actual` state of the property once the previous operation is complete and sends the actual state to the cloud. Offline device interaction in the edge is possible via twin properties for control/command operations.|
+|State                  | indicates the current device state ( unknown, running, offline  etc.)|
+
+The Device CRD needs to be installed in a running Kubernetes cluster which allows us to perform standard CRUD operations on devices. The group and domain need to be specified under which the APIs will reside.
+
+## Device Sample
+```yaml
+apiVersion: crd.kubeedge.io/v1alpha1
```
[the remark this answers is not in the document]
rohitsardesai83: done

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+        value: OFF
+```
+
+### Device Types Definition
+
+```go
+// DeviceSpec defines the desired state of Device
+type DeviceSpec struct {
+	DeviceInstances []DeviceInstance  `json:"deviceInstances,omitempty"`
+	DeviceModels []DeviceModel        `json:"deviceModels,omitempty"`
```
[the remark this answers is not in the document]
rohitsardesai83: done.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+## Device Sample
+```yaml
+apiVersion: crd.kubeedge.io/v1alpha1
+kind: Device
+metadata:
+  labels:
+    description: 'TI Simplelink SensorTag 2.0 with Bluetooth 4.0'
+    manufacturer: 'Texas Instruments'
+    model: CC2650
+  name: SensorTag
```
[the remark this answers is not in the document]
rohitsardesai83: done

rohitsardesai83:
@kevin-wangzefeng , @CindyXing , @qizha , I have updated the design with your review comments.
Some points are still open like secure device provisioning , scalability , conflict resolution. 
I hope we can address these concerns in the next version. Kindly provide your comments. Thanks a lot in advance.

docs/proposals/device-crd.md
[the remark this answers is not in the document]
rohitsardesai83: > It looks much better. Thanks Rohit!
> 
> There are three things I'd like us to think thoroughly.
> 
>     1. Persona for changing the expected device state and what kind of state we are talking about. In IOT scenarios, often customers' application would set the expected state and make things happen. This also includes the "actions" a device can support.  We should separate the attributes & state for IT admin and customers can change. For application ones, we probably don't want to sync those device twin data from edge back to cloud.
> 
>     2. I am not clear on what the "visitor" field is for; and how to use. can you give an example?
> 
>     3. Can you help understand why configmap is used for devicespec? In that case, why we create this CRD?


1. For point 1, I am not sure how do we segregate attributes versus state. Maybe we need some kind of RBAC here to classify what roles can do what actions ? Need  to do more investigation on this. Maybe device twin attributes can be consumed by developers who write mappers to connect to device and want to control the logical attributes. Adminstrative actions like reboot , upgrade , rollback etc could be WOT style actions ?

2. For point 2, I have added more details for the visitor. Please take a look and let me know if it is still not clear.

3. For point 3, this is also explained in a bit more detail in the device controller design. This was discussed with Kevin and Haibo. If we don't go with config map way , then we need to send all such metadata as separate events to edge and then probably the edge side code needs to be modified to handle these extra metadata. With the config map approach , we avoid changing any code on the edge side as we propagate the updates using existing config map sync mechanism. Thoughts ?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,207 @@
...
+deployed on the edge node, but there is no way to manage devices from the cloud.
+Device management is a key feature required for an edge computing platform.
+This proposal addresses how can we manage devices from the cloud and synchronize
+the device updates between edge nodes and cloud.
+
+### Goals
+
+Device management must:
+* provide APIs for managing devices from the cloud.
+* synchronize the device updates between cloud and edge nodes.
```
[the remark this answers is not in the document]
rohitsardesai83: With the device properties / visitors / protocol metadata stored and pushed to edge as config maps , very little or no changes are expected  in device twin code.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+|-----------------------|-------------------------|
+|Group                  | devices.kubeedge.io     |
+|APIVersion             | v1alpha1                |
+|Kind                   | Device                  |
+
+### Device Model and Instance Types Definition
+
+```go
+// DeviceModelSpec defines the desired state of the device model
+type DeviceModelSpec struct {
```
edisonxiang: It is better to adjust the indent for the fields and the fields' comments of the struct.
https://github.com/kubeedge/kubeedge/blob/f906d97f18048ded697dceb5c8b2d526eb4addba/docs/proposals/device-crd.md#device-model-and-instance-types-definition

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+<img src="../images/device-crd/device-model-crd.png">
+
+A `device model` describes the device properties exposed by the device and property visitors to access these properties. A device model is like a reusable template using which many devices can be created and managed.
+
+The `Spec` contains the following fields
+
+| Field                 | Description                                                                      |
+|-----------------------|----------------------------------------------------------------------------------|
+| Properties            | A list of device properties with details like data type , access mode, units etc.|
+| Property Visitors     | A list of property visitors. Each visitor describes how to access a device property. e.g. to access the temperature of a CC2650 SensorTag , the uuid of the register to access is specified.|
```
edisonxiang: Could the `Property Visitors` field be also called by `Twin Property` or `Dynamic Property`?
https://support.huaweicloud.com/api-ief/ief_02_0039.html
sorry currently there is no english document, but the example can make sense:)

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+  // protocol configuration to talk to the device
+	Protocol    Protocol              `json:"protocol,omitempty"`
+  // node to which the device belongs
+	NodeName    string                `json:"nodeName,omitempty"`
+}
+
+// Protocol configuration
+type Protocol struct {
+  // protocol name e.g. ble , modbus, opcua etc.
+  Name string              `json:"name,omitempty"`
```
kevin-wangzefeng: Probably make protocol name to pre-defined string, or use similar design with kubernetes pod.spec.volumes to make easier enabling validation for protocol config

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+A `device model` describes the device properties exposed by the device and property visitors to access these properties. A device model is like a reusable template using which many devices can be created and managed.
+
+The `Spec` contains the following fields
+
+| Field                 | Description                                                                      |
+|-----------------------|----------------------------------------------------------------------------------|
+| Properties            | A list of device properties with details like data type , access mode, units etc.|
+| Property Visitors     | A list of property visitors. Each visitor describes how to access a device property. e.g. to access the temperature of a CC2650 SensorTag , the uuid of the register to access is specified.|
+
+### Device model sample
```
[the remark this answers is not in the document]
m1093782566: +1 We want to see the full API definition first.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+  labels:
+    description: 'TI Simplelink SensorTag Device Attributes Model'
+    manufacturer: 'Texas Instruments'
+    model: CC2650
+  name: sensor-tag-model
+spec:
+  properties:
+  - name: temperature
+    description: 'Temperature from Temperature Sensor in Degrees Fahrenheit'
+    metadata:
```
m1093782566: I am not sure if we really need the metadata field. Theoretically, `name` can also be called metadata. Or can we come up with a better name?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+
+### Validation
+[Open API v3 Schema based validation](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation) can be used to guard against bad requests.
+
+## Synchronizing Device Twin Updates
+
+The below illustrations describe the flow of events that would occur when device twin expected / actual property values are updated from the cloud / edge.
+
+### Syncing Expected Device Twin Property Update From Cloud To Edge
+<img src="../images/device-crd/device-updates-cloud-edge.png">
```
edisonxiang: It lack of `step 3` in this png image

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,390 @@
...
+
+### Modelling Devices as CRD’s
+<img src="../images/device-crd/device-crd-model.png">
+
+[Custom resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) allows us to define declarative APIs for custom resources. Devices attached to edge nodes can be described using a custom resource and managed using APIs installed in a Kubernetes cluster. Users can interact with the device CRD’s using standard `kubectl` operations or using client libraries. The Device controller is a CRD controller whose job is to ensure the device updates are synced between the cloud and the edge node.
+
+### Device Custom Resource Overview
+The device custom resource layout is shown below. It contains standard fields ( Name, Namespace, Labels , Annotations etc. ) which help the Kubernetes ecosystem to manage resources. In addition, it contains custom fields in the Spec and Status detailed further below.
```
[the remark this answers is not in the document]
rohitsardesai83: removed references to the metadata fields.

docs/proposals/device-crd.md
rohitsardesai83: applied correct indentation. thanks.

docs/proposals/device-crd.md
rohitsardesai83: The naming has been put after discussing with zhenghaibo.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+
+## Motivation
+
+Device management is a key feature required for IoT use-cases in edge computing.
+This proposal addresses how can we manage devices from the cloud and synchronize
+the device updates between edge nodes and cloud.
+
+### Goals
+
+Device management must :
```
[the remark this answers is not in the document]
rohitsardesai83: done.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+  * Mapper applications running on the edge can report the device state as well the actual values of the control attributes.
+
+## Design Details
+
+### Device model CRD
+<img src="../images/device-crd/device-model-crd.png">
+
+A `device model` describes the device properties exposed by the device and property visitors to access these properties. A device model is like a reusable template using which many devices can be created and managed.
+
+The `Spec` contains the following fields
```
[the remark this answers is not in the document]
rohitsardesai83: done.

docs/proposals/device-crd.md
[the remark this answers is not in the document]
rohitsardesai83: done.

docs/proposals/device-crd.md
rohitsardesai83: The property has a name which can be referenced by the property visitor to define additional access mechanisms. The metadata is a generic key -value map which can contain additional stuff like what is the data type , what is the min / max range , what is the units etc. Since we could have different types of devices with varying properties , this metadata can contain any of such information.

docs/proposals/device-crd.md
rohitsardesai83: fixed. thanks.

docs/proposals/device-crd.md
CindyXing: Like what Kevin suggested. It is like a predefined enum.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+type DeviceModelSpec struct {
+  // list of device properties
+	DeviceProperties []DeviceProperty              `json:"properties,omitempty"`
+  // list of device property visitors
+  DevicePropertyVisitors []DevicePropertyVisitor `json:"propertyVisitors,omitempty"`
+}
+
+// Device property
+type DeviceProperty struct {
+  // name of the device property
```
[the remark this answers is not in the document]
rohitsardesai83: done.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,457 @@
...
+  // device property visitor description
+  Description string                    `json:"description,omitempty"`
+  // additional metadata about the how to access the device property
+  Metadata    map[string]string         `json:"metadata,omitempty"`
+}
+
+// DeviceSpec defines the desired state of Device
+type DeviceSpec struct {
+  // device model used to create the device
+	DeviceModel *core.ObjectReference `json:"deviceModel,omitempty"`
```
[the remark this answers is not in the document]
rohitsardesai83: I have changed it to LocalObjectReference. Please take a look.

docs/proposals/device-crd.md
rohitsardesai83: I have introduced constants for the protocol types. Thanks.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,532 @@
...
+          "name": "BluetoothLE",
+          "protocolConfig": {}
+        }
+      ],
+      "property_visitors": [
+        {
+          "name": "temperature",
+          "model_name": "SensorTagModel",
+          "protocol": "BluetoothLE",
+          "metadata": {
```
zhenghaibo1008: Maybe we can change metadata to protocol_config or visitor_config. I don't think the mean of metadata is very clear.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,532 @@
...
+              "datatype": "float",
+              "accessMode": "r",
+              "unit": "degree fahrenheit"
+            }
+          ]
+        }
+      ],
+      "protocols": [
+        {
+          "name": "BluetoothLE",
```
zhenghaibo1008: In our user scenario， the dpl will include more than one device. So we use name as a indentifier, and add protocol property to show which protocol the device used.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,532 @@
...
+        {
+          "name": "temperature",
+          "model_name": "SensorTagModel",
+          "protocol": "BluetoothLE",
+          "metadata": {
+            "uuid": "F000AA01-0451-4000-B000-000000000000"
+          }
+        },
+        {
+          "name": "temperatureEnable",
```
zhenghaibo1008: as the same we use name as indentifer.  Add property_name to show which property in the model.

docs/proposals/device-crd.md
rohitsardesai83: I have updated. please check.

docs/proposals/device-crd.md
rohitsardesai83: i have renamed to visitor_config , please check.

docs/proposals/device-crd.md
rohitsardesai83: added propertyName. please check.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+// ProtocolType describes the protocol type used to communicate with the device instance.
+type ProtocolType string
+
+const (
+	// Bluetooth Low Energy(BLE) Protocol
+	ProtocolTypeBluetoothLE ProtocolType = "BluetoothLE"
+	// Zigbee Protocol
+	ProtocolTypeZigbee      ProtocolType = "Zigbee"
+	// Modbus Protocol
+	ProtocolTypeModbus      ProtocolType = "Modbus"
```
zhenghaibo1008: modbus has two types. modbus-rtu and modbus-tcp

rohitsardesai83: i have added the two variants of modbus. thanks @zhenghaibo1008

CindyXing:
For the protocol data struct, the name string and type enum const seems a dup. Otherwise, the design looks good.

rohitsardesai83:
> For the protocol data struct, the name string and type enum const seems a dup. Otherwise, the design looks good.

Assume 2 devices A and B wish to use bluetooth protocol with different configurations , device A could refer to the first protocol with name  say "ble-01" , and device B could refer to the second protocol with the name say "ble-02". The type is the same for both that is BluetoothLE. I had discussed with Haibo about this earlier.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+  // Required.
+  Visitors     []PropertyVisitor `json:"visitors,omitempty"`
+}
+
+// ProtocolType describes the protocol type used to communicate with the device instance.
+type ProtocolType string
+
+const (
+	// Bluetooth Low Energy(BLE) Protocol
+	ProtocolTypeBluetoothLE  ProtocolType = "BluetoothLE"
```
m1093782566: Remove the `ProtocolType` prefix here and below?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+// PropertyVisitor contains details like which property can be accessed and the
+// access mechanisms are described via key-value pairs in the metadata.
+type PropertyVisitor struct {
+  // The device property visitor name.
+  // Required.
+  Name         string            `json:"name,omitempty"`
+  // The device property name to be accessed. This should refer to one of the
+  // device properties defined in the device model.
+  // Required.
+  PropertyName string            `json:"propertyName,omitempty"`
```
m1093782566: DevicePropertyName?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+  // The property name.
+  // Required.
+  Name        string            `json:"name,omitempty"`
+  // The property description.
+  // +optional
+  Description string            `json:"description,omitempty"`
+  // Additional metadata about the device property like datatype, range, min, max etc.
+  // Optional: If no metadata is provided, it is assumed that the property is an empty read-only string.
+  // +optional
+  Metadata    map[string]string `json:"metadata,omitempty"`
```
m1093782566: A silly question, who will update the `Metadata` field?

m1093782566: metadata -> parameters?

A better name?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+    metadata:
+      dataType: float
+      accessMode: r
+      unit: 'degree fahrenheit'
+  - name: temperature-enable
+    description: 'Enable Temperature Sensor'
+    metadata:
+      dataType: string
+      accessMode: rw
+  propertyVisitors:
```
m1093782566: Please give an example on metadata in propertyVisitors.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+```
+
+The above example is for a [CC2650 SensorTag](http://processors.wiki.ti.com/index.php/CC2650_SensorTag_User's_Guide) device. The SensorTag exposes all its sensors via [Bluetooth Smart (Bluetooth Low Energy)](https://en.wikipedia.org/wiki/Bluetooth_Low_Energy) services. The sensor drivers interface to a GATT server running on TI BLE-Stack v2. The GATT server contains a primary service for each sensor for configuration and data collection. Each service exposes charactersitics which can be accessed via 128 bit UUIDs.
+
+The sample shows two charactersitics being exposed to the user. Access modes describe whether the charactersitic is read-only or can be written to. For the temperature sensor, the temperature is a read-only charactersitic . The sensor can be configured to enable data collection via another charactersitic.
+
+Property visitors provide such additional details like how to access these device properties. For example, the temperature can be read from charactersitic with UUID `F000AA01-0451-4000-B000-000000000000`. If we need to enable data collection, we can write to the charactersitic with UUID `F000AA02-0451-4000-B000-000000000000`.
```
m1093782566: Visitor seems very tricky to me and the meaningful field of Visitor are only propertyName and a map structure. Can we reduce the concepts?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+	ProtocolTypeOPCUA        ProtocolType = "OPCUA"
+)
+
+// PropertyVisitor contains details like which property can be accessed and the
+// access mechanisms are described via key-value pairs in the metadata.
+type PropertyVisitor struct {
+  // The device property visitor name.
+  // Required.
+  Name         string            `json:"name,omitempty"`
+  // The device property name to be accessed. This should refer to one of the
```
m1093782566: device property -> device attributes.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+	ProtocolTypeModbusTCP    ProtocolType = "ModbusTCP"
+	// BACnet Protocol
+	ProtocolTypeBACnet       ProtocolType = "BACnet"
+	// OPC UA Protocol
+	ProtocolTypeOPCUA        ProtocolType = "OPCUA"
+)
+
+// PropertyVisitor contains details like which property can be accessed and the
+// access mechanisms are described via key-value pairs in the metadata.
+type PropertyVisitor struct {
```
m1093782566: PropertyVisitor -> AttributesVisitor?

I still think visitor is not a good name, can we come up with a better name?

Is visitor read-only? Fecther? Getter?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+          uuid: F000AA01-0451-4000-B000-000000000000
+      - name: temperatureEnable
+        propertyName: temperature-enable
+        visitorConfig:
+          uuid: F000AA02-0451-4000-B000-000000000000
+```
+
+The above example is for a [CC2650 SensorTag](http://processors.wiki.ti.com/index.php/CC2650_SensorTag_User's_Guide) device. The SensorTag exposes all its sensors via [Bluetooth Smart (Bluetooth Low Energy)](https://en.wikipedia.org/wiki/Bluetooth_Low_Energy) services. The sensor drivers interface to a GATT server running on TI BLE-Stack v2. The GATT server contains a primary service for each sensor for configuration and data collection. Each service exposes charactersitics which can be accessed via 128 bit UUIDs.
+
+The sample shows two charactersitics being exposed to the user. Access modes describe whether the charactersitic is read-only or can be written to. For the temperature sensor, the temperature is a read-only charactersitic . The sensor can be configured to enable data collection via another charactersitic.
```
m1093782566: > Access modes describe whether the charactersitic is read-only or can be written to

I can't see the access mode in the above example.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+  // +optional
+  Description string            `json:"description,omitempty"`
+  // Additional metadata about the device property like datatype, range, min, max etc.
+  // Optional: If no metadata is provided, it is assumed that the property is an empty read-only string.
+  // +optional
+  Metadata    map[string]string `json:"metadata,omitempty"`
+}
+
+// DevicePropertyVisitor describes the specifics of accessing a particular device
+// property. Visitors are intended to be consumed by device mappers which connect to devices
```
m1093782566: > Visitors are intended to be consumed by device mappers which connect to devices

Can you give more detailed information on how device mapper relies on the visitors?

docs/proposals/device-crd.md
zhenghaibo1008: @m1093782566 visitor is not read-only. We use the visitor to access the device's property. if  we want to access the property, we should know what we need with the protocol driver. Such as modbus, we should konw which type of register or  from which one register. and in opc-ua, we should konw the node-id of property.

m1093782566: Okay, So I suggest to give a more descriptive API name, e.g. DeviceAtributesVisitor?

WDYT?

docs/proposals/device-crd.md
rohitsardesai83: In property visitors , we don't have metadata, we have `visitorConfig` which is there in example.

docs/proposals/device-crd.md
rohitsardesai83: These are not parameters to the property, but additional information about the property. 
we could make it similar to visitorConfig ?  like propertyConfig ? thoughts ?

rohitsardesai83: these could be updated by the user who creates the device model via kubectl. The update would be a kubectl patch operation.

docs/proposals/device-crd.md
rohitsardesai83: since this is part of the DevicePropertyVisitor , it is ok to just keep it as PropertyName ?

m1093782566: Indeed, I suggest change DevicePropertyVisitor to DeviceAttributeVisitor.

rohitsardesai83: we wish to align with https://iot.mozilla.org/wot/. Hence we have named them as properties and not attributes.

docs/proposals/device-crd.md
rohitsardesai83: @zhenghaibo1008 , please provide your thoughts ?

docs/proposals/device-crd.md
rohitsardesai83: as pointed out earlier , we are trying alignment with mozilla web of things.

docs/proposals/device-crd.md
rohitsardesai83:
```yaml
metadata:
      dataType: string
      accessMode: rw
```

docs/proposals/device-crd.md
rohitsardesai83: DevicePropertyVisitor is for a device with this struct : 
```go
type DevicePropertyVisitor struct {
  // The protocol type to connect to the device and access the property.
  // Required.
  Protocol     ProtocolType      `json:"protocol,omitempty"`
  // A list of visitors describing the properties which can be accessed with the protocol.
  // Required.
  Visitors     []PropertyVisitor `json:"visitors,omitempty"`
}
```

It has a list of PropertyVisitor's : 
```go
// PropertyVisitor contains details like which property can be accessed and the
// access mechanisms are described via key-value pairs in the metadata.
type PropertyVisitor struct {
  // The device property visitor name.
  // Required.
  Name         string            `json:"name,omitempty"`
  // The device property name to be accessed. This should refer to one of the
  // device properties defined in the device model.
  // Required.
  PropertyName string            `json:"propertyName,omitempty"`
  // Additional metadata about the how to access the device property.
  // Required.
  Metadata     map[string]string `json:"metadata,omitempty"`
}
```
We are referring to device properties as DeviceProperty types and a means to visit them as DevicePropertyVisitor. 

Mozilla Web Of Things also has concept of "properties" rather than "attributes".

m1093782566: "properties" VS. "attributes", it's an open question. I think the API name should be simple and easy to understand. At least, attribute is simpler than property in English. Please consider people who are not good at English :)

docs/proposals/device-crd.md
m1093782566: We can keep coming up with better names. I am sure metadata is not a good name.

docs/proposals/device-crd.md
rohitsardesai83: removed.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+A `device` instance represents an actual device object. It is like an instantiation of the `device model` and references properties defined in the model. It additionally reports the expected and actual values of the properties, protocol details like server url etc.
+
+### Device instance type definition
+```go
+// DeviceSpec represents a single device instance. It is an instantation of a device model.
+type DeviceSpec struct {
+  // DeviceModelRef is reference to the device model used as a template
+  // to create the device instance.
+  // Required.
+  DeviceModelRef *core.LocalObjectReference `json:"deviceModelRef,omitempty"`
```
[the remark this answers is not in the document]
rohitsardesai83: As per Kevin's suggestion , I have named it as DeviceModelRef: https://github.com/kubeedge/kubeedge/pull/181#discussion_r265109905

Also I observed similar convention in Kubernetes for SecretRef : 
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L702

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,556 @@
...
+// ProtocolType describes the protocol type used to communicate with the device instance.
+type ProtocolType string
+
+const (
+	// Bluetooth Low Energy(BLE) Protocol
+	ProtocolTypeBluetoothLE  ProtocolType = "BluetoothLE"
+	// Zigbee Protocol
+	ProtocolTypeZigbee       ProtocolType = "Zigbee"
+	// Modbus RTU Protocol
+	ProtocolTypeModbusRTU    ProtocolType = "ModbusRTU"
```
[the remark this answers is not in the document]
rohitsardesai83: updated to single protocol.

docs/proposals/device-crd.md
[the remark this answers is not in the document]
m1093782566: Okay, fair enough. I buy it.

docs/proposals/device-crd.md
rohitsardesai83: renamed to `propertyConfig` , to be consistent with `visitorConfig`.

docs/proposals/device-crd.md
rohitsardesai83: I am not really clear how "attribute" is simpler than "property" in English :).

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+<img src="../images/device-crd/device-model-crd.png">
+
+A `device model` describes the device properties exposed by the device and property visitors to access these properties. A device model is like a reusable template using which many devices can be created and managed.
+
+### Device Model Type Definition
+```go
+// DeviceModelSpec defines the model / template for a device.It is a blueprint which describes the device
+// capabilities and access mechanism via property visitors.
+type DeviceModelSpec struct {
+  // List of device attributes.
```
m1093782566: If you like the word property, why you comment in the word "attributes"? Looks wired.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+  // The property name.
+  // Required.
+  Name            string            `json:"name,omitempty"`
+  // The property description.
+  // +optional
+  Description     string            `json:"description,omitempty"`
+  // Additional configuration about the device property like datatype, range, min, max etc.
+  // Optional: If no configuration details are provided, it is assumed that the property is an empty read-only string.
+  // +optional
+  PropertyConfig  map[string]string `json:"propertyConfig,omitempty"`
```
m1093782566: Since `PropertyConfig` is in `DeviceProperty`, why not simply say config? We can discuss before changing :)

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+  // Required.
+  DeviceProperties       []DeviceProperty        `json:"properties,omitempty"`
+  // List of property visitors which describe how to access the device attributes.
+  // Required.
+  DevicePropertyVisitors []DevicePropertyVisitor `json:"propertyVisitors,omitempty"`
+}
+
+// DeviceProperty describes an individual device property / attribute like temperature / humidity etc.
+type DeviceProperty struct {
+  // The property name.
```
m1093782566: The device property name.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+  // Required.
+  Name            string            `json:"name,omitempty"`
+  // The property description.
```
m1093782566: The device property description.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+  // +optional
+  Description     string            `json:"description,omitempty"`
+  // Additional configuration about the device property like datatype, range, min, max etc.
+  // Optional: If no configuration details are provided, it is assumed that the property is an empty read-only string.
```
m1093782566: > If no configuration details are provided, it is assumed that the property is an empty read-only string

Sorry I don't quite understand. Please explain it.

docs/proposals/device-crd.md
m1093782566: I think `property.additionalConfigs` is better than `property.propertyConfig`.

docs/proposals/device-crd.md
m1093782566: So why not just don't mention the information on device mapper here?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+// We can define multiple visitors per protocol for a device. For e.g If a device supports Bluetooth
+// protocol , the visitors would describe the 128-bit UUIDs to read / write charactersitics of a Bluetooth
+// service exposed by the device.
+type DevicePropertyVisitor struct {
+  // The protocol type to connect to the device and access the property.
+  // Required.
+  Protocol     ProtocolType      `json:"protocol,omitempty"`
+  // A list of visitors describing the properties which can be accessed with the protocol.
+  // Required.
+  Visitors     []PropertyVisitor `json:"visitors,omitempty"`
```
m1093782566: Define a visitor array in visitor is not a good idea.

docs/proposals/device-crd.md
zhenghaibo1008: both attribute and property are fine.In c#,  attribute is very simple, just with the type definition. But property is complicate， it has getter and setter. In our scenario, maybe property is more suitable

docs/proposals/device-crd.md
rohitsardesai83: there is no "liking" for the word "property". we were just aliging with standards to some extent. I will update the description if it's so much confusing :)

docs/proposals/device-crd.md
rohitsardesai83: Kevin suggested we could have a separate design for how mapper will consume this. The current mapper design ( which is already approved ), is not in sync with this design. we should have done the device management design first and then introduce mappers.

docs/proposals/device-crd.md
qizha: I vote for property since that can align with WOT standard(https://iot.mozilla.org/wot/#properties-member)

docs/proposals/device-crd.md
m1093782566: OK, I buy it :)

docs/proposals/device-crd.md
m1093782566: OK, Thanks!

docs/proposals/device-crd.md
rohitsardesai83: :+1:

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+      accessMode: r
+      unit: 'degree fahrenheit'
+  - name: temperature-enable
+    description: 'Enable Temperature Sensor'
+    propertyConfig:
+      dataType: string
+      accessMode: rw
+  propertyVisitors:
+    - protocol:  BluetoothLE # valid values : [BluetoothLE, Zigbee, ModbusTCP, ModbusRTU, BACnet, OPCUA]
+      visitors:
```
m1093782566: So one `propertyVisitor` can have multiple `visitors`? What's the user cases?

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+      - name: temperature
+        propertyName: temperature
+        visitorConfig:
+          uuid: F000AA01-0451-4000-B000-000000000000
+      - name: temperatureEnable
+        propertyName: temperature-enable
+        visitorConfig:
```
m1093782566: visitorConfig -> visitConfigs? means we are going to visit the configs?

docs/proposals/device-crd.md
rohitsardesai83: for example , the temperature property of a device could be accessed using different protocols , so each visitor has the protocol to be used to access the property.

m1093782566: Okay, sounds fair.

docs/proposals/device-crd.md
rohitsardesai83: in this example , it means that if you want to enable data collection for temperature from a sensor tag, you need to access a register with a specific uuid. there could be many more details other than uuid like at what index , offset to start reading data , what is the byte-size of the data etc.

m1093782566: As there must be multiple configurations, I suggest add `s` in this field. e.g `visitorConfigs` or `visitConfigs`

docs/proposals/device-crd.md
rohitsardesai83: the top level visitor is per protocol , within each protocol visitor , we have property visitors which can be read using that protocol.

```yaml
    - protocolName: opc-ua    // this is the top level visitor supporting opc-ua
      visitors:
      - name: temperature
        propertyName: Temperature
      - name: pressure
        propertyName: Pressure
      - name: humidity
        propertyName: Humidity
        
    - protocolName: modebus // this is the top level visitor supporting modbus
      visitors:
      - name: humidity
        propertyName: Humidity
```

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+  // The device property visitor name.
+  // Required.
+  Name         string            `json:"name,omitempty"`
+  // The device property name to be accessed. This should refer to one of the
+  // device properties defined in the device model.
+  // Required.
+  PropertyName string            `json:"propertyName,omitempty"`
+  // Additional metadata about the how to access the device property.
+  // Required.
+  Metadata     map[string]string `json:"metadata,omitempty"`
```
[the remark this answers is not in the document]
rohitsardesai83: additional means something more about the device property.
in the comment I have mentioned an example as well : 

Additional configuration about the device property **like datatype, range, min, max etc.**

docs/proposals/device-crd.md
rohitsardesai83: for example , if property is "temperature" and no property config is provided , it will be assumed that the data type is a string and value is ""

kevin-wangzefeng:
Please not squash or force push before it's solid.

docs/proposals/device-crd.md
rohitsardesai83: the new API struct doesn't have the map.

docs/proposals/device-crd.md
rohitsardesai83: updated APIs don't have a map now.

docs/proposals/device-crd.md
rohitsardesai83: updated APIs don't have additional config.

docs/proposals/device-crd.md
rohitsardesai83: updated APIs don't have this struct now.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,554 @@
...
+### Device Model Type Definition
+```go
+// DeviceModelSpec defines the model / template for a device.It is a blueprint which describes the device
+// capabilities and access mechanism via property visitors.
+type DeviceModelSpec struct {
+  // List of device attributes.
+  // Required.
+  DeviceProperties       []DeviceProperty        `json:"properties,omitempty"`
+  // List of property visitors which describe how to access the device attributes.
+  // Required.
```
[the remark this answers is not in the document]
rohitsardesai83: updated API struct doesn't have per protocol visitors.

docs/proposals/device-crd.md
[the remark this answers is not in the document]
rohitsardesai83: removed Optional as it's not needed.

docs/proposals/device-crd.md
rohitsardesai83: updated APIs don't have this struct

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,650 @@
...
+        register: DiscreteInputRegister
+        offset: 2
+        limit: 1
+        scale: 1
+        isSwap: true
+        isRegisterSwap: true
+```
+
+The device model shown above describes the temperature property of a sensor. Access modes describe whether the property is read-only or can be written to.
+Property visitors provide details like how to access device properties. In the above example , the visitor is for Modbus protocol and describes the register to read data , the offset to apply etc.
```
kevin-wangzefeng: Please remove all the whitespaces before ",":`s/\ ,/,/g`

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,655 @@
...
+kind: Device
+metadata:
+  name: sensor-tag01
+  labels:
+    description: 'TI Simplelink SensorTag 2.0 with Bluetooth 4.0'
+    manufacturer: 'Texas Instruments'
+    model: CC2650
+spec:
+  deviceModelRef:
+    Name: sensor-tag-model
```
kevin-wangzefeng: s/Name/name/

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,650 @@
...
+  propertyVisitors:
+  - propertyName: temperature
+    modbus:
+      register: CoilRegister
+      offset: 2
+      limit: 1
+      scale: 1
+      isSwap: true
+      isRegisterSwap: true
+   - propertyName: temperature-enable
```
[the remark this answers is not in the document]
rohitsardesai83: fixed.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,650 @@
...
+Property visitors provide details like how to access device properties. In the above example , the visitor is for Modbus protocol and describes the register to read data , the offset to apply etc.
+
+### Device instance CRD
+<img src="../images/device-crd/device-crd.png">
+
+A `device` instance represents an actual device object. It is like an instantiation of the `device model` and references properties defined in the model. It additionally reports the expected and actual values of the properties, protocol details like server url etc.
+
+### Device instance type definition
+```go
+// DeviceSpec represents a single device instance. It is an instantation of a device model.
```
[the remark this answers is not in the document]
rohitsardesai83: fixed.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,650 @@
...
+	Parity     string `json:"parity,omitempty"`
+	// Required. Bit that stops 1|2
+	StopBits   int64  `json:"stopBits,omitempty"`
+	// Required. 0-255
+	SlaveID    int64  `json:"slaveID,omitempty"`
+}
+
+// DeviceStatus reports the device state and the expected/actual values of twin attributes.
+type DeviceStatus struct {
+	// A list of device twins containing expected/actual states of control properties.
```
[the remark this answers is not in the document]
rohitsardesai83: fixed.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,650 @@
...
+	// Optional: A passive device won't have control attributes and this list could be empty.
+	// +optional
+	Twins []Twin `json:"twins,omitempty"`
+}
+
+// A Twin provides a logical representation of control properties (writable properties in the
+// device model). The properties can have an Expected state and an Actual state. The cloud configures
```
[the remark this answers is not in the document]
rohitsardesai83: fixed.

docs/proposals/device-crd.md
rohitsardesai83: fixed.

docs/proposals/device-crd.md
rohitsardesai83: fixed.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,650 @@
...
+    * [Device model CRD](#device-model-crd)
+    * [Device model type definition](#device-model-type-definition)
+    * [Device model sample](#device-model-sample)
+    * [Device instance CRD](#device-instance-crd)
+    * [Device instance type definition](#device-instance-type-definition)
+    * [Device instance sample](#device-instance-sample)
+    * [CRD API Group and Version](#crd-api-group-and-version)
+    * [Validation](#validation)
+    * [Synchronizing Device Twin Updates](#synchronizing-device-twin-updates)
+      * [Syncing Expected Device Twin Property Update From Cloud To Edge](#syncing-expected-device-twin-property-update-from-cloud-to-edge)
```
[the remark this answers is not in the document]
rohitsardesai83: fixed.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,650 @@
...
+          timestamp: '1550049403598'
+          type: string
+        value: OFF
+      expected:
+        metadata:
+          timestamp: '1550049403598'
+          type: string
+        value: OFF
+```
+### CRD API Group and Version
```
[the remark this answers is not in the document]
rohitsardesai83: fixed.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,650 @@
...
+|Group                  | devices.kubeedge.io     |
+|APIVersion             | v1alpha1                |
+|Kind                   | Device                  |
+
+### Validation
+[Open API v3 Schema based validation](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation) can be used to guard against bad requests.
+
+## Synchronizing Device Twin Updates
+
+The below illustrations describe the flow of events that would occur when device twin expected/actual property values are updated from the cloud/edge.
```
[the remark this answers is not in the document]
rohitsardesai83: fixed.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,670 @@
...
+  nodeSelector:
+    nodeSelectorTerms:
+    - matchExpressions:
+      - key: ''
+        operator: In
+        values:
+        - node1
+status:
+  twins:
+    - name: temperature-enable
```
kevin-wangzefeng: s/name/propertyName/

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,670 @@
...
+  name: device1-config
+  namespace: foo
+data:
+  deviceProfile.json: |-
+    {
+      "deviceInstances": [
+        {
+          "id": "1",
+          "name": "device1",
+          "protocol": "ble-01", // TODO: not supported in API definition
```
kevin-wangzefeng: It's actually the name of protocolConfig, and only needed when converting from devices and deviceModels to the mappers' config. You can make it generated by controller.

rohitsardesai83: ok got it.

docs/proposals/device-crd.md
rohitsardesai83: fixed.

rohitsardesai83:
@qizha , @CindyXing , @m1093782566 , please review. Kevin and myself have reworked on the API definitions etc.

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,671 @@
...
+	// Defaults to "none".
+	// +optional
+	SecurityMode   string `json:"securityMode,omitempty"`
+	// Certificate for access opc server.
+	// +optional
+	Certificate    string `json:"certificate,omitempty"`
+	// PrivateKey for access opc server.
+	// +optional
+	PrivateKey     string `json:"privateKey,omitempty"`
+	// Timeout seconds for the opc server connection.???
```
kevin-wangzefeng: Remove "???"

docs/proposals/device-crd.md
```diff
@@ -0,0 +1,671 @@
...
+
+type ProtocolConfigOpcUA struct {
+	// Required: The URL for opc server endpoint.
+	Url            string `json:"url,omitempty"`
+	// Username for access opc server.
+	// +optional
+	UserName       string `json:"userName,omitempty"`
+	// Password for access opc server.
+	// +optional
+	Password       string `json:"password,omitempty"`
```
m1093782566: Is it reasonable to give password in the config? If we have to, should we encode it with base64 or other cryptographic algorithm? @zhenghaibo1008

qizha:
LGTM

docs/proposals/device-crd.md
m1093782566: Just have a chat with @zhenghaibo1008 offline, we will store users' credentials with some cryptographic algorithms in cloud, so the credentials only visible for users.

I buy it :)

m1093782566:
Once @CindyXing approved, I think we can merge it. Thanks for the big efforts!

CindyXing:
Lgtm

PR #209, opened 2019-03-07
DeviceTwin: Modifying testcases of dttype

lidiyag:
This commit modifies some of testcases in types_helper.go and types.go. Also it removes some unwanted code from edgehub/common/http/http_test.go

**What type of PR is this?**
/kind test

This PR fixes some flaky test cases in devicetwin/dttype and edgehub/common/http

### DIFF
--- a/edge/pkg/devicetwin/dttype/types_helper_test.go
+++ b/edge/pkg/devicetwin/dttype/types_helper_test.go
@@ -599,33 +599,42 @@ func TestBuildDeviceTwinResult(t *testing.T) {
 
 // TestBuildErrorResult is function to test BuildErrorResult().
 func TestBuildErrorResult(t *testing.T) {
-	result := Result{BaseMessage: BaseMessage{Timestamp: time.Now().UnixNano() / 1e6,
+	result := Result{BaseMessage: BaseMessage{
 		EventID: ""},
 		Code:   1,
 		Reason: ""}
-	bytesResult, _ := json.Marshal(result)
 	tests := []struct {
 		name    string
 		para    Parameter
-		want    []byte
+		want    Result
 		wantErr error
 	}{
 		{
 			name:    "BuildErrorResultTest",
 			para:    Parameter{EventID: "", Code: 1, Reason: ""},
-			want:    bytesResult,
+			want:    result,
 			wantErr: nil,
 		},
 	}
 	for _, test := range tests {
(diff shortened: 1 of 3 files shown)


edge/pkg/devicetwin/dttype/types_helper_test.go
```diff
@@ -599,33 +599,42 @@ func TestBuildDeviceTwinResult(t *testing.T) {
 
 // TestBuildErrorResult is function to test BuildErrorResult().
 func TestBuildErrorResult(t *testing.T) {
-	result := Result{BaseMessage: BaseMessage{Timestamp: time.Now().UnixNano() / 1e6,
```
rohitsardesai83: any reason why the timestamp is now removed ?

lidiyag: That field was not used previously. Spotted it and just removed.

lidiyag: Timestamp is assigned from the source code function. So once we assign a timestamp from test case , there is very less chance that the same may get assigned in source code as well which causes test case to fail randomly. So removed the timestamp check and checking only other fields in the struct.

edge/pkg/devicetwin/dttype/types_helper_test.go
```diff
@@ -761,7 +770,8 @@ func TestBuildDeviceTwinDocument(t *testing.T) {
 	twinDoc := make(map[string]*TwinDoc)
 	doc := TwinDoc{LastState: &MsgTwin{Metadata: &TypeMetadata{"updated"}}, CurrentState: &MsgTwin{Metadata: &TypeMetadata{"deleted"}}}
 	twinDoc["SensorTag"] = &doc
-	devTwinDoc := DeviceTwinDocument{BaseMessage: BaseMessage{EventID: "", Timestamp: time.Now().UnixNano() / 1e6}, Twin: twinDoc}
+	timeStamp := time.Now().UnixNano() / 1e6
```
rohitsardesai83: based on previous comment , whether we need to assign timestamp in the test ? is it assigned in the source code ?
will this cause random failures ?

lidiyag: Here the timestamp is a field of BaseMessage which is directly passed as a parameter to the source code and is not getting modified. So this will not cause random failures.

PR #227, opened 2019-03-12
Document pod spec limitation on edge node

### LINKED ISSUE #195: Document limitations of edge
**What happened**:
Tried to deploy a pod with host and container port different. Found out that it was a limitation of edged.

Pod to be created normally.

We need to document the limitations of edged so that users/developers can keep in mind those limitations while creating deployments.

nkwangleiGIT:
**What type of PR is this?**
/kind documentation

Document the pod spec limitation about hostPort on edge node

### DIFF
--- a/docs/setup/setup.md
+++ b/docs/setup/setup.md
@@ -218,6 +218,9 @@ Try out a sample application deployment by following below steps
 ```shell
 kubectl apply -f $GOPATH/src/github.com/kubeedge/kubeedge/build/deployment.yaml
 ```
+**Note:**
+Currently, for edge node, we must use hostPort in the Pod container spec to make it started normally, or the pod will be always in *ContainerCreating* status. And the hostPort must be equal to containerPort and can not be 0.
+
 ## Run Edge Unit Tests
 
  ```shell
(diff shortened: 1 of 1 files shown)


sids-b:
@nkwangleiGIT , This is one of the limitation in edged. We don't support many other features supported by full Kubernetes kubelet. We intend to document all of them as a fix for issue #195 . @samy2019 can collaborate with you on this. What say ?

Also it would be better if we have a separate document for this :)
cc @m1093782566 @rohitsardesai83

nkwangleiGIT:
Yes, I notice the description of #195, but I'm not sure all limitations of current edged, do we have any internal docs or some hints to help on this issue?

Or we can run k8s conformance tests against edge nodes to find the potential limitations?

If it'll take time to finish all these, maybe we can add this note for now for this common limitation.

docs/setup/setup.md
```diff
@@ -218,6 +218,9 @@ Try out a sample application deployment by following below steps
 ```shell
 kubectl apply -f $GOPATH/src/github.com/kubeedge/kubeedge/build/deployment.yaml
 ```
+**Note:**
+Currently, for edge node, we must use hostPort in the Pod container spec to make it started normally, or the pod will be always in *ContainerCreating* status. And the hostPort must be equal to containerPort and can not be 0.
```
rohitsardesai83: to make it started normally  --> so that the pod comes up normally.

And the hostPort  -> ~~And~~ the hostPort

PR #236, opened 2019-03-14
Add build files to build cloud image and manifests to run in k8s cluster

nkwangleiGIT:
**What type of PR is this?**
/kind feature

Allow users to easily run edgecontroller inside k8s cluster

Will add the instruction about how to use kubectl to run edgecontroller in k8s cluster, and will also add doc about how to run edged in container later.


Makefile
```diff
@@ -25,3 +25,6 @@ edge_cross_build:
 edgecontroller:
 	cd cloud/edgecontroller && $(MAKE)
 
+.PHONY: cloudimage
+cloudimage:
+	docker build -t kubeedge/edgecontroller:v0.2 -f build/cloud/Dockerfile .
```
m1093782566: make image tag configurable?

build/cloud/03-clusterrole.yaml
```diff
@@ -0,0 +1,20 @@
...
+  resources:
+  - nodes
+  - configmaps
+  - pods
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
+  - update
```
m1093782566: Please add a newline please.

build/cloud/06-secret.yaml
```diff
@@ -0,0 +1,100 @@
+apiVersion: v1
```
m1093782566: Can we generate the cert files at runtime?

build/cloud/08-service.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+    k8s-app: kubeedge
+    kubeedge: edgecontroller
+spec:
+  ports:
+  - port: 10000
+    name: cloudhub
+  selector:
+    k8s-app: kubeedge
+    kubeedge: edgecontroller
```
m1093782566: Add a newline please.

build/cloud/07-deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+    matchLabels:
+      k8s-app: kubeedge
+      kubeedge: edgecontroller
+  template:
+    metadata:
+      labels:
+        k8s-app: kubeedge
+        kubeedge: edgecontroller
+      annotations:
+        sidecar.istio.io/inject: "false"
```
rohitsardesai83: why is this annotation specified ?

nkwangleiGIT: We can remove this, it's just used to disable istio sidecar injection

rohitsardesai83: better we remove. thanks.

tedli:
https://github.com/tenxcloud/kubeedge/pull/2

here is the up coming fixs.

nkwangleiGIT:
Some questions here:
1）Is there any registry server that we can use to upload all KubeEdge images?
2）Is it enough to only mount docker.sock to run edged inside a container? for k8s kubelet, we have more to configure to make it run inside a container.

edge/pkg/eventbus/event_bus.go
```diff
@@ -35,13 +36,24 @@ type eventbus struct {
 
 func init() {
 	mode := config.CONFIG.GetConfigurationByKey("mqtt.mode")
+	var mqttMode int
+	var ok bool
 	if mode == nil {
-		mode = internalMqttMode
+		mqttMode = internalMqttMode
+	} else {
+		if mqttMode, ok = mode.(int); !ok {
```
m1093782566: @nkwangleiGIT 

Are you checking if mqttMode is an integer? Can we simply do something like:

```
i, err := strconv.Atoi("mqttMode")
if err != nil {
  panic("mqtt.mode should be an integer")
}
```

tedli: The story is that, the original logic get the mqtt.mode from yaml, which the whole yaml content is parsed into
```golang
map[string]interface{}
```
so the underneath mqtt.mode value type is `int` what can cast to int value directly.

However, when deploying using container, we use environment variable to overwrite these config entries, and mqtt.mode parsed into string, so it panics when casting to int.

So adds the conditions to parse mqtt.mode no matter it is a string or int value.

m1093782566:
@nkwangleiGIT 

> 1）Is there any registry server that we can use to upload all KubeEdge images?

Yes, we have created the kubeedge org and edgecontroller repository in dockerhub.

edge/pkg/eventbus/event_bus.go
m1093782566: Okay, we need some UTs for this change.

build/edge/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM golang:1.12-alpine3.9 AS builder
```
m1093782566: @sids-b 

Did we test if edge-core can run inside a container?

nkwangleiGIT:
> @nkwangleiGIT
> 
> > 1）Is there any registry server that we can use to upload all KubeEdge images?
> 
> Yes, we have created the kubeedge org and edgecontroller repository in dockerhub.

Can we get the write access to the repositories? or you have the owner to do this? We can start to simplify the document once the images are ready on dockerhub.

m1093782566:
@nkwangleiGIT 

No worries, I will manage the docker registry.

tedli:
> 2）Is it enough to only mount docker.sock to run edged inside a container? for k8s kubelet, we have more to configure to make it run inside a container.

Please mark this PR as `WIP` since edgecore can not fully functioned with just docker.sock mounted.

I'm working on it now.

m1093782566:
@nkwangleiGIT @tedli 

I would be appreciated if you can cut this PR into smaller ones for the sake of easily review and merge.

nkwangleiGIT:
@tedli OK, let's close this one and split it to some small PRs

PR #245, opened 2019-03-18
use value interface of goarchaius config framework to get mqtt mode

### LINKED ISSUE #249: panics if using environment variable (mqtt.mode) to overwrite yaml config of edge core
**What happened**:

edge core process panics. (if mqtt.mode config entry is overwritten by using environment variable)

parse configs succeeded without any panic.

```bash
export mqtt.mode=0
./edge_core
```

the goarchaius config framework parse yaml config into a `map[string]interface{}`. so if a config value in yaml is int, then the parsed object is an int what can directly cast to int. But the overwritten value of environment variable is always a string.

**Environment**:
- KubeEdge version: v0.2
- Others: irrelevant

tedli:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Use value interface of goarchaius config framework to get mqtt mode. The original logic cast directly to int, it may panic if the underneath value is not an int, this happens when using environment variable overwriting config entry, the value always be a string. The value interface of goarchaius provides many handy `ToXX` methods that can handle this well.

Fixes #249 

The `mqtt.mode` config entry was config by yaml by default, the yaml content is parsed into a

```golang
map[string]interface{}
```

so with `mqtt.mode` key, we can get the int value. However if we overwrite this config entry by environment variable, with the same key, we got a string literal of the number.

### DIFF
--- a/edge/pkg/eventbus/event_bus.go
+++ b/edge/pkg/eventbus/event_bus.go
@@ -34,14 +34,11 @@ type eventbus struct {
 }
 
 func init() {
-	mode := config.CONFIG.GetConfigurationByKey("mqtt.mode")
-	if mode == nil {
+	mode, err := config.CONFIG.GetValue("mqtt.mode").ToInt()
+	if err != nil || mode > externalMqttMode || mode < internalMqttMode {
 		mode = internalMqttMode
 	}
-	if mode.(int) > externalMqttMode || mode.(int) < internalMqttMode {
-		panic("mqtt.mode should be one of [0,1,2]")
-	}
-	edgeEventHubModule := eventbus{mqttMode: mode.(int)}
+	edgeEventHubModule := eventbus{mqttMode: mode}
 	core.Register(&edgeEventHubModule)
 }
 


tedli:
```golang
case nil:
	return 0, nil
```

`ToInt` returns `0`, if value of a config entry is `nil`, while `internalMqttMode` constant is also `0`.

edge/pkg/eventbus/event_bus.go
```diff
@@ -3,7 +3,6 @@ package eventbus
 import (
 	"encoding/json"
 	"fmt"
-
```
sids-b: nit: Please don't remove this blank line. It is used to seperate golang and vendor packages

tedli: OK

sids-b:
> ```go
> case nil:
> 	return 0, nil
> ```
> 
> `ToInt` returns `0`, if value of a config entry is `nil`, while `internalMqttMode` constant is also `0`.

You would hit the error case if there is no entry in the config file right ? Also we want default mode to be internalmqtt mode.

edge/pkg/eventbus/event_bus.go
```diff
@@ -34,14 +34,13 @@ type eventbus struct {
...
 
 func init() {
-	mode := config.CONFIG.GetConfigurationByKey("mqtt.mode")
-	if mode == nil {
-		mode = internalMqttMode
-	}
-	if mode.(int) > externalMqttMode || mode.(int) < internalMqttMode {
+	mode, err := config.CONFIG.GetValue("mqtt.mode").ToInt()
+	if err != nil {
+		panic(err)
```
sids-b: Is there a need to panic here ? If mode is not specified in config file, we can safely start with internalMqttMode

tedli: OK

tedli:
> > ```go
> > case nil:
> > 	return 0, nil
> > ```
> > 
> > 
> > `ToInt` returns `0`, if value of a config entry is `nil`, while `internalMqttMode` constant is also `0`.
> 
> You would hit the error case if there is no entry in the config file right ? Also we want default mode to be internalmqtt mode.

Indeed. Would write condition cases by hand be a good idea? Somethine like this:

```golang
mode := config.CONFIG.GetConfigurationByKey("mqtt.mode")
var mqttMode int
var ok bool
if mode == nil {
	mode = internalMqttMode
	mqttMode = internalMqttMode
} else {
	if mqttMode, ok = mode.(int); !ok {
		var err error
		if modeStr, ok := mode.(string); !ok {
			panic("mqtt.mode should be one of [0,1,2]")
		} else if mqttMode, err = strconv.Atoi(modeStr); err != nil {
			panic(err)
		}
	}
}
```

sids-b:
> > > ```go
> > > case nil:
> > > 	return 0, nil
> > > ```
> > > 
> > > 
> > > `ToInt` returns `0`, if value of a config entry is `nil`, while `internalMqttMode` constant is also `0`.
> > 
> > 
> > You would hit the error case if there is no entry in the config file right ? Also we want default mode to be internalmqtt mode.
> 
> Indeed. Would write condition cases by hand be a good idea? Somethine like this:
> 
> ```go
> mode := config.CONFIG.GetConfigurationByKey("mqtt.mode")
> var mqttMode int
> var ok bool
> if mode == nil {
> 	mode = internalMqttMode
> 	mqttMode = internalMqttMode
> } else {
> 	if mqttMode, ok = mode.(int); !ok {
> 		var err error
> 		if modeStr, ok := mode.(string); !ok {
> 			panic("mqtt.mode should be one of [0,1,2]")
> 		} else if mqttMode, err = strconv.Atoi(modeStr); err != nil {
> 			panic(err)
> 		}
> 	}
> }
> ```
How about this ? I guess this solves the problem ?
```
mode, err := config.CONFIG.GetValue("mqtt.mode").ToInt()
	if err != nil || (mode > externalMqttMode || mode < internalMqttMode) {
		mode = 0
         }
```

tedli:
> > > > ```go
> > > > case nil:
> > > > 	return 0, nil
> > > > ```
> > > > 
> > > > 
> > > > `ToInt` returns `0`, if value of a config entry is `nil`, while `internalMqttMode` constant is also `0`.
> > > 
> > > 
> > > You would hit the error case if there is no entry in the config file right ? Also we want default mode to be internalmqtt mode.
> > 
> > 
> > Indeed. Would write condition cases by hand be a good idea? Somethine like this:
> > ```go
> > mode := config.CONFIG.GetConfigurationByKey("mqtt.mode")
> > var mqttMode int
> > var ok bool
> > if mode == nil {
> > 	mode = internalMqttMode
> > 	mqttMode = internalMqttMode
> > } else {
> > 	if mqttMode, ok = mode.(int); !ok {
> > 		var err error
> > 		if modeStr, ok := mode.(string); !ok {
> > 			panic("mqtt.mode should be one of [0,1,2]")
> > 		} else if mqttMode, err = strconv.Atoi(modeStr); err != nil {
> > 			panic(err)
> > 		}
> > 	}
> > }
> > ```
> 
> How about this ? I guess this solves the problem ?
> 
> ```
> mode, err := config.CONFIG.GetValue("mqtt.mode").ToInt()
> 	if err != nil || (mode > externalMqttMode || mode < internalMqttMode) {
> 		mode = 0
>          }
> ```

I've corrected as you told.

m1093782566:

edge/pkg/eventbus/event_bus.go
```diff
@@ -34,14 +34,11 @@ type eventbus struct {
 }
 
 func init() {
-	mode := config.CONFIG.GetConfigurationByKey("mqtt.mode")
-	if mode == nil {
+	mode, err := config.CONFIG.GetValue("mqtt.mode").ToInt()
+	if err != nil || (mode > externalMqttMode || mode < internalMqttMode) {
```
m1093782566: what about 

```
if err != nil || mode > externalMqttMode || mode < internalMqttMode {
  mode = internalMqttMode
}
```

tedli: Done.

PR #246, opened 2019-03-18
Dockerfiles and scripts to containerize edge controller

### LINKED ISSUE #251: Containerize edge controller to further orchestrated with k8s

Dockerfile and scripts to run edge controller as a container.

**Why is this needed**:

Simplify the deployment of edge controller.

tedli:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

Containerize edge controller so that we could deploy it into k8s cluster in next step.

Fixes #251 

Part of #236


m1093782566:

sids-b:
cc @rohitsardesai83 @Baoqiang-Zhang for review

rohitsardesai83:
@tedli , running edge_core in a container is a good idea. But since edge_core contains edged which is a simplified kubelet version and it manages container lifecycle, we need to analyze the impact whether edged can mount volumes inside other container namespaces. I saw few issues logged previously : 

- https://github.com/kubernetes/kubernetes/issues/6848
- https://github.com/kubernetes/kubernetes/pull/6936

 in Kubernetes, need to ensure whether edged also supports it.

tedli:
> @tedli , running edge_core in a container is a good idea. But since edge_core contains edged which is a simplified kubelet version and it manages container lifecycle, we need to analyze the impact whether edged can mount volumes inside other container namespaces. I saw few issues logged previously :
> 
> * [kubernetes/kubernetes#6848](https://github.com/kubernetes/kubernetes/issues/6848)
> * [kubernetes/kubernetes#6936](https://github.com/kubernetes/kubernetes/pull/6936)
> 
> in Kubernetes, need to ensure whether edged also supports it.

Indeed, now I found edged uses `/var/lib/edged` to prepare configmaps and secrets for pod to be mounted. In my dev environment, run edge_core container with `--privilege` and `-v /var/lib/edged:/var/lib/edged` works well with basic cases.

But if I kubectl get node, the internal ip of edge node is docker default network's IP, if we care about this we can append `--net=host`. Since edge_core only support `hostPort`, the docker engine api handles port forwarding for the pod container well.

Further, new features may added, we can bypass the 'container' layer, just use docker to manage the edge_core's lifecycle, with `--pid=host` and inside the container use `nsenter`, somethine like this:

```bash
nsenter --target=1 --mount --wd=/ -- edge_core
```

Makefile
```diff
@@ -25,3 +25,12 @@ edge_cross_build:
 edgecontroller:
 	cd cloud/edgecontroller && $(MAKE)
 
+IMAGE_TAG ?= $(shell git describe --tags)
+
+.PHONY: cloudimage
+cloudimage:
+	docker build -t kubeedge/edgecontroller:${IMAGE_TAG} -f build/cloud/Dockerfile .
+
+.PHONY: certgenimage
```
m1093782566: Do we really need the `make certgenimage` command? When do we need this command?

tedli: The `certgen` is an image to generate certs. Because the script inside the image depends on `pr` command which may behaves weird in different linux dirstributions. To avoid this image comes.

It just an option not an necessary. I can remove this make rule.

build/tools/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.9
```
rohitsardesai83: maybe we can avoid creating a Docker image for generating certs ? This could be done manually or as a step of the cloud Dockerfile ?

rohitsardesai83:
@tedli , please resolve conflicts.

Issue #208, opened 2019-03-07
Reduce edge-core binary size


m1093782566:
**What would you like to be added**:

Reduce edge-core binary size

**Why is this needed**:

edge-core's current binary file size is about 70MB, which is bigger than k3s. As less resource consuming is one of the declared advantages of kubeedge, we need to optimize the binary file size.

sids-b:
cc @edisonxiang

tedli:
go build option `-ldflags="-w -s -extldflags -static"` along with `CGO_ENABLED=1` (for sqlite), produce an 48M binary.

If we `upx` the built binary with flags `--best`, it gives us a 14M binary.

For now on the upxed one works well with basic cases.

But if we append ` --ultra-brute` to upx, the produced 11M binary won't run.

m1093782566:
> If we upx the built binary with flags --best, it gives us a 14M binary.

Thanks for your advice, we will investigate it.

sids-b:
Closed via #262

PR #247, opened 2019-03-18
Dockerfiles and scripts to containerize edge core

### LINKED ISSUE #250: Containerize edge core so docker can manage the lifecycle and simplify the deployment

Dockerfile and scripts to run edge core as a container.

**Why is this needed**:

Simplify edge side deployment.

tedli:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

Containerize edge core so we can handle both edge core's logs and lifecycle with docker engine.

Fixes #250 

Part of #236 
Depends on #246


m1093782566:

sids-b:
cc @Baoqiang-Zhang @rohitsardesai83 for review

rohitsardesai83:
@tedli , running edge_core in a container is a good idea. But since edge_core contains edged which is a simplified kubelet version and it manages container lifecycle, we need to analyze the impact whether edged can mount volumes inside other container namespaces. I saw few issues logged previously : 

- https://github.com/kubernetes/kubernetes/issues/6848
- https://github.com/kubernetes/kubernetes/pull/6936

We need to ensure whether edged also supports it. Have we tested deploying pods consuming volumes after running edge_core in a  container ?

tedli:
> @tedli , running edge_core in a container is a good idea. But since edge_core contains edged which is a simplified kubelet version and it manages container lifecycle, we need to analyze the impact whether edged can mount volumes inside other container namespaces. I saw few issues logged previously :
> 
> * [kubernetes/kubernetes#6848](https://github.com/kubernetes/kubernetes/issues/6848)
> * [kubernetes/kubernetes#6936](https://github.com/kubernetes/kubernetes/pull/6936)
> 
> We need to ensure whether edged also supports it. Have we tested deploying pods consuming volumes after running edge_core in a container ?

Indeed, now I found edged uses `/var/lib/edged` to prepare configmaps and secrets for pod to be mounted. In my dev environment, run edge_core container with `--privilege` and `-v /var/lib/edged:/var/lib/edged` works well with basic cases.

But if I kubectl get node, the internal ip of edge node is docker default network's IP, if we care about this we can append `--net=host`. Since edge_core only support `hostPort`, the docker engine api handles port forwarding for the pod container well.

Further, new features may added, we can bypass the 'container' layer, just use docker to manage the edge_core's lifecycle, with `--pid=host` and inside the container use `nsenter`, somethine like this:

```bash
nsenter --target=1 --mount --wd=/ -- edge_core
```

nkwangleiGIT:
@rohitsardesai83 yes, I understand your concern. Actually we already gained lots of experience to run kubelet inside a container when we built PaaS based on Kubernetes, so I think we can fix the potential issues if find later. 
And we just want to provide one deployment option for the user, they can still run edged on the host directly if they prefer.  But the container approach can let the user try kubeedge more easily if they have docker installed, just 'docker run' can bring edged up. So we can simplify the 'get start' document using these container approach later.

m1093782566:
I agree we can add an option for running edged inside a container.

rohitsardesai83:
@tedli , @nkwangleiGIT , thanks for your comments and work !  I think its good to have another option as an end user.  Glad to know the basic tests with configmap and secrets work fine.  I was thinking maybe in the future if we wish to replace docker with some other container runtime like containerd , what would be the impact.

build/edge/run_daemon.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
rohitsardesai83: We need some documentation in the readme as to how to run edge_core in a container.  Can we have the documentation as a part of this PR ?

build/edge/run_daemon.sh
```diff
@@ -0,0 +1,39 @@
...
+    local cloudHub=$2
+    docker run -d --name edgecore --restart always \
+    --cpu-period=50000 --cpu-quota=100000 --memory=1g --privileged \
+    -e mqtt.server=${mqttBroker} \
+    -e edgehub.websocket.url=${cloudHub} \
+    -v /etc/kubeedge/edge/certs:/etc/kubeedge/edge/certs:ro \
+    -v /var/lib/edged:/var/lib/edged \
+    -v /var/lib/kubeedge:/var/lib/kubeedge \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    kubeedge/edgecore:v0.2
```
rohitsardesai83: the Makefile uses the ${IMAGE_TAG}  to tag the edgecore docker image.
```sh
docker build -t kubeedge/edgecore:${IMAGE_TAG} -f build/edge/Dockerfile .
```
This script uses a hard-coded version v0.2 ? can we pass the image name as a parameter ?

build/edge/run_daemon.sh
tedli: The documentation is along with the #248 . If it makes more sense to commit docs along with this PR, I'm glad to do a correction.

build/edge/run_daemon.sh
tedli: OK

build/edge/run_daemon.sh
rohitsardesai83: oh sorry I missed it. Its ok to have it as part of #248 . thanks !

build/edge/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM golang:1.12.1-alpine3.9 AS builder
+
+COPY . /go/src/github.com/kubeedge/kubeedge
+
+RUN apk --no-cache update && \
+apk --no-cache upgrade && \
+apk --no-cache add build-base linux-headers sqlite-dev && \
+CGO_ENABLED=1 go build -v -o /usr/local/bin/edge_core -ldflags="-w -s -extldflags -static" \
```
rohitsardesai83: there is an open PR to compile a miniature edge_core binary using upx. Just for your reference : https://github.com/kubeedge/kubeedge/pull/262

tedli: Yes, we'v shared the idea at #208 .

tedli:
議事録

PR：edge_core をコンテナ化すること

rohitsardesai83：
a. いま複数コンテナソリューションがあるゆえ、このPRは、docker のみ利用可能ので、将来に他のコンテナソリューションを利用しようとするなら、工夫がどのぐらいかかるのか、検討必要だ。
b. edge_core コンテナを起動するスクリプトに、イメージタグがハードコードしているので、パラミタで入力するのように修正すること。
c. edge_core をコンテナで稼働する手順を提供すること。（tedli：#248 と供に提出する）
d. edge_core バイナリを upx するのを検討すること。

build/edge/Dockerfile
m1093782566: upx is just an option, it's not default. I don't like upx very much as well.

build/edge/run_daemon.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+if [ ! -d /etc/kubeedge/edge/certs ] || [ ! -e /etc/kubeedge/edge/certs/edge.crt ] || [ ! -e /etc/kubeedge/edge/certs/edge.key ]; then
+    mkdir -p /etc/kubeedge/edge/certs
+    mkdir -p /etc/kubeedge/ca
+    docker run --rm \
+    -v /etc/kubeedge/ca:/etc/kubeedge/ca \
+    -v /etc/kubeedge/edge/certs:/etc/kubeedge/certs \
+    kubeedge/certgen:v0.2 genCertAndKey edge
```
rohitsardesai83: as we had discussed in other PR , we may not need a special docker image for certificate generation etc. Please see if we can incorporate this via some common script utilities .

tedli: OK. I correct here to use a script not the image.

rohitsardesai83:
@tedli , please resolve conflicts.

sids-b:
@tedli , Please resolve the conflict so that we can go ahead with merging the PR

PR #248, opened 2019-03-18
manifests to deploy edge controller into k8s

### LINKED ISSUE #252: K8s manifests and docs to deploy control panel (edge controller) into cluster

k8s manifests and related docs to guide deploying control panel (edge controller) into cluster.

**Why is this needed**:

k8s enhances high level business logics, while managing cloud part in a more elastic way.

tedli:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

Deploy edge controller into k8s cluster. K8s enhances high level business logic of deploying and managing the edge controller.

Fixes #252 

Part of #236


m1093782566:

sids-b:
cc @Baoqiang-Zhang @rohitsardesai83 for review

README.md
```diff
@@ -250,6 +250,38 @@ Please click [Cross Compilation](docs/setup/cross-compilation.md) for the instru
 
 ### Run Cloud
 
+#### Method A
```
rohitsardesai83: we can rename to something like : Run as Kubernetes deployment

README.md
```diff
@@ -250,6 +250,38 @@ Please click [Cross Compilation](docs/setup/cross-compilation.md) for the instru
...
+Second, we create k8s resources from the manifests in name order. Before
+creating, check the content of each manifest to make sure it meets your
+environment.
+
+```bash
+for resource in $(ls *.yaml); do kubectl create -f $resource; done
+```
+
+
+#### Method B
```
rohitsardesai83: we can rename to maybe "Run as a binary"

README.md
```diff
@@ -268,6 +300,28 @@ kubectl apply -f $GOPATH/src/github.com/kubeedge/kubeedge/build/node.json
 
 Run Edge
 
+#### Method A
+
+This method will guide you to deploy the edge part running in docker
+container, so make sure that docker engine listening on
+`/var/run/docker.sock` which will then mount into the edge container.
+
+Before start the edge part container, check the content of this script
```
rohitsardesai83: nit: start -> starting

tedli: Thanks. I will correct the points mentioned above.

README.md
```diff
@@ -250,6 +250,38 @@ Please click [Cross Compilation](docs/setup/cross-compilation.md) for the instru
...
+
+This method will guide you to deploy the cloud part into a k8s cluster,
+so you need to login to the k8s master node (or where else if you can
+operate the cluster with `kubectl`).
+
+The manifests and scripts in `github.com/kubeedge/kubeedge/build/cloud`
+will be used, so place these files to somewhere you can kubectl with.
+
+Then, first, we need to generate the tls certs (and CA if we don't have
+one) for CloudHub to host websocket server. It then will give us
```
rohitsardesai83: for CloudHub to host websocket server.   --> what does this mean ? can you clarify a little more ?

README.md
```diff
@@ -250,6 +250,38 @@ Please click [Cross Compilation](docs/setup/cross-compilation.md) for the instru
...
+`06-secret.yaml` if succeeded.
+
+```bash
+# this script uses a docker image to generate secrets
+./00-generate_secret.sh
+
+# or use tools/certgen.sh directly if openssl and coreutils are installed
+tools/certgen.sh buildSecret | tee ./06-secret.yaml
```
rohitsardesai83: tools -> build/tools ?

README.md
```diff
@@ -250,6 +250,38 @@ Please click [Cross Compilation](docs/setup/cross-compilation.md) for the instru
...
+```
+
+Second, we create k8s resources from the manifests in name order. Before
```
rohitsardesai83: i think we  need to specify the step to build the docker image for the cloud before creating the below resources ?

nkwangleiGIT: Actually, if we have a public registry that the user can pull docker image directly, I don't think the user will care about how to build it :-)
and we can just add the dockerfile here, so the user can build it if they'd like to.

m1093782566: What we add to README should be complete and accurate.

nkwangleiGIT: ok, that's fine, we can add a 'docker build' step before create k8s resources using cloud image.

build/cloud/08-service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
rohitsardesai83: Should we delay this for now since support for K8S service isn't there in edge_core yet?

README.md
tedli: Internal cloud hub module, there is a websocket server object, to start the server for accepting edge (client connection), a config object which contains tls certs is need.
It means cloud hub module need tls certs. I will ask someone good at English to help me clarify these words.

README.md
tedli: OK, I will correct this.

build/cloud/08-service.yaml
tedli: This service definition is used to expose edge controller ( port 10000 [cloud hub]), user should modify the service type to `LoadBalancer` or use `ExternalIPs` or `NodePort` to let edge core connect to.

https://kubernetes.io/docs/concepts/services-networking/#loadbalancer
https://kubernetes.io/docs/tutorials/stateless-application/expose-external-ip-address/

tedli: I think I finally got your point. Your mean in a k8s cluster, there is no normal node at all. Only edge node. So the `service` won't work.

In this scenario, It's not a good idea to deploy edge controller as deployment.

rohitsardesai83: for now since we don't support service , we can avoid the service definition for the cloud module. maybe in future we can add this if we add the support in edge_core.

README.md
rohitsardesai83: :+1:

rohitsardesai83:
@tedli , please resolve the conflicts . Thanks !

PR #264, opened 2019-03-20
remove tmp code for ut

### LINKED ISSUE #135: Code Fixing need to done for isModuleEnabled() in db.go
The temporary change made in file db.go(pkg/common/dbm/db.go) needs to be fixed. The function isModuleEnabled() returns true in the beginning of the function.
dbgoerror

edisonxiang:
What type of PR is this?
/kind bug

* remove the hard code for unit test
* clean the temp file in unit test

Which issue(s) this PR fixes:
remove the hard code for unit test

Fixes #135

### DIFF
--- a/edge/pkg/common/dbm/db.go
+++ b/edge/pkg/common/dbm/db.go
@@ -1,6 +1,7 @@
 package dbm
 
 import (
+	"os"
 	"strings"
 
 	"github.com/kubeedge/kubeedge/common/beehive/pkg/common/config"
@@ -72,9 +73,25 @@ func InitDBManager() {
 	DBAccess.Using(dbName)
 }
 
+// Cleanup cleans up resources
+func Cleanup() {
+	cleanDBFile(dataSource)
+}
+
+// cleanDBFile removes db file
+func cleanDBFile(fileName string) {
+	// Remove db file
+	err := os.Remove(fileName)
+	if err != nil {
+		if os.IsNotExist(err) {
+			log.LOGGER.Infof("DB file %s is not existing", fileName)
+		} else {
+			log.LOGGER.Errorf("Failed to remove DB file %s: %v", fileName, err)
+		}
+	}
+}
+
 func isModuleEnabled(m string) bool {
-	// TODO: temp change for ut
-	return true
 	modules := config.CONFIG.GetConfigurationByKey("modules.enabled")
 	if modules != nil {
 		for _, value := range modules.([]interface{}) {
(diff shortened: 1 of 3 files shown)


edge/pkg/common/dbm/db.go
```diff
@@ -72,9 +73,22 @@ func InitDBManager() {
 	DBAccess.Using(dbName)
 }
 
+// Clean db file if it is neccessary
+func Clean() {
```
m1093782566: CleanDBFile(file string) ??

edisonxiang: the file name is defined in db.go, what about `CleanDBFile()`?

edge/pkg/common/dbm/db.go
```diff
@@ -72,9 +73,22 @@ func InitDBManager() {
...
+	// Check if db file is existing
+	_, err := os.Stat(defaultDataSource)
```
m1093782566: Can't we directly do `os.Remove()` and then handle the `*PathError`?

edge/pkg/common/dbm/db.go
edisonxiang: Done, Thanks

edge/pkg/common/dbm/db.go
edisonxiang: Done, Thanks

edge/pkg/common/dbm/db.go
```diff
@@ -72,9 +73,20 @@ func InitDBManager() {
 	DBAccess.Using(dbName)
 }
 
+// CleanDBFile removes db file if it is neccessary
+func CleanDBFile() {
```
m1093782566: This function is only used to remove `defaultDataSource` file?

Why not pass a file argument in?

edge/pkg/common/dbm/db.go
```diff
@@ -72,9 +73,20 @@ func InitDBManager() {
 	DBAccess.Using(dbName)
 }
 
+// CleanDBFile removes db file if it is neccessary
```
sids-b: nit: neccessary -> necessary 

This is causing CI failure.

edge/pkg/common/dbm/db.go
edisonxiang: I will Thanks

edge/pkg/common/dbm/db.go
edisonxiang: I will Thanks

PR #270, opened 2019-03-20
Update docs of "What is KubeEdge"

junxu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Fixes #

### DIFF
--- a/docs/modules/kubeedge.md
+++ b/docs/modules/kubeedge.md
@@ -1,6 +1,6 @@
 # What is KubeEdge
 
-**KubeEdge** is an open source system extending native containerized application orchestration and device management to hosts at the Edge. It is built upon Kubernetes and provides core infrastructure support for networking, application deployment and metadata synchronization between cloud and edge. It also supports MQTT and allows developers to author custom logic and enable resource constrained device communication at the Edge. Kubeedge consists of a cloud part and an edge part. The edge part has already been open sourced and the cloud part is coming soon!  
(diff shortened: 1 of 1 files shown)


docs/modules/kubeedge.md
```diff
@@ -1,6 +1,6 @@
 # What is KubeEdge
 
-**KubeEdge** is an open source system extending native containerized application orchestration and device management to hosts at the Edge. It is built upon Kubernetes and provides core infrastructure support for networking, application deployment and metadata synchronization between cloud and edge. It also supports MQTT and allows developers to author custom logic and enable resource constrained device communication at the Edge. Kubeedge consists of a cloud part and an edge part. The edge part has already been open sourced and the cloud part is coming soon!  
+**KubeEdge** is an open source system extending native containerized application orchestration and device management to hosts at the Edge. It is built upon Kubernetes and provides core infrastructure support for networking, application deployment and metadata synchronization between cloud and edge. It also supports MQTT and allows developers to author custom logic and enable resource constrained device communication at the Edge. Kubeedge consists of a cloud part and an edge part. Both edge and cloud parts are now opensourced.
```
m1093782566: Thanks for contribution @junxu 

Would you please update the Chinese README as well? https://github.com/kubeedge/kubeedge/blob/master/README_zh.md

rohitsardesai83: Do we need to mention this now : " Both edge and cloud parts are now opensourced."
I feel its ok to remove this now. earlier since cloud part was not opensourced , it was ok to mention.

junxu: Thanks for you reply @m1093782566 !
The Chinese README is corresponding to https://github.com/kubeedge/kubeedge/blob/master/README.md, I think it's doesn't need to be updated, what's your opinion?

m1093782566: Okay, fair enough.

PR #277, opened 2019-03-21
WIP: notify edge to delete pod again if edge didn't handle pod delete event

### LINKED ISSUE #260: If edge core didn't handle pod delete event from cloud, pod remains with status Terminating
**What happened**:

Pod (running at edge) remain with status Terminating.

image

For temporary go over this problem, I have to delete meta entry of the pod in edge.db and delete the pod object in etcd manully.

Pod should be deleted.

Deploy pod to edge, after pod up and running, kill edge core, delete the pod, again start the edge core.

Then the pod we intend to delete get never deleted, ramains with status Terminating.

The downstream controller (response for trigger event to edge side) just try to notice edge once, and if failed did nothing but log an error.

https://github.com/kubeedge/kubeedge/blob/6878d86b9ec33ca716933575b93ef70663504ba5/cloud/edgecontroller/pkg/controller/controller/downstream.go#L51-L67

**Environment**:
- KubeEdge version: v0.2
- Others: irrelevant

tedli:
**What type of PR is this?**

/kind bug

If edge side didn't handle pod delete event (for network off or edge core down). The pod remains with Terminating.

Fixes #260 


sids-b:
@tedli , One more reason could be for podStatus to be in terminating state could be that edge_core was offline and it terminated the Pod but was not able to send the message to Cloud. So in this scenario, when you send a delete pod message to edge, does edged discard the message or send response message saying pod is deleted ?

Also, even though we solve this issue for updatePod status this way, I feel it will be difficult for us to solve similar issue in updating configmaps and secrets when edge is offline in similar fashion. So i think the idea of @Baoqiang-Zhang to have a cache/sync message mechanism in cloudHub for reliable message delivery makes sense.
Thoughts @m1093782566 @rohitsardesai83 @edisonxiang @CindyXing @qizha @Baoqiang-Zhang

tedli:
Terminating status is because `pod.metadata.deletionTimestamp` is not `nil`, it is a soft delete of k8s.

https://github.com/kubernetes/kubernetes/blob/b000cd46bd749e24fbee4fcc0593747b5e0b8818/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L178-L197

A terminating pod can also have a status Running. Edge don't know the pod should be deleted, so it start up the pod's container, and by syncing status to cloud, cloud update the status of the pod to Running.

If a pod's status is not healthy (Failed, Terminated...), it doesn't affect deletion timestamp, so kubectl gives `CrashLoopBackOff` not terminating.

image

So, a pod with deletion timestamp not nil should be deleted. The k8s kubelet did the same:

https://github.com/kubernetes/kubernetes/blob/b000cd46bd749e24fbee4fcc0593747b5e0b8818/pkg/kubelet/kubelet.go#L1836-L1868

https://github.com/kubernetes/kubernetes/blob/b000cd46bd749e24fbee4fcc0593747b5e0b8818/pkg/kubelet/kubelet.go#L1977-L1989

One difference is that kubelet pulls pods, edgecore pushs podStatuses.

cc @nkwangleiGIT

tedli:
Not only the Delete event, but also do the Add, Modify event, if edge missed the event, it will never know it happend. Whether to use a cache or not, this should take a deeper consideration.

sids-b:
> Not only the Delete event, but also do the Add, Modify event, if edge missed the event, it will never know it happend. Whether to use a cache or not, this should take a deeper consideration.

@tedli . This is the reason I think we need to come up with a generalized solution for reliable delivery of messages between cloud and edge than solving it just for pod delete scenario :)

tedli:
@CindyXing Hi, I made an update of the code of this PR, I didn't run nor test the code. I push the code just for review the idea.

Now the idea is like this, at edge side, every time it connected to cloud, it send an request to pull all resources that should be on this node. And then cloud response all the latest resources that should be on this node. At edge side, diff out changes from meta db, and send message to corresponding modules to handle all events edge had missed.

What's your take?

Baoqiang-Zhang:
@tedli I think your idea can solve most issues of message losing, but what we need is a reliable message delivery mechanism.  since pull message is an async message, trigger resource resync from cloud to edge by node connected event does not solve the issue of unreliable message delivery.

Baoqiang-Zhang:
> Terminating status is because `pod.metadata.deletionTimestamp` is not `nil`, it is a soft delete of k8s.
> 
> https://github.com/kubernetes/kubernetes/blob/b000cd46bd749e24fbee4fcc0593747b5e0b8818/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L178-L197
> 
> A terminating pod (deletionTimestamp != nil) can also have a status Running (pod.status.containerStatuses[?].state == running, phase == running). Edge don't know the pod should be deleted, so it start up the pod's container, and by syncing status to cloud, cloud update the status of the pod to Running.
> 
> If a pod's status is not healthy (Failed, Terminated...), it doesn't affect deletion timestamp, so kubectl gives `CrashLoopBackOff` not terminating.
> 
> image
> 
> So, a pod with deletion timestamp not nil should be deleted. The k8s kubelet did the same:
> 
> https://github.com/kubernetes/kubernetes/blob/b000cd46bd749e24fbee4fcc0593747b5e0b8818/pkg/kubelet/kubelet.go#L1836-L1868
> 
> https://github.com/kubernetes/kubernetes/blob/b000cd46bd749e24fbee4fcc0593747b5e0b8818/pkg/kubelet/kubelet.go#L1977-L1989
> 
> One difference is that kubelet pulls pods, edgecore pushs podStatuses.
> 
> cc @nkwangleiGIT

https://github.com/kubernetes/kubernetes/blob/7dfcacd1cfcbdfe74b28f2473fb107e9a47ec905/pkg/registry/core/pod/strategy.go#L135-L138

deletion timestamp is not only condition for pod deletion, kube-apiserver will check pod status phase to delete pod record in etcd.

CindyXing:
@tedli , @Baoqiang-Zhang , I think deletiontimestamp might be able to solve this issue. We should address the reliable message delivery issue. 

The benefit of pulling metadata from cloud when an edge node re-connect is to make sure syncing down customers' intention as a batch and edge taking actions correctly. However there can be quite some data pulling down. This can be an issue provided the network bandwidth & reliability. Another challenge is to diff efficiently on the edge side.  

Ideally we'd like the client be able to know & pull the exact diff from cloud in one shot at any time; and without worrying about message loss.

sids-b:
> @tedli I think your idea can solve most issues of message losing, but what we need is a reliable message delivery mechanism. since pull message is an async message, trigger resource resync from cloud to edge by node connected event does not solve the issue of unreliable message delivery.

+1 I agree with @Baoqiang-Zhang . Pulling all resources will have an impact on traffic. Every time there is a cloud disconnect/connect(which can be very frequent in edge scenarios) pulling all resources will increase the traffic alot.

nkwangleiGIT:
OK, so should we introduce a messaging middleware here? like rabbitmq etc...? so we can persist and consume the events in some reliable approach?

sids-b:
> OK, so should we introduce a messaging middleware here? like rabbitmq etc...? so we can persist and consume the events in some reliable approach?

@nkwangleiGIT , We would love to see a design. Reliable message delivery will be one of the very important features. If you guys have bandwidth, maybe you can own both design and implementation ?

PR #279, opened 2019-03-21
Service support cross cloud and edge

### LINKED ISSUE #172: Service support and data communication between nodes
A mechanism to support interaction between nodes. The interaction include both data transport and application calling.
There are two factors for the data transportation:
1. Define the endpoint which include data source and destination in a data transportation process
2. Define a router rule to connect the pre-defined source and destination points 
Once the router created, the data will be able to transport from the source to destination which may corss the nodes. In the data transportation, there may have a proxy in each nodes to execute the router rule.

As for the application calling, I would like to call it EdgeMesh which is similar with the Service Mesh idea. Compare to the Service Mesh, there are three factors need to be considered:
* Service discovery
We can leverage istio as the control panel, to define services in kubernetes and use istio to define the service discovery rule. The service is defined base on the deployment then we are able to know the entry point(address/port).
* Data communication
We can leverage our edgehub act like envoy to proxy the service communication. By doing that we need to add other protocol support in edgehub and find a way to let the applications know the service address. On cloud, user can use DNS or environment variable to visit another service address. But on edge, we need to consider whether we can use the same mechanism. 

**Why is this needed**:
Edge node is more and more powful to run a big distributed system. It is import to support such distribution system running cross edges

qizha:
**What type of PR is this?**
/kind design

The design proposal of the service support cross cloud and edge

Fixes #172


docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+The service discovery, commuunication should be considered in this feature.
+Developers can define services on top of the ddeployments deployed on the cloud or edges in a unified manner without caring about their location. The service name is the only thing they need to know to do the communication. For example, the following address can be used for a REST  request:
+`http://{service_name}/resources`
+
+## Motivation
+The cloud native and microservice architecture becoming more and more popular and the edge node is becoming more and more powerful. The user want to separate their application to multiple microservice and some of them need to be deployed in edge node to get close to the data generated by deivces. By simply the development, the user need to use the same michanism to do the service discovery and communication with what they are doing on cloud.
+
+## Constraints and Assumptions
+1. Since the edge environment can be some non-standard system to running on some specific devices so we should not rely on the kernel network capability. In that way, we need to use userspace proxy to implement the data proxy on edge.
```
CindyXing: I assume the userspace is for the local container network routing like what kubeproxy does. For that, I think we should make this pluggable, say a kubeedge agent startup flag. Userspace in general doesn't perform well.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+2. With the same reason, clusterIP is not supported on Edge.
```
CindyXing: Agree on headless service. Our assumption is that edge nodes are behind firewall that cloud can't access directly. So services to be deployed onto edge can't be the type of clusterip, nor nodeport. I don't think we would support load balancer either considering the scale of edge nodes; geo distribution; multi-tenancy, etc.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+3. With the same reason, no DNS services available for edge.
```
CindyXing: Can you help clarify why?

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+4. The identification and authentication should be handled by microservices.
```
CindyXing: Agree. Though the underlining data plane, we should provide security guard on the routing and access control.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+5. When service communication between edge to edge, they may work offline. With the purpose to deliver a "expected result" to the user system, the service definition should be push down to the edge right after the definition operation. That is to say, the istio model may not be suitable for edge, in istio model, the service definition is pull from K8S master when the service calling happens.
+6. Only support HTTP communication in the first step
+
+## Use cases
+<img src="../images/proposals/service-design.png">
+
+### Register the service located on cloud
+* CS0.1: User create service(type ClusterIP) in K8S API Server
+* CS0.2: EC get the service definition and create router rules with the edge servicebus as source and service clusterIP as target.
```
CindyXing: Can we elaborate the router rule in detail? Which component creates the IP for the service?

I'd prefer separating the routing part from EC. Each internal service should be functional cohesive.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+### Service discovery on cloud
+* Application access service with the same mechanism no matter it located on cloud or edge
+
+### Service discovery on edge
+* Need to setup http_proxy in the container(can be done automatically) with the address of ServiceBus. Then the application can vist the service provider via `http://{service_name}/resource`. ServiceBus get the service name part from the request as a reverse proxy, then pick up one POD to redirect the request.
+
+### Service communication from cloud to edge
+* ESD1: Application on cloud visit service with clusterIP and request goes to EC(refer to step CS2, CS3)
+* ESD2: Router find out the PODs in this service and pick one with some kind of load balance mechanism. Then send the request to the choosen edge with the router rule set in CS3.
+* ESD3: EdgeHub/ServiceBus get the request and redirect to the POD
```
CindyXing: In my mind, cloudhub/edgehub is the channel for control plane. For data plane, we build the mesh.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+|ServiceBus|Source|Resutst Forward, Service Discovery, Load Balance|
+|ServiceBus|Target|Request Forward|
+|Router|Source|Resutst Forward, Service Discovery, Load Balance|
+
+### EdgeHub
+* Listen to the edge communication
+* Package the message and send to the target
+
+### EdgeController
+* Listen the service definition and create routing rule
```
CindyXing: As mentioned above, I'd recommend separating the routing functionality from EC.

docs/proposals/service-design.md
qizha: well, Service Mesh proxy(like Envoy and CSE Mesher) works in userspace. From now on, I saw some cases to using them in the production environment and they all work well. So balance the complexity, environment requirement and the performance, I think it works for most of the case.

docs/proposals/service-design.md
qizha: I am afraid LB can not be missed. In most of our current production usage scenario, user deploy video analysis applications on edge. They usually deploy multiple instances in the same edge node since each one can only handle few tasks. So, I think in most of the case, the LB capability here works for the multi-instances located in the same edge node.

docs/proposals/service-design.md
qizha: right now, we do have certification based authentication for the cross edge communication. In application level, maybe we can leverage SPIFFE to work with our Mesh proxy(EdgeHub/ServiceBus it is) to do such works in future.

docs/proposals/service-design.md
qizha: cloudhub/edgehub act as the data gateway for each node so both control and data channel data will go with them. But in those components, we have different data wrapper for the control/data packages so we do distinguish them logically.

docs/proposals/service-design.md
CindyXing: Not sure if we are talking about the same thing or not... When K8s service is specified with loadbalancer type, it meant to have an external load balancer. Do you mean that? 

Otherwise, with a service backed by endpoints on the same edge node, client accessing the service, it'll balance among them. I think Istio will define the rule on how to load balance.

docs/proposals/service-design.md
qizha: I agree, router can run separatly as a individual process besides EC. A service can be created for the router when deploy the KubeEdge cloud part then.

docs/proposals/service-design.md
qizha: I agree

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+# Service support on edge
+
+## Abstract
+To support service capabilities on edge to support microservice communication cross cloud and edge.
+The service discovery, commuunication should be considered in this feature.
+Developers can define services on top of the ddeployments deployed on the cloud or edges in a unified manner without caring about their location. The service name is the only thing they need to know to do the communication. For example, the following address can be used for a REST  request:
+`http://{service_name}/resources`
+
+## Motivation
+The cloud native and microservice architecture becoming more and more popular and the edge node is becoming more and more powerful. The user want to separate their application to multiple microservice and some of them need to be deployed in edge node to get close to the data generated by deivces. By simply the development, the user need to use the same michanism to do the service discovery and communication with what they are doing on cloud.
```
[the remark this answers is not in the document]
rohitsardesai83: becoming -> is becoming.

docs/proposals/service-design.md
qizha: coz the clusterIP is needed for config the kubelet to connect with DNS service. However, there is no ClusterIP for edge nodes. In additional, DNS works with a pull model, when DNS name resolver needed, the DNS client will pull the address with this DNS name from the DNS server. However, the edge node may work offline, it will be unable to get the DNS name. So we can not rely on DNS to do service discovery.

qizha: @sids-b can you add the design for the router when you open that commponent?

docs/proposals/service-design.md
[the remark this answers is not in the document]
rohitsardesai83: The user want to separate their application to multiple microservice  --> Users want to decompose their application into multiple microservices.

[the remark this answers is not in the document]
rohitsardesai83: By simply the development  -> By simplifying the development ?

docs/proposals/service-design.md
rohitsardesai83: some non-standard system to running ~~on~~ some specific devices,  ~~so~~ we should not rely on the kernel network capability.

docs/proposals/service-design.md
sids-b: Sure @qizha

docs/proposals/service-design.md
CindyXing: What if customers don't apply Istio? We should allow services on K8s be able to route among the cluster. I think we need something like kubeproxy at the edge side to enable the service/endpoint routing.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+### Service discovery on edge
+* Need to setup http_proxy in the container(can be done automatically) with the address of ServiceBus. Then the application can vist the service provider via `http://{service_name}/resource`. ServiceBus get the service name part from the request as a reverse proxy, then pick up one POD to redirect the request.
+
+### Service communication from cloud to edge
+* ESD1: Application on cloud visit service with clusterIP and request goes to EC(refer to step CS2, CS3)
+* ESD2: Router find out the PODs in this service and pick one with some kind of load balance mechanism. Then send the request to the choosen edge with the router rule set in CS3.
+* ESD3: EdgeHub/ServiceBus get the request and redirect to the POD
+
+### Service communication between edges
+* ESR1: Application send the request to ServiceBus(refer to the "Service discovery on edge")
```
rohitsardesai83: what are the user scenarios for service communication between edges ? since for now it is REST based communication , like service A on edge node 1 will invoke some REST API of another service B running on another edge node. these APIs are supposed to perform some action on some devices running on another edge node or get device status from the other edge node ? 
Could we give some real time user scenarios with some example for all the three types of communication we are targeting ?

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Constraints and Assumptions
+1. Since the edge environment can be some non-standard system to running on some specific devices so we should not rely on the kernel network capability. In that way, we need to use userspace proxy to implement the data proxy on edge.
+2. With the same reason, clusterIP is not supported on Edge.
+3. With the same reason, no DNS services available for edge.
+4. The identification and authentication should be handled by microservices.
+5. When service communication between edge to edge, they may work offline. With the purpose to deliver a "expected result" to the user system, the service definition should be push down to the edge right after the definition operation. That is to say, the istio model may not be suitable for edge, in istio model, the service definition is pull from K8S master when the service calling happens.
```
rohitsardesai83: when the service calling happens -> when the service is invoked ?

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+### Register the service located on edge
+* CS1: User create service(headless service without ClusterIP) in K8S API Server
+* CS2, CS3: EC get the service definiton and update the service ClusterIP to itself. Then create router rule with the EC(router) address as the source and edge servicebus as target.
+* CS4: Send the service definition to edge
+
+### Service discovery on cloud
+* Application access service with the same mechanism no matter it located on cloud or edge
+
+### Service discovery on edge
+* Need to setup http_proxy in the container(can be done automatically) with the address of ServiceBus. Then the application can vist the service provider via `http://{service_name}/resource`. ServiceBus get the service name part from the request as a reverse proxy, then pick up one POD to redirect the request.
```
rohitsardesai83: can be done automatically -> this needs to added as an environment variable by edged when the container comes up ?
since servicebus will a beehive module which doesn't expose any endpoint , how will the address of the servicebus look like ?

docs/proposals/service-design.md
rohitsardesai83: do we mean to say edgehub will contain two separate data wrapper modules to handle control and data packets separately ? currently edgehub only has websocket based channel for control plane right ? we plan to put service bus as a separate module parallel to edgehub right ? I am not sure how will we distinguish the packets and where will be do this.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+### Add servicebus on Edge to handle request proxy
+ServiceBus is in the same process with EdgeHub as a module
+
+### Router
+Add two endpoints definition:
+
+|Endpoint|Type|Capability|
+|---|---|---|
+|ClusterIP|Target|Request Forward|
+|ServiceBus|Source|Resutst Forward, Service Discovery, Load Balance|
```
rohitsardesai83: Resutst -> Request

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+|ServiceBus|Target|Request Forward|
+|Router|Source|Resutst Forward, Service Discovery, Load Balance|
```
rohitsardesai83: Resutst -> Request

docs/proposals/service-design.md
CindyXing: For the REST communication among edges, we assume edge nodes are among the same subnet, am I right? If so, we need to list it.

docs/proposals/service-design.md
qizha: Yes, kubeedge here act like the kubeproxy. The only difference here is it is using the iptables to route the request to kubeproxy in a regular K8S cluster. However, in edge, I suggest to use http_proxy or environment/hosts file to redirect the communication.

docs/proposals/service-design.md
qizha: You are right, I am going to add this in the assumption.

docs/proposals/service-design.md
qizha: there will be a environment variable as well. The servicebus itself listensing the docker0 address or 127.0.0.1 to those containers using host network.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,85 @@
...
+## Use cases
+<img src="../images/proposals/service-design.png">
+
+### Register the service located on cloud
+* CS0.1: User create service(type ClusterIP) in K8S API Server
+* CS0.2: EC get the service definition and create router rules with the edge servicebus as source and service clusterIP as target.
+* CS0.3: Send the service definition to edge
+
+### Register the service located on edge
+* CS1: User create service(headless service without ClusterIP) in K8S API Server
```
m1093782566: Any reason we don't support ClusterIP on edge and I am wondering how headless service work without DNS servers.

docs/proposals/service-design.md
```diff
@@ -0,0 +1,85 @@
...
+* ESU1: Application send the request to ServiceBus(refer to the "Service discovery on edge")
+* ESU2: ServiceBus pick one POD in this service with some kind of load balance mechanism. The target is on cloud, EdgeHub send the request to cloud.
+* ESU3: Router send the request to the corresponding POD on cloud following the rule created in CS0.2
+
+## Proposal Changes
+### Add servicebus on Edge to handle request proxy
+ServiceBus is in the same process with EdgeHub as a module
+
+### Router
+Add two endpoints definition:
```
m1093782566: four endpoints definition?

docs/proposals/service-design.md
qizha: 1. I want to keep the edge part simple especially for the network. And it is a websocket connection between cloud and edge, considering some services are located on cloud and some on the edge. A overlay network will be required to build the overall clusterIP system. It will be too complicate. 
2. This way we can leverage the service mesh ideas to build some more flow control capabilities on top of it
For the headless service, we just need a service name with a set of selected POD. Application can visit another one with the service name just like what istio provided.

docs/proposals/service-design.md
qizha: well, two is enough since each one can be both source and target

docs/proposals/service-design.md
m1093782566: Okay, thanks for clarification.

docs/proposals/service-design.md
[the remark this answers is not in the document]
qizha: no need, clusterIP is enough

docs/proposals/service-design.md
```diff
@@ -0,0 +1,86 @@
...
+<img src="../images/proposals/service-design.png">
+
+### Register the service located on cloud
+* CS0.1: User create service(type ClusterIP) in K8S API Server
+* CS0.2: EC get the service definition and create router rules with the edge servicebus as source and service clusterIP as target.
+* CS0.3: Send the service definition to edge
+
+### Register the service located on edge
+* CS1: User create service(headless service without ClusterIP) in K8S API Server
+* CS2, CS3: EC get the service definiton and update the service ClusterIP to itself. Then create router rule with the EC(router) address as the source and edge servicebus as target.
```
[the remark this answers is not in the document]
qizha: that's why we need a LB in the servicebus to pick the right target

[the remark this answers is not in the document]
CindyXing: So the servicebus will handle service LB across nodes as well? Would like to see the details when it is ready.

docs/proposals/service-design.md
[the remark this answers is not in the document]
CindyXing: As we discussed, router at cloud side will handle the routing to the right node & then endpoint. We need to add those details in this design doc. Thanks.

PR #286, opened 2019-03-22
remove beehive from vendor

kexun:
beehive is a common lib, we can import beehive form github.com/kubeedge/beehive(in the same GOPATH), not from vendor.


m1093782566:
cc @lidiyag @sids-b

sids-b:
Hi @kexun , as discussed can we close this PR ??

PR #307, opened 2019-03-27
Change the doc title from service support to edgemesh

### LINKED ISSUE #172: Service support and data communication between nodes

qizha:
**What type of PR is this?**
/kind design

The document is actually talking about the service mesh support instead of the native K8S service support. To clarify the concept, rename the document and update the title

Fixes #172

### DIFF
--- a/docs/proposals/edgemesh-design.md
+++ b/docs/proposals/edgemesh-design.md
@@ -7,13 +7,13 @@ approvers:
   - "@kevin-wangzefeng"
   - "@CindyXing"
 creation-date: 2019-03-20
-last-updated: 2019-03-20
+last-updated: 2019-03-27
 ---
 
-# Service support on edge
+# Edge Mesh on edge
 
 ## Abstract
-To support service capabilities on edge to support microservice communication cross cloud and edge.
+To support service mesh capabilities on edge to support microservice communication cross cloud and edge.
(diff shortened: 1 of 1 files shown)


sids-b:
Squashing all commits and merging it.

PR #316, opened 2019-04-01
Design: EdgeSite

CindyXing:
**What type of PR is this?**
   /kind design

Customers want to deploy and run clusters at edge location. This is to allow one to deploy a lightweight and autonomous worker node k8s cluster at edge.

Fixes #

This is a feature idea that we do not need it in v0.3.


docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,77 @@
...
+2. Edge worker node autonomy in case of network disconnection/reconnection
+3. All benefits of edge computing including latency, data locality, etc.
+4. Support large scale of edge clusters without consuming too much resources for control plane
+
+## Protocol 
+K8s client library interface will be used. The edgecontroller on each edge node only watches against k8s types for the node itself. 
+
+The informer programming model will be used between EdgeController and APIServer. 
+For example:
+
```
rohitsardesai83: please embed the code in this format : 
```go
<code>
```

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,77 @@
...
+4. The same K8s HA mechanism is to be used to enable HA
+
+## Architecture Design
+<img src="../images/EdgeSite_arch.PNG"/>
+
+## Advantages
+With the integration, the following can be enabled
+
+1. Light weight control plane and agent
+2. Edge worker node autonomy in case of network disconnection/reconnection
```
rohitsardesai50: since the entire edge cluster is at customer site, what are the scenarios where network disconnects / reconnects ? I assume the network is local within the edge site.

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,77 @@
...
+Here we assume a cluster is deployed at edge location including the management control plane. 
+For the management control plane to manage some scale of edge worker nodes, the hosting master node needs to have sufficient resources. 
+The assumptions are
+1. EdgeSite cluster master node is of >2 CPUs and >1GB memories
+2. If high availability is required, 2-3 master nodes are needed at different edge locations
+3. The same Kubernetes security (authN and authZ) mechanisms are used to ensure the secure handshake between master and slave
+4. The same K8s HA mechanism is to be used to enable HA
+
+## Architecture Design
+<img src="../images/EdgeSite_arch.PNG"/>
```
rohitsardesai50: 1. In the architecture, the other components of kubeedge on edge side are missing like devicetwin ,  metamanager , eventbus etc. Do we have a plan to integrate them as well ? 
2. Do we plan to run edgecontroller and kube-proxy as beehive modules on edge side ?
3. With kube-proxy on the edge node , will our edge service mesh design get affected ? Will we disable it ?
4. In the architecture, we intend to reuse the tunnel proxy from k3s ? What about the k3s agent -- does edged replace it ?
5. In the diagram , edged doesn't have a connection to the tunnel proxy ?

rohitsardesai83: 1. In the architecture, the other components of kubeedge on edge side are missing like devicetwin , metamanager , eventbus etc. Do we have a plan to integrate them as well ? 
2. Do we plan to run edgecontroller and kube-proxy as beehive modules on edge side ? 
3. With kube-proxy on the edge node , will our edge service mesh design get affected ? Will we disable it ? 
4. In the architecture, we intend to reuse the tunnel proxy from k3s ? What about the k3s agent -- does edged replace it ? 
5. In the diagram , edged doesn't have a connection to the tunnel proxy ?

docs/proposals/EdgeSite.md
rohitsardesai83: since the entire edge cluster is at customer site, what are the scenarios where network disconnects/reconnects? I assume the network is local within the edge site.

docs/proposals/EdgeSite.md
m1093782566: A silly question, can we simply deploy cloud part of kubeedge + K8s master in the edge side without modifying the current architecture?

CindyXing: Good questions. Thanks folks. 

To @m1093782566 , we could do that but those components are unnecessary. Really we just need to store the metadata targeted for the current node, right? 

To @rohitsardesai83 , for 1. good point. Devicetwin, eventbus don't need to change.  metamanager will be combined into edgecontroller and send data directly to tunnelproxy then apiserver. For 2, given edgecontroller and kube-proxy are must have, I am not seeing the value to put them as beehive modules. For 3, it depends on how we design the service mesh. If edgemesh is to use envoy like proxy, then kubeproxy mechanism will be bypassed. However, we can't force customers to use envoy/istio, kubeproxy is still needed. This is the same for data center k8s cluster. For 4, the reason to re-use the tunnel proxy is for scenarios when edge nodes behind firewall; and here it is to take advantages of lightweight edged. 5. Edged doesn't send data back to apiserver directly so no connection.

docs/proposals/EdgeSite.md
CindyXing: Agree that the network here is more reliable than KubeEdge scenarios. However, there can be network glitch that agent doesn't need to keep on reporting heartbeat like kubelet does. Otherwise we wouldn't need the local store at edge side.

sids-b:
hi @CindyXing , looks like there is some issue with PR commits, can you please squash them ?

yongwww:
Actually, I do think proposal/design could be posted as a RFC issue, or open a separate repo like kubeedge/RFCs. Since design/proposal needs more discussion. No need to keep them is code base

yongwww:
opened an issue #347

CindyXing:
Hi Jun and Rohit, can you please help reviewing the PR? I sent an email with e2e working as well. Thanks.

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+  - "@cindyxing"
+approvers:
+  - "@qizha"
+  - "@kevin-wangzefeng"
+  - "@m1093782566"
+---
+# EdgeSite: Lightweight Cluster at edge
+
+## Abstract
+In Edge computing, there are scenarios where customers would like to have an whole cluster installed at edge location. As a result,
```
m1093782566: nit: s/an whole cluster/a whole cluster

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+In Edge computing, there are scenarios where customers would like to have an whole cluster installed at edge location. As a result, 
+admins/users can leverage the local control plane to implement management functionalities and take advantages of all edge computing's benefits. 
+
+This design doc is to enable customers deploy and run lightweight clusters at edge. 
+
+## Motivation
+For these use cases, when considering large set of edge locations, the resource usage of management plane adding together will be high. 
+K3s is an opensource project enabling lightweight kubernetes cluster including the light control plane and worker components. 
+KubeEdge is the other opensource project enabling lightweight k8s client on the edge nodes. The KubeEdge agent runtime memory footprint is 
+10MB and it supports kubernetes types. Meanwhile KubEdge is targeted for Edge/IOT computing with many functionality enablements.
```
m1093782566: s/kubernetes types/Kubernetes primitives ?

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+## Advantages
+With the integration, the following can be enabled
+
+1. Light weight control plane and agent
+2. Edge worker node autonomy in case of network disconnection/reconnection
+3. All benefits of edge computing including latency, data locality, etc.
+4. Support large scale of edge clusters without consuming too much resources for control plane
+
+## Protocol 
+K8s client library interface will be used. The edgecontroller on each edge node only watches against k8s types for the node itself.
```
m1093782566: s/k8s types/K8s resource ?

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+    },
+})
+```
+
+And the data can be written to the client side store. 
+
+## Work Items
+1. Port current EdgeController code to KubeEdge agent side
+2. Remove cloudhub/edgehub 
+3. Come up with lightweight etcd
```
m1093782566: AFAIK, k3s server replace etcd with sqlite, if we are leverage the whole k3s server binary, do we still need to create a lightweight etcd?

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+
+This design doc is to enable customers deploy and run lightweight clusters at edge. 
+
+## Motivation
+For these use cases, when considering large set of edge locations, the resource usage of management plane adding together will be high. 
+K3s is an opensource project enabling lightweight kubernetes cluster including the light control plane and worker components. 
+KubeEdge is the other opensource project enabling lightweight k8s client on the edge nodes. The KubeEdge agent runtime memory footprint is 
+10MB and it supports kubernetes types. Meanwhile KubEdge is targeted for Edge/IOT computing with many functionality enablements. 
+
+By integrating KubeEdge and K3s, this proposal is to allow customers to run an efficient kubernetes cluster for Edge/IOT computing.
```
m1093782566: I don't think edgesite should decouple with K3s, K3s can be one of the control plane backend and K8s can be another one as well. So we can create a more open framework for integrating K8s/ligthweight K8s and K3s?

PS. In our test, K3S Server still consume 300-400 MB memory.

kevin-wangzefeng: @m1093782566 I think you mean the edgesite solution should not be coupled with K3s.

docs/proposals/EdgeSite.md
CindyXing: Good catch. Thanks.

docs/proposals/EdgeSite.md
CindyXing: Sounds good.

docs/proposals/EdgeSite.md
CindyXing: If K3s is to be used as the server, then no need to create the lightweight etcd. 

However, sqlite doesn't support HA. For EdgeSite scenarios when HA is needed, we would still need a lightweight etcd.

docs/proposals/EdgeSite.md
CindyXing: Edgesite is not coupled with K3S. I can update the wording a bit. 
The new edgesite client just watches a API server. The server can be k3s or a normal k8s.

docs/proposals/EdgeSite.md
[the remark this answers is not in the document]
CindyXing: > Sorry for joining the review late.
> The edge-site cases is quite important scenario in edge computing, big +1 to the overall idea.
> 
> For the design detail, I have 2 major questions here:
> 
> 1. It's not clear in the architecture that what components are from k3s, what can be conformant Kubernetes components. One of the overall external dependencies so far of KubeEdge is just a conformant Kubernetes, it's better we keep this and not tied to a specific implementation when dealing with edge-site cases.
> 2. As for the location of edgecontroller and removal of cloudhub / edgehub, why not keep edgecontroller in the control plane at edge and keep cloudhub / edgehub to deal with behind-firewall discovery / communication issue? I would like to see pros and cons listed in comparison between these two solutions.

Yes, in principal Edgesite doesn't couple with K3s. In fact, the Edgesite does conform to the K8s API that it can work with a normal K8s control plane. Let me try to update the architecture a bit and make it clearer. Thanks for pointing this out. 

For the 2nd comment, not sure what you meant. The controller is at edge side. For edgesite scenario, it is to deploy a whole cluster at edge. Why would there be a need of  firewall? Maybe the code PR can explain. I'll update the architecture diagram to make it more obvious.

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+Here we assume a cluster is deployed at edge location including the management control plane. 
+For the management control plane to manage some scale of edge worker nodes, the hosting master node needs to have sufficient resources. 
+The assumptions are
+1. EdgeSite cluster master node is of no less than 2 CPUs and no less than 1GB memories
+2. If high availability is required, 2-3 master nodes are needed at different edge locations
+3. The same Kubernetes security (authN and authZ) mechanisms are used to ensure the secure handshake between master and worker nodes
+4. The same K8s HA mechanism is to be used to enable HA
+
+## Architecture Design
+<img src="../images/EdgeSite_arch.PNG"/>
```
m1093782566: nits:

1. edged instead of kube-proxy will call CNI plugins.

2. missing a connecting line between device and eventbus in the 2nd diagram.

CindyXing: Good catch! Thanks Jun.

CindyXing: Good catch! Thanks Jun.

CindyXing: Good catch. Thanks Jun.

docs/proposals/EdgeSite.md
rohitsardesai83: lightweight etcd would support HA ?

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+KubeEdge is the other opensource project enabling lightweight k8s client on the edge nodes. The KubeEdge agent runtime memory footprint is 
+10MB and it supports kubernetes primitives. Meanwhile KubEdge is targeted for Edge/IOT computing with many functionality enablements. 
+
+By integrating KubeEdge and K3s, this proposal is to allow customers to run an efficient kubernetes cluster for Edge/IOT computing. 
+
+## Assumptions
+Here we assume a cluster is deployed at edge location including the management control plane. 
+For the management control plane to manage some scale of edge worker nodes, the hosting master node needs to have sufficient resources. 
+The assumptions are
+1. EdgeSite cluster master node is of no less than 2 CPUs and no less than 1GB memories
```
rohitsardesai83: s/memories/Memory

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+In Edge computing, there are scenarios where customers would like to have a whole cluster installed at edge location. As a result, 
+admins/users can leverage the local control plane to implement management functionalities and take advantages of all edge computing's benefits. 
+
+This design doc is to enable customers deploy and run lightweight clusters at edge. 
+
+## Motivation
+For these use cases, when considering large set of edge locations, the resource usage of management plane adding together will be high. 
+K3s is an opensource project enabling lightweight kubernetes cluster including the light control plane and worker components. 
+KubeEdge is the other opensource project enabling lightweight k8s client on the edge nodes. The KubeEdge agent runtime memory footprint is 
+10MB and it supports kubernetes primitives. Meanwhile KubEdge is targeted for Edge/IOT computing with many functionality enablements.
```
rohitsardesai83: s/kubernetes/Kubernetes

rohitsardesai83: s/KubEdge/KubeEdge

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Abstract
+In Edge computing, there are scenarios where customers would like to have a whole cluster installed at edge location. As a result, 
+admins/users can leverage the local control plane to implement management functionalities and take advantages of all edge computing's benefits. 
+
+This design doc is to enable customers deploy and run lightweight clusters at edge. 
+
+## Motivation
+For these use cases, when considering large set of edge locations, the resource usage of management plane adding together will be high. 
+K3s is an opensource project enabling lightweight kubernetes cluster including the light control plane and worker components.
```
rohitsardesai83: s/kubernetes/Kubernetes

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+## Abstract
+In Edge computing, there are scenarios where customers would like to have a whole cluster installed at edge location. As a result, 
+admins/users can leverage the local control plane to implement management functionalities and take advantages of all edge computing's benefits. 
+
+This design doc is to enable customers deploy and run lightweight clusters at edge. 
+
+## Motivation
+For these use cases, when considering large set of edge locations, the resource usage of management plane adding together will be high. 
+K3s is an opensource project enabling lightweight kubernetes cluster including the light control plane and worker components. 
+KubeEdge is the other opensource project enabling lightweight k8s client on the edge nodes. The KubeEdge agent runtime memory footprint is
```
rohitsardesai83: is the other -> is another ?

docs/proposals/EdgeSite.md
```diff
@@ -0,0 +1,78 @@
...
+
+This design doc is to enable customers deploy and run lightweight clusters at edge. 
+
+## Motivation
+For these use cases, when considering large set of edge locations, the resource usage of management plane adding together will be high. 
+K3s is an opensource project enabling lightweight kubernetes cluster including the light control plane and worker components. 
+KubeEdge is the other opensource project enabling lightweight k8s client on the edge nodes. The KubeEdge agent runtime memory footprint is 
+10MB and it supports kubernetes primitives. Meanwhile KubEdge is targeted for Edge/IOT computing with many functionality enablements. 
+
+By integrating KubeEdge and K3s, this proposal is to allow customers to run an efficient kubernetes cluster for Edge/IOT computing.
```
rohitsardesai83: proposal is to allow customer  -> proposal enables customers

docs/proposals/EdgeSite.md
CindyXing: Thanks

docs/proposals/EdgeSite.md
CindyXing: :)

docs/proposals/EdgeSite.md
CindyXing: I'd prefer keeping as is. Thanks for the review comments.

PR #328, opened 2019-04-08
Fixed flaky unit test in edgehub/config

### LINKED ISSUE #263: test: Edgehub Config Unit test failed
**Which jobs are failing**:
https://travis-ci.org/kubeedge/kubeedge/jobs/508213930

edgehub config unit test.

**Reason for failure**:
2019/03/19 04:52:10 DEBUG: file event [/tmp/kubeedge/testData REMOVE], operation is %!d(MISSING). reload it.
2019/03/19 04:52:10 WARN: the file change mode: ["/tmp/kubeedge/testData": REMOVE], continue
2019/03/19 04:52:10 DEBUG: watch file error:%!(EXTRA []interface {}=[<nil>])
2019/03/19 04:52:10 DEBUG: file event [/tmp/kubeedge/testData/edge.yaml REMOVE], operation is %!d(MISSING). reload it.
2019/03/19 04:52:10 WARN: the file change mode: ["/tmp/kubeedge/testData/edge.yaml": REMOVE], continue
2019/03/19 04:52:10 DEBUG: file event [/tmp/kubeedge/testData REMOVE], operation is %!d(MISSING). reload it.
2019/03/19 04:52:10 WARN: the file change mode: ["/tmp/kubeedge/testData": REMOVE], continue
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x73f8d3]
goroutine 83 [running]:
github.com/kubeedge/kubeedge/vendor/github.com/ServiceComb/go-archaius/sources/file-source.(*watch).startWatchPool(0xc0003729c0)
	/home/travis/gopath/src/github.com/kubeedge/kubeedge/vendor/github.com/ServiceComb/go-archaius/sources/file-source/yamlconfigurationsource.go:384 +0x73
created by github.com/kubeedge/kubeedge/vendor/github.com/ServiceComb/go-archaius/sources/file-source.(*yamlConfigurationSource).DynamicConfigHandler
	/home/travis/gopath/src/github.com/kubeedge/kubeedge/vendor/github.com/ServiceComb/go-archaius/sources/file-source/yamlconfigurationsource.go:361 +0x91
FAIL	github.com/kubeedge/kubeedge/edge/pkg/edgehub/config	0.019s
ok  	github.com/kubeedge/kubeedge/edge/pkg/edgehub	(cached)
ok  	github.com/kubeedge/kubeedge/edge/pkg/edgehub/clients	0.006s
ok  	github.com/kubeedge/kubeedge/edge/pkg/edgehub/clients/wsclient	300.117s
ok  	github.com/kubeedge/kubeedge/edge/pkg/eventbus/common/util	0.018s
ok  	github.com/kubeedge/kubeedge/edge/pkg/metamanager/dao	0.007s
ok  	github.com/kubeedge/kubeedge/edge/pkg/metamanager	64.101s

sujithsimon22:
**What type of PR is this?**
 /kind failing-test

This PR fixes a flaky unit test that was present in edgehub/config. This will improve our code quality by ensuring that this test case does not fail in the CI while PRs are being raised.

Fixes #263

The issue seems to have been caused when the configuration file mentioned within the test is trying to be loaded before the successful creation of the config file. This issue was occurring randomly since, in the normal execution scenario the file is created pretty quickly and hence this panic is not observed. This is solved by ensuring that the file is loaded only after providing sufficient time for its creation.

### DIFF
--- a/edge/pkg/common/util/config.go
+++ b/edge/pkg/common/util/config.go
@@ -22,7 +22,6 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
-	"io/ioutil"
 	"math/big"
 	"os"
 	"path"
@@ -78,8 +77,18 @@ func GenerateTestYaml(test interface{}, path, filename string) error {
 	if err != nil {
 		return err
 	}
-	err = ioutil.WriteFile(path+"/"+filename+".yaml", data, 0777)
+	fp, err := os.Create(path + "/" + filename + ".yaml")
 	if err != nil {
+		return err
+	}
+	_, err = fp.Write(data)
+	if err != nil {
+		return err
+	}
+	if err = fp.Sync(); err != nil {
+		return err
+	}
+	if err = fp.Close(); err != nil {
 		return err
 	}
 	return nil


sujithsimon22:
cc @rohitsardesai83 @sids-b @m1093782566, Pls review :)

m1093782566:
Probably this change will fix the issue but I have to say it's ugly.

sujithsimon22:
@m1093782566 I have taken your advice and made use of `func (f *File) Sync() (err error)`. The issue does not seem to replicate itself after making this change, so i guess this could work as well. Does it look good to you ?

edge/pkg/common/util/config.go
```diff
@@ -78,7 +77,19 @@ func GenerateTestYaml(test interface{}, path, filename string) error {
 	if err != nil {
 		return err
 	}
-	err = ioutil.WriteFile(path+"/"+filename+".yaml", data, 0777)
+	fp, err:=os.Create(path+"/"+filename+".yaml")
+	if err != nil {
+		return err
+	}
+	_,err=fp.Write(data)
```
m1093782566: Please run go fmt against this file to fix the go fmt issues.

m1093782566:
@sujithsimon22 

I think we can use `sync()` and keep watching the CI statuses to see if it really fixes the issue.

edge/pkg/common/util/config.go
sujithsimon22: oops, my bad. Done, thanks :)

edge/pkg/common/util/config.go
```diff
@@ -78,7 +77,19 @@ func GenerateTestYaml(test interface{}, path, filename string) error {
...
-	err = ioutil.WriteFile(path+"/"+filename+".yaml", data, 0777)
+	fp, err := os.Create(path + "/" + filename + ".yaml")
+	if err != nil {
+		return err
+	}
+	_, err = fp.Write(data)
+	if err != nil {
+		return err
+	}
+	err = fp.Sync()
```
m1093782566: What about:
```
if err := fp.Sync(); err != nil {
  return err
}
```

for the sake of saving one line?

edge/pkg/common/util/config.go
```diff
@@ -78,7 +77,19 @@ func GenerateTestYaml(test interface{}, path, filename string) error {
...
+	if err != nil {
+		return err
+	}
+	err = fp.Close()
```
m1093782566: What about:
```
if err := fp.Close(); err != nil {
  return err
}
```

for the sake of saving one line?

sujithsimon22:
> @sujithsimon22
> 
> I think we can use `sync()` and keep watching the CI statuses to see if it really fixes the issue.

Sure, in case we see this problem repeat itself in the future, we can revisit this and look for another solution

edge/pkg/common/util/config.go
sujithsimon22: Done, thanks

edge/pkg/common/util/config.go
sujithsimon22: Done, thanks

sids-b:
@sujithsimon22 Does this change make sure that the test-code does not try to load the file before it is created ? We could have simply used wait group's if we were using goroutines to create and load files ?

sujithsimon22:
> is change make sure that the test-code does not try to load the file before it is created ? We could have simply used wait group's if we were using goroutines to create and load files ?

@sids-b Currently the functions to create and load the configuration file are not running as go routines, that was the reason why we did not make use of wait groups

PR #349, opened 2019-04-13
Update kubeedge arch pic in README

m1093782566:
**What type of PR is this?**
>

/kind documentation

Update kubeedge arch pic in README to add a new device layer.

Fixes #

```
NONE
```


m1093782566:
> I think the devices should be connected to the app(mapper) instead of directly connecting to the MQTT broker.

Right now we are only MQTT broker way, right?

sids-b:
> > I think the devices should be connected to the app(mapper) instead of directly connecting to the MQTT broker.
> 
> Right now we only support the MQTT broker way, right?

Umm, yes... Lets change this pic once mappers are merged

PR #354, opened 2019-04-15
Modifying device.go for code fix

### LINKED ISSUE #353: Code fixing necessary in device.go
DeviceIssue

**What happened**:
Here msgMetaJSON and attrMetaJSON has got the same content. Hence strings.Compare will always result in 0 and thus code inside if() never hits.
attrMetaJSON,_:=json.marshal(attr.Metadata)
The comparison should happen between the value in context as well as the value that come from cloud.

lidiyag:
**What type of PR is this?**
 /kind bug

This fixes a minor code error in device.go

Fixes #353


lidiyag:
cc @sids-b , @rohitsardesai83 , @m1093782566

PR #364, opened 2019-04-16
clean files after e2e test

edisonxiang:
What type of PR is this?
/kind cleanup

To clean up files which are produced by e2e test.

### DIFF
--- a/tests/e2e/scripts/cleanup.sh
+++ b/tests/e2e/scripts/cleanup.sh
@@ -40,5 +40,25 @@ kill_edgecontroller() {
     fi
 }
 
+cleanup_files(){
+    workdir=$GOPATH/src/github.com/kubeedge/kubeedge
+    cd $workdir
+
+    sudo rm -rf cloud/edgecontroller/edgecontroller
+    sudo rm -rf cloud/edgecontroller/tmp/
+    sudo rm -rf edge/edge.db
+    sudo rm -rf edge/edge_core
+    sudo rm -rf edge/tmp/
+    sudo rm -rf tests/e2e/config.json
+    sudo rm -rf tests/e2e/kubeedge.crt
+    sudo rm -rf tests/e2e/kubeedge.csr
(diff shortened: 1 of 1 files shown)


tests/e2e/scripts/cleanup.sh
```diff
@@ -40,5 +40,28 @@ kill_edgecontroller() {
     fi
 }
 
+cleanup_files(){
+    echo $PWD
+    cd $PWD
+    cd ../../
```
m1093782566: Probably we can define the gopath and then find the proper files to delete base on gopath.

edisonxiang: Thanks.

sids-b:
As far as I know, travis will spinup a new container for running each job, and when the job completes, the container will be destroyed. Do we need cleanup script here ??

m1093782566:
I think this PR make sense for non-CI test environment.

pavan187:
> As far as I know, travis will spinup a new container for running each job, and when the job completes, the container will be destroyed. Do we need cleanup script here ??

Yes, it needs a cleanup, when you run locally framework is generating lot of files, its better if we cleanup those things.

tests/e2e/scripts/cleanup.sh
```diff
@@ -40,5 +40,28 @@ kill_edgecontroller() {
...
+
+    sudo rm -rf cloud/edgecontroller/edgecontroller
+    sudo rm -rf cloud/edgecontroller/edgecontroller.log
+    sudo rm -rf cloud/edgecontroller/tmp/
+    sudo rm -rf edge/edge.db
+    sudo rm -rf edge/edge_core
+    sudo rm -rf edge/edge_core.log
```
pavan187: @edisonxiang , do not cleanup module logs, as we may need them for issue debug in case if any failures.

edisonxiang:
Yep, we are running in local env.

edisonxiang:
Done, PTAL, Thanks:)

PR #377, opened 2019-04-17
certs: use the script to generate the keys and certs

chendave:
There is script which can be reused to generated the keys and certs,
no need to manually generated those files, and manual steps for
certs generation missed some inputs and is error prone.

Also update some configuration for easier installtion, so that
there is no need to update them manually

**What type of PR is this?**
/kind cleanup
/kind documentation

make it easy to create the certs/keys for installation


m1093782566:
Great! That's what I have queried :)

PR #378, opened 2019-04-17
Skipping e2e tests when the change only docs(*.md) #371

### LINKED ISSUE #371: Speed up CI by skipping e2e tests when the change obviously not relevant to that

skipping e2e tests when the change obviously not relevant to that, i.e. document changes.

**Why is this needed**:

Speed up CI.

u5surf:
**What type of PR is this?**
/kind test

speed up ci
**Which issue(s) 
Fixes #371 

 I’d like to test on travis-ci, that’s why I made this PR. Previously, I could execute this ci how I just only push my branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,13 @@ jobs:
       name: "integration test edge"     # names the fourth Tests stage job
     - script: make edge_cross_build
       name: "cross build edge"     # names the fifth Tests stage job
-    - script: make e2e_test
+    - script:
+      | if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(.md$)'
+        then
+          echo "Only doc files were updated, skip e2e test."
+        else
+          make e2e_test
+        fi
       name: "e2e_test"     # names the seventh Tests stage job
     - script: make cloud_lint
       name: "cloud_lint"     # names the eighth Tests stage job


.travis.yml
```diff
@@ -32,7 +32,13 @@ jobs:
       name: "integration test edge"     # names the fourth Tests stage job
     - script: make edge_cross_build
       name: "cross build edge"     # names the fifth Tests stage job
-    - script: make e2e_test
+    - script:
+      | if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(.md)'
```
m1093782566: What will happen if a commit includes two changes: one for `foo.go` and the other one for `bar.md`?

m1093782566:
@pavan187 Seems CI got stuck?

pavan187:
> @pavan187 Seems CI got stuck?

Seems rebase is not done.

.travis.yml
u5surf: grep is quiet, invert-match, and extend-regexp mode. 
I think invert-match means if match except (.md) then it returns true.
And list of filenames reads OR conditions that even only one NOT .md file returns true. 
e.g.
grep -qvE foo.md is false
grep -qvE foo1.md is false
grep -qvE bar.go is true
OR conditions of filelist is true
So, invert-if statement pass through the else conditions that executes e2e test.
c.f. https://reflectoring.io/skip-ci-build/

m1093782566: Okay, thanks for clarification. 

Per the blog, why not

 `if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(.md$)'`?

u5surf: sorry, typo found.`$`

PR #383, opened 2019-04-18
fix wrong way when get node id in funciton dispatchMessage

ugvddm:
we need the length of slice instead of string when get node id
from  msg.Router.Resource

**What type of PR is this?**

> /kind bug

### DIFF
--- a/cloud/edgecontroller/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/edgecontroller/pkg/cloudhub/channelq/channelq.go
@@ -72,9 +72,10 @@ func (q *ChannelEventQueue) dispatchMessage() {
 		}
 		resource := msg.Router.Resource
 		tokens := strings.Split(resource, "/")
+		numOfTokens := len(tokens)
 		var nodeID string
 		for i, token := range tokens {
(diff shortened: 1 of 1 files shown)


cloud/edgecontroller/pkg/cloudhub/channelq/channelq.go
```diff
@@ -72,9 +72,10 @@ func (q *ChannelEventQueue) dispatchMessage() {
 		}
 		resource := msg.Router.Resource
 		tokens := strings.Split(resource, "/")
+		tokenLen := len(tokens)
```
m1093782566: nit: tokenLen means the length of a given token which I think it's not suitable for our case.

I suggest to use `numOfTokens` instead. Thoughts?

m1093782566:
@ugvddm Please rebase to the HEAD for the sake of running the test jobs as we just fix the travis CI error

cloud/edgecontroller/pkg/cloudhub/channelq/channelq.go
ugvddm: nice

PR #387, opened 2019-04-19
[do not merge] test ci

edisonxiang:
test ci

### DIFF
--- a/test.md
+++ b/test.md
@@ -1,0 +1,6 @@
+## Motivation
+
+### Test
+
+This is a test, please do not merge
+


m1093782566:
I expected the CI result to be green. However, it keeps being pending state though it REALLY skips all test jobs.

edisonxiang:
A moment ago, it was in the queue.
Now it is running.
https://travis-ci.org/kubeedge/kubeedge/builds/521996156

edisonxiang:
Seem good. It saves about 30 mins

edisonxiang:
It works we can close this pr.
If we have better solution to skip the docs-only commit, we will reopen it

PR #397, opened 2019-04-22
output time duration per test case

edisonxiang:
**What type of PR is this?**
/kind test

output the time duration per test case for the E2E tests.
It can be used to measure the pod starup time and so on.

### DIFF
--- a/tests/e2e/deployment/deployment_test.go
+++ b/tests/e2e/deployment/deployment_test.go
@@ -33,13 +33,25 @@ const (
 	DeploymentHandler = "/apis/apps/v1/namespaces/default/deployments"
 )
 
+var DeploymentTestTimerGroup *utils.TestTimerGroup = utils.NewTestTimerGroup()
+
 //Run Test cases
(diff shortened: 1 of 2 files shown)


tests/e2e/deployment/deployment_test.go
```diff
@@ -76,8 +78,14 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
 				}
 			}
 			utils.CheckPodRunningState(ctx.Cfg.ApiServer+AppHandler, podlist)
+			// End test timer
+			testTimer.End()
+			// Print result
+			testTimer.PrintResult()
```
m1093782566: Do we need to store the time duration in somewhere?

edisonxiang: OK. Thanks

edisonxiang: Done, PTAL. Thanks

edisonxiang:
@sids-b It makes sense. Thanks. Updated:)

PR #398, opened 2019-04-22
Adding UT for controller/manager/location.go

omung-g:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Fixes #

/uncc @sids-b

### DIFF
--- a/cloud/edgecontroller/pkg/controller/manager/location_test.go
+++ b/cloud/edgecontroller/pkg/controller/manager/location_test.go
@@ -26,14 +26,6 @@ import (
 
 var nodes = []string{"Node1", "Node2"}
 
-//locationCache returns a new LocationCache
-func locationCache(configMapKey string, configMapValue interface{}, secretKey string, SecretValue interface{}) LocationCache {
-	lc := LocationCache{}
-	lc.configMapNode.Store(configMapKey, configMapValue)
-	lc.secretNode.Store(secretKey, SecretValue)
-	return lc
-}
-
 //TestAddOrUpdatePod is function to test AddOrUpdatePod
 func TestAddOrUpdatePod(t *testing.T) {
 	pod := v1.Pod{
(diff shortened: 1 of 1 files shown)


cloud/edgecontroller/pkg/controller/manager/location_test.go
```diff
@@ -0,0 +1,161 @@
...
+	locationCache.DeleteConfigMap(namespace, configMapName)
+}
+
+//TestDeleteSecret is function to test DeleteSecret
+func TestDeleteSecret(t *testing.T) {
+	locationCache := LocationCache{}
+	locationCache.secretNode.Store("ObjectMeta1/VolumeSecret1", nodes)
+	namespace := "ObjectMeta1"
+	secretName := "VolumeSecret1"
+	locationCache.DeleteSecret(namespace, secretName)
```
rohitsardesai83: we should do a test to check if the secret is deleted by doing a get from the cache, it should return false / nil.

cloud/edgecontroller/pkg/controller/manager/location_test.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+//TestDeleteConfigMap is function to test DeleteConfigMap
+func TestDeleteConfigMap(t *testing.T) {
+	locationCache := LocationCache{}
+	locationCache.configMapNode.Store("ObjectMeta1/VolumeConfig1", nodes)
+	namespace := "ObjectMeta1"
+	configMapName := "VolumeConfig1"
+	locationCache.DeleteConfigMap(namespace, configMapName)
```
rohitsardesai83: we should do a test to check if the config map is deleted by doing a get from the cache, it should return false / nil.

cloud/edgecontroller/pkg/controller/manager/location_test.go
```diff
@@ -0,0 +1,161 @@
...
+			}},
+			ImagePullSecrets: []v1.LocalObjectReference{{Name: "ImageSecret1"}},
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "ObjectMeta1",
+			Name:      "Object1",
+		},
+	}
+	locationCache := LocationCache{}
+	locationCache.configMapNode.Store("ObjectMeta1/VolumeConfig1", "Node1")
```
rohitsardesai83: let's define constants for all the keys we are reusing throughout the tests.

rohitsardesai83:
@omung-g , please check CI failure.

PR #402, opened 2019-04-23
Device CRD: Fix #384, make device property and visitor config pointer types

### LINKED ISSUE #384: The DataTypes in the PropertyType of device_model_types.go should be of pointer type.
**What happened**:
Since the data types in the PropertyType are not pointers, cannot check if Int exists or String exists. 

Expected it to be a pointer if run following code
``` go
if deviceModel.Spec.Properties[0].Type.Int == nil{
}

rohitsardesai83:
**What type of PR is this?**
/kind bug

Changed the visitor config and property types to pointers for nil validation checks in the device controller.

Fixes #384 

NONE


rohitsardesai83:
@m1093782566 , this is the same PR as #401 , with clean commit history.

PR #408, opened 2019-04-25
change poll into watch

edisonxiang:
**What type of PR is this?**
/kind cleanup

change poll into watch for check if pods are running status


m1093782566:
Nice catch! Thanks!

Issue #396, opened 2019-04-22
Unable to run cloud as k8s deployments


chendave:
**What happened**:
Follow the guide to deploy cloud as k8s deployments,  "edgecontroller" is started on host successfully, then create a edge node, but the edge node never come to be ready.

**What you expected to happen**:
node should be ready just like how to deploy with binaries, and there should be a way to know what happened on controller node.

**How to reproduce it (as minimally and precisely as possible)**:
- Run as Kubernetes deployment, follow the below guide https://github.com/kubeedge/kubeedge/blob/master/build/cloud/README.md

- edgecontroller process detectable on the controller node.
```bash
# ps -ef | grep edge
root     21894 13963  0 18:10 pts/1    00:00:00 grep --color=auto edge
root     24088 24053  0 13:49 ?        00:00:00 edgecontroller
```
- Run edge with binary, follow the below guide, 
https://github.com/kubeedge/kubeedge/blob/master/docs/getting-started/usage.md#run-as-a-binary-1

- mosquitto is started, edge_core is started

- node is created with below template,
```
{
  "kind": "Node",
  "apiVersion": "v1",
  "metadata": {
    "name": "10.169.41.219",
    "labels": {
      "name": "edge-node"
    }
  }
}
```
- `kubectl get nodes` shows the node is not ready and always not ready.
```bash
# kubectl get nodes
NAME            STATUS     ROLES    AGE    VERSION
10.169.41.219   NotReady   <none>   158m
dave-desktop    Ready      master   6d2h   v1.14.1
```

BTW, If master is deployed with binary everything should be fine. 
And I also want to know how to debug for this case? where I can get the *log* if the controller is deployed in k8s cluster?

**Anything else we need to know?**:

**Environment**:
- KubeEdge version: latest
- Hardware configuration: x84_64 / arm64
- OS (e.g. from /etc/os-release): ubuntu18.04
- Kernel (e.g. `uname -a`):
- Others:

m1093782566:
Is there any interesting log to show? i.e. the log of cloudhub or edgecontroller or edge-core?

cc @sids-b @nkwangleiGIT

sids-b:
also can you please share the config file of edge_core ??

sids-b:
I think ``` kubectl logs <pod-name>``` should fetch you the logs of controller in k8s deployment.

chendave:
@sids-b  do you mean this fie?
[...]

PR #410, opened 2019-04-25
fix cloudpart cannot run in container

edisonxiang:
**What type of PR is this?**
/kind bug

Currently the kubeedge cloud part image can not be running in container.
Firstly it is because the cloud part is refering to some device types definition in the edge codes.
when we build the cloud part image, the edge part codes will be invovled into the cloud part image.
The sqlite is required for edge but not for cloud.
currently becase of some edge device types are referred by cloud part
so that the cloud part image will require sqlite.
we need to let the common device type definition into the common folder.

Another reason is about the deployment yamls like the modified files in this pr.

**Which issue(s) this PR fixes**: #396

### DIFF
--- a/build/cloud/05-configmap.yaml
+++ b/build/cloud/05-configmap.yaml
@@ -20,9 +20,9 @@ data:
     cloudhub:
       address: 0.0.0.0
       port: 10000
-      ca: /etc/kubeedge/cloud/certs/ca.crt
-      cert: /etc/kubeedge/cloud/certs/cloud.crt
-      key: /etc/kubeedge/cloud/certs/cloud.key
+      ca: /etc/kubeedge/certs/rootCA.crt
+      cert: /etc/kubeedge/certs/edge.crt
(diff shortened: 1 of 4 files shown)


edisonxiang:
@sids-b I guess we need to do further work except this pr.

https://github.com/kubeedge/kubeedge/blob/master/cloud/edgecontroller/pkg/devicecontroller/types/device.go#L4

we should not refer edge part code from the cloud part directly.

m1093782566:
> we should not refer edge part code from the cloud part directly.

+1

m1093782566:
cc @chendave

sids-b:
My bad @edisonxiang .I forgot to delete them.

sids-b:
@edisonxiang , can you please delete all the unused structures in device.go in this PR ?

edisonxiang:
@sids-b Thanks, I will update later

edisonxiang:
@sids-b Done. PTAL. Thanks very much.

PR #411, opened 2019-04-25
[Device CRD] Fix #409: Change uint to int for bluetooth read properties, add MAC address in device instance protocol config

### LINKED ISSUE #409: Change type of ShiftLeft and ShiftRight fields present in device_model_types.go from uint to int
**What happened**:
We are unable to create device model for bluetooth protocol device, since uint values are expected for the fields ShiftLeft and ShiftRight, but while passing values from yaml, the numbers are considered as integers, which leads to an error in creation

We should be expecting integer values for fields ShiftLeft and ShiftRight instead of uint
 
Create a device model for a bluetooth protocol device which contains values for ShiftLeft and ShiftRight attributes.

rohitsardesai83:
**What type of PR is this?**

/kind bug

Change uint to int for bluetooth visitor properties, add MAC address in device instance protocol config

Fixes #409 

NONE


sids-b:
@sujithsimon22 is blocked because of this. Just a small fix changing uint to int. Merging it to unblock.

PR #412, opened 2019-04-26
remove keepalive

### LINKED ISSUE #256: CloudHub logs warning when handling keepalive message from edge
**What happened**:

cloud hub logs warning, with error message "resource type not found"

image

no warning message

common (default) configuration would stable reproduce.

EdgeHub side heartbeat to CloudHub:

https://github.com/kubeedge/kubeedge/blob/6878d86b9ec33ca716933575b93ef70663504ba5/edge/pkg/edgehub/controller.go#L253-L266

CloudHub forward message through beehive:
https://github.com/kubeedge/kubeedge/blob/6878d86b9ec33ca716933575b93ef70663504ba5/cloud/edgecontroller/pkg/cloudhub/wsserver/server.go#L102-L106

Upstream controller dispatch message:

https://github.com/kubeedge/kubeedge/blob/6878d86b9ec33ca716933575b93ef70663504ba5/cloud/edgecontroller/pkg/controller/controller/upstream.go#L141-L145

**Environment**:
- KubeEdge version: v0.2
- Others: irrelevant

subpathdev:
the keepalive message of the websocket client is not used and creates a warning in the cloud part.
fix #256

/kind cleanup

Fixes #256

### DIFF
--- a/edge/pkg/edgehub/controller.go
+++ b/edge/pkg/edgehub/controller.go
@@ -104,7 +104,6 @@ func (ehc *Controller) Start(ctx *context.Context) {
 		ehc.pubConnectInfo(true)
 		go ehc.routeToEdge()
 		go ehc.routeToCloud()
-		go ehc.keepalive()
 
 		// wait the stop singal
 		// stop authinfo manager/websocket connection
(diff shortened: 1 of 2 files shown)


m1093782566:
@subpathdev Please fix the UT failures.

m1093782566:
cc @sids-b @rohitsardesai83 @Baoqiang-Zhang

m1093782566:
Or do we need to handle the heartbeat in cloud side? Thoughts?

sids-b:
@subpathdev my bad, I think ping was created with a idea that there will be a placement server in cloud when we support cloud HA. There is a config parameter called placement, if it is true we use the ping. Can we keep the function call ```go go ehc.keepalive() ``` inside if placement true statement ?
Sorry for trouble.

subpathdev:
@sids-b yes of course this is possible but the message failed in parsing on the cloud side (because static parameter). The functionality of a keepalive is managed by the heartbeat (which synchronize the node state to the cloud).

sids-b:
@subpathdev Sounds Fair. Thanks

PR #424, opened 2019-04-28
judge faild whether is nil for SetPods func

kadisi:

**What type of PR is this?**
>

### DIFF
--- a/edge/pkg/edged/podmanager/pod_manager.go
+++ b/edge/pkg/edged/podmanager/pod_manager.go
@@ -48,18 +48,9 @@ func NewPodManager() pod.Manager {
 	pm := &podManager{
 		MockManager: new(testing.MockManager),
 	}
-	pm.SetPods(nil)
-	return pm
-}
-
-// Set the internal pods based on the new pods.
-func (pm *podManager) SetPods(newPods []*v1.Pod) {
-	pm.lock.Lock()
-	defer pm.lock.Unlock()
-
 	pm.podByUID = make(map[types.UID]*v1.Pod)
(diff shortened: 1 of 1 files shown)


m1093782566:
@kadisi A general comment, better to provide how to reproduce if you think it's a bug.

BTW, I feel `pm.podByUID` and `pm.podByFullName` should be ALWAYS nil in this case?

kadisi:
@m1093782566 yes,  it should be always nil, but you can't guarantee that developers will use it incorrectly, i think SetPods functions is no usefull.   SetPods `s function can be directly integrated into NewPodManager function。 SetPods function only be called once in this project.

`
//NewPodManager creates new pod manager object
func NewPodManager() pod.Manager {
	pm := &podManager{
		MockManager: new(testing.MockManager),
	}
	pm.podByUID = make(map[types.UID]*v1.Pod)
	pm.podByFullName = make(map[string]*v1.Pod)
	return pm
}
`

m1093782566:
@kadisi Thanks for your information.

>SetPods function only be called once in this project.

If so, I would like to move the logic of `SetPods()` into ` NewPodManager()`.

sids-b:
@kadisi , thanks. SetPods() is only called once in edged. 
As @m1093782566 suggested. The logic of setPods can be moved inside NewPodManager().

kadisi:
@sids-b @m1093782566 PTAL

m1093782566:
LGTM, please squash the commits.

kadisi:
@m1093782566  i have squashed the commits

PR #428, opened 2019-04-28
Revert "setup script changes for IEF"

sids-b:
Reverts kubeedge/kubeedge#418

Since these changes were made after ping was removed. We are bringing back ping so these won't be needed.


sids-b:
cc @m1093782566 @naveensriram

PR #433, opened 2019-04-29
change break to continue , for iterate through the remaining container records

kadisi:

**What type of PR is this?**

 /kind bug

Fixes #

change break to continue, for iterate through the remaining container records

### DIFF
--- a/edge/pkg/edged/containers/container_manager.go
+++ b/edge/pkg/edged/containers/container_manager.go
@@ -1122,7 +1122,7 @@ func (cm *containerManager) freeContainer(freeTime time.Time, gcPolicy kubeconta
 		log.LOGGER.Infof("Evaluating Container ID %s for possible garbage collection", record.containerID)
 		if record.lastUsed.Equal(freeTime) || record.lastUsed.After(freeTime) {
 			log.LOGGER.Infof("Container ID %s has lastUsed=%v which is >= freeTime=%v, not eligible for garbage collection", record.containerID, record.lastUsed, freeTime)
-			break
+			continue
 		}
 
 		if freeTime.Sub(record.firstDetected) < gcPolicy.MinAge {


edge/pkg/edged/containers/container_manager.go
```diff
@@ -1122,7 +1122,7 @@ func (cm *containerManager) freeContainer(freeTime time.Time, gcPolicy kubeconta
 		log.LOGGER.Infof("Evaluating Container ID %s for possible garbage collection", record.containerID)
 		if record.lastUsed.Equal(freeTime) || record.lastUsed.After(freeTime) {
 			log.LOGGER.Infof("Container ID %s has lastUsed=%v which is >= freeTime=%v, not eligible for garbage collection", record.containerID, record.lastUsed, freeTime)
-			break
+			continue
```
m1093782566: I think continue here make no sense.

rohitsardesai83: since we are looping over container records , at each step we evaluate if the container is eligible for GC and if not , then we should continue to process the next container. The break statement would force termination of the loop with just the first container processed which is wrong.

m1093782566: Yes, you are right.

kadisi: my pod can not sync status in my test env.  i find this break will force terminate the loop.  but this change maybe can not fix sync problem.。 by the way, i found that `detectTime` param  is useless in  `detectContainers` funtions.  i do not know why paas `detectTime` param to `detectContainers` funtions.

m1093782566:
cc @Baoqiang-Zhang

edge/pkg/edged/containers/container_manager.go
sids-b: I think @rohitsardesai83 is right

rohitsardesai83:
lgtm

edge/pkg/edged/containers/container_manager.go
Baoqiang-Zhang: @rohitsardesai83  you are right, it should be continue here.

PR #465, opened 2019-05-08
change event channel to non-blocking channel

kadisi:

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

Fixes #

change event channel to non-blocking channel to Increase processing speed

### DIFF
--- a/cloud/pkg/controller/config/buffer.go
+++ b/cloud/pkg/controller/config/buffer.go
@@ -18,32 +18,63 @@ var QueryConfigMapBuffer int
 // QuerySecretBuffer is the size of channel which save query secret message from edge
 var QuerySecretBuffer int
 
+// PodEventBuffer is the size of channel which save pod event from k8s
+var PodEventBuffer int
+
+// ConfigMapEventBuffer is the size of channel which save configmap event from k8s
+var ConfigMapEventBuffer int
+
+// SecretEventBuffer is the size of channel which save secret event from k8s
+var SecretEventBuffer int
+
 func init() {
 	if psb, err := config.CONFIG.GetValue("controller.buffer.update-pod-status").ToInt(); err != nil {
 		UpdatePodStatusBuffer = constants.DefaultUpdatePodStatusBuffer
 	} else {
 		UpdatePodStatusBuffer = psb
 	}
-	log.LOGGER.Infof("update pod status buffer: %d", UpdatePodStatusBuffer)
+	log.LOGGER.Infof("Update controller.buffer.update-pod-status: %d", UpdatePodStatusBuffer)
 
 	if nsb, err := config.CONFIG.GetValue("controller.buffer.update-node-status").ToInt(); err != nil {
 		UpdateNodeStatusBuffer = constants.DefaultUpdateNodeStatusBuffer
 	} else {
 		UpdateNodeStatusBuffer = nsb
 	}
-	log.LOGGER.Infof("Update node status buffer: %d", UpdateNodeStatusBuffer)
+	log.LOGGER.Infof("Update controller.buffer.update-node-status: %d", UpdateNodeStatusBuffer)
 
 	if qcb, err := config.CONFIG.GetValue("controller.buffer.query-configmap").ToInt(); err != nil {
 		QueryConfigMapBuffer = constants.DefaultQueryConfigMapBuffer
 	} else {
 		QueryConfigMapBuffer = qcb
 	}
-	log.LOGGER.Infof("query config map buffer: %d", QueryConfigMapBuffer)
+	log.LOGGER.Infof("Update controller.buffer.query-configmap: %d", QueryConfigMapBuffer)
 
 	if qsb, err := config.CONFIG.GetValue("controller.buffer.query-secret").ToInt(); err != nil {
 		QuerySecretBuffer = constants.DefaultQuerySecretBuffer
 	} else {
 		QuerySecretBuffer = qsb
 	}
-	log.LOGGER.Infof("query secret buffer: %d", QuerySecretBuffer)
+	log.LOGGER.Infof("Update controller.buffer.query-secret: %d", QuerySecretBuffer)
+
+	if peb, err := config.CONFIG.GetValue("controller.buffer.pod-event").ToInt(); err != nil {
+		PodEventBuffer = constants.DefaultPodEventBuffer
+	} else {
+		PodEventBuffer = peb
+	}
+	log.LOGGER.Infof("Update controller.buffer.pod-event: %d", PodEventBuffer)
+
+	if cmeb, err := config.CONFIG.GetValue("controller.buffer.configmap-event").ToInt(); err != nil {
+		ConfigMapEventBuffer = constants.DefaultConfigMapEventBuffer
+	} else {
+		ConfigMapEventBuffer = cmeb
+	}
+	log.LOGGER.Infof("Update controller.buffer.configmap-event: %d", ConfigMapEventBuffer)
+
+	if seb, err := config.CONFIG.GetValue("controller.buffer.secret-event").ToInt(); err != nil {
+		SecretEventBuffer = constants.DefaultSecretEventBuffer
+	} else {
+		SecretEventBuffer = seb
+	}
+	log.LOGGER.Infof("Update controller.buffer.secret-event: %d", SecretEventBuffer)
+
 }
--- a/cloud/pkg/controller/manager/configmap.go
+++ b/cloud/pkg/controller/manager/configmap.go
@@ -1,6 +1,8 @@
 package manager
 
 import (
+	"github.com/kubeedge/kubeedge/cloud/pkg/controller/config"
+
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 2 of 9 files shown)


rohitsardesai83:
@kadisi , thanks for the contribution , can you please check the CI failures ?

kadisi:
does this ci have `/retest` function like kubernetes?

sids-b:
can we have the same change in all the managers in controller and deviceController ??

kadisi:
@sids-b , ok, i think we need change. but is it ok to set channel buff to 20?

sids-b:
I am not sure, Maybe make it configuration based and set default to 10 ?

sids-b:
It maybe not be a required config, we have many such fields in controller. You can have a look at the config folder of controller.

kadisi:
@sids-b think you,

kadisi:
@sids-b by the way, what is the deference between `-` and `.` in the Key  when i use `config.CONFIG.GetValue()`.  i know `.` come from yaml config. and  does `-` come from cmd or ENV ?

m1093782566:
Sorry, I am not against chaging the size of channle buffer but I need to know the effect after changing the size. Does it help improving the performance? We need to see the data.

kadisi:
@m1093782566 do you have some comparative suggestions about test case.

m1093782566:
I think we care about the e2e performance data comparisions. @edisonxiang is working on e2e performance test, you may ask him for suggestions.

kadisi:
@m1093782566  thank you, @edisonxiang  I contact you by email

edisonxiang:
Hey @kadisi Thanks your pr.
I guess this is something important for the performance.
We need to make more tests for the kubeedge performance.

Currently we only add channel buffer in the upstream like the following.
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/controller/controller/upstream.go#L93-L122
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/controller/upstream.go#L64-L79

There are some places like below which is using  `make(chan watch.Event)`
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/controller/manager/configmap.go#L24
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/controller/manager/pod.go#L83-L84
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/controller/manager/secret.go#L24
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/manager/device.go#L31
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/manager/devicemodel.go#L31

Could you help to add the buffer config like below so that we can avoid to use the fixed value like `20`, perhaps we can set the default value like `1`. if the users want to change these buffer, they can modify the conf of kubeedge to adjust them.

https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/config/buffer.go
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/controller/config/buffer.go

@kadisi what do you think about that?
@rohitsardesai83 @m1093782566 @sids-b

sids-b:
@kadisi , in a configuration file which is yaml, there is heirarchy. So ``` . ``` is used to traverse down the heirarchy. for eg:
```
kube:
  master: 127.0.0
```
So to get the value of master, you need to use the key kube.master as master is in the heirarchy of kube
As per my knowledge ``` - ``` doesn't hold any significance. If the key has a - for eg: master-ip, they we use it as kube.master-ip.

Also, environmental variable has higher preference than config file. If same key is present as environmental variable and in config file, The one in config file will be overwritten during runtime.

kadisi:
@sids-b thank for your help.      @edisonxiang ok, i will set it

cloud/pkg/controller/config/buffer.go
```diff
@@ -46,4 +55,26 @@ func init() {
 		QuerySecretBuffer = qsb
 	}
 	log.LOGGER.Infof("query secret buffer: %d", QuerySecretBuffer)
+
+	if peb, err := config.CONFIG.GetValue("pod-event-buffer").ToInt(); err != nil {
```
edisonxiang: perhaps we need to replace 'pod-event-buffer' into 'controller.pod-event-buffer'
The below config should be that too.

cloud/pkg/devicecontroller/config/buffer.go
```diff
@@ -9,11 +9,32 @@ import (
...
+
 func init() {
 	if psb, err := config.CONFIG.GetValue("devicecontroller.update-device-status-buffer").ToInt(); err != nil {
 		UpdateDeviceStatusBuffer = constants.DefaultUpdateDeviceStatusBuffer
 	} else {
 		UpdateDeviceStatusBuffer = psb
 	}
 	log.LOGGER.Infof("Update device status buffer: %d", UpdateDeviceStatusBuffer)
+
+	if deb, err := config.CONFIG.GetValue("device-event-buffer").ToInt(); err != nil {
```
edisonxiang: perhaps we need to replace 'device-event-buffer' into 'devicecontroller.device-event-buffer'
The below config should be that too.

cloud/pkg/controller/config/buffer.go
sids-b: Can we refer to PR #490 and add these configurations under buffer for both controller and deviceController ?
Lets wait till pr #490 is approved though

kadisi: ok

sids-b: @kadisi , the PR is merged. Please take a look and add configurations under buffer

rohitsardesai83:
@kadisi , can you please check ?

cloud/pkg/controller/config/buffer.go
kadisi: ok

cloud/pkg/devicecontroller/config/buffer.go
kadisi: ok

cloud/pkg/controller/manager/configmap.go
```diff
@@ -1,6 +1,7 @@
 package manager
 
 import (
+	"github.com/kubeedge/kubeedge/cloud/pkg/controller/config"
```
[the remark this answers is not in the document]
kadisi: ok

cloud/pkg/controller/config/buffer.go
kadisi: ok

cloud/pkg/devicecontroller/config/buffer.go
```diff
@@ -9,11 +9,32 @@ import (
...
 func init() {
 	if psb, err := config.CONFIG.GetValue("devicecontroller.buffer.update-device-status").ToInt(); err != nil {
 		UpdateDeviceStatusBuffer = constants.DefaultUpdateDeviceStatusBuffer
 	} else {
 		UpdateDeviceStatusBuffer = psb
 	}
-	log.LOGGER.Infof("Update device status buffer: %d", UpdateDeviceStatusBuffer)
+	log.LOGGER.Infof("devicecontroller Update device status buffer: %d", UpdateDeviceStatusBuffer)
+
+	if deb, err := config.CONFIG.GetValue("devicecontroller.device-event-buffer").ToInt(); err != nil {
```
edisonxiang: As @sids-b suggests, it is better to add configurations under buffer like the line 19 mentions
"devicecontroller.buffer.update-device-status"
So the folloing modification is welcome:
Replace "devicecontroller.device-event-buffer" with "devicecontroller.buffer.device-event"
Replace "devicecontroller.device-model-event-buffer" with "devicecontroller.buffer.device-model-event"

cloud/pkg/controller/config/buffer.go
```diff
@@ -18,32 +18,63 @@ var QueryConfigMapBuffer int
...
 
 	if qsb, err := config.CONFIG.GetValue("controller.buffer.query-secret").ToInt(); err != nil {
 		QuerySecretBuffer = constants.DefaultQuerySecretBuffer
 	} else {
 		QuerySecretBuffer = qsb
 	}
-	log.LOGGER.Infof("query secret buffer: %d", QuerySecretBuffer)
+	log.LOGGER.Infof("controller query secret buffer: %d", QuerySecretBuffer)
+
+	if peb, err := config.CONFIG.GetValue("controller.pod-event-buffer").ToInt(); err != nil {
```
edisonxiang: The following modifications are welcome:
Replace "controller.pod-event-buffer" with "controller.buffer.pod-event"
Replace "controller.configmap-event-buffer" with "controller.buffer.configmap-event"
Replace "controller.secret-event-buffer" with "controller.buffer.secret-event"

@sids-b @kadisi How about your ideas?

sids-b: Agree with @edisonxiang

cloud/pkg/controller/config/buffer.go
```diff
@@ -18,32 +18,63 @@ var QueryConfigMapBuffer int
...
+var SecretEventBuffer int
+
 func init() {
 	if psb, err := config.CONFIG.GetValue("controller.buffer.update-pod-status").ToInt(); err != nil {
 		UpdatePodStatusBuffer = constants.DefaultUpdatePodStatusBuffer
 	} else {
 		UpdatePodStatusBuffer = psb
 	}
-	log.LOGGER.Infof("update pod status buffer: %d", UpdatePodStatusBuffer)
+	log.LOGGER.Infof("controller update pod status buffer: %d", UpdatePodStatusBuffer)
```
sids-b: nit : Please start log with capital letters
applicable to similar places

cloud/pkg/controller/manager/secret.go
```diff
@@ -1,6 +1,7 @@
 package manager
 
 import (
+	"github.com/kubeedge/kubeedge/cloud/pkg/controller/config"
```
sids-b: nit: please seperate kubeedge and vendor packages with a blank line.

cloud/pkg/devicecontroller/config/buffer.go
sids-b: +1

cloud/pkg/controller/config/buffer.go
kadisi: ok

kadisi:
@sids-b @edisonxiang

m1093782566:
/lgtm

leave approve to @edisonxiang @sids-b

edisonxiang:
@kadisi please rebase your pr, and then we will merge it. Thanks very much.

PR #467, opened 2019-05-08
Fix GetContent format []byte output string

lou-lan:
**What type of PR is this?**
/kind bug

Fix GetContent format []byte output string

### DIFF
--- a/cloud/pkg/cloudhub/common/model/types.go
+++ b/cloud/pkg/cloudhub/common/model/types.go
@@ -170,5 +170,5 @@ func (event *Event) IsToEdge() bool {
 
 // GetContent dumps the content to string
 func (event *Event) GetContent() string {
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudhub/common/model/types.go
```diff
@@ -170,5 +170,5 @@ func (event *Event) IsToEdge() bool {
 
 // GetContent dumps the content to string
 func (event *Event) GetContent() string {
-	return fmt.Sprintf("%v", event.Content)
+	return fmt.Sprintf("%s", event.Content)
```
sids-b: I think event.Content is of type interface{}, so we may need %v here?

lou-lan: Yes, event.Content is of type interface{}.

PR #455, opened 2019-05-07
add quic design doc

kexun:
**What type of PR is this?**

/kind design

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

PR #473, opened 2019-05-08
when edge register into cloud, cloud send all info to edge

### LINKED ISSUE #260: If edge core didn't handle pod delete event from cloud, pod remains with status Terminating

kadisi:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind bug

when edgecontroller first started , and  edge_core sencond started after many hours or days.
edgecontroller first get all pods and then send to the relevant host channelPool
```
type ChannelEventQueue struct {
	ctx         *context.Context
	channelPool sync.Map
}
```
but if now , edge_core not started, the relevant host channelpool can not find, so edgecontroller can not send full ammount event to edge_core, this will result in loss of pod information.

so we need save these events in advance.

but maybe we need merge this events to Remove old duplicates.

Fixes #455 
FIxes #260

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -84,6 +84,11 @@ func (q *ChannelEventQueue) dispatchMessage() {
 			log.LOGGER.Warnf("node id is not found in the message")
 			continue
 		}
+		_, ok := q.channelPool.Load(nodeID)
+		if !ok {
+			rChannel := make(chan model.Event, rChanBufSize)
+			q.channelPool.LoadOrStore(nodeID, rChannel)
+		}
 		rChannel, err := q.getRChannel(nodeID)
 		if err != nil {
 			log.LOGGER.Infof("fail to get dispatch channel for %s", nodeID)
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -115,15 +115,14 @@ func (q *ChannelEventQueue) getRChannel(nodeID string) (chan model.Event, error)
 // Connect allocates rChannel for given project and group
 func (q *ChannelEventQueue) Connect(info *model.HubInfo) error {
 	_, ok := q.channelPool.Load(info.NodeID)
-	if ok {
-		return fmt.Errorf("edge node %s is already connected", info.NodeID)
-	}
-	// allocate a new rchannel with default buffer size
-	rChannel := make(chan model.Event, rChanBufSize)
-	_, ok = q.channelPool.LoadOrStore(info.NodeID, rChannel)
-	if ok {
-		// rchannel is already allocated
-		return fmt.Errorf("edge node %s is already connected", info.NodeID)
+	if !ok {
+		// allocate a new rchannel with default buffer size
+		rChannel := make(chan model.Event, rChanBufSize)
+		_, ok = q.channelPool.LoadOrStore(info.NodeID, rChannel)
+		if ok {
+			// rchannel is already allocated
+			return fmt.Errorf("edge node %s is already connected", info.NodeID)
+		}
 	}
 	return nil
 }


cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -110,15 +115,14 @@ func (q *ChannelEventQueue) getRChannel(nodeID string) (chan model.Event, error)
...
-	}
-	// allocate a new rchannel with default buffer size
-	rChannel := make(chan model.Event, rChanBufSize)
-	_, ok = q.channelPool.LoadOrStore(info.NodeID, rChannel)
-	if ok {
-		// rchannel is already allocated
-		return fmt.Errorf("edge node %s is already connected", info.NodeID)
+	if !ok {
+		// allocate a new rchannel with default buffer size
+		rChannel := make(chan model.Event, rChanBufSize)
```
sids-b: There is a problem here, if ok is true, we will return nil which is wrong. ok means edgenode with same id already exists and it should return an error. Please undo this change

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -84,6 +84,11 @@ func (q *ChannelEventQueue) dispatchMessage() {
 			log.LOGGER.Warnf("node id is not found in the message")
 			continue
 		}
+		_, ok := q.channelPool.Load(nodeID)
+		if !ok {
+			rChannel := make(chan model.Event, rChanBufSize)
```
sids-b: We should not create channel of node which are not connected.

CindyXing:
@kadisi , thanks for raising the issue. Can you clarify the customer impact here? Are we having a data loss? Can the edgenode get pod information later when channelpool created?

kadisi:
@CindyXing , in cloud, controller will create DownstreamController which create many manager, such as PodManager, ConfigManager. In the case of PodManager,  it will get realEvents by SharedInformer and ResourceEventHander 
```
// NewPodManager create PodManager from config
func NewPodManager(kubeClient *kubernetes.Clientset, namespace string) (*PodManager, error) {
	lw := cache.NewListWatchFromClient(kubeClient.CoreV1().RESTClient(), "pods", namespace, fields.Everything())
	realEvents := make(chan watch.Event)
	mergedEvents := make(chan watch.Event)
	rh := NewCommonResourceEventHandler(realEvents)
	si := cache.NewSharedInformer(lw, &v1.Pod{}, 0)
	si.AddEventHandler(rh)

	pm := &PodManager{realEvents: realEvents, mergedEvents: mergedEvents}

	stopNever := make(chan struct{})
	go si.Run(stopNever)
	go pm.merge()

	return pm, nil
}
```
when edgecontroller start, it will get all pod events(first list all pods, and then watch them) , and merge them into `mergedEvents channel`.

and in `syncPod` function, it will Consuming these events。and send them by messageLayer.Send
```
func (dc *DownstreamController) syncPod(stop chan struct{}) {
...
			if err := dc.messageLayer.Send(*msg); 
...
}

```

after send , cloudhub module will receive this msg by Context.
```
func (q *ChannelEventQueue) dispatchMessage() {
	for {
		msg, err := q.ctx.Receive("cloudhub")
		if err != nil {
			log.LOGGER.Infof("receive not Message format message")
			continue
		}
		resource := msg.Router.Resource
		tokens := strings.Split(resource, "/")
		numOfTokens := len(tokens)
		var nodeID string
		for i, token := range tokens {
			if token == "node" && i+1 < numOfTokens {
				nodeID = tokens[i+1]
				break
			}
		}
		if nodeID == "" {
			log.LOGGER.Warnf("node id is not found in the message")
			continue
		}
		rChannel, err := q.getRChannel(nodeID)
		if err != nil {
			log.LOGGER.Infof("fail to get dispatch channel for %s", nodeID)
			continue
		}
		event := model.MessageToEvent(&msg)
		select {
		case rChannel <- event:
		default:
		}
	}
}
```

but in this function,  it need get host channel , but  it can not get that host channel, because edge_core not started.
if we continue, and do nothing , we will lost that whole amount of information. only get the watched information after edge_core started.

sids-b:
@kadisi , till edge_core starts, the node will be in notReady state and hence scheduler will not schedule any event for that node. So I think there will not be any events for that node. Thoughts ?

kadisi:
@sids-b yes , scheduler will not schedule any event for that node.  but if we keep the old logic,，edge_core only get the watched information after  edge_core starts.    edge_core can not get enough pod information , and can not clean up the dirty pod

sids-b:
@kadisi , to get cleanup pod information, we need something like reliable message delivery in cloud. Also one more reason I am suggesting to not make a channel if node is not connected is we may not know for how long a node is disconnected, so the channel buffer may get full and we may end up blocked ? Thoughts ?

CindyXing:
> @CindyXing , in cloud, controller will create DownstreamController which create many manager, such as PodManager, ConfigManager. In the case of PodManager, it will get realEvents by SharedInformer and ResourceEventHander
> 
> ```
> // NewPodManager create PodManager from config
> func NewPodManager(kubeClient *kubernetes.Clientset, namespace string) (*PodManager, error) {
> 	lw := cache.NewListWatchFromClient(kubeClient.CoreV1().RESTClient(), "pods", namespace, fields.Everything())
> 	realEvents := make(chan watch.Event)
> 	mergedEvents := make(chan watch.Event)
> 	rh := NewCommonResourceEventHandler(realEvents)
> 	si := cache.NewSharedInformer(lw, &v1.Pod{}, 0)
> 	si.AddEventHandler(rh)
> 
> 	pm := &PodManager{realEvents: realEvents, mergedEvents: mergedEvents}
> 
> 	stopNever := make(chan struct{})
> 	go si.Run(stopNever)
> 	go pm.merge()
> 
> 	return pm, nil
> }
> ```
> 
> when edgecontroller start, it will get all pod events(first list all pods, and then watch them) , and merge them into `mergedEvents channel`.
> 
> and in `syncPod` function, it will Consuming these events。and send them by messageLayer.Send
> 
> ```
> func (dc *DownstreamController) syncPod(stop chan struct{}) {
> ...
> 			if err := dc.messageLayer.Send(*msg); 
> ...
> }
> ```
> 
> after send , cloudhub module will receive this msg by Context.
> 
> ```
> func (q *ChannelEventQueue) dispatchMessage() {
> 	for {
> 		msg, err := q.ctx.Receive("cloudhub")
> 		if err != nil {
> 			log.LOGGER.Infof("receive not Message format message")
> 			continue
> 		}
> 		resource := msg.Router.Resource
> 		tokens := strings.Split(resource, "/")
> 		numOfTokens := len(tokens)
> 		var nodeID string
> 		for i, token := range tokens {
> 			if token == "node" && i+1 < numOfTokens {
> 				nodeID = tokens[i+1]
> 				break
> 			}
> 		}
> 		if nodeID == "" {
> 			log.LOGGER.Warnf("node id is not found in the message")
> 			continue
> 		}
> 		rChannel, err := q.getRChannel(nodeID)
> 		if err != nil {
> 			log.LOGGER.Infof("fail to get dispatch channel for %s", nodeID)
> 			continue
> 		}
> 		event := model.MessageToEvent(&msg)
> 		select {
> 		case rChannel <- event:
> 		default:
> 		}
> 	}
> }
> ```
> 
> but in this function, it need get host channel , but it can not get that host channel, because edge_core not started.
> if we continue, and do nothing , we will lost that whole amount of information. only get the watched information after edge_core started.

Thanks @kadisi for the details. This is a great scenario and can be a common one. My thought is that after edgecore is started, send something to edgecontroller and be able to get the delta; afterwards is the watch. It can be timestamp of last successful sync. 

In general, we want to make sure edge and cloud in sync and control the message load between.

CindyXing:
> @kadisi , to get cleanup pod information, we need something like reliable message delivery in cloud. Also one more reason I am suggesting to not make a channel only for cleanup is if node is not connected is we may not know for how long a node is disconnected, so the channel buffer may get full and we may end up blocked ? Thoughts ?

Agree with @sids-b that the problem is how we can ensure the reliable message delivery between cloud and edge; and be in sync. Let's address the big one than target specifics.

kadisi:
> This is a great scenario and can be a common one. My thought is that after edgecore is started, send something to edgecontroller and be able to get the delta; afterwards is the watch. It can be timestamp of last successful sync.

@sids-b @CindyXing  you are right,   as @CindyXing  said, when edgecore is started, she need tell edgecontroller that she has started and she need get the detail information about that host.

CindyXing:
> > This is a great scenario and can be a common one. My thought is that after edgecore is started, send something to edgecontroller and be able to get the delta; afterwards is the watch. It can be timestamp of last successful sync.
> 
> @sids-b @CindyXing you are right, as @CindyXing said, when edgecore is started, she need tell edgecontroller that she has started and she need get the detail information about that host.

@kadisi The scenario is valid. I think we all agree. It is just the current approach needs to be discussed. Do you attend the bi-weekly meeting? We can chat there or through github.

kadisi:
@CindyXing  is bi-weekly meeting on Wednesday at 11:00 ? i think i am ok

kadisi:
@CindyXing @sids-b i think we need to cache these changing events on cloudhub side.
in cloudhub, we store change events in `channelPool` 
```
// ChannelEventQueue is the channel implementation of EventQueue
type ChannelEventQueue struct {
	ctx         *context.Context
	channelPool sync.Map
}
```

but `channelPool` is map, it`s value is no-buffer-channel,   we need to change this channel to a queue to catch this changeing event.

 and this queue need support upstate event function 。 if we receive a new event which has the same attributes(Group, Source,UserGroup, ID, ParenetID) as an event int he queue. we only update the event in the queue.  If the new event is different from all the event properties in the queue, we need push it to the queue.

PR #475, opened 2019-05-08
Changed exit status bluetooth mapper on bad configuration and Fixed a typo in bluetooth mapper README

### LINKED ISSUE #474: Change exit code of bluetooth mapper when it exits due to wrong configuration
**What happened**:
Bluetooth mapper container indicates exit status 0 when mapper exits due to bad configuration (like config map volume not containing crct config file).
 
It would make it easier for the user to debug issues if the exit status was 1 instead of 0 when the mapper exits due to wrong  configuration

Run bluetooth mapper without the right config map

sujithsimon22:
**What type of PR is this?**
 /kind bug

This PR fixes a typo in bluetooth mapper README and ensures that the mapper returns status 1 instead of status report when the program exits due to bad configuration, which will make it easier for user to debug incase something goes wrong.
 
Fixes #474


sids-b:
cc @rohitsardesai83 for review

PR #479, opened 2019-05-09
Device CRD: remove configmap when no device is bound with

### LINKED ISSUE #478: configmap should be cleaned when no device is bound with
**What happened**:
Currently, the configmap will be kept even there is no device bound with, this will cause many useless leftover configmap, which has an empty profile with it.

Those configmap mentioned above should be clean up.

- create a device
- delete a device
- list the configmap and check the content

**Environment**:
- KubeEdge version: latest
- Hardware configuration: ALL
- OS (e.g. from /etc/os-release): UBUTNU16.04
- Kernel (e.g. `uname -a`):
- Others:

chendave:
Fix #478

**What type of PR is this?**

/kind bug

Fixes #478

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -584,6 +584,16 @@ func (dc *DownstreamController) deleteFromConfigMap(device *v1alpha1.Device) {
 		}
 
 		dc.deleteFromDeviceProfile(device, nodeConfigMap)
+
+		// no device is bound to the configMap, then remove the configMap directly.
+		if nodeConfigMap.Data[DeviceProfileJSON] == "{}" {
+			deleteOptions := &metav1.DeleteOptions{}
+			dc.kubeClient.CoreV1().ConfigMaps(device.Namespace).Delete(nodeConfigMap.Name, deleteOptions)
+			// remove from cache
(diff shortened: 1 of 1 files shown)


m1093782566:
cc @rohitsardesai83

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -586,6 +586,15 @@ func (dc *DownstreamController) deleteFromConfigMap(device *v1alpha1.Device) {
 		dc.deleteFromDeviceProfile(device, nodeConfigMap)
 		// store new config map
 		dc.configMapManager.ConfigMap.Store(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0], nodeConfigMap)
+
+		// no device is bound to the configMap, then remove the configMap directly.
```
sids-b: Can we move line line 588 below this if to reduce 1 extra store operation before delete ?

sids-b:
@chendave , would you like to join KubeEdge as a collaborator ?

cloud/pkg/devicecontroller/controller/downstream.go
chendave: sure, will do later.

chendave:
@sids-b , that would be great!  it's my honor!

sids-b:
> @sids-b , that would be great! it's my honor!

Thanks for your contributions. Looking forward for active contributions :) 
@kevin-wangzefeng will send you invite soon

PR #490, opened 2019-05-09
Fix Configuration keys of controller and deviceController

sids-b:
**What type of PR is this?**
/kind bug

Fix the configurations keys. Controller and deviceController related configurations should be grouped below them. Maintain similar structure in controller and deviceController


sids-b:
@pavan187 , @edisonxiang , the bot should merge after we have 2 approvals.

PR #509, opened 2019-05-14
Adding containername as a flag

### LINKED ISSUE #507: ConatinerName should be given as a flag in run_daemon.sh
IssueContainerName
Here by default the name of container is always edgecore(github.com/kubeedge/kubeedge/build/edge/run_daemon.sh). The script should be able to take container name as a flag

**Why is this needed**:
This is required to run multiple containerized edgcores in a single host.

lidiyag:
**What type of PR is this?**
kind feature
This PR takes modifies the script to take containername as a flag
Fixes #507

### DIFF
--- a/build/edge/run_daemon.sh
+++ b/build/edge/run_daemon.sh
@@ -163,6 +163,7 @@ docker_only_run_edge(){
     edgename=${edgename:-$(hostname)}
     edgehubWebsocketUrl=wss://${cloudhub}/e632aba927ea4ac2b575ec1603d56f10/${edgename}/events 
     image=${image:-"kubeedge/edgecore:latest"}
+    containername=${containername:-"edgecore"}
 
     docker run -d --name ${containername} --restart always \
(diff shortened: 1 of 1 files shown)


lidiyag:
cc @rohitsardesai83 @sids-b @edisonxiang @pavan187

build/edge/run_daemon.sh
```diff
@@ -164,7 +164,7 @@ docker_only_run_edge(){
     edgehubWebsocketUrl=wss://${cloudhub}/e632aba927ea4ac2b575ec1603d56f10/${edgename}/events 
     image=${image:-"kubeedge/edgecore:latest"}
 
-    docker run -d --name edgecore --restart always \
+    docker run -d --name ${containername} --restart always \
```
m1093782566: What's the default conatinername?

lidiyag: edgecore

sids-b: Please add a default name like we have added for other parameters. edgecore can be default name as it was earlier

PR #510, opened 2019-05-14
adjust hubtest params and fix bugs

edisonxiang:
**What type of PR is this?**
/kind test

This pr includes the following modifications:
* adjust the parameters in the hubtest so that it can be used to test under 500 edge nodes
* add logs so that it can be more clear to see what happen in the perf test
* fix the bug in the controller stub which may cause the go channel is stopped.

### DIFF
--- a/tests/performance/common/common.go
+++ b/tests/performance/common/common.go
@@ -77,8 +77,8 @@ func HandleEdgeDeployment(cloudhub, depHandler, nodeHandler, cmHandler, imgURL, 
 	//Create edgecore configMaps based on the users choice of edgecore deployment.
 	for i := 0; i < numOfNodes; i++ {
 		nodeName := "perf-node-" + utils.GetRandomString(10)
-		nodeSelector := "node-" + utils.GetRandomString(3)
-		configmap := "edgecore-configmap-" + utils.GetRandomString(3)
+		nodeSelector := "node-" + utils.GetRandomString(5)
+		configmap := "edgecore-configmap-" + utils.GetRandomString(5)
 		//Register EdgeNodes to K8s Master
 		go utils.RegisterNodeToMaster(nodeName, nodeHandler, nodeSelector)
 		cmd := exec.Command("bash", "-x", "scripts/update_configmap.sh", "create_edge_config", nodeName, cloudhub, configmap)
@@ -113,7 +113,7 @@ func HandleEdgeDeployment(cloudhub, depHandler, nodeHandler, cmHandler, imgURL, 
 			}
 		}
 		return count
-	}, "60s", "2s").Should(Equal(numOfNodes), "Nodes register to the k8s master is unsuccessfull !!")
+	}, "1200s", "2s").Should(Equal(numOfNodes), "Nodes register to the k8s master is unsuccessfull !!")
 
 	return podlist
 }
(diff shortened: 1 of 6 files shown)


tests/performance/common/common.go
```diff
@@ -113,7 +113,7 @@ func HandleEdgeDeployment(cloudhub, depHandler, nodeHandler, cmHandler, imgURL,
 			}
 		}
 		return count
-	}, "60s", "2s").Should(Equal(numOfNodes), "Nodes register to the k8s master is unsuccessfull !!")
+	}, "1200s", "2s").Should(Equal(numOfNodes), "Nodes register to the k8s master is unsuccessfull !!")
```
sids-b: 1200s = 20mins, The test will be stuck for 20 mins if node registration has failed. can we keep short time duration ?

tests/performance/scripts/runperf.sh
```diff
@@ -32,7 +32,7 @@ cat >config.json<<END
 {
         "image_url": ["nginx", "hello-world"],
         "k8smasterforkubeedge": "$K8SMasterForKubeEdge",
-        "node_num": 10,
+        "node_num": 500,
```
sids-b: It will be great that we will be running our performance tests against 500 edge nodes :)

tests/performance/common/common.go
edisonxiang: @sids-b I tested about 50 edge nodes. it will take more than 240s.
if we test 500 edge nodes, i guess 1200s is better so that it will not be timeout:(

tests/performance/scripts/runperf.sh
edisonxiang: Thanks @sids-b The testing is in progress

tests/performance/common/common.go
sids-b: OK, Thanks :)

sids-b:
ping @rohitsardesai83 @m1093782566 @pavan187 for review

edisonxiang:
@sids-b Thanks very much

tests/performance/common/common.go
```diff
@@ -283,10 +300,12 @@ func SendHttpRequest(method, reqApi string, body io.Reader) (error, *http.Respon
 	req.Header.Set("Content-Type", "application/json")
 	t := time.Now()
 	resp, err = client.Do(req)
-	utils.Info("%s %s %v in %v", req.Method, req.URL, resp.Status, time.Now().Sub(t))
 	if err != nil {
 		utils.Failf("HTTP request is failed :%v", err)
 		return err, resp
 	}
+	if resp != nil {
+		utils.Info("%s %s %v in %v", req.Method, req.URL, resp.Status, time.Now().Sub(t))
```
rohitsardesai83: can we improve this message ?

edisonxiang: Thank @rohitsardesai83 your suggestion is?

edisonxiang: `GET http://172.16.0.18:8080/api/v1/nodes/perf-node-h33cjb8q9v 200 OK in 1.345807ms`

Currently it is like the above message:)

tests/performance/scripts/runperf.sh
pavan187: can we keep this as 10 itself ?

edisonxiang: Let's keep it 500. It is the default max nodes,  not the running nodes:)

PR #514, opened 2019-05-14
Enhancements to KubeEdge installer

samy2019:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
1. Separate flags for "kubeedge join" command for controller and k8s apiserver
2. Creation of cloud and edge node configuration files dynamically using json and structures, so that in future if any new configuratino field is added then it can be taken care efficiently.
3. Uses "curl" to add KubeEdge edge nodes to K8S api-server, rather than kubectl. It helps in reduce memory footprint in edge side.
4. A new flag is added for "kubeedge join" command to set KubeEdge edge node ID by user. If the user doesn't uses this flag, then the installer assigns a unique ID and displays it over the terminal after successful addition to api-server
5. K8S Apiserver flag made optional. User shall be able to add a KubeEdge edge node either from cloud (using kubectl) or from kubeedge join's "--k8sserverip" flag

Why is this needed:
Above listed issues help in customer usage enhancements and reducing memory foot print.

**Which issue(s) this PR fixes**:
Fixes #513 

**Special notes for your reviewer**:
README.md will be updated soon with commands and how to use.

PR #535, opened 2019-05-17
Enhancements to KubeEdge installer

### LINKED ISSUE #513: Enhancements to kubeEdge installer
1. It is better to have separate Flag options controller and k8s apiserver
2. Creation of cloud and edge node configuration files should be dynamic using json and structures
3. Use Curl to add KubeEdge edge nodes to K8S api-server, rather than kubectl. It helps in reduce memory footprint in edge side.
4. Flag to set KubeEdge edge node ID by user, if not provided, then create a unique ID and display it over the terminal.
5. K8S Apiserver flag should not be mandatory. User shall be able to add a KubeEdge edge node either from cloud (using kubectl) or from kubeedge join's "--k8sserverip" flag

**Why is this needed**:
Above listed issues help in customer usage enhancements and reducing memory foot print.

samy2019:
**What type of PR is this?**
/kind feature

1. Separate flags for "kubeedge join" command for controller and k8s apiserver
2. Creation of cloud and edge node configuration files dynamically using json and structures, so that in future if any new configuratino field is added then it can be taken care efficiently.
3. Uses "curl" to add KubeEdge edge nodes to K8S api-server, rather than kubectl. It helps in reduce memory footprint in edge side.
4. A new flag is added for "kubeedge join" command to set KubeEdge edge node ID by user. If the user doesn't uses this flag, then the installer assigns a unique ID and displays it over the terminal after successful addition to api-server
5. K8S Apiserver flag made optional. User shall be able to add a KubeEdge edge node either from cloud (using kubectl) or from kubeedge join's "--k8sserverip" flag

Why is this needed:
Above listed issues help in customer usage enhancements and reducing memory foot print.

Fixes #513 

The code is already been reviewed in PR #514. Had some issues while squashing he commits. Hence raising the PR again.
README.md will be updated soon with commands and how to use.


samy2019:
@m1093782566, @sids-b, @rohitsardesai83, please check and take it forward

PR #536, opened 2019-05-17
support quic and latency in hubtest

edisonxiang:
**What type of PR is this?**
/kind test

* support quic in hubtest
   quic is passing the data by bytes[] and websocket is passing the data by json
* support latency calculation in hubtest
   support 50%, 90%, 99% and 100% throughputs latency
* remove the Resource Limits for EdgeController
   avoid the perf test bottleneck

### DIFF
--- a/tests/e2e/utils/common.go
+++ b/tests/e2e/utils/common.go
@@ -154,10 +154,6 @@ func cloudcoreDeploymentSpec(imgURL, configmap string, replicas int) *apps.Deplo
 								v1.ResourceName(v1.ResourceCPU):    resource.MustParse("100m"),
 								v1.ResourceName(v1.ResourceMemory): resource.MustParse("512Mi"),
 							},
-							Limits: v1.ResourceList{
-								v1.ResourceName(v1.ResourceCPU):    resource.MustParse("200m"),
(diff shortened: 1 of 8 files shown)


tests/e2e/utils/common.go
```diff
@@ -144,10 +144,10 @@ func cloudcoreDeploymentSpec(imgURL, configmap string, replicas int) *apps.Deplo
 								v1.ResourceName(v1.ResourceCPU):    resource.MustParse("100m"),
 								v1.ResourceName(v1.ResourceMemory): resource.MustParse("512Mi"),
 							},
-							Limits: v1.ResourceList{
+							/*Limits: v1.ResourceList{
```
sids-b: This is a good change. EdgeController is a cloud component and may not resource constraint environment

sids-b:
ping @m1093782566 @pavan187 for review

tests/e2e/utils/common.go
m1093782566: If we don't need these test scenarios, why not delete them?

edisonxiang: OK. will update later.

PR #537, opened 2019-05-17
Supports automatic deletion of edge containers if their pods do not exist in the cloud

### LINKED ISSUE #332: Supports automatic deletion of edge containers if their pods do not exist in the cloud.

Supports automatic deletion of edge containers if their pods do not exist in the cloud.

**Why is this needed**:

When we delete pods in the cloud, if the network is disconnected, the containers on the edge will not be deleted.This will lead to inconsistencies between the cloud and the edge, even if the network is connected.That will create a series of problems, such as the inability to  update the services in the cloud before logging in  edge nodes to remove the old containers.

So I think it's necessary to support the ability to automatically delete edge containers when pod not found in the cloud.

the code in github.com/kubeedge/kubeedge/cloud/edgecontroller/pkg/controller/controller

```GO
func (uc *UpstreamController) updatePodStatus(stop chan struct{}) {
	running := true
	for running {
		select {
		case msg := <-uc.podStatusChan:
			log.LOGGER.Infof("message: %s, operation is: %s, and resource is: %s", msg.GetID(), msg.GetOperation(), msg.GetResource())

			namespace, podStatuses := uc.unmarshalPodStatusMessage(msg)
			switch msg.GetOperation() {
			case model.UpdateOperation:
				for _, podStatus := range podStatuses {
					getPod, err := uc.kubeClient.CoreV1().Pods(namespace).Get(podStatus.Name, metaV1.GetOptions{})
					if errors.IsNotFound(err) {
						log.LOGGER.Warnf("message: %s, pod not found, namespace: %s, name: %s", msg.GetID(), namespace, podStatus.Name)
						continue
					}
					if err != nil {
						log.LOGGER.Warnf("message: %s, pod is nil, namespace: %s, name: %s, error: %s", msg.GetID(), namespace, podStatus.Name, err)
						continue
					}
          
          //……
}
```

shouhong:
**What type of PR is this?**
/kind bug

Supports automatic deletion of edge containers if their pods do not exist in the cloud.

Fixes #332

### DIFF
--- a/cloud/pkg/controller/controller/upstream.go
+++ b/cloud/pkg/controller/controller/upstream.go
@@ -173,6 +173,29 @@ func (uc *UpstreamController) updatePodStatus(stop chan struct{}) {
 					getPod, err := uc.kubeClient.CoreV1().Pods(namespace).Get(podStatus.Name, metaV1.GetOptions{})
 					if errors.IsNotFound(err) {
 						log.LOGGER.Warnf("message: %s, pod not found, namespace: %s, name: %s", msg.GetID(), namespace, podStatus.Name)
+
+						// Send request to delete this pod on edge side
+						delMsg := model.NewMessage("")
+						nodeID, err := messagelayer.GetNodeID(msg)
+						if err != nil {
+							log.LOGGER.Warnf("Get node ID failed with error: %s", err)
+							continue
+						}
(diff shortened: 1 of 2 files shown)


sids-b:
I think we can have this as a temporary fix. The correct solution to this would be reliable message delivery. But there maybe long offline scenarios where apiserver may delete a pod after grace time. Hence have this solution makes sense.

sids-b:
ping @m1093782566 @edisonxiang  for review

PR #542, opened 2019-05-19
Add keepalive check logic for quic protocol

### LINKED ISSUE #528: Pod Deployment failed when EdgeCore re-initialized to use Quic from websocket
**What happened**:
Pod Deployment is failing when re-initializing edgecore to use QUIC protocol from Websocket protocol.

Pod deployment should be successful.
1. Setup Kubeedge Env
2. Register EdgeNode to K8s Master
2. Do the cloudcore deployment.
3. Do the Edgecore deployment with protocol: websocket (in edge.yaml)
4. Check is node is Ready.
5. Chnage the config(configMap) to  protocol:quic (node name should be same)
6. delete edgecore pod(not deployment), pod should be deleted and new pod should bring-up by k8s deployment object.
7. check node is in Ready state
8. do the application deployment on kubeedge edgenode.
9.Check the deployment is success and pod is in Running state.
10. Repeat steps 6 - 9 again 
11. check application deployment is successfull 2nd time.
**Actual result:**
2nd Application deployment is not success.
**Expected result:**
Application deployment should be successfull for all the time.

This tests are performed by deploying cloudcore and edgecores as k8s deployment.

**Environment**:
- KubeEdge version: v0.3.0-beta.0
- Hardware configuration:NA
- OS (e.g. from /etc/os-release):Ubuntu16.04
- Kernel (e.g. `uname -a`):NA
- Others:NA

shouhong:
**What type of PR is this?**
/kind bug

Add keepalive check logic for quic protocol

Fixes #528

The reason of this issue is edge node disconnection event cannot be detected by cloudhub when using quic protocol. So the obsolete connection information is not cleaned up in cloudhub. When the edge node starts again, the new connection cannot be properly setup because the old connection information is still there.

### DIFF
--- a/cloud/pkg/cloudhub/handler/quichandler.go
+++ b/cloud/pkg/cloudhub/handler/quichandler.go
@@ -1,6 +1,8 @@
 package handler
 
 import (
+	"time"
+
 	bhLog "github.com/kubeedge/beehive/pkg/common/log"
 	hubio "github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/io"
 	emodel "github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/model"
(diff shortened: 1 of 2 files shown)


sids-b:
@shouhong , In websocket we can detect the disconnect event. So when moving for websocket to quic should not be a problem right ? Issue #528 is for moving from websocket to quic

sids-b:
cc @kexun for review

shouhong:
> @shouhong , In websocket we can detect the disconnect event. So when moving for websocket to quic should not be a problem right ? Issue #528 is for moving from websocket to quic

@sids-b Issue #528 title maybe a little confusing. In the reproducing steps, it repeats steps 6 - 9, both rounds of repeated steps are using quic.

kexun:
it looks good.  i am confused with the issure528. it has nothing to do with websocket. This is quic's problem  for not detect the disconnect event.

sids-b:
Ohh.... I get it now. The title is little confusing but the issue which is solved here is corerct. Thanks for pointing that out @shouhong @kexun

PR #552, opened 2019-05-22
Print message content as string instead of uint8 array

### LINKED ISSUE #547: logs are inconsistent in both edgecontroller and edgecore modules
**What happened**:
Logs generation in both edgecontroller and edgecore are inconsistent(prints are in byte format).

logs should be consistent across the modules

1. Deploy applications on EdgeEdge node and check the both edgecontroller and edge_core logs.
NA
**Environment**: GCP
- KubeEdge version: v0.3.0-beta.0
- Hardware configuration: NA
- OS (e.g. from /etc/os-release):ubuntu 16.04
- Kernel (e.g. `uname -a`):NA
- Others:NA

**Log Attachment**
[logs.zip](https://github.com/kubeedge/kubeedge/files/3197531/logs.zip)

shouhong:
**What type of PR is this?**
/kind bug

Print message content as string instead of uint8 array

Fixes #547

### DIFF
--- a/edge/pkg/edgehub/controller.go
+++ b/edge/pkg/edgehub/controller.go
@@ -196,7 +196,7 @@ func (ehc *Controller) routeToEdge() {
 			return
 		}
 
-		log.LOGGER.Infof("received msg from cloud-hub:%#v", message)
+		log.LOGGER.Infof("received msg from cloud-hub:%+v", message)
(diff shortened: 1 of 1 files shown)


m1093782566:
Nice catch! Thanks @shouhong 

/lgtm

/approve

PR #554, opened 2019-05-22
Consolidate the kube config at controller

### LINKED ISSUE #525: Consolidate the kube config at controller
**What happened**:
In both edgecontroller and edgesite controller code, there are many fields to define the kubernetes configuration. 

This is code cleanup/enhancement. We should:
1. wrap all KubeX fields into a struct
2. wrap all default values into several structs
3. use struct copy / deep copy to set default values for KubeX fields, and then rewrite those user indicated ones.

**Environment**:
- KubeEdge version:
- Hardware configuration:
- Kernel (e.g. `uname -a`):
- Others:

shouhong:
**What type of PR is this?**
/kind cleanup

Consolidate the kube config at controller

Fixes #525


m1093782566:
I like this change.

cc @rohitsardesai83 @sids-b for review.

PR #557, opened 2019-05-27
KubeEdge Installer: Adding support to establish secure connectivity b…

### LINKED ISSUE #520: KubeEdge Installer: Adding support to establish secure connectivity between edgecontroller and K8S api-server
Insecure mode of connectivity between edgecontroller and K8S api-server is supported now so that end user can start using KubeEdge installer and explore KubeEdge. 
Now we should also support "Secure" connectivity. 

**Why is this needed**:
Secure connectivity is always preferred than insecure mode. Hence it is required to be supported.

kadisi:
…etween edgecontroller and K8S api-server

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind feature

Fixes #520 

### DIFF
--- a/keadm/app/cmd/util/cloudinstaller.go
+++ b/keadm/app/cmd/util/cloudinstaller.go
@@ -58,7 +58,7 @@ func (cu *KubeCloudInstTool) InstallTools() error {
 	}
 
 	//Create controller.yaml
-	if err = common.WriteControllerYamlFile(KubeEdgeControllerYaml); err != nil {
+	if err = common.WriteControllerYamlFile(KubeEdgeControllerYaml, cu.KubeConfig); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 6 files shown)


keadm/app/cmd/util/cloudinstaller.go
```diff
@@ -18,6 +18,7 @@ import (
 //It implements ToolsInstaller interface
 type KubeCloudInstTool struct {
 	Common
+	KubeConfig string
```
m1093782566: Can't we put the `KubeConfig` into `Common`?

sids-b:
Looks good to me. Can merge post @m1093782566 review comments addressed.

kadisi:
@sids-b PTAL

PR #558, opened 2019-05-27
Mismatch in access mode names

### LINKED ISSUE #545: Action not performing as expected in Bluetooth Mapper
**What happened**:
While using bluetooth mapper with CC2650-sensortag , it is observed that the Action Manager is not performing action as specified in the config file. 

It is expected to perform action as specified in the config file. 

Run bluetooth mapper and test using CC3650-SensorTag

It was observed that the struct "Action" was not getting proper values upon running the mapper.

**Environment**:
- KubeEdge version:
- Hardware configuration:
- Kernel (e.g. `uname -a`):
- Others:

sujithsimon22:
**What type of PR is this?**
 /kind bug

This PR fixes a mismatch in the Access mode names due to which the action manager is not performing as expected in the bluetooth mapper

Fixes #545

### DIFF
--- a/device/bluetooth_mapper/configuration/config.go
+++ b/device/bluetooth_mapper/configuration/config.go
@@ -43,6 +43,8 @@ var Config *BLEConfig
 //Blutooth protocol name
 const (
 	Protocol_Name string = "BLUETOOTH"
+	READWRITE     string = "ReadWrite"
+	READ          string = "Read"
 )
 
 //BLEConfig is the main structure that stores the configuration information read from both the config file as well as the config map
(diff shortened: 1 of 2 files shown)


device/bluetooth_mapper/configuration/config.go
```diff
@@ -182,7 +184,7 @@ func (b *BLEConfig) Load() error {
 			if strings.ToUpper(deviceModel.Name) == strings.ToUpper(b.Device.Name) {
 				for _, property := range deviceModel.Properties {
 					if strings.ToUpper(property.Name) == strings.ToUpper(actionConfig.PropertyName) {
-						if strings.ToUpper(property.AccessMode) == "RW" {
+						if strings.ToUpper(property.AccessMode) == strings.ToUpper(READWRITE){
```
rohitsardesai83: can't we just compare without comparing to upper case ? since our schema is defined , users are expected to provide the correct enum values.

sujithsimon22: Removed case insensitive check and replaced it with a case sensitive one. Going forward when we have API validation we can prevent the users from accidentally giving the wrong case for enum values, thus eliminating the need for a case insensitive check. Thanks

sujithsimon22:
cc @m1093782566

m1093782566:
Let's get it in.

/approve

PR #561, opened 2019-05-27
Added nil pointer handling in watcher,go

### LINKED ISSUE #560: Bluetooth mapper panics when desired state values are not set before starting the watcher
**What happened**:
A panic is caused when the watcher component of the  bluetooth mapper is started without setting the desired state of the device is set

The watcher should not panic and should wait for the desired state to be set .

1. Start Kubeedge Cloud and Edge
2. Create device CRD for bluetooth device without including desired state
3. Start bluetooth mapper with Watcher enabled 

**Environment**:
- KubeEdge version: v0.3 beta
- Hardware configuration: 16GB/500GB
- OS (e.g. from /etc/os-release): Ubuntu 18.04

sujithsimon22:
**What type of PR is this?**
 /kind bug

This PR fixes the issue of bluetooth mapper panicing when watcher is started without creating desired state for the device

Fixes #560 

### DIFF
--- a/device/bluetooth_mapper/controller/controller.go
+++ b/device/bluetooth_mapper/controller/controller.go
@@ -79,7 +79,7 @@ func (c *ControllerConfig) Start() {
 
 	<-watcher.DeviceConnected
 	for _, action := range c.ActionManager.Actions {
-		if action.PerformImmediately{
+		if action.PerformImmediately {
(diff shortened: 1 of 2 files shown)


m1093782566:
/lgtm

leave approve to @sids-b @edisonxiang

PR #565, opened 2019-05-27
using extra function to generate csr and certificate

### LINKED ISSUE #564: gencerten.sh generate key and csr, generate certificate from csr
 It should be possible to use gencerts.sh to generate only an csr. And generate a certificate from the csr + CA certificate + CA key

**Why is this needed**:
This can be used to generate the certificates, when cloud and edge part are on different systems. In the future the typical usage is not that the edge and the cloud are on the same system so we should support this feature.

subpathdev:
update certgen.sh tool to have seperate function to generate csr and certificates
**What type of PR is this?**
/kind feature

support generating cert and csr on different systems

Fixes #564

### DIFF
--- a/build/tools/certgen.sh
+++ b/build/tools/certgen.sh
@@ -26,14 +26,24 @@ ensureFolder() {
     fi
 }
 
+genCsr() {
+    local name=$1
+    openssl genrsa -out ${certPath}/${name}.key 2048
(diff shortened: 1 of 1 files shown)


m1093782566:
/lgtm

leave approve to @sids-b @edisonxiang

PR #600, opened 2019-05-29
Correcting nits in sample device model for bluetooth

sujithsimon22:
**What type of PR is this?**
 /kind documentation

This PR corrects the problem of not placing string keys of the map within quotes in the sample CRD device-model that has been given for bluetooth

### DIFF
--- a/build/crd-samples/devices/CC2650-device-model.yaml
+++ b/build/crd-samples/devices/CC2650-device-model.yaml
@@ -57,8 +57,8 @@ spec:
     bluetooth:
      characteristicUUID: f000aa0204514000b000000000000000
      dataWrite:
-      ON: [1]
-      OFF: [0]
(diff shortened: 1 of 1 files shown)


build/crd-samples/devices/CC2650-device-model.yaml
```diff
@@ -57,8 +57,8 @@ spec:
     bluetooth:
      characteristicUUID: f000aa0204514000b000000000000000
      dataWrite:
-      ON: [1]
-      OFF: [0]
+      "ON": [1]
```
m1093782566: Sorry I just forget what does the [1] [2] [3] represent? So I think it's worth giving some comments here and below?

sids-b: +1

sujithsimon22: Here [1] , [2], [3] etc represents the []byte value to be written into the device for the corresponding key value of the property. Have given a brief explanation in the comment. Thanks

PR #602, opened 2019-05-29
Added mapper component to the kubeedge architecture diagram

sujithsimon22:
**What type of PR is this?**
 /kind documentation

This PR introduces the mapper component into the architecture diagram of kubeedge. This would give users a basic idea regarding the placement of the mapper


m1093782566:
/lgtm

cc @sids-b

PR #620, opened 2019-05-30
Added support to update the protocol configmap

### LINKED ISSUE #619: Update in protocol config should update the configmap
**What happened**:
Update the protocol config of a device, it didn't update in configmap

Protocol config should be updated

1) Create a device instance
2) Update the protocol config
3) check configmap

**Environment**:
- KubeEdge version: v0.3.0

sids-b:
**What type of PR is this?**
 /kind bug

This PR adds support to update the protocol in configmap if protocol is updated in devceinstance

Fixes #619

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -438,6 +438,101 @@ func isNodeSelectorUpdated(old *v1.NodeSelector, new *v1.NodeSelector) bool {
 	return !reflect.DeepEqual(old.NodeSelectorTerms, new.NodeSelectorTerms)
 }
 
+// isProtocolConfigUpdated checks if protocol is updated
+func isProtocolConfigUpdated(old *v1alpha1.ProtocolConfig, new *v1alpha1.ProtocolConfig) bool {
+	return !reflect.DeepEqual(old, new)
+}
+
+// updateProtocolInConfigMap updates the protocol in the deviceProfile in configmap
+func (dc *DownstreamController) updateProtocolInConfigMap(device *v1alpha1.Device) {
+	if len(device.Spec.NodeSelector.NodeSelectorTerms) != 0 && len(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions) != 0 && len(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values) != 0 {
+		configMap, ok := dc.configMapManager.ConfigMap.Load(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0])
+		if ok {
+			nodeConfigMap, ok := configMap.(*v1.ConfigMap)
+			if !ok {
+				log.LOGGER.Errorf("Failed to assert to configmap")
+				return
+			}
+			dp, ok := nodeConfigMap.Data[DeviceProfileJSON]
+			if !ok || dp == "{}" {
+				// This case should never be hit as we delete empty configmaps
+				log.LOGGER.Errorf("Failed to get deviceProfile from configmap data or deviceProfile is empty")
+				return
+			}
+
+			deviceProfile := &types.DeviceProfile{}
+			if err := json.Unmarshal([]byte(dp), deviceProfile); err != nil {
+				log.LOGGER.Errorf("Failed to unmarshal due to error: %v", err)
+				return
+			}
+			var oldProtocol string
+			for _, devInst := range deviceProfile.DeviceInstances {
+				if device.Name == devInst.Name {
+					oldProtocol = devInst.Protocol
+					break
+				}
+			}
+
+			// delete the old protocol
+			for i, ptcl := range deviceProfile.Protocols {
+				if ptcl.Name == oldProtocol {
+					deviceProfile.Protocols = append(deviceProfile.Protocols[:i], deviceProfile.Protocols[i+1:]...)
+					break
+				}
+			}
+
+			// add new protocol
+			deviceProtocol := &types.Protocol{}
+			var protocol string
+			if device.Spec.Protocol.OpcUA != nil {
+				protocol = OPCUA + "-" + device.Name
+				deviceProtocol.Name = protocol
+				deviceProtocol.Protocol = OPCUA
+				deviceProtocol.ProtocolConfig = device.Spec.Protocol.OpcUA
+			} else if device.Spec.Protocol.Modbus != nil && device.Spec.Protocol.Modbus.RTU != nil {
+				protocol = ModbusRTU + "-" + device.Name
+				deviceProtocol.Name = protocol
+				deviceProtocol.Protocol = ModbusRTU
+				deviceProtocol.ProtocolConfig = device.Spec.Protocol.Modbus.RTU
+			} else if device.Spec.Protocol.Modbus != nil && device.Spec.Protocol.Modbus.TCP != nil {
+				protocol = ModbusTCP + "-" + device.Name
+				deviceProtocol.Name = protocol
+				deviceProtocol.Protocol = ModbusTCP
+				deviceProtocol.ProtocolConfig = device.Spec.Protocol.Modbus.TCP
+			} else if device.Spec.Protocol.Bluetooth != nil {
+				protocol = Bluetooth + "-" + device.Name
+				deviceProtocol.Name = protocol
+				deviceProtocol.Protocol = Bluetooth
+				deviceProtocol.ProtocolConfig = device.Spec.Protocol.Bluetooth
+			} else {
+				log.LOGGER.Warnf("Unsupported device protocol")
+			}
+
+			// update the protocol in deviceInstance
+			for _, devInst := range deviceProfile.DeviceInstances {
+				if device.Name == devInst.Name {
+					devInst.Protocol = protocol
+					break
+				}
+			}
+			deviceProfile.Protocols = append(deviceProfile.Protocols, deviceProtocol)
+
+			bytes, err := json.Marshal(deviceProfile)
+			if err != nil {
+				log.LOGGER.Errorf("Failed to marshal deviceprofile: %v", deviceProfile)
+				return
+			}
+			nodeConfigMap.Data[DeviceProfileJSON] = string(bytes)
+			// store new config map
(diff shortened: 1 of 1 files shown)


cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -438,6 +438,103 @@ func isNodeSelectorUpdated(old *v1.NodeSelector, new *v1.NodeSelector) bool {
...
+	if len(device.Spec.NodeSelector.NodeSelectorTerms) != 0 && len(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions) != 0 && len(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values) != 0 {
+		configMap, ok := dc.configMapManager.ConfigMap.Load(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0])
+		if ok {
+			nodeConfigMap, ok := configMap.(*v1.ConfigMap)
+			if !ok {
+				log.LOGGER.Errorf("Failed to assert to configmap")
+				return
+			}
+			dp, ok := nodeConfigMap.Data[DeviceProfileJSON]
+			if !ok || dp == "{}" {
```
samy2019: "{}" could be nil. If so, please change it.

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -438,6 +438,103 @@ func isNodeSelectorUpdated(old *v1.NodeSelector, new *v1.NodeSelector) bool {
...
+}
+
+// updateProtocolInConfigMap updates the protocol in the deviceProfile in configmap
+func (dc *DownstreamController) updateProtocolInConfigMap(device *v1alpha1.Device) {
+	if len(device.Spec.NodeSelector.NodeSelectorTerms) != 0 && len(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions) != 0 && len(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values) != 0 {
+		configMap, ok := dc.configMapManager.ConfigMap.Load(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0])
+		if ok {
+			nodeConfigMap, ok := configMap.(*v1.ConfigMap)
+			if !ok {
+				log.LOGGER.Errorf("Failed to assert to configmap")
```
samy2019: All error's should start with small letters. Like wise for all ErrorF() below

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -438,6 +438,103 @@ func isNodeSelectorUpdated(old *v1.NodeSelector, new *v1.NodeSelector) bool {
...
+				deviceProtocol.Name = protocol
+				deviceProtocol.Protocol = ModbusTCP
+				deviceProtocol.ProtocolConfig = device.Spec.Protocol.Modbus.TCP
+			} else if device.Spec.Protocol.Bluetooth != nil {
+				protocol = Bluetooth + "-" + device.Name
+				deviceProtocol.Name = protocol
+				deviceProtocol.Protocol = Bluetooth
+				deviceProtocol.ProtocolConfig = device.Spec.Protocol.Bluetooth
+			} else {
+				log.LOGGER.Warnf("Unsupported device protocol")
```
samy2019: Is there a check before this routine to add only supported devices? If yes, then above default case is fine. If not, then it should return here with possible rollover of any previous configuration done.

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: it is a string not a pointer. So it cannot be verified against nil

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: We follow the convention that all Log's start with capital letters and all errors start with small letters

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: There is no check before this routine. the protocol can be empty but it should not stop updation of configmap. It will just delete the old protocol config as it is not associated with the device anymore.

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -457,8 +554,10 @@ func (dc *DownstreamController) deviceUpdated(device *v1alpha1.Device) {
 				}
 				dc.deviceDeleted(deletedDevice)
 
+			} else if isProtocolConfigUpdated(&cachedDevice.Spec.Protocol, &device.Spec.Protocol) {
+				dc.updateProtocolInConfigMap(device)
```
m1093782566: A quick question, when an error occurs in updateProtocolInConfigmap, will the following procedure continue?

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -438,6 +438,103 @@ func isNodeSelectorUpdated(old *v1.NodeSelector, new *v1.NodeSelector) bool {
 	return !reflect.DeepEqual(old.NodeSelectorTerms, new.NodeSelectorTerms)
 }
 
+// isProtocolConfigUpdated checks if protocol is updated
+func isProtocolConfigUpdated(old *v1alpha1.ProtocolConfig, new *v1alpha1.ProtocolConfig) bool {
+	return !reflect.DeepEqual(old, new)
+}
+
+// updateProtocolInConfigMap updates the protocol in the deviceProfile in configmap
+func (dc *DownstreamController) updateProtocolInConfigMap(device *v1alpha1.Device) {
```
m1093782566: I suggest to add a return error for `updateProtocolInConfigMap()`.

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -438,6 +438,103 @@ func isNodeSelectorUpdated(old *v1.NodeSelector, new *v1.NodeSelector) bool {
...
+			}
+			dp, ok := nodeConfigMap.Data[DeviceProfileJSON]
+			if !ok || dp == "{}" {
+				// This case should never be hit as we delete empty configmaps
+				log.LOGGER.Errorf("Failed to get deviceProfile from configmap data or deviceProfile is empty")
+				return
+			}
+
+			deviceProfile := &types.DeviceProfile{}
+			err := json.Unmarshal([]byte(dp), deviceProfile)
```
m1093782566: I like the format of

```
if err := xxx; err != nil {
  xxx
}
```

to save one line :)

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -438,6 +438,103 @@ func isNodeSelectorUpdated(old *v1.NodeSelector, new *v1.NodeSelector) bool {
...
+			var oldProtocol string
+			for _, devInst := range deviceProfile.DeviceInstances {
+				if device.Name == devInst.Name {
+					oldProtocol = devInst.Protocol
+					break
+				}
+			}
+
+			// delete the old protocol
+			for i, ptcl := range deviceProfile.Protocols {
```
m1093782566: This function is easy to make mistake, please add tests.

cloud/pkg/devicecontroller/controller/downstream.go
m1093782566: Should not we return error here?

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: There cannot be multiple protocols with same name as protocol is suffixed with device name which will always be unique(enforced by apiserver).
for eg: protocol = protocolName + "-" + deviceName

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: The function only has if else cases, so when this else if ends, the function exits.

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: The functions in which calls this is called doesn't do anything after calling this function. Hence even if we return error here, we will just be logging it. Hence I am logging errors in this function itself.

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: We will soon be adding e2e tests though in next PR's

m1093782566: So we can break the for loop immediately once there is a match?

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: We need to update the configmap by deleting the old protocol even if the new protocol is not supported to keep our cache consistent with apiserver

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: Yess, that's a nice catch. I will add a break here.

cloud/pkg/devicecontroller/controller/downstream.go
m1093782566: Also I think we can handle deleting one element in slice via one line code.

```
a = append(a[:i], [i+1:]...)
```

Welcome trying it :)

cloud/pkg/devicecontroller/controller/downstream.go
m1093782566: Also I think we can handle deleting one element in slice via one line code.

```
a = append(a[:i], [i+1:]...)
```

Welcome trying it :)

cloud/pkg/devicecontroller/controller/downstream.go
m1093782566: OK.

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: Super !! Done Thanks

cloud/pkg/devicecontroller/controller/downstream.go
sids-b: Done, Thanks

PR #621, opened 2019-05-30
Update crd design with validation

rohitsardesai83:
**What type of PR is this?**
/kind design

Added basic validations we need to support while creating a device model or device instance.
We are not handling runtime validations (e.g deleting a device instance which is being referred by mapper , delete device model which is being referred by mapper etc .)

### DIFF
--- a/docs/proposals/device-crd.md
+++ b/docs/proposals/device-crd.md
@@ -570,6 +570,28 @@ status:
 
 ### Validation
 [Open API v3 Schema based validation](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation) can be used to guard against bad requests.
+Invalid values for fields ( example string value for a boolean field etc) can be validated using this.
+In some cases , we also need custom validations (e.g create a device instance which refers to a non -existent device model ) .
+[Validation admission web hooks](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) can be used to implement such custom validation rules.
+
+Here is a list of validations we need to support :
+
+#### Device Model Validations
+- Don't allow model creation if any `Required` fields are missing ( like property name , type, access mode etc.)
+- Property type can be only string or int64. Other types are not supported.
+- The Property access mode for a property can be one of `ReadWrite` or `ReadOnly`. Other access modes are not supported.
+- Property visitors are currently supported for bluetooth, modbus and opcua protocols only. Other protocol visitors are not supported.
+- The BluetoothArithmeticOperationType for Bluetooth read converter can have values one of `[Add, Subtract, Multiply, Divide]`
+- The ModbusRegisterType for Modbus visitor config can have values one of `[CoilRegister, DiscreteInputRegister, InputRegister, HoldingRegister]`. Other register types are not supported.
+- Don't allow model deletion if there is at least one device instance referring to it.
+- Don't allow deletion of a device property from the model if there is a corresponding device twin property in a device instance which refers to this model.
+- Don't allow deletion of a device property visitor if the corresponding device property exists.
+
+#### Device Instance Validations
+- Don't allow device instance creation if any `Required` fields are missing ( like devicemodel ref , twin property value, twin property name etc.)
+- Don't allow device instance creation if it refers to a device model reference which doesn't exist.
+- Don't allow device instance creation if a desired twin property's name cannot be matched to a device property in the device model it refers to.
+- Protocol configs are optional , but if provided , they can be one of `[opcua, modbus and bluetooth]`. Other protocol configs are not supported.
 
 ## Synchronizing Device Twin Updates
 


docs/proposals/device-crd.md
```diff
@@ -570,6 +570,24 @@ status:
...
+- Property type can be only string or int64. Other types are not supported.
+- The Property access mode for a property can be one of `ReadWrite` or `ReadOnly`. Other access modes are not supported.
+- Property visitors are currently supported for bluetooth, modbus and opcua protocols only. Other protocol visitors are not supported.
+- The BluetoothArithmeticOperationType for Bluetooth read converter can have values one of `[Add, Subtract, Multiply, Divide]`
+- The ModbusRegisterType for Modbus visitor config can have values one of `[CoilRegister, DiscreteInputRegister, InputRegister, HoldingRegister]`. Other register types are not supported.
+
+#### Device Instance Validations
+- Don't allow device instance creation if any `Required` fields are missing ( like devicemodel ref , protocol config, twin property value, twin property name etc.)
+- Don't allow device instance creation if it refers to a device model reference which doesn't exist.
+- Don't allow protocol configs other than opcua, modbus and bluetooth.
```
sids-b: Maybe we can support missing protocol config for wired devices ?

docs/proposals/device-crd.md
```diff
@@ -570,6 +570,24 @@ status:
...
+- Don't allow model creation if any `Required` fields are missing ( like property name , type, access mode etc.)
+- Property type can be only string or int64. Other types are not supported.
+- The Property access mode for a property can be one of `ReadWrite` or `ReadOnly`. Other access modes are not supported.
+- Property visitors are currently supported for bluetooth, modbus and opcua protocols only. Other protocol visitors are not supported.
+- The BluetoothArithmeticOperationType for Bluetooth read converter can have values one of `[Add, Subtract, Multiply, Divide]`
+- The ModbusRegisterType for Modbus visitor config can have values one of `[CoilRegister, DiscreteInputRegister, InputRegister, HoldingRegister]`. Other register types are not supported.
+
+#### Device Instance Validations
+- Don't allow device instance creation if any `Required` fields are missing ( like devicemodel ref , protocol config, twin property value, twin property name etc.)
+- Don't allow device instance creation if it refers to a device model reference which doesn't exist.
```
sids-b: Need to add a point that dont't allow device instance creation if any of the twin name is not present in devicemodel property list ?

docs/proposals/device-crd.md
```diff
@@ -570,6 +570,24 @@ status:
...
+
+Here is a list of validations we need to support :
+
+#### Device Model Validations
+- Don't allow model creation if any `Required` fields are missing ( like property name , type, access mode etc.)
+- Property type can be only string or int64. Other types are not supported.
+- The Property access mode for a property can be one of `ReadWrite` or `ReadOnly`. Other access modes are not supported.
+- Property visitors are currently supported for bluetooth, modbus and opcua protocols only. Other protocol visitors are not supported.
+- The BluetoothArithmeticOperationType for Bluetooth read converter can have values one of `[Add, Subtract, Multiply, Divide]`
+- The ModbusRegisterType for Modbus visitor config can have values one of `[CoilRegister, DiscreteInputRegister, InputRegister, HoldingRegister]`. Other register types are not supported.
```
sids-b: not allow deletion of a model if any device refers to it ?

sids-b: Need to add validations for updation of a device model.
1) Cannot allow deletion of a property if some deviceTwin refers to it.
2) Need to property visitor cannot be deleted if property exists ?

rohitsardesai83: added this below.

rohitsardesai83: done. thanks.

docs/proposals/device-crd.md
rohitsardesai83: added.

docs/proposals/device-crd.md
rohitsardesai83: currently in the api types , the protocol config is required , but we can see if we can make it optional. updated the doc.

sids-b: https://github.com/kubeedge/examples/blob/master/led-raspberrypi/sample-crds/led-light-device-instance.yaml

I guess this instance works with current types without protocol

sids-b: cc @sujithsimon22

sids-b:
ping @kevin-wangzefeng @m1093782566 for review

docs/proposals/device-crd.md
sujithsimon22: @rohitsardesai83 What @sids-b mentioned is true, in the  current example of LED light we do not specify any protocol and the device CRD feature for it seems to be working fine

rohitsardesai83: yes what I meant to say was that the api types has a comment saying that is Required, but somehow the openapi schema validation doesn't seem to kick in and so its working. So we can modify the comment in the api types to make it optional. The validation point has already been modified accordingly.

sujithsimon22: I think it might be a good idea for us to have a protocol for wired devices, maybe all types of wired devices can come under such a protocol, but for such a protocol i'm not sure if we need to provide  protocol config or if we do support protocol configs it should be very generic. thoughts?

rohitsardesai83: This is what the validation says now : 

- Protocol configs are optional , but if provided , they can be one of `[opcua, modbus and bluetooth]`. Other protocol configs are not supported.

docs/proposals/device-crd.md
```diff
@@ -570,6 +570,28 @@ status:
...
+- Don't allow model deletion if there is at least one device instance referring to it.
+- Don't allow deletion of a device property from the model if there is a corresponding device twin property in a device instance which refers to this model.
+- Don't allow deletion of a device property visitor if the corresponding device property exists.
+
+#### Device Instance Validations
+- Don't allow device instance creation if any `Required` fields are missing ( like devicemodel ref , twin property value, twin property name etc.)
+- Don't allow device instance creation if it refers to a device model reference which doesn't exist.
+- Don't allow device instance creation if a desired twin property's name cannot be matched to a device property in the device model it refers to.
+- Protocol configs are optional , but if provided , they can be one of `[opcua, modbus and bluetooth]`. Other protocol configs are not supported.
```
chendave: This is not much clear, the schema shows `Protocol`  here is  `Required`.
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/apis/devices/v1alpha1/device_instance_types.go#L29

docs/proposals/device-crd.md
```diff
@@ -570,6 +570,28 @@ status:
 
 ### Validation
 [Open API v3 Schema based validation](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#validation) can be used to guard against bad requests.
+Invalid values for fields ( example string value for a boolean field etc) can be validated using this.
```
chendave: for example, .... etc.

docs/proposals/device-crd.md
sujithsimon22: I guess that makes sense, in the sense that even if we do support wired protocol they would not require a protocol config, in which case the validation would remain unaffected

sujithsimon22:
> I am okay with most of the change, good documentation, just want to make it clear whether `Protocol` is optional or not.

@chendave I guess 'Protocol Config' can be made optional since not all protocol would require a protocol config (for example wired protocol). I guess as @rohitsardesai83 has mentioned above we can change the comment in the api types to make it optional

docs/proposals/device-crd.md
rohitsardesai83: Yes , we had thought of this section as required , but given certain scenarios and examples like led etc , we may need to make it optional. We need to update the api types accordingly.

chendave:
hmm... it was saying at least two approval is needed :(
sorry guys for the abusing the approve label.

sids-b:
LGTM

sids-b:
No worries @chendave , looks like some issue with the bot.
cc @edisonxiang @pavan187

kevin-wangzefeng:
It's the bot didn't enforce prerequisite restriction, we should hurry up moving to prow.
@sids-b

sids-b:
cc @naveensriram @omung-g

PR #644, opened 2019-06-03
added edge node role label while running installer

### LINKED ISSUE #643: KubeEdge didn't add a role label while creating a node
**What happened**:
Kubedge Installer didn't add the role label while creating a node.

KubeEdge installer should add the role as edge for a node

run kubeedge join

The addition of role has become mandatory as KubeEdge uses role to differentiate between edge nodes and other nodes.
XREF : #244 #294

sids-b:
**What type of PR is this?**
/kind bug

This PR adds role label while creating a node in kubeedge installer code 

Fixes #643


sids-b:
@samy2019 @srivatsav123 , please let me know if we need to add this label in any other place. I will take care of it in this PR

samy2019:
@sids-b , Only 2 places to change where node json is created.

PR #653, opened 2019-06-04
removing race condition in dtcontext

### LINKED ISSUE #640: race condition @ DTContext.Lock / DTContext.Unlock
**What happened**:
We have a potential race @ locking / unlocking the device mutex. (https://github.com/kubeedge/kubeedge/blob/0ddc4b3f7752fb527c73bec4d5d2adbe5f3635f2/edge/pkg/devicetwin/dtcontext/dtcontext.go#L107-L128)

EXAMPLE CASE:
- 2 messages from the same device
- start handling first message
  - run dtc.Lock('xyz')
      - success on dtc.Mutex.Lock()
      - success on deviceMutex.Lock()
      - success on dtc.Mutex.Unlock()
- start handling second message
   - run dtc.Lock('xyz')
      - success on dtc.Mutex.Lock()
      - bounded waiting on deviceMutex.Lock() # beause the first message has this mutex
- continue handling first message
  - run dtc.Unlock('xyz')
     - bounded waiting on dtc.Mutex.Unlock() # because the second message has the mutex

no race condition

**Environment**:
not relevant

subpathdev:
**What type of PR is this?**
 /kind bug

fix race condition which is described in #640 

**Which issue(s) this PR fixes:
Fixes #640 

### DIFF
--- a/edge/pkg/devicetwin/dtcontext/dtcontext.go
+++ b/edge/pkg/devicetwin/dtcontext/dtcontext.go
@@ -31,7 +31,7 @@ type DTContext struct {
 	ModulesContext *context.Context
 	DeviceList     *sync.Map
 	DeviceMutex    *sync.Map
-	Mutex          *sync.Mutex
+	Mutex          *sync.RWMutex
 	// DBConn *dtclient.Conn
 	State string
 }
@@ -49,7 +49,7 @@ func InitDTContext(context *context.Context) (*DTContext, error) {
 	var confirm sync.Map
 	var deviceList sync.Map
 	var deviceMutex sync.Map
-	var mutex sync.Mutex
+	var mutex sync.RWMutex
 	// var deviceVersionList sync.Map
 
 	return &DTContext{
@@ -107,9 +107,8 @@ func (dtc *DTContext) GetMutex(deviceID string) (*sync.Mutex, bool) {
 func (dtc *DTContext) Lock(deviceID string) bool {
 	deviceMutex, ok := dtc.GetMutex(deviceID)
 	if ok {
-		dtc.Mutex.Lock()
+		dtc.Mutex.RLock()
 		deviceMutex.Lock()
-		dtc.Mutex.Unlock()
 		return true
 	}
 	return false
@@ -119,9 +118,8 @@ func (dtc *DTContext) Lock(deviceID string) bool {
 func (dtc *DTContext) Unlock(deviceID string) bool {
 	deviceMutex, ok := dtc.GetMutex(deviceID)
 	if ok {
-		dtc.Mutex.Lock()
 		deviceMutex.Unlock()
-		dtc.Mutex.Unlock()
+		dtc.Mutex.RUnlock()
 		return true
 	}
 	return false
--- a/edge/pkg/devicetwin/dtcontext/dtcontext_test.go
+++ b/edge/pkg/devicetwin/dtcontext/dtcontext_test.go
@@ -245,7 +245,7 @@ func TestGetMutex(t *testing.T) {
 //TestLock is function to test Lock().
 func TestLock(t *testing.T) {
 	dtc := &DTContext{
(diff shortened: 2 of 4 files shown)


subpathdev:
cc @sids-b

edge/pkg/devicetwin/dtcontext/dtcontext.go
```diff
@@ -107,9 +107,8 @@ func (dtc *DTContext) GetMutex(deviceID string) (*sync.Mutex, bool) {
 func (dtc *DTContext) Lock(deviceID string) bool {
 	deviceMutex, ok := dtc.GetMutex(deviceID)
 	if ok {
-		dtc.Mutex.Lock()
+		dtc.Mutex.RLock()
 		deviceMutex.Lock()
-		dtc.Mutex.Unlock()
```
sids-b: @subpathdev can we have a RUnlock here ?

edge/pkg/devicetwin/dtcontext/dtcontext.go
```diff
@@ -119,9 +118,8 @@ func (dtc *DTContext) Lock(deviceID string) bool {
 func (dtc *DTContext) Unlock(deviceID string) bool {
 	deviceMutex, ok := dtc.GetMutex(deviceID)
 	if ok {
-		dtc.Mutex.Lock()
```
sids-b: can we have a RLock here ?

subpathdev:
Why we should directly unlock the dtc.Mutex? (I think that is the main point here) 

What are the benefits from leaving away:
  - dealMembershipDetail can only run if no reader is running (=> when dealMembershipDetail is running no changes will be made on any device)

What are the losses about this?
  - dealMembership must wait till all readers have unlocked the RWMutex

sids-b:
> Why we should directly unlock the dtc.Mutex? (I think that is the main point here)
> 
> What are the benefits from leaving away:
> 
> * dealMembershipDetail can only run if no reader is running (=> when dealMembershipDetail is running no changes will be made on any device)
> 
> What are the losses about this?
> 
> * dealMembership must wait till all readers have unlocked the RWMutex

Sounds Fair

sids-b:
cc @rohitsardesai83 for review

sids-b:
cc @m1093782566 for review

rohitsardesai83:
@subpathdev , please mark conversations as resolved if you have addressed them.

subpathdev:
@rohitsardesai83 sry I will do this in the future

sids-b:
@subpathdev , please give me sometime. I am thinking that if readers keep on acquiring lock on after the other, dealMembershipDetail may acquire the lock after a long time. Give me sometime to think if something can be done to solve this.

subpathdev:
@sids-b please share your thoughts. It is a pleasure to resolve the problem with you.

If I understand the problem correct you are thinking that the read locks stop executing dealMembershipDetail for a very long time, because they are acquire the read lock again and again and ...?

This is problem should be fixed by sync.RWLock. Here an output of the documentation (https://golang.org/pkg/sync/#RWMutex):
> If a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released. In particular, this prohibits recursive read locking. This is to ensure that the lock eventually becomes available; a blocked Lock call excludes new readers from acquiring the lock.

If I am understanding the documentation correctly, after acquire a rwmutex.Lock() no new read access on the critical section will be granted till rwmutex.Unlock() is called.
To make it concrete if dealMembershipDetail runs "dtcontex".LockAll() all new readers have to wait till dealMembershipDeatil runs "dtcontext".UnlockAll().

On the other hand we process only one message on one rwmutex.RLock call and unlock the read after finishing the process.

sids-b:
Thanks @subpathdev

kevin-wangzefeng:
Thanks for the fix!
/lgtm
/approve

PR #664, opened 2019-06-07
set minimum go version

subpathdev:
**What type of PR is this?**
/kind documentation

set a minimum go version to avoid frustrating users

I set the go version to 1.11 because our travis is using this version. On my own machine I am using 1.12 and it works fine.

### DIFF
--- a/docs/setup/setup.md
+++ b/docs/setup/setup.md
@@ -18,6 +18,8 @@
     - --insecure-bind-address=0.0.0.0
     ```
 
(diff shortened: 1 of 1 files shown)


subpathdev:
Is there a reason, that the links in the documentation are not usable in github?

docs/setup/setup.md
```diff
@@ -18,6 +18,8 @@
     - --insecure-bind-address=0.0.0.0
     ```
 
++ **Go** The minimum require go version is 1.11. You can install this version by using [this website.](https://golang.org/dl/)
```
rohitsardesai83: require  -> required.

PR #667, opened 2019-06-09
expected value twins 

### LINKED ISSUE #662: pointer error in cloud/pkg/devicecontroller/controller/downstream.go
**What happened**:
If you set a desired value for only one parameter the value will used by all twins.
the value should only be used by the expected twin
git diff to add more debugging:
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -629,7 +629,11 @@ func addUpdatedTwins(new []v1alpha1.Twin, twin map[string]*types.MsgTwin, versio
        opt := false
        optional := &opt
        for _, dtwin := range new {
\+               log.LOGGER.Debugf("twin actual value: %v", dtwin.Reported.Value)
\+               log.LOGGER.Debugf("twin expected value: %v", dtwin.Desired.Value)
\+               log.LOGGER.Debugf("twin expected value: %v", &dtwin.Desired.Value)
                expected := &types.TwinValue{}
\+               log.LOGGER.Debugf("twin expected: %v", expected)
                expected.Value = &dtwin.Desired.Value
                timestamp := time.Now().UnixNano() / 1e6
## Device Instance
```yaml
---
apiVersion: devices.kubeedge.io/v1alpha1
kind: Device
metadata:
        name: cpu-sensor-tag01
        labels:
                description: CPU-Sensor
                model: cpu-sensor
spec:
        deviceModelRef:
                name: cpu-sensor
        nodeSelector:
                nodeSelectorTerms:
                        - matchExpressions:
                                  - key: ''
                                    operator: IN
                                    values:
                                            - edge
status:
        twins:
                - propertyName: CPU_Temperatur
                  reported:
                          metadata:
                                  type: int
                - propertyName: cpu_id
                  reported:
                          value: "0"
                          metadata:
                                  type: int

```
## Device Model
```yaml
---
apiVersion: devices.kubeedge.io/v1alpha1
kind: DeviceModel
metadata:
        label:
                description: CPU-Sensor
        name: cpu-sensor
spec:
        properties:
                - name: CPU_Temperatur
                  description: temperatur in degree celsius
                  type:
                          int:
                                  accessMode: ReadOnly
                                  unit: Degree Celsius
                - name: cpu_id
                  description: set the id of the cpu
                  type:
                          int:
                                  accessMode: ReadWrite
                                  defaultValue: 0

```

- at this point you need an running instance
- kubectl apply -f deviceModel + deviceInstance
- kubectl edit device cpu-sensor-tag01; #and adding the following to cpu_id:
  desired:
    value: "12313"

By running the above example I produce the above the following output:
2019-06-06 23:11:34.785 +02:00 DEBUG controller/downstream.go:632 twin actual value:
2019-06-06 23:11:34.786 +02:00 DEBUG controller/downstream.go:633 twin expected value:
2019-06-06 23:11:34.786 +02:00 DEBUG controller/downstream.go:634 twin expected value: 0xc0003c47d0
2019-06-06 23:11:34.786 +02:00 DEBUG controller/downstream.go:636 twin expected: &{<nil> <nil>}
2019-06-06 23:11:34.786 +02:00 DEBUG controller/downstream.go:632 twin actual value: 0
2019-06-06 23:11:34.787 +02:00 DEBUG controller/downstream.go:633 twin expected value: 99887766
2019-06-06 23:11:34.787 +02:00 DEBUG controller/downstream.go:634 twin expected value: 0xc0003c47d0
2019-06-06 23:11:34.787 +02:00 DEBUG controller/downstream.go:636 twin expected: &{<nil> <nil>}

**Environment**:
not relevant

subpathdev:
**What type of PR is this?**
/kind bug

you can use different expected values for each twin of an device. 

Fixes #662

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -397,9 +397,9 @@ func createDevice(device *v1alpha1.Device) types.Device {
 	opt := false
 	optional := &opt
 	twin := make(map[string]*types.MsgTwin)
-	for _, dtwin := range device.Status.Twins {
+	for i, dtwin := range device.Status.Twins {
 		expected := &types.TwinValue{}
-		expected.Value = &dtwin.Desired.Value
+		expected.Value = &device.Status.Twins[i].Desired.Value
 		timestamp := time.Now().UnixNano() / 1e6
 
 		metadata := &types.ValueMetadata{Timestamp: timestamp}
@@ -588,10 +588,10 @@ func (dc *DownstreamController) deviceUpdated(device *v1alpha1.Device) {
 func addDeletedTwins(oldTwin []v1alpha1.Twin, newTwin []v1alpha1.Twin, twin map[string]*types.MsgTwin, version string) {
 	opt := false
 	optional := &opt
-	for _, dtwin := range oldTwin {
+	for i, dtwin := range oldTwin {
 		if !ifTwinPresent(dtwin, newTwin) {
 			expected := &types.TwinValue{}
-			expected.Value = &dtwin.Desired.Value
+			expected.Value = &oldTwin[i].Desired.Value
 			timestamp := time.Now().UnixNano() / 1e6
 
 			metadata := &types.ValueMetadata{Timestamp: timestamp}
@@ -628,9 +628,9 @@ func ifTwinPresent(twin v1alpha1.Twin, newTwins []v1alpha1.Twin) bool {
 func addUpdatedTwins(newTwin []v1alpha1.Twin, twin map[string]*types.MsgTwin, version string) {
 	opt := false
 	optional := &opt
-	for _, dtwin := range newTwin {
+	for i, dtwin := range newTwin {
 		expected := &types.TwinValue{}
-		expected.Value = &dtwin.Desired.Value
+		expected.Value = &newTwin[i].Desired.Value
 		timestamp := time.Now().UnixNano() / 1e6
 
 		metadata := &types.ValueMetadata{Timestamp: timestamp}
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -122,13 +122,13 @@ func (dc *DownstreamController) deviceModelAdded(deviceModel *v1alpha1.DeviceMod
 }
 
 // isDeviceModelUpdated is function to check if deviceModel is actually updated
-func isDeviceModelUpdated(old *CacheDeviceModel, new *v1alpha1.DeviceModel) bool {
+func isDeviceModelUpdated(oldTwin *CacheDeviceModel, newTwin *v1alpha1.DeviceModel) bool {
 	// does not care fields
-	old.ObjectMeta.ResourceVersion = new.ObjectMeta.ResourceVersion
-	old.ObjectMeta.Generation = new.ObjectMeta.Generation
+	oldTwin.ObjectMeta.ResourceVersion = newTwin.ObjectMeta.ResourceVersion
+	oldTwin.ObjectMeta.Generation = newTwin.ObjectMeta.Generation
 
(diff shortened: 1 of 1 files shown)


cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -122,13 +122,13 @@ func (dc *DownstreamController) deviceModelAdded(deviceModel *v1alpha1.DeviceMod
 }
 
 // isDeviceModelUpdated is function to check if deviceModel is actually updated
-func isDeviceModelUpdated(old *CacheDeviceModel, new *v1alpha1.DeviceModel) bool {
+func isDeviceModelUpdated(old *CacheDeviceModel, newTwin *v1alpha1.DeviceModel) bool {
```
rohitsardesai83: old -> oldTwin just to be consistent with new -> newTwin .

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -625,12 +625,12 @@ func ifTwinPresent(twin v1alpha1.Twin, newTwins []v1alpha1.Twin) bool {
...
 // addUpdatedTwins is function of add updated twins to send to edge
-func addUpdatedTwins(new []v1alpha1.Twin, twin map[string]*types.MsgTwin, version string) {
+func addUpdatedTwins(newTwin []v1alpha1.Twin, twin map[string]*types.MsgTwin, version string) {
 	opt := false
 	optional := &opt
-	for _, dtwin := range new {
+	for i, dtwin := range newTwin {
 		expected := &types.TwinValue{}
-		expected.Value = &dtwin.Desired.Value
+		expected.Value = &newTwin[i].Desired.Value
```
sids-b: use dtwin here ?

sids-b:
@subpathdev , I think i am missing something. Can you please tell me how this PR fixes #662

subpathdev:
> @subpathdev , I think i am missing something. Can you please tell me how this PR fixes #662

This will resolve the problem because dtwin has the same address space in every loop cycle. But in the original array is using different address space for each value.

I have created a minimal example: https://play.golang.org/p/YkvsOsFUJgO

cloud/pkg/devicecontroller/controller/downstream.go
subpathdev: That is the fix which resolve the problem. Please have a look on this comment: https://github.com/kubeedge/kubeedge/pull/667#issuecomment-500704525

subpathdev:
I have to make more changes because this problem occurs in different functions.

subpathdev:
@sids-b I will try and go deeper and explain you the problem and how this change will fix this.  (It was helping me to understand how pointer works in C)

y is an array of something
```go
for _, x := range y {
```
this part will create an an new object x. X has his own address space and the values of y[i] will be copied to x. 
To make it concrete y is an array of strings and has the length of two a address of the values could be: 80f4 (first element of the array) and 8102 (second element of the array). X has it owns address space so we could say 90ff is the address of x.


z is a variable which is declared before the for loop; (we are in the for loop) (j is an index variable)
```go
z[j] = &x
```
in z the address of x will be stored. 
On the concrete example this will mean in z stands always 90ff. But x will stores on the first run the value of y[0] and on the second run y[1]. So the values on which the pointer shows is alaways y[1].

I have created a little show case of my concrete example. You will find it here: https://play.golang.org/p/zKjyMCgdQ8o

( I don't know but the value which is behind z[i] could be undefined outside the loop. So it could be luck that we aren't run into an exception. (Also comparing this to C and access to an address which was allocated and freed before) 

The solution:
the array has its own address space so we could take the address from this to go ahead. I also created a little show case. https://play.golang.org/p/Y9VahG5Kwmk

Can you understand this? My English is not well enough to write it better and it is complicated to describe.

sids-b:
@subpathdev The explanation is quite clear :)

Issue #631, opened 2019-05-31
devicecrd: add basic validation for device model and device instance management.


rohitsardesai83:
**What would you like to be added**:
An issue  #405   was created to track pending issues with respect to device crd implementation and stability. We cannot cover all the items in 1.0 release. So we can address basic validations first and then address the remaining items. This issue tracks what basic validations can be addressed for 1.0

**Why is this needed**:
To provide a solid implementation of device crds and guard against incorrect actions from users when using device models and instances to manage their devices.

**Validations targeted for 1.0.0**:

**Device Model Validations**
- [x] Don't allow model creation if any Required fields are missing ( like property name , type, access mode etc.)
- [x] The Property access mode for a property can be one of ReadWrite or ReadOnly. Other access modes are not supported.
- [x] The BluetoothArithmeticOperationType for Bluetooth read converter can have values one of [Add, Subtract, Multiply, Divide]
- [x] The ModbusRegisterType for Modbus visitor config can have values one of [CoilRegister, DiscreteInputRegister, InputRegister, HoldingRegister]. Other register types are not supported.

**Device Instance Validations**
- [x] Don't allow device instance creation if any Required fields are missing ( like devicemodel ref , twin property value, twin property name etc.)

rohitsardesai83:
Came across this : https://swagger.io/docs/specification/data-models/data-types/#required.
Required properties can be validated like this. Our device model and device instance crd yamls need to have such enforcements.

PR #669, opened 2019-06-10
Issue #631: Add validation of required fields for device model and device instance

rohitsardesai83:
**What type of PR is this?**
/kind api-change

This PR introduces very basic validation of required fields while creating device models and device instances. It doesn't take care of the rest of the validations mentioned in #631.

xref #631 : 

### DIFF
--- a/build/crds/devices/devices_v1alpha1_device.yaml
+++ b/build/crds/devices/devices_v1alpha1_device.yaml
@@ -57,14 +57,41 @@ spec:
                           description: Required. BaudRate 115200|57600|38400|19200|9600|4800|2400|1800|1200|600|300|200|150|134|110|75|50
                           format: int64
                           type: integer
+                          enum:
+                            - 115200
+                            - 57600
+                            - 38400
+                            - 19200
+                            - 9600
+                            - 4800
+                            - 2400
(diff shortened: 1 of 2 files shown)


rohitsardesai83:
@chendave , added a few more validations with enums to validate valid values for certain fields like access modes , baud rate etc.

rohitsardesai83:
@sids-b , @kevin-wangzefeng , please review.

kevin-wangzefeng:
Can you add e2e tests as well?

chendave:
Looks like there is no good way to reject an unknown property which is a kind of object, I saw "additionalProperties" is banned.

rohitsardesai83:
> Looks like there is no good way to reject an unknown property which is a kind of object, I saw "additionalProperties" is banned.

We can reject unknown properties in the custom validation hooks , if usage of `additionalProperties:false` is not recommended.

rohitsardesai83:
> Can you add e2e tests as well?

@kevin-wangzefeng , yes we can add , but there is another PR dealing with e2e tests for device management already : https://github.com/kubeedge/kubeedge/pull/669.

Once this PR gets merged , I can add some tests for validation.

Issue #673, opened 2019-06-10
EdgeMesh implementation


sids-b:
**What would you like to be added**:
To support service mesh capabilities on edge to support microservice communication cross cloud and edge. The service discovery, communication should be considered in this feature. Developers can define services on top of the deployments deployed on the cloud or edges in a unified manner without caring about their location. The service name is the only thing they need to know to do the communication. For example, the following address can be used for a REST request: http://{service_name}/resources

**Why is this needed**:
The cloud native and microservice architecture is becoming more and more popular and the edge node is becoming more and more powerful. The user want to decompose their application into multiple microservices. Some of those microservices need to be deployed in edge node to get close to the data generated by devices. By simplify the development, the user need to use the same mechanism to do the service discovery and communication with what they are doing on cloud.

I am taking some points from @qizha in edgemesh design document to create this issue

sids-b:
creating this issue to track edgemesh implementation
cc @qizha @anyscoding

sids-b:
#807 is the second phase for edgemesh implementation

PR #674, opened 2019-06-10
Support to watch on service and endpoints and send to edge for EdgeMesh Cloud Side Impl

sids-b:
**What type of PR is this?**
/kind feature

This PR adds cloud side code to watch on services and endpoints and send to edge nodes

XREF: #673 

### DIFF
--- a/cloud/pkg/controller/controller/downstream.go
+++ b/cloud/pkg/controller/controller/downstream.go
@@ -17,6 +17,7 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/pkg/controller/manager"
 	"github.com/kubeedge/kubeedge/cloud/pkg/controller/messagelayer"
 	"github.com/kubeedge/kubeedge/cloud/pkg/controller/utils"
+	common "github.com/kubeedge/kubeedge/common/constants"
 )
 
 // DownstreamController watch kubernetes api server and send change to edge
@@ -35,6 +36,9 @@ type DownstreamController struct {
 
 	nodeManager *manager.NodesManager
 	nodesStop   chan struct{}
+
+	serviceManager *manager.ServiceManager
+	serviceStop    chan struct{}
 
 	lc *manager.LocationCache
 }
@@ -189,8 +193,31 @@ func (dc *DownstreamController) syncEdgeNodes(stop chan struct{}) {
 			switch e.Type {
 			case watch.Added:
 				dc.lc.UpdateEdgeNode(node.ObjectMeta.Name)
+				fallthrough
 			case watch.Modified:
-				continue
+				// TODO: keep node cache for edge node to check if status has changed to Running from not running
+				// When node comes to running, send all the service information to edge
+				for _, nsc := range node.Status.Conditions {
+					if nsc.Type == "Ready" {
+						if nsc.Status == "True" {
+							msg := model.NewMessage("")
+							// TODO: what should in place of namespace and service when we are sending service list ?
+							resource, err := messagelayer.BuildResource(node.Name, "namespace", common.ResourceTypeServiceList, "service")
+							if err != nil {
+								log.LOGGER.Warnf("Built message resource failed with error: %s", err)
+								break
+							}
+							msg.BuildRouter(constants.EdgeControllerModuleName, constants.GroupResource, resource, model.UpdateOperation)
+							msg.Content = dc.lc.Services
+							if err := dc.messageLayer.Send(*msg); err != nil {
+								log.LOGGER.Warnf("Send message failed with error: %s, operation: %s, resource: %s", err, msg.GetOperation(), msg.GetResource())
+							} else {
+								log.LOGGER.Infof("Send message successfully, operation: %s, resource: %s", msg.GetOperation(), msg.GetResource())
+							}
+						}
+						break
+					}
+				}
 			case watch.Deleted:
 				dc.lc.DeleteNode(node.ObjectMeta.Name)
 			default:
@@ -200,6 +227,56 @@ func (dc *DownstreamController) syncEdgeNodes(stop chan struct{}) {
 			}
 		case <-stop:
 			log.LOGGER.Infof("Stop syncNodes")
+			running = false
+		}
+	}
+}
+
+func (dc *DownstreamController) syncService(stop chan struct{}) {
+	running := true
+	var operation string
+	for running {
+		select {
+		case e := <-dc.serviceManager.Events():
+			svc, ok := e.Object.(*v1.Service)
+			if !ok {
+				log.LOGGER.Warnf("Object type: %T unsupported", svc)
+				continue
+			}
+			switch e.Type {
+			case watch.Added:
+				dc.lc.AddService(*svc)
+				operation = model.InsertOperation
+			case watch.Modified:
+				dc.lc.UpdateService(*svc)
+				operation = model.UpdateOperation
+			case watch.Deleted:
+				dc.lc.DeleteService(*svc)
+				operation = model.DeleteOperation
+			default:
+				// unsupported operation, no need to send to any node
+				log.LOGGER.Warnf("Service event type: %s unsupported", e.Type)
+				continue
+			}
+
+			// send to all nodes
+			for _, nodeName := range dc.lc.EdgeNodes {
+				msg := model.NewMessage("")
+				resource, err := messagelayer.BuildResource(nodeName, svc.Namespace, common.ResourceTypeService, svc.Name)
+				if err != nil {
+					log.LOGGER.Warnf("Built message resource failed with error: %s", err)
+					break
+				}
+				msg.BuildRouter(constants.EdgeControllerModuleName, constants.GroupResource, resource, operation)
+				msg.Content = svc
+				if err := dc.messageLayer.Send(*msg); err != nil {
+					log.LOGGER.Warnf("Send message failed with error: %s, operation: %s, resource: %s", err, msg.GetOperation(), msg.GetResource())
+				} else {
+					log.LOGGER.Infof("Send message successfully, operation: %s, resource: %s", msg.GetOperation(), msg.GetResource())
+				}
+			}
+		case <-stop:
+			log.LOGGER.Infof("Stop sync services")
 			running = false
 		}
 	}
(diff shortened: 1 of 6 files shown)


cloud/pkg/controller/controller/downstream.go
```diff
@@ -189,8 +193,31 @@ func (dc *DownstreamController) syncEdgeNodes(stop chan struct{}) {
 			switch e.Type {
 			case watch.Added:
 				dc.lc.UpdateEdgeNode(node.ObjectMeta.Name)
+				fallthrough
 			case watch.Modified:
-				continue
+				// TODO: keep node cache for edge node to check if status has changed to Running from not running
```
sids-b: The current cache of edgeNodes is just the name. I will update it to save status for comparison

cloud/pkg/controller/controller/downstream.go
```diff
@@ -189,8 +193,31 @@ func (dc *DownstreamController) syncEdgeNodes(stop chan struct{}) {
...
+				// When node comes to running, send all the service information to edge
+				for _, nsc := range node.Status.Conditions {
+					if nsc.Type == "Ready" {
+						if nsc.Status == "True" {
+							msg := model.NewMessage("")
+							// TODO: what should in place of namespace and service when we are sending service list ?
+							resource, err := messagelayer.BuildResource(node.Name, "namespace", common.ResourceTypeServiceList, "service")
```
sids-b: we will send the entire list of services to the edgenode when the nodes comes to ready state from notReady. What should be the namespace and servicename field here ? as there will be  multiple services from different namespaces here

sids-b:
cc @qizha @anyscoding for review

sids-b:
@omung-g  @pavan187 we need to configure two 2 approvers from OWNERS file to give approval before adding approved label

cloud/pkg/controller/controller/downstream.go
```diff
@@ -205,6 +232,56 @@ func (dc *DownstreamController) syncEdgeNodes(stop chan struct{}) {
...
+					log.LOGGER.Infof("Send message successfully, operation: %s, resource: %s", msg.GetOperation(), msg.GetResource())
+				}
+			}
+		case <-stop:
+			log.LOGGER.Infof("Stop sync services")
+			running = false
+		}
+	}
+}
+
```
anyscoding: we need sync endpoint to edge too

cloud/pkg/controller/manager/location.go
```diff
@@ -10,12 +10,14 @@ import (
 
 // LocationCache cache the map of node, pod, configmap, secret
 type LocationCache struct {
-	//edgeNodes is a list of valid edge nodes
-	edgeNodes []string
+	// EdgeNodes is a list of valid edge nodes
+	EdgeNodes []string
```
anyscoding: Using sync.Map(key: nodename; value: state) instead of this "EdgeNodes" string slice here would be much better. Reasons:
1. Using sync.Map makes the time complexity of function UpdateEdgeNode and IsEdgeNode O(1), while using string slice with for loop makes it O(n).
2. Later we could be checking node state whether it's "ready" or still "not ready" by directly comparing "value" of sync.Map and doing the node state updating.

cloud/pkg/controller/manager/location.go
```diff
@@ -113,18 +115,18 @@ func (lc *LocationCache) SecretNodes(namespace, name string) (nodes []string) {
 
 //IsEdgeNode checks weather node is edge node or not
 func (lc *LocationCache) IsEdgeNode(nodeName string) bool {
-	for _, node := range lc.edgeNodes {
+	for _, node := range lc.EdgeNodes {
```
anyscoding: code as follow if we use sync.Map:
```
_, ok := lc.EdgeNodes.Load(nodeName)
return ok
```

cloud/pkg/controller/manager/location.go
```diff
@@ -113,18 +115,18 @@ func (lc *LocationCache) SecretNodes(namespace, name string) (nodes []string) {
...
 	}
 	return false
 }
 
-//UpdateEdgeNode is to maintain edge nodes name upto-date by querying kubernetes client
+// UpdateEdgeNode is to maintain edge nodes name upto-date by querying kubernetes client
 func (lc *LocationCache) UpdateEdgeNode(nodeName string) {
-	lc.edgeNodes = append(lc.edgeNodes, nodeName)
-	log.LOGGER.Infof("Edge nodes updated : %v \n", lc.edgeNodes)
+	lc.EdgeNodes = append(lc.EdgeNodes, nodeName)
```
anyscoding: UpdateEdgeNode with two args: nodeName and state and the lc.EdgeNodes.Store(nodeName, state)

cloud/pkg/controller/manager/location.go
sids-b: yess, I have added this TODO https://github.com/kubeedge/kubeedge/pull/674/files#diff-55277caec6421504e2848e75c588c384R198 ... I will make this change in this PR itself

cloud/pkg/controller/controller/downstream.go
sids-b: Sure, I will add

cloud/pkg/controller/manager/location.go
```diff
@@ -139,9 +141,27 @@ func (lc *LocationCache) DeleteSecret(namespace, name string) {
 
 // DeleteNode from cache
 func (lc *LocationCache) DeleteNode(name string) {
-	for i, v := range lc.edgeNodes {
+	for i, v := range lc.EdgeNodes {
```
anyscoding: lc.EdgeNodes.Delete(name)

cloud/pkg/controller/manager/location.go
sids-b: I have used map[string]string with lock instead of sync.Map as sync.Map doesn't allow to range(sync.Map)  and we need to range over all nodes to send services and endpoints to all nodes.

cloud/pkg/controller/controller/downstream.go
sids-b: Done

cloud/pkg/controller/manager/location.go
sids-b: Used map instead of sync.Map because of range issue mentioned above

cloud/pkg/controller/manager/location.go
sids-b: used delete(map,key) as i have used map instead of sync.Map

cloud/pkg/controller/manager/location.go
sids-b: used map instead of syncMap so saving directly

cloud/pkg/controller/manager/location.go
anyscoding: sync.Map allow range, continue when f return true, break when f return false
```
// Range calls f sequentially for each key and value present in the map.
// If f returns false, range stops the iteration.
//
// Range does not necessarily correspond to any consistent snapshot of the Map's
// contents: no key will be visited more than once, but if the value for any key
// is stored or deleted concurrently, Range may reflect any mapping for that key
// from any point during the Range call.
//
// Range may be O(N) with the number of elements in the map even if f returns
// false after a constant number of calls.
func (m *Map) Range(f func(key, value interface{}) bool) {
}
```

cloud/pkg/controller/manager/location.go
```diff
@@ -4,18 +4,23 @@ import (
...
-	"github.com/kubeedge/beehive/pkg/common/log"
 	"k8s.io/api/core/v1"
 )
 
 // LocationCache cache the map of node, pod, configmap, secret
 type LocationCache struct {
-	//edgeNodes is a list of valid edge nodes
-	edgeNodes []string
+	// EdgeNodes is a map, key is nodeName, value is Status
+	EdgeNodes map[string]string
```
anyscoding: sync.Map.Range allow range

```go
// Range calls f sequentially for each key and value present in the map.
// If f returns false, range stops the iteration.
//
// Range does not necessarily correspond to any consistent snapshot of the Map's
// contents: no key will be visited more than once, but if the value for any key
// is stored or deleted concurrently, Range may reflect any mapping for that key
// from any point during the Range call.
//
// Range may be O(N) with the number of elements in the map even if f returns
// false after a constant number of calls.
func (m *Map) Range(f func(key, value interface{}) bool) {
```

cloud/pkg/controller/manager/location.go
```diff
@@ -4,18 +4,23 @@ import (
...
+	// EdgeNodesLock is lock for edgeNodes
+	EdgeNodesLock sync.RWMutex
 	// configMapNode is a map, key is namespace/configMapName, value is nodeName
 	configMapNode sync.Map
 	// secretNode is a map, key is namespace/secretName, value is nodeName
 	secretNode sync.Map
+	// Services is an array of services
+	Services []v1.Service
```
anyscoding: 1. if we use []string, time complexity of function UpdateService and DeleteService is O(n), but sync service to edge when nodestate from ```not ready``` to ```ready``` is O(1)

2. if we use sync.Map, time complexity of function UpdateService and DeleteService is O(1), but sync service to edge when nodestate from ```not ready``` to ```ready``` is O(n)

Usually, service count is more than node count. service update event is more than node state update event.
on the other hand, add or update service in map is clearer than in []string

So i think, method 2 is better than method 1, can be considered, no mandatory

*note: sync.Map(key: namespace/serviceName, value: v1.Service)*

cloud/pkg/controller/manager/location.go
```diff
@@ -4,18 +4,23 @@ import (
...
+	// Endpoints is an array of endpoints
+	Endpoints []v1.Endpoints
```
anyscoding: same as Services

cloud/pkg/controller/manager/location.go
sids-b: Thanks Yushun. Used syncMap

cloud/pkg/controller/manager/location.go
sids-b: Thanks for pointing it out. Used sync.Map

cloud/pkg/controller/manager/location.go
sids-b: Done, Thanks

anyscoding:
lgtm

rohitsardesai83:
@kevin-wangzefeng , please review

PR #687, opened 2019-06-12
Adding visio file for KubeEdge architecture diagram

sujithsimon22:
**What type of PR is this?**

 /kind documentation

This PR contains the source file for KubeEdge architecture document. Users can use this to update the architecture diagram


sujithsimon22:
cc @sids-b @rohitsardesai83

yamt:
@sujithsimon22 

thank you!

a question; can we have a Makefile, bat file, or something to convert it to png?
or does it always require manual operation?
(probably a dumb question as i haven't used visio.)

sujithsimon22:
> @sujithsimon22
> 
> thank you!
> 
> a question; can we have a Makefile, bat file, or something to convert it to png?
> or does it always require manual operation?
> (probably a dumb question as i haven't used visio.)

@yamt I understand what you are trying to say. If I'm not wrong, we need to use visio software to view and edit this file and visio provides an option to convert it into PNG file. If you have any kind of script or makefile which could convert these visio source files to png, pls feel free to contribute :), but i guess we still we need to use visio to edit though

sids-b:
I think we should move our architecture picture to some online tool like draw.io, etc and share the link so it can be  made available to everyone for updating and converting to PNG.

kevin-wangzefeng:
Let's get this accepted first and then find alternative solution.

PR #701, opened 2019-06-13
Make device/bluetooth_mapper

subpathdev:
**What type of PR is this?**
/kind test

All parts on the master branch should buildable.
This will also add lint checks and go vet checks to the device/bluetooth_mapper part.

I have to change parts to run successfully "golangcli-lint" and "go vet".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,3 +50,7 @@ jobs:
       name: "cloud_lint"     # names the eighth Tests stage job
     - script: make keadm_lint
       name: "keadm_lint"
+    - script: make bluetoothdevice
+      name: "build bluetooth_device"
+    - script: make bluetoothdevice_lint
+      name: "bluetooth_device_lint"
--- a/Makefile
+++ b/Makefile
@@ -84,3 +84,14 @@ edgeimage:
 .PHONY: depcheck
 depcheck:
 	dep check
+
+.PHONY: bluetoothdevice
+bluetoothdevice:
+	make -C device/bluetooth_mapper
+
+.PHONY: bluetoothdevice_image
+	make -C device/bluetooth_mapper_docker
+
+.PHONY: bluetoothdevice_lint
+bluetoothdevice_lint:
+	make -C device/bluetooth_mapper lint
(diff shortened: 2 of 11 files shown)


subpathdev:
@sujithsimon22
I have fixed all lint problems but one go vet problem is still there. My question is how should I fix this:
>go/src/github.com/kubeedge/kubeedge/device/bluetooth_mapper/watcher/watcher.go:131:6: Infof format %s has arg helper.TwinResult.Twin[twinAttribute.Name].Actual of wrong type *github.com/kubeedge/kubeedge/device/bluetooth_mapper/helper.TwinValue

TwinValue is here always nil so I set it static or I replace %s with %v. (I like the second one but I am not sure about this)

sujithsimon22:
> @sujithsimon22
> I have fixed all lint problems but one go vet problem is still there. My question is how should I fix this:
> 
> > go/src/github.com/kubeedge/kubeedge/device/bluetooth_mapper/watcher/watcher.go:131:6: Infof format %s has arg helper.TwinResult.Twin[twinAttribute.Name].Actual of wrong type *github.com/kubeedge/kubeedge/device/bluetooth_mapper/helper.TwinValue
> 
> TwinValue is here always nil so I set it static or I replace %s with %v. (I like the second one but I am not sure about this)

@subpathdev The second option sounds better to me as well :) I think it is better if we replace %s with %v

subpathdev:
> > @sujithsimon22
> > I have fixed all lint problems but one go vet problem is still there. My question is how should I fix this:
> > > go/src/github.com/kubeedge/kubeedge/device/bluetooth_mapper/watcher/watcher.go:131:6: Infof format %s has arg helper.TwinResult.Twin[twinAttribute.Name].Actual of wrong type *github.com/kubeedge/kubeedge/device/bluetooth_mapper/helper.TwinValue
> > 
> > 
> > TwinValue is here always nil so I set it static or I replace %s with %v. (I like the second one but I am not sure about this)
> 
> @subpathdev The second option sounds better to me as well :) I think it is better if we replace %s with %v

@sujithsimon22 
thanks

sids-b:
/lgtm

Thanks @subpathdev

PR #710, opened 2019-06-17
Edgemesh dev

Arcanique:
**What type of PR is this?**
>
> /kind feature

>
> In the edgemesh scenario, implement application traffic hijacking using init container and iptables tecnology and contains the init container image creation script containing the Dockerfile for deb and rpm. Currently supports scenarios which the POD and the host's network namespace are isolated (for example, container networks such as bridges and flannels). We support IPv4 messages now .

Fixes #


subpathdev:
please write your commit messages in English

Issue #681, opened 2019-06-11
adding type support


subpathdev:
**What would you like to be added**:
 The type of value of a device should not only by string.

As you can see we are setting the type static to string. ([cloud/pkg/devicecontroller/downstream.go Line 417](https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/controller/downstream.go#L417) OR [cloud/pkg/devicecontroller/downstream.go Line 648](https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/controller/downstream.go#L648)) 

**Why is this needed**:
We say, that we are supporting string and int64 ...

rohitsardesai83:
:+1: , let's target this post 1.0 release ?

subpathdev:
@rohitsardesai83 Yes, can you put this in the backlog?

subpathdev:
thanks

subpathdev:
is fixed by pr #712

subpathdev:
The type will be overwritten somewhere => reopening issue

subpathdev:
OK I think this is only a documentation problem. By creating a new device we have to set the value and the type in the desired field and not in the reported field. (The reported field will be ignored)

Also the ValueType which is set by the DeviceModel.yaml, will be ignored.

sids-b:
@subpathdev , does this issue still persists ?

subpathdev:
no -> I am closing this issue

PR #712, opened 2019-06-17
Retaining metadata type value given by user for device

### LINKED ISSUE #711: Device Type Metadata getting overwritten to "string" type
**What happened**:
Device instance Metadata gets overwritten to "string" value even if it is created with "int" value

The metadata type given by user should be retained instead of overwriting it

1. Setup kubeedge cloud and edge
2. Create device model and a device instance with desired metadata type as int
3. Check the device value at edge using MQTT (it would have converted the metadata value to string)

**Environment**:
- KubeEdge version: 0.3

sujithsimon22:
**What type of PR is this?**
 /kind bug

This PR retains the value of metadata type if given by user else assigns it as string

Fixes #711 #681

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -400,6 +400,10 @@ func createDevice(device *v1alpha1.Device) types.Device {
 	for i, dtwin := range device.Status.Twins {
 		expected := &types.TwinValue{}
 		expected.Value = &device.Status.Twins[i].Desired.Value
+		metadataType, ok := device.Status.Twins[i].Desired.Metadata["type"]
+		if !ok {
+			metadataType = "string"
+		}
 		timestamp := time.Now().UnixNano() / 1e6
 
 		metadata := &types.ValueMetadata{Timestamp: timestamp}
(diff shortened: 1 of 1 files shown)


sujithsimon22:
cc @rohitsardesai83 @sids-b for review

subpathdev:
This will also fix #681

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -400,6 +400,10 @@ func createDevice(device *v1alpha1.Device) types.Device {
 	for i, dtwin := range device.Status.Twins {
 		expected := &types.TwinValue{}
 		expected.Value = &device.Status.Twins[i].Desired.Value
+		metadataType, ok := device.Status.Twins[i].Desired.Metadata["type"]
```
sids-b: Does CRD definition support "Type" as well ??

subpathdev: no we are using a map there.
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/apis/devices/v1alpha1/device_instance_types.go#L145

sujithsimon22: No the CRD definition is generic and accepts map[string]string 
But we given an examples in sample crds which mentions the same https://github.com/kubeedge/kubeedge/blob/1a0d4e230e23036fc584991a95f06c968422e268/build/crd-samples/devices/create-device-instance.yaml#L34

sids-b: OK

sids-b: We need to mention all these points in usage docs. It can be done in future PRs

sujithsimon22: +1

PR #713, opened 2019-06-17
Edgemesh dev

Arcanique:
…,including init container mirroring tools.

**What type of PR is this?**
>
> /kind feature

>
> In the edgemesh scenario, implement application traffic hijacking which contain DNS inqury using init container and iptables tecnology and contains the init container image creation script containing the Dockerfile for deb and rpm. Currently supports scenarios which the POD and the host's network namespace are isolated (for example, container networks such as bridges and flannels). We support IPv4 messages now .
>

Fixes #


sids-b:
@Arcanique , can you please fix the CI ?

sids-b:
@Arcanique please use dep to update the vendor packages. https://github.com/golang/dep

Arcanique:
> @Arcanique please use dep to update the vendor packages. https://github.com/golang/dep

ok，I will do that，thx！

PR #714, opened 2019-06-17
minor fixes message build / parsing

### LINKED ISSUE #709: updating device values
**What happened**:
```sh
controller/upstream.go:169 Failed to patch device status &{{[{CPU_Temperatur { map[]} {161 map[timestamp:1560756072724 type:string]}} {cpu_id { map[]} {0 map[timestamp:1560756072724 type:string]}}]}} of device cpu-sensor-tag01 in namespace default
```

the concrete problem seems to be this: 
```sh
status.twins.desired.value in body is required
```

I've create an minimal example to understand this problem: https://play.golang.org/p/JepBWmp_YxD
I think the problem is this:
 - json.Marshal doesn't remove the desired field

An possible solution is could the Usage of Struct-Pointer as TwinValues. 

Updating the values of a device in the kuberentes API should run successfully

1. setting up kuberenetes api-server
2. kubectl apply -f build/crd.samples/devices/led-light-device-model.yaml
3. kubectl apply -f build/crd.samples/devices/led-light-device-instance.yaml
4. kubectl edit device led-light-instance-01 
   a. adding desired: {} to power-status

**Environment**:
- KubeEdge version: git commit hash: 1a0d4e230e23036fc584991a95f06c968422e268
- OS (e.g. from /etc/os-release): not relevant
- kubectl version:
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:44:30Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:36:19Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}

subpathdev:
**What type of PR is this?**
/kind bug

first commit see issue #709 
second commit fix parsing message from cloud to edge on updating device twin
        -> we have already restrictions on empty string, so this is only another one
third commit create only a delta message if no not an empty string is set
        -> we have already restrictions on empty string, so this is only another one

Fixes #709

### DIFF
--- a/cloud/pkg/devicecontroller/apis/devices/v1alpha1/device_instance_types.go
+++ b/cloud/pkg/devicecontroller/apis/devices/v1alpha1/device_instance_types.go
@@ -139,7 +139,7 @@ type Twin struct {
 // TwinProperty represents the device property for which an Expected/Actual state can be defined.
 type TwinProperty struct {
 	// Required: The value for this property.
-	Value string `json:"value,omitempty"`
+	Value string `json:"value,"`
 	// Additional metadata like timestamp when the value was reported etc.
 	// +optional
 	Metadata map[string]string `json:"metadata,omitempty"`
--- a/edge/pkg/devicetwin/dttype/types.go
+++ b/edge/pkg/devicetwin/dttype/types.go
@@ -263,17 +263,21 @@ func UnmarshalDeviceTwinUpdate(payload []byte) (*DeviceTwinUpdate, error) {
 		if value != nil {
 			if value.Expected != nil {
 				if value.Expected.Value != nil {
-					match := dtcommon.ValidateTwinValue(*value.Expected.Value)
-					if !match {
-						return &deviceTwinUpdate, errors.New(errorValue)
+					if *value.Expected.Value != "" {
+						match := dtcommon.ValidateTwinValue(*value.Expected.Value)
+						if !match {
+							return &deviceTwinUpdate, errors.New(errorValue)
+						}
 					}
 				}
 			}
 			if value.Actual != nil {
 				if value.Actual.Value != nil {
-					match := dtcommon.ValidateTwinValue(*value.Actual.Value)
-					if !match {
-						return &deviceTwinUpdate, errors.New(errorValue)
+					if *value.Actual.Value != "" {
+						match := dtcommon.ValidateTwinValue(*value.Actual.Value)
+						if !match {
+							return &deviceTwinUpdate, errors.New(errorValue)
+						}
 					}
 				}
 			}
--- a/edge/pkg/devicetwin/dttype/types_helper.go
+++ b/edge/pkg/devicetwin/dttype/types_helper.go
@@ -307,7 +307,9 @@ func BuildDeviceTwinDelta(baseMessage BaseMessage, twins map[string]*MsgTwin) ([
 				}
 				if strings.Compare(expectedValue, actualValue) != 0 {
 					value := expectedValue
-					delta[k] = value
+					if expectedValue != "" {
+						delta[k] = value
+					}
 				}
 			} else {
 				if expectedValue != "" {


cloud/pkg/devicecontroller/apis/devices/v1alpha1/device_instance_types.go
```diff
@@ -139,7 +139,7 @@ type Twin struct {
 // TwinProperty represents the device property for which an Expected/Actual state can be defined.
 type TwinProperty struct {
 	// Required: The value for this property.
-	Value string `json:"value,omitempty"`
+	Value string `json:"value,"`
```
sujithsimon22: @subpathdev does this change warrant a change in https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/apis/devices/v1alpha1/zz_generated.deepcopy.go
Could you please run this script https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/apis/devices/v1alpha1/zz_generated.deepcopy.go and verify if the deepcopy file is changed ?

subpathdev: I didn't change https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/apis/devices/v1alpha1/zz_generated.deepcopy.go. Why we have to change this file? I have a running example with the changes which are provided by this pr. (Only this changes)

I have also executed the update-codegen.sh script and nothing has changed.

sujithsimon22: Usually when we modify the structures in either the device_instance_types or device_model_types, it requires changes in the deepcopy file which is a generated file.

sujithsimon22: If the script has not changed your deepcopy file I guess there is no change required in the deepcopy file for this change

sids-b:
@subpathdev , can you squash the commits please ?

subpathdev:
@sids-b yes of course

subpathdev:
@sids-b @rohitsardesai83 please merge this into release-1.0. If you don't like to merge there could be the problem that the device reported value in the cloud could not be updated and error in the cloud will be thrown.

rohitsardesai83:
@sujithsimon22 , please check if you faced a similar problem as described.

sids-b:
@subpathdev , status.twins.reported.value in body is required right ? without that value it is currently not able to patch correct ?

subpathdev:
@sids-b no status.twins.desired.value in body is required. The problem is described in issue #709.

sids-b:
Actually if we mention desired, the value field is compulory. If we want to keep it empty, remove the entire desired section in yaml. This is added as part of validation.

sids-b:
Can you please check that if we remove the entire desired section, does it work ? If no, we will merge this PR in release-1.0 branch. Else we can open the topic for discussion

sujithsimon22:
> Actually if we mention desired, the value field is compulory. If we want to keep it empty, remove the entire desired section in yaml. This is added as part of validation.

+1

subpathdev:
the problem is not in the yaml!  I didn't add the desired section in the deviceInstance yaml.  The problem occurs if the cloud are patching the device to update the reported value.

subpathdev:
To explain the behavior in go an struct is never empty, so we have to use pointers as TwinProperty or we are setting the Value in the desired field. (You can see a minimal example on this here: https://play.golang.org/p/JepBWmp_YxD)

subpathdev:
> Can you please check that if we remove the entire desired section, does it work ? If no, we will merge this PR in release-1.0 branch. Else we can open the topic for discussion

Yes this is also a possible solution on this. But If we are removing the complete desired field from we are loosing an important feature of kubeedge. (Sending a value from cloud to edge to device)

sids-b:
@subpathdev , you are right

Issue #341, opened 2019-04-11
UT is missing in cloud part


m1093782566:
**What would you like to be added**:

Add UTs for cloud part and run them in the CI.

**Why is this needed**:

UT is very basic tests.

m1093782566:
cc @lidiyag @pavan187 @naveensriram

m1093782566:
Also what can we do for integration tests?

pavan187:
@m1093782566 , Integration is not required for this, as the cloud part test already covering in E2E framework tests. we do not need have cloud part seperately.

To do only cloud part integration test we need to mock K8s API-Server, which is difficult to handle as we need to study entire k8s API-server code. so we decided to handle the cloud part tests along with E2E tests.
cc @sids-b @rohitsardesai83

woshiyezitang:
膜拜大佬

rohitsardesai83:
we need to see how kubernetes handles such tests and try to attempt and write whatever we can .

PR #717, opened 2019-06-18
adding unit tests cloud

subpathdev:
**What type of PR is this?**
/kind test

we have already unit tests in the cloud part this should be tested on every pr.
updating #341 

### DIFF
--- a/cloud/Makefile
+++ b/cloud/Makefile
@@ -1,11 +1,14 @@
-
 # make edgecontroller
-.PHONY: default edgecontroller
+TEST_DIR ?= "./"
+.PHONY: default edgecontroller test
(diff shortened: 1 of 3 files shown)


cloud/Makefile
```diff
@@ -7,6 +6,9 @@ lint:
 	golangci-lint run --disable-all -E gofmt -E golint ./...
 	go vet ./...
 
+test:
+	find ${TEST_DIR} -name "*_test.go"| xargs -I{} dirname {} | uniq | xargs -I{} go test ${T} {}
```
kevin-wangzefeng: Better set a default value for $TEST_DIR.

PR #727, opened 2019-06-19
Service

sids-b:
**What type of PR is this?**
/kind feature

This PR will get the cloud side impl of edgemesh in the edgemesh branch. 

Note: the PR has some extra commits as my branch is more new that the edgemesh branch. But it should not cause conflict while raising PR from edgemeshdev to master


sids-b:
UT is failing in this PR because the vendor has been updated...

PR #735, opened 2019-06-20
Handled case in bluetooth mapper with no converter information while reading

### LINKED ISSUE #734: Bluetooth mapper trying to use data converter for reading even when data converter not specified
**What happened**:
Bluetooth mapper trying to use data converter for reading even when data converter is not specified in the device model

Bluetooth mapper should report the data that is read as it is without converting it if no conversion data is specified

Setup Kubeedge
Apply device model and device for bluetooth without any converter information
Run bleutooth mapper

sujithsimon22:
**What type of PR is this?**
/kind bug

This PR handles case in bluetooth mapper when no data converter is given while reading

Fixes #734

### DIFF
--- a/device/bluetooth_mapper/configuration/config.go
+++ b/device/bluetooth_mapper/configuration/config.go
@@ -147,7 +147,8 @@ func (b *BLEConfig) Load() error {
 					return errors.New("Error in unmarshalling data property visitor configuration: " + err.Error())
 				}
 				action.Operation.CharacteristicUUID = bluetoothPropertyVisitor.CharacteristicUUID
-				if !reflect.DeepEqual(bluetoothPropertyVisitor.BluetoothDataConverter, nil) {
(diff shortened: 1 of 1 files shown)


device/bluetooth_mapper/configuration/config.go
```diff
@@ -147,7 +147,8 @@ func (b *BLEConfig) Load() error {
 					return errors.New("Error in unmarshalling data property visitor configuration: " + err.Error())
 				}
 				action.Operation.CharacteristicUUID = bluetoothPropertyVisitor.CharacteristicUUID
-				if !reflect.DeepEqual(bluetoothPropertyVisitor.BluetoothDataConverter, nil) {
+				newBluetoothVisitorConfig:=VisitorConfigBluetooth{}
```
rohitsardesai83: gofmt please.

sujithsimon22: done

PR #784, opened 2019-06-24
using std. way to query nodes

subpathdev:
**What type of PR is this?**
/kind cleanup

Using the provided way to query nodes of client-go.
This will also remove issues with pod deployment. In the last few days I had issues to deploy pods, they could be referenced that no Added, Modified or Deleted event could be recognized.

Fixes #


subpathdev:
this will create more issues

PR #809, opened 2019-06-25
Refactor bluetooth mapper to discover characteristics only once

### LINKED ISSUE #808: Bluetooth mapper should only discover the characteristics once
Bluetooth mapper currently discovers characteristics each time before performing operations, this is not an efficient method. It should only discover it once at the beginning and reuse the same information 

**Why is this needed**:
Bluetooth mapper currently discovers characteristics each time before performing operations, this is not an efficient method and may cause performance issues.

sujithsimon22:
**What type of PR is this?**
 /kind cleanup

This PR contains code to ensure that the  bluetooth mapper discovers characteristics only once at the beginning and reuse the same information to perform actions

Fixes #808

### DIFF
--- a/device/bluetooth_mapper/watcher/watcher.go
+++ b/device/bluetooth_mapper/watcher/watcher.go
@@ -18,6 +18,7 @@ package watcher
 
 import (
 	"errors"
+	"os"
 	"reflect"
 	"strings"
 	"time"
@@ -100,8 +101,22 @@ func onPeripheralDisconnected(p gatt.Peripheral, err error) {
 //onPeripheralConnected contains the operations to be performed as soon as the peripheral device is connected
 func (w *Watcher) onPeripheralConnected(p gatt.Peripheral, err error) {
 	actionmanager.GattPeripheral = p
-	DeviceConnected <- true
 	helper.ChangeDeviceState("online", deviceID)
+	ss, err := p.DiscoverServices(nil)
+	if err != nil {
+		glog.Errorf("Failed to discover services, err: %s\n", err)
+		os.Exit(1)
+	}
+	for _, s := range ss {
+		// Discovery characteristics
+		cs, err := p.DiscoverCharacteristics(nil, s)
+		if err != nil {
+			glog.Errorf("Failed to discover characteristics for service %s, err: %v\n", s.Name(), err)
+			continue
+		}
+		actionmanager.CharacteristicsList = append(actionmanager.CharacteristicsList, cs...)
+	}
+	DeviceConnected <- true
 	for {
 		newWatcher := &Watcher{}
 		if !reflect.DeepEqual(w, newWatcher) {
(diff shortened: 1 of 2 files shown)


sujithsimon22:
cc @sids-b @rohitsardesai83  for review

device/bluetooth_mapper/watcher/watcher.go
```diff
@@ -100,8 +100,21 @@ func onPeripheralDisconnected(p gatt.Peripheral, err error) {
 //onPeripheralConnected contains the operations to be performed as soon as the peripheral device is connected
 func (w *Watcher) onPeripheralConnected(p gatt.Peripheral, err error) {
 	actionmanager.GattPeripheral = p
-	DeviceConnected <- true
 	helper.ChangeDeviceState("online", deviceID)
+	ss, err := p.DiscoverServices(nil)
+	if err != nil {
+		glog.Errorf("Failed to discover services, err: %s\n", err)
```
rohitsardesai83: we should return on error if we fail to discover services.

device/bluetooth_mapper/watcher/watcher.go
```diff
@@ -100,8 +100,21 @@ func onPeripheralDisconnected(p gatt.Peripheral, err error) {
...
+	}
+	for _, s := range ss {
+		// Discovery characteristics
+		cs, err := p.DiscoverCharacteristics(nil, s)
+		if err != nil {
+			glog.Errorf("Failed to discover characteristics, err: %s\n", err)
```
rohitsardesai83: %s -> %v. also we can log the service name for which the characteristics could not be discovered.

sujithsimon22: Done, thanks

device/bluetooth_mapper/watcher/watcher.go
sujithsimon22: Done, thanks

device/bluetooth_mapper/watcher/watcher.go
```diff
@@ -100,8 +101,22 @@ func onPeripheralDisconnected(p gatt.Peripheral, err error) {
 //onPeripheralConnected contains the operations to be performed as soon as the peripheral device is connected
 func (w *Watcher) onPeripheralConnected(p gatt.Peripheral, err error) {
 	actionmanager.GattPeripheral = p
-	DeviceConnected <- true
 	helper.ChangeDeviceState("online", deviceID)
+	ss, err := p.DiscoverServices(nil)
+	if err != nil {
+		glog.Errorf("Failed to discover services, err: %s\n", err)
+		os.Exit(1)
```
rohitsardesai83: right now our mapper manages only one device at a time right ? if we support multiple devices , then the watcher should not exit if it fails to connect to one device.  are we tracking future mapper improvements via open issues ?
please create issues for all such improvements.

sujithsimon22: we have an open umbrella issue to track future improvements in the mapper. We can add support for multiple devices as one item in it

sids-b:
@sujithsimon22 squash the commits please

sujithsimon22:
> @sujithsimon22 squash the commits please

Done :)

PR #812, opened 2019-06-25
updating error message with keadm

subpathdev:
**What type of PR is this?**
/kind bug

we have now three commands with `make WHAT=cloud` `make WHAT=edge` `make WHAT=keadm` 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,12 +16,15 @@ all:
 	cd edge && $(MAKE)
 else ifeq ($(WHAT),keadm)
 all:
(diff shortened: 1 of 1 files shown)


sids-b:
I think make currently builds edgesite as well

subpathdev:
> I think make currently builds edgesite as well

Yes you are right. We should add `make WHAT=edgesite`.

subpathdev:
I added `make WHAT=edgesite` as an extra option.

PR #835, opened 2019-06-30
Return nil instead the err as it must be nil

hainesc:
**What type of PR is this?**
/kind cleanup

Cleanup

Fixes #

Cleanup


hainesc:
/assign @skdwriting
as suggestion from @kubeedge-bot

PR #838, opened 2019-07-01
Use switch instead of if-else

hainesc:
**What type of PR is this?**
/kind cleanup

It looks more professional

Fixes #


hainesc:
> /approve

Thanks, but it does not work.

PR #840, opened 2019-07-01
Improve the poor code

hainesc:
**What type of PR is this?**
/kind cleanup

The poor code quality...

Fixes #

### DIFF
--- a/edge/pkg/edgehub/controller.go
+++ b/edge/pkg/edgehub/controller.go
@@ -26,9 +26,13 @@ const (
 
 var (
 	authEventType = "auth_info_event"
-	groupMap      = map[string]string{"resource": modules.MetaGroup,
-		"twin": modules.TwinGroup, "app": "sync",
-		"func": modules.MetaGroup, "user": modules.BusGroup}
+	groupMap      = map[string]string{
+		"resource": modules.MetaGroup,
+		"twin":     modules.TwinGroup,
+		"app":      "sync",
+		"func":     modules.MetaGroup,
+		"user":     modules.BusGroup,
+	}
 
 	// clear the number of data of the stop channel
 	times = 2


hainesc:
BTW, I am surveying an edge solution, but the poor code quality makes me hesitate to use KubeEdge.

sids-b:
> BTW, I am looking for an edge solution, but the poor code quality makes me hesitate to use KubeEdge.

Looking forward for more contributions from you which will help enhance the quality :)

hainesc:
I don't like the `context.Context` used in KubeEdge, it is `"github.com/kubeedge/beehive/pkg/core/context"` not `context` of golang library. It is confuse and makes the code less readable.

Why not `context` of the golang library?

sids-b:
> I don't like the `context.Context` used in KubeEdge, it is `"github.com/kubeedge/beehive/pkg/core/context"` not `context` of golang library. It is confuse and makes the code less readable.
> 
> Why not `context` of the golang library?

Beehive is an in-process cross module communication framework. The modules in the same beehive context can send messages to each other and is different from golang "context".

hainesc:
> Beehive is an in-process cross module communication framework. The modules in the same beehive context can send messages to each other and is different from golang "context".

So, we should use another name. `context` in golang is used widely.

qizha:
We should named the package to make it describe itself. There is no rule to avoid using the same name with standard package. Actually, looking into some other golang project, you will be able to find there are lots of packages with the same name with standard package. For example, "github.com/google/certificate-transparency/go/asn1", "github.com/google/certificate-transparency/go/x509", they all have the same name with standard package. Moreover, it is the best practice to keep your models in "types" package and API in "api" package. So you will find lots of packages with such names in different projects, like Docker, Kubernetes.
The name "context" within the "beehive" is the best name to describe this package which represent a cross module communication. Refer to "https://rakyll.org/style-packages/" what we should do is to rename this package or the go standard package when importing them. Let's find where we introduced two package(beehive/context and standard context) and rename the standard one to gocontext or the beehive one to beehivecontext.

You are always welcomed to join the community to improve the code quality. Or you can open issues to indicate the bad ones you found so that anyone else can also fix them. Thanks

@hainesc

PR #844, opened 2019-07-02
Keadm error handling improvements

### LINKED ISSUE #698: keadm should not ignore errors
**What happened**:

in many cases, keadm ignores errors and continues.

it should stop processing and prompt the user.

**Environment**:
- KubeEdge version: 0.3.0
- Hardware configuration:
- Kernel (e.g. `uname -a`):
- Others:

yamt:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

xref #698 
(partially)

### DIFF
--- a/keadm/app/cmd/cloud/init.go
+++ b/keadm/app/cmd/cloud/init.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cmd
 
 import (
-	"fmt"
 	"io"
 
 	"github.com/spf13/cobra"
@@ -62,13 +61,13 @@ func NewCloudInit(out io.Writer, init *types.InitOptions) *cobra.Command {
 		Short:   "Bootstraps cloud component. Checks and install (if required) the pre-requisites.",
 		Long:    cloudInitLongDescription,
 		Example: cloudInitExample,
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			checkFlags := func(f *pflag.Flag) {
 				util.AddToolVals(f, flagVals)
 			}
 			cmd.Flags().VisitAll(checkFlags)
 			Add2ToolsList(tools, flagVals, init)
-			Execute(tools)
+			return Execute(tools)
 		},
 	}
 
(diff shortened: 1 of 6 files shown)


sids-b:
@pavan187 @srivatsav123 @subpathdev @kadisi , any idea why CI does not run for installer code ?

kadisi:
perhaps we need check `TRAVIS_COMMIT_RANGE `variable.

```
before_install:
  - |
      echo "TRAVIS_COMMIT_RANGE:"$TRAVIS_COMMIT_RANGE
      if ! git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(.md)|(.png)|(.pdf)|(.html)|^(LICENSE)|^(docs)'
      then
        echo "Only doc files were updated, skip running the CI."
        exit
      fi

```

subpathdev:
Yes, the problem is that the commit range is to short. I will create a fix for this problem. 
(First possible solution: comparing to origin/master) 
(Second: finding the commit, which is in origin/master and the current branch and comparing with them)

keadm/app/cmd/cloud/init.go
```diff
@@ -62,13 +61,13 @@ func NewCloudInit(out io.Writer, init *types.InitOptions) *cobra.Command {
 		Short:   "Bootstraps cloud component. Checks and install (if required) the pre-requisites.",
 		Long:    cloudInitLongDescription,
 		Example: cloudInitExample,
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
```
rohitsardesai83: why is this renamed as RunE ?

yamt: Run and RumE are separate fields of the struct. See the Cobra documentation.
I guess E stands for error.

keadm/app/cmd/util/k8sinstaller.go
```diff
@@ -49,7 +49,8 @@ func (ks *K8SInstTool) InstallTools() error {
 	case types.VersionNAInRepo:
 		return fmt.Errorf("Expected %s version is not available in OS repo", component)
 	case types.AlreadySameVersionExist:
-		return fmt.Errorf("Same version of %s already installed in this host", component)
+		fmt.Println("Same version of %s already installed in this host", component)
```
rohitsardesai83: Same version of %s  *is* already installed .
similarly at other places .

yamt: do you mean to improve the message? i feel it's a separate change.

rohitsardesai83: ok , we can have a separate PR for this.

PR #848, opened 2019-07-02
fix WHAT string edgesite

subpathdev:
**What type of PR is this?**
/kind bug

possibility to execute ```make WHAT=edgesite```

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ else ifeq ($(WHAT),edge)
 all:
 # make all what=edge, build edge binary
 	cd edge && $(MAKE)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,7 +14,7 @@ else ifeq ($(WHAT),edge)
 all:
 # make all what=edge, build edge binary
 	cd edge && $(MAKE)
-else ifeq ($(WHAT),keadm)
+else ifeq ($(WHAT),eadgesite)
```
sids-b: s/eadgesite/edgesite

PR #854, opened 2019-07-03
Add stop channel to controller&cloudhub

fisherxu:
**What type of PR is this?**

> /kind bug

Fixes #

/cc @rohitsardesai83 @sids-b

### DIFF
--- a/cloud/pkg/controller/controller.go
+++ b/cloud/pkg/controller/controller.go
@@ -12,7 +12,9 @@ import (
 )
 
 // Controller use beehive context message layer
-type Controller struct{}
+type Controller struct {
+	stopChan chan bool
+}
 
 func init() {
 	edgeController := Controller{}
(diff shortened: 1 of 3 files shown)


cloud/pkg/controller/controller.go
```diff
@@ -32,30 +34,33 @@ func (ctl *Controller) Group() string {
...
+		log.LOGGER.Warnf("new downstream controller failed with error: %s", err)
+		os.Exit(1)
 	}
+	downstream.Start()
+
+	<-ctl.stopChan
+	log.LOGGER.Warnf("stopping upstream controller...")
+	upstream.Stop()
+	log.LOGGER.Warnf("upstream controller stopped")
+	log.LOGGER.Warnf("elected slave, stopping downstream controller...")
```
rohitsardesai83: I don't think we have any leader election mechanism , probably we can delete this log.

fisherxu: Agree, Done, ptal. We can add the leader election in the future.

PR #859, opened 2019-07-04
Add feature : Pass through the DNS request if the client does not access the services inside the cluster

arcanique:
…ess the services inside the cluster

**What type of PR is this?**
>
> /kind feature

>
>  Pass through the DNS request if the client does not access the services inside the cluster

Fixes #

### DIFF
--- a/edge/pkg/metamanager/msg_processor.go
+++ b/edge/pkg/metamanager/msg_processor.go
@@ -115,7 +115,7 @@ func parseResource(resource string) (string, string, string) {
 
 // is resource type require remote query
 func requireRemoteQuery(resType string) bool {
-	return resType == model.ResourceTypeConfigmap || resType == model.ResourceTypeSecret || resType == constants.ResourceTypeService || resType == constants.ResourceTypeEndpoints
+	return resType == model.ResourceTypeConfigmap || resType == model.ResourceTypeSecret || resType == constants.ResourceTypeEndpoints
 }
 
 func isConnected() bool {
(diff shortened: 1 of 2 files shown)


edgemesh/pkg/server/dns.go
```diff
@@ -257,19 +272,30 @@ func (q *dnsQuestion) getQName(req []byte, offset uint16) uint16 {
 
 // lookupFromMetaManager implement confirm the service exists
 // for earlier version returns true by default
```
anyushun: Now we need to delete this line

arcanique: k

edge/pkg/metamanager/msg_processor.go
```diff
@@ -115,7 +115,7 @@ func parseResource(resource string) (string, string, string) {
 
 // is resource type require remote query
 func requireRemoteQuery(resType string) bool {
-	return resType == model.ResourceTypeConfigmap || resType == model.ResourceTypeSecret || resType == constants.ResourceTypeService || resType == constants.ResourceTypeEndpoints
+	return resType == model.ResourceTypeConfigmap || resType == model.ResourceTypeSecret || resType == constants.ResourceTypeEndpoints
```
sids-b: We support process remoteQuery for Service. Why are we removing it ?

arcanique: There are several reasons for the changes here. First，we have to get the service endpoint every time. if we get it from the cloud every time, it is very time consuming and extremely inefficient.  Second, the update of service and endpoint is actually notified by the cloud, we did not take the initiative to watch. Every time we go to the cloud to query, and the related resource information has not changed, the operation of the active query is actually redundant. Therefore, the operation of supporting the service to query on the cloud is removed

sids-b: OK

PR #870, opened 2019-07-05
fix bug #866

### LINKED ISSUE #866: edge_core cannot connect to edgecontroller after disconnecting once
**What happened**:
Stopped and started edge_core, it caused problem while creating websocket connection

Connect successfully via websocket

1) Start edgecontroller
2) Start edge_core
3) Stop edge_core
4) Start edge_core again
Error logs will come in both edgecontroller and edge_corethen

Looks like some issue in closing of websocket connection

**Environment**:
- KubeEdge version: latest master

shouhong:
**What type of PR is this?**
/kind bug

1. When edge node disconnecting from edgecontroller, the edgecontroller does not aware this until the next keepalive message is not received from the edge node. 
2. The edge node connection information is not cleaned up until the edgecontroller aware the edge node is not alive.
3. If the edge node connect to edgecontroller again before the previous obsolete connection information was cleaned up, it will get an error.
4. To fix this problem, add retry mechanism to wait for the previous obsolete connection information to be cleanup.

Fixes #866

### DIFF
--- a/cloud/pkg/cloudhub/handler/eventhandler.go
+++ b/cloud/pkg/cloudhub/handler/eventhandler.go
@@ -244,7 +244,14 @@ func (eh *EventHandle) ServeConn(hi hubio.CloudHubIO, info *emodel.HubInfo) {
 
 // EnrollNode enroll node for the incoming connection
 func (eh *EventHandle) EnrollNode(hi hubio.CloudHubIO, info *emodel.HubInfo) error {
-	err := eh.EventQueue.Connect(info)
+	// Wait for the previous connection to be cleaned up
+	var err error
+	for i := 0; i <= eh.KeepaliveInterval; i++ {
+		if err = eh.EventQueue.Connect(info); err == nil {
+			break
+		}
+		time.Sleep(time.Duration(1) * time.Second)
+	}
 	if err != nil {
 		bhLog.LOGGER.Errorf("fail to connect to event queue for node %s, reason %s", info.NodeID, err.Error())
 		notifyEventQueueError(hi, eventQueueDisconnect, info.NodeID)


sids-b:
just restarted the test

sids-b:
@kexun @shouhong , I just had 1 question. In websocket, if a client or server loses connection with other, I think it immediately gets notified and err is returned. The previous implementation was something similar. Can we not have something similar in viaduct as well ? I remember that in previous implementation it didn't need to wait for keepalive interval to remove the connection. Any thoughts ?

shouhong:
> @kexun @shouhong , I just had 1 question. In websocket, if a client or server loses connection with other, I think it immediately gets notified and err is returned. The previous implementation was something similar. Can we not have something similar in viaduct as well ? I remember that in previous implementation it didn't need to wait for keepalive interval to remove the connection. Any thoughts ?

@sids-b 
1. In previous implementation, there is a "EventReadLoop" function in which there is loop to **actively** read data from edge node (the timeout value is keepalive interval).  So if the edge node loses connection, the read function will immediately get an error. 
2. After switching to viaduct, we follow the implementation patten in the viaduct example (https://github.com/kubeedge/viaduct/blob/master/examples/chat/server.go). In this example, the server side register a callback function named "handleServer" to **passively** read data from client side. If the client loses connection, the server is not notified immediately. 
3. Do you think we need to follow the previous pattern to actively read data from edge node?

sids-b:
@kexun your thoughts ? I think we may lose some messages in the KeepAliveInterval so an active read loop would be a good option for websocket ??

sids-b:
@rohitsardesai83 @qizha , thoughts ??

sids-b:
I think lets get this fix in first and later we can think about comparing implementation.

/lgtm

PR #876, opened 2019-07-08
Add dockerfile for edgesite

### LINKED ISSUE #875: Add Dockerfile for Edgesite
Currently we do not have a dockerfile present for edgesite. There needs to be a Dockerfile added for edgesite which would be helpful to users who want to run edgesite as a container as opposed to running it as a binary.

**Why is this needed**:
A Dockerfile for edgesite would be helpful to users who want to run edgesite as a container as opposed to running it as a binary.

sujithsimon22:
**What type of PR is this?**
 /kind feature

This PR adds a Dockerfile for edgesite and a provision in makefile using which the user will be able to build a docker image for edgesite easily

Fixes #875 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,17 @@ edgeimage:
 	--build-arg RUN_FROM=${ARCH}/docker:dind \
 	-f build/edge/Dockerfile .
 
+.PHONY: edgesiteimage
+edgesiteimage:
+	mkdir -p ./build/edgesite/tmp
+	rm -rf ./build/edgesite/tmp/*
+	curl -L -o ./build/edgesite/tmp/qemu-${QEMU_ARCH}-static.tar.gz https://github.com/multiarch/qemu-user-static/releases/download/v3.0.0/qemu-${QEMU_ARCH}-static.tar.gz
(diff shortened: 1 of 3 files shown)


sujithsimon22:
cc @kevin-wangzefeng  @rohitsardesai83 @sids-b

Makefile
```diff
@@ -91,6 +91,17 @@ edgeimage:
 	--build-arg RUN_FROM=${ARCH}/docker:dind \
 	-f build/edge/Dockerfile .
 
+.PHONY: edgesiteimage
+edgesiteimage:
+	mkdir -p ./build/edgesite/tmp
+	rm -rf ./build/edgesite/tmp/*
```
rohitsardesai83: first we create the tmp and then delete nothing inside it ?

sujithsimon22: This is to handle cases in which the image has already been built before i.e. the tmp folder has contents inside. Something similar to this has been done while building edge_core image as well  https://github.com/kubeedge/kubeedge/blob/4d44da1ce4d9c001a814beaa44d0b32e257554db/Makefile#L86

rohitsardesai83: then why are creating the directory ?

sujithsimon22: the tmp folder does not exist by default, we create it the first we build the image and used for 
 downloading and using QEMU

PR #882, opened 2019-07-10
keadm support to fix panic error when port field is not mentioned during keadm join and reset

srivatsav123:
**What type of PR is this?**

 /kind bug

keadm support to fix panic error when port field is not mentioned during keadm join and reset

Fixes #

keadmjoin
keadmreset


rohitsardesai83:
@srivatsav123 , please create an issue to track this.

subpathdev:
https://github.com/kubeedge/kubeedge/blob/4315a2ce5f5772c4c719e2a00038f11fcdedae32/keadm/app/cmd/util/edgecoreinstaller.go#L155
has the same if condition as line 150; please merge the if body

rohitsardesai83:
@srivatsav123 , please address the review comments.

sids-b:
Lets get this one merged as it is a non-blocking code issue and we can fix it in further PRs

PR #885, opened 2019-07-10
Fix the bug #884

### LINKED ISSUE #884: Edge_core crashing if there is no bridge decvice when starting
**What happened**:
Edge_core crashing if there is no bridge decvice when starting

Edge_core waiting for creating bridge device when starting

delete the bridge device "docker0", and restarting the edge_core program.

**Environment**:
- KubeEdge version:
- Hardware configuration:
- Kernel (e.g. `uname -a`):
- Others:

arcanique:
**What type of PR is this?**
>
> /kind bug

>
> Fix the bug #884

Fixes #884 

### DIFF
--- a/edgemesh/pkg/server/dns.go
+++ b/edgemesh/pkg/server/dns.go
@@ -76,20 +76,26 @@ func DnsStart() {
 
 // getDnsServer returns the specific interface ip of version 4
 func getDnsServer() (net.IP, error) {
-	ifaces, err := net.InterfaceByName(inter)
-	if err != nil {
-		return nil, err
-	}
-
-	addrs, _ := ifaces.Addrs()
-
-	for _, addr := range addrs {
-		if ip, inet, _ := net.ParseCIDR(addr.String()); len(inet.Mask) == 4 {
-			return ip, nil
+	for {
+		ifaces, err := net.InterfaceByName(inter)
+		if err != nil {
+			log.LOGGER.Warnf("get interface error : %s", err)
+			time.Sleep(time.Second * 3)
+			continue
 		}
+
+		addrs, _ := ifaces.Addrs()
+
+		for _, addr := range addrs {
+			if ip, inet, _ := net.ParseCIDR(addr.String()); len(inet.Mask) == 4 {
+				return ip, nil
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


edgemesh/pkg/server/dns.go
```diff
@@ -76,20 +76,26 @@ func DnsStart() {
...
 func getDnsServer() (net.IP, error) {
-	ifaces, err := net.InterfaceByName(inter)
-	if err != nil {
-		return nil, err
-	}
+	for {
+		ifaces, err := net.InterfaceByName(inter)
+		if err != nil {
+			log.LOGGER.Warnf("get interface error : %s", err)
+			time.Sleep(time.Second * 5)
```
Baoqiang-Zhang: why sleep here?

edgemesh/pkg/server/dns.go
```diff
@@ -76,20 +76,26 @@ func DnsStart() {
...
-		if ip, inet, _ := net.ParseCIDR(addr.String()); len(inet.Mask) == 4 {
-			return ip, nil
+		for _, addr := range addrs {
+			if ip, inet, _ := net.ParseCIDR(addr.String()); len(inet.Mask) == 4 {
+				return ip, nil
+			}
 		}
+
+		log.LOGGER.Warnf("the interface " + inter + " has not config ip of version 4")
+		time.Sleep(time.Second * 5)
```
Baoqiang-Zhang: ditto

edgemesh/pkg/server/dns.go
arcanique: The main reason for this bug is that containerd will only create network devices when the Pod is first created. Therefore, during the period of edge_core startup, it will always loop execution. The purpose of adding sleep is to make the cpu not too high. When you deploy the first Pod, this function will detect the corresponding device and exit normally.

edgemesh/pkg/server/dns.go
arcanique: The reason is the same as above.

edgemesh/pkg/server/dns.go
rohitsardesai83: can we reduce the sleep time ? 5 seconds looks high.

arcanique: I can reduce this time to 3 seconds.

PR #886, opened 2019-07-10
Adding extended header parameters

lidiyag:
**What type of PR is this?**
kind bug

This PR adds some external header parameters to the websocket request send to cloud.


lidiyag:
These headers are no longer used in current code and hence closing this PR.

PR #891, opened 2019-07-11
add change model message to protobuf message design

### LINKED ISSUE #399: Change model message and its content to protobuf based

Currently, in beehive framework, model message is defined and serialized/deserialized using json format. With it, for parsing the message, each field index is hardcoded (see https://github.com/kubeedge/beehive/blob/master/pkg/common/util/parse_resource.go). 

**Why is this needed**:
This adds a strong limit on how the message will look like and prevent any future extension. For example, we want to add clusterID info., etc. 

On the other hand, protobuf is extensible with field type/name defined. 
Plus the json serialization/deserialization performance is slower than protobuf

kadisi:

**What type of PR is this?**

/kind feature

Fixes #399 

ref: https://github.com/kubeedge/beehive/pull/24

### DIFF
--- a/docs/proposals/message-proto-design.md
+++ b/docs/proposals/message-proto-design.md
@@ -1,0 +1,298 @@
+---
+title: change model message to  protobuf message 
+authors:
+    - "@kadisi"
+    - "@rohitsardesai83"
+approvers:
+  - "@CindyXing"
+  - "@sids-b"
+  - "@qizha"
+  - "kevin-wangzefeng"
+creation-date: 2019-07-11
+last-updated: 2019-07-24
+status: alpha
+---
+
+# change model message to protobuf message 
+
+* [change model message to protobuf message](#change-model-message-to-protobuf-message)
+   * [Abstract](#abstract)
+   * [Need to do](#need-to-do)
+   * [How to do](#how-to-do)
+   * [Change the model message to protobuf message  directly](#change-the-model-message-to-protobuf-message--directly)
+   * [Solve the filed index problem of Resource](#solve-the-filed-index-problem-of-resource)
+   * [Consider compatibility](#consider-compatibility)
+* [How to use in KubeEdge project](#how-to-use-in-KubeEdge-project)
+* [Which projects need to be modified](#which-projects-need-to-be-modified)
+
+## Abstract 
+
+Currently, in beehive framework, model message is defined and serialized using json format. The performance is slower than protobuf.
+
+And for parsing the message resource, each field index is hardcoded, such as (https://github.com/kubeedge/beehive/blob/master/pkg/common/util/parse_resource.go).
+
+In KubeEdge project, we can see many `Split(string, "/")` function which to parse the resource path string,  these are hard code. when we deal will the message resource after we receive message, we need to know the format of the assembly in advance, then we need use `Split()` function to get the field we want, it`s not easy to expand, it's not so extensible, so we need a more extended form.
+
+Please see the #399 issue of kubeeedge 
+
+In addition, currently when we use quic or websocket protocol to deliver message, we need to convert model message to probuf message (which defined in `github.com/kubeedge/viaduct/pkg/protos/message/message.pb.go`), and then send it, this also affects the efficiency of the transmission, rather than simply turning the model mesage into proto message.
+
+## Need to do
+
+*  Change the model message to protobuf message  directly
+*  Solve the filed index problem of `Resource`
+*  Consider compatibility  
+
+## How to do 
+
+##  Change the model message to protobuf message  directly
+
+we need to create proto file in `model` dir to define message struct and use `go generate`(will move to makefile) to create message.pb.go file 
+
+```
+syntax = "proto3";
+package model;
+import "google/protobuf/any.proto";
+
+message MessageHeader {
+    string ID = 1;
+    string ParentID = 2;
+    int64 Timestamp = 3;
+    bool Sync = 4;
+}
+message MessageRoute {
+    string Source = 1;
+    string Group = 2;
+    string Operation = 3;
+    google.protobuf.Any Resource = 4;
+}
+
+message Message {
+    MessageHeader Header = 1;
+    MessageRoute Router = 2;
+    google.protobuf.Any Data = 3;
+}
+```
+
+We use `Any type` to define `Resource` and `Data` field,  `Data` field is the original `Content` field.
+
+##  Solve the filed index problem of `Resource`
+
+this problem is the most important, currently, beehive define path string to stand for resource, such as `/node/{nodeID}/{namespace}/{resourceType}/{resource}`, different path stand for different resource.  but we need hardcord to parse `nodeid`,`resourceType`,`resourceid`, it's not so extensible, so we need a more extended form. we need change `Resource` to different struct type to stand for different resource path string, but in protobuf, there is not `interface` definition, so  we need to use `Any` type.
+
+We can see `MessageRoute ` definition.
+
+We use `github.com/gogo/protobuf`  package not offical `github.com/golang/protobuf` package, becasue many famous projects(like kubernetes, etcd, containerd) use it. 
+
+`gogo/protobuf/types` package provides function to encodes `Any` types 
+```
+	func MarshalAny(pb proto.Message) (*Any, error) 
+	func UnmarshalAny(any *Any, pb proto.Message) error 
+```
+But in gogo package, i can only support convert `proto.Message` and `Any` type to each other, it can not support convert custom type and `Any` type to each other.
+
+In `pkg/core/typeurl` package of `github.com/kubeedge/beehive` project, we support convert custom types and Any types to each other. there are some functions:
+
+* `func RegisterType(v interface{}, args ...string)`:
+
+this function Register custom types, if we want to support convert string types and Any types to each other. we need to add:
+ ```
+ RegisterType("", "string")
+ ``` 
+
+if we want to support convert bool types and `Any` types to each other. we need to add
+```
+RegisterType(true, "bool")
+```
+
+if we want to convert v1.Pod types ("k8s.io/api/core/v1") and `Any` types to each other. we need to add
+
+``` 
+RegisterType(v1.Pod{}, "v1.pod")
+```
+
+the second parameter used to generate TypeUrl of `Any` type. the TypeUrl must be unique, otherwise it will panic.
+
+* `func MarshalAny(v interface{}) (\*types.Any, error) `, this function takes interface and encodes it into google.protobuf.Any, and this interface must be Registerd by `RegisterType` function.
+* `func UnmarshalAny(any \*types.Any) (interface{}, error)`, this function parses the `google.protobuf.Any` message into a interface.
+
+## Consider compatibility  
+
+In order to be more compatible, we try to avoid modifying methods of message. Only the following method has been modified, but I don't think it will have much impact。
+```
+func (msg *Message) BuildRouter(source, group, res, opr string) *Message {
+func (msg *Message) BuildRouter(source, group string, res interface{}, opr string) *Message {
+```
+
+```
+func (msg *Message) SetResourceOperation(res, opr string) *Message {
+func (msg *Message) SetResourceOperation(res interface{}, opr string) *Message {
+```
+
+```
+func (msg *Message) GetResource() string {
+func (msg *Message) GetResource() interface{} {
+```
+
+let me just mention `FillBody()` function again:
+```go
+func (msg *Message) FillBody(content interface{}) *Message {
+	msg.Content = content
+	real, err := typeurl.MarshalAny(content)
+	if err != nil {
+		panic(errors.Errorf("Marshal content interface to any type error %v", err))
+	}
+	msg.Data = real
+	return msg
+}
+```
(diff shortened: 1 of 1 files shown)


kadisi:
i have sumit a pr for beehive project: https://github.com/kubeedge/beehive/pull/23

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,346 @@
...
+
+```
+and finnally, we convert Message to Event :
+```
+event := model.MessageToEvent(&msg)
+		select {
+		case rChannel <- event:
+		}
+```
+
```
kadisi: i think `Event` struct is useless.   Shall we consider using `Message` directly ??

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,346 @@
...
+    string ID = 1;
+    string ParentID = 2;
+    int64 Timestamp = 3;
+    bool Sync = 4;
+}
+message MessageRoute {
+    string Source = 1;
+    string Group = 2;
+    string Operation = 3;
+    google.protobuf.Any Resource = 4;
```
rohitsardesai83: can't this be of type string ?

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,346 @@
...
+```
+
+we use `Any type` to define `Resource` and `Data` field.  `Data` field is the original `Content` field.
+
+##  Solve the filed index problem of `Resource`
+
+this problem is the most important, currently, beehive define path string to stand for resource, such as `/node/{nodeID}/{namespace}/{resourceType}/{resource}`, different path stand for different resource.  but we need hardcord to parse `nodeid` ,`resourceType`,`resourceid`. It's not so extensible, so we need a more extended form. we need change `Resource` to golang `interface{}`. but in protobuf, there is not `interface` define. we need use `Any` type.
+
+```
+message MessageRoute {
```
rohitsardesai83: why is this struct repeated ?

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,346 @@
...
+if we want bool types can conversion to Any types. we need to add
+```
+RegisterType(true, "bool.content")
+```
+
+if we want v1.Pod types ("k8s.io/api/core/v1") can conversion to Any types. we need to add
+``` 
+RegisterType(v1.Pod{}, "v1.pod.content")
+```
+the second parameter used to generate TypeUrl of `Any` type. the TypeUrl must be unique, otherwise it will panic.
```
rohitsardesai83: what is TypeUrl ?

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,346 @@
...
+```
+but they can only support convert `proto.Message` into `Any` type, or convert `Any` type into `proto.Message`.  they can not convert custom type into `Any` type.
+
+so We need to support convert custom type into `Any` types, and convert `Any` types into custom types.
+
+in `pkg/core/typeurl` package of `github.com/kubeedge/beehive` project, we support custom types conversion to Any types. there are some functions:
+
+* `func RegisterType(v interface{}, args ...string)` , this function Register custom types . if we want string types can conversion to Any types. we need to add 
+ ```
+ RegisterType("", "string.content")
```
rohitsardesai83: why do we need to use .content suffix ?

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,346 @@
...
+* `github.com/kubeedge/kubeedge`
+    
+    Mainly process `Resource` problem
+* `github.com/kubeedge/beehive`
+    
+    define message proto file, and define `typeurl` package to support convert custom type into `Any` type.
+* `github.com/kubeedge/viaduct`
+    
+    no need `translator.NewTran().Encode()`
+
```
rohitsardesai83: remove extra new lines.

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,346 @@
...
+# How to use in kubeedge project
+
+take `DownstreamController` as an example.
+
+In `DownstreamController` , it will send 6 kind of message. `Pod`, `CongfigMap`,`secret`,`node`,`service`,`endpoint`. They have similar resource formats(`/node/{nodeID}/{namespace}/{resourceType}/{resource}`).
+
+so we need define a interface :
+```
+type ToEdgeResourcer interface {
```
rohitsardesai83: why do we need an interface ?

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,346 @@
...
+	return this.ID
+}
+```
+
+for `ConfigMap` resource, we can define:
+```
+const (
+	ResourceTypeConfigmap  = "configmap"
+)	
+type ConfigMapResource struct {
```
rohitsardesai83: both the PodResource and ConfigMapResource structs are same except the type.

docs/proposals/message-proto-design.md
lzx258: can't this be of type bytes?
```suggestion
    bytes Resource = 4;
```

kadisi: @rohitsardesai83  if we use string type, we need use `Split(string, "/")` function to parse the resource path string.  these are so many hard code.

kadisi: @lzx258  bytes types is ok, but if we use bytes, We don't know what type to convert bytes to。  `google.protobuf.Any` contains 
```
 string type_url = 1;

  // Must be a valid serialized protocol buffer of the above specified type.
  bytes value = 2;
}
```
it also contains an arbitrary serialized message as bytes. `Any` message type lets you use messages as embedded types without having their .proto definition.

docs/proposals/message-proto-design.md
kadisi: ok

docs/proposals/message-proto-design.md
kadisi: @rohitsardesai83  it is not repeated.  the are the seem definition.

docs/proposals/message-proto-design.md
kadisi: `.content` suffix is not necessary.

docs/proposals/message-proto-design.md
kadisi: TypeUrl is A URL/resource name that uniquely identifies the type of the serialized protocol buffer message.

https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto

docs/proposals/message-proto-design.md
kadisi: yes, you are right, for this example,  interface is not necessary. we can only define a  particular structure:
```
type ToEdgeResource struct {
	ResourceType string
	NameSpace    string
	NodeID       string
	ID           string
}
```
and let it registerd by `RegisterType()` function.

docs/proposals/message-proto-design.md
kadisi: yes , we can define one struct not different struct. such as 
```
type ToEdgeResource struct {
	ResourceType string
	NameSpace    string
	NodeID       string
	ID           string
}

```

docs/proposals/message-proto-design.md
rohitsardesai83: do we need to register primitive types like string, bool , int etc ? or only custom types need to be registered ?

docs/proposals/message-proto-design.md
rohitsardesai83: can we make these changes in the design ?

docs/proposals/message-proto-design.md
rohitsardesai83: please make the required changes in the proposal. thanks.

docs/proposals/message-proto-design.md
kadisi: @rohitsardesai83 , yes,  we need to register `string`, `bool` types.  because we also change  the original `Content` field type to `Any` type.

```
message Message {
    MessageHeader Header = 1;
    MessageRoute Router = 2;
    google.protobuf.Any Data = 3;
}
```
`Data` field is the original `Content` field.

in kubeedge,  there are some places where we need `string` as Content of message. such as:
https://github.com/kubeedge/kubeedge/blob/5ca1f17b87d070ccc37ea72879354355ae65857b/edge/pkg/edgehub/controller.go#L239   , so we need register it.

docs/proposals/message-proto-design.md
kadisi: ok

docs/proposals/message-proto-design.md
kadisi: ok

glenn-lg:
it is a good optimization proposal to me

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,297 @@
...
+* [How to use in kubeedge project](#how-to-use-in-kubeedge-project)
+* [Which projects need to be modified](#which-projects-need-to-be-modified)
+
+## Abstract 
+
+currently, In beehive framework, model message is defined and serialized using json format. the performance is slower than protobuf.
+
+and for parsing the message resource, each field index is hardcoded. such as (https://github.com/kubeedge/beehive/blob/master/pkg/common/util/parse_resource.go)
+
+in kubeedge project , we can see mush `Split(string, "/")` function which to parse the resource path string.  these are hard code.   when we deal will the message resource after we receive message  . we need to know the format of the assembly In advance。 then use `Split()` function to get the field we want.    it`s not easy to expand. It's not so extensible. so we need a more extended form.
```
rohitsardesai83: mush -> ?

rohitsardesai83: kubeedge -> KubeEdge, applies to all such occurences.

kadisi: sorry, it is many

kadisi: Ok

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,294 @@
...
+}
+```
+The parameter `content interface{}` is not changed, but the parameter passed in must be registered by `RegisterType()`.
+so if you want to fill string,  map or struct as message Data, the map or struct must registerd by  `RegisterType()`. 
+
+in KubeEdge, there are some places where we need `string` as Content of message. such as https://github.com/kubeedge/kubeedge/blob/5ca1f17b87d070ccc37ea72879354355ae65857b/edge/pkg/edgehub/controller.go#L239
+so  we pre-registered `string` and `bool` type in `pkg/core/typeurl` package of beehive project。
+
+it also has registerd some frequently used types.
+```
```
rohitsardesai83: use the go tag for the code sample.

kadisi: ok

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,296 @@
...
+   * [How to do](#how-to-do)
+   * [Change the model message to protobuf message  directly](#change-the-model-message-to-protobuf-message--directly)
+   * [Solve the filed index problem of Resource](#solve-the-filed-index-problem-of-resource)
+   * [Consider compatibility](#consider-compatibility)
+* [How to use in KubeEdge project](#how-to-use-in-KubeEdge-project)
+* [Which projects need to be modified](#which-projects-need-to-be-modified)
+
+## Abstract 
+
+currently, In beehive framework, model message is defined and serialized using json format. the performance is slower than protobuf.
```
kevin-wangzefeng: Currently, in beehive framework

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,296 @@
...
+
+and for parsing the message resource, each field index is hardcoded. such as (https://github.com/kubeedge/beehive/blob/master/pkg/common/util/parse_resource.go)
```
kevin-wangzefeng: And for parsing the message resource, each field index is hardcoded, such as (https://github.com/kubeedge/beehive/blob/master/pkg/common/util/parse_resource.go).

docs/proposals/message-proto-design.md
```diff
@@ -0,0 +1,296 @@
...
+
+we use `Any type` to define `Resource` and `Data` field.  `Data` field is the original `Content` field.
+
+##  Solve the filed index problem of `Resource`
+
+this problem is the most important, currently, beehive define path string to stand for resource, such as `/node/{nodeID}/{namespace}/{resourceType}/{resource}`, different path stand for different resource.  but we need hardcord to parse `nodeid` ,`resourceType`,`resourceid`. It's not so extensible, so we need a more extended form. we need change `Resource` to golang `interface{}`. but in protobuf, there is not `interface` define. we need use `Any` type.
+
+we can see `MessageRoute ` definition.
+
+we use `github.com/gogo/protobuf`  package not offical `github.com/golang/protobuf` package .
```
kevin-wangzefeng: Pelease add reasons(benifits) of why gogo/protobuf is suggested here instead of golang/protobuf

docs/proposals/message-proto-design.md
kadisi: ok

docs/proposals/message-proto-design.md
kadisi: ok

PR #893, opened 2019-07-11
modify the cni-plugin version , recommend using version 0.2.0

### LINKED ISSUE #892: version information in the cni-plugin configuration file causes a exception
**What happened**:
Follow the instructions in the guide, if the cni-plugin version is set to 0.3.0 , there will be an error with  "failed to find network info for sandbox".

Follow the instructions in the guide,  the system is all right

use any of the 0.1.0 to 0.4.1versions of  cni-plugin ,and set the version feild to  "0.3.0". and then deploy an app. It will be reproduce. 

**Environment**:
- KubeEdge version:
- Hardware configuration:
- Kernel (e.g. `uname -a`):
- Others:

arcanique:
**What type of PR is this?**
>
> /kind bug

>
> Fixes #892
>

Fixes #892

### DIFF
--- a/docs/guides/edgemesh_test_env_guide.md
+++ b/docs/guides/edgemesh_test_env_guide.md
@@ -18,19 +18,20 @@ edged:
 > <https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd>
 
 ## Install CNI plugin 
-+ get cni plugin source code of version (0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0) from the github.com. Compile and install.
++ get cni plugin source code of version (0.2.0) from the github.com. Compile and install.
+> It is recommended to use version 0.2.0, which have tested to ensure stability and availability.
 > make sure to run on the Go development environment.
 ```bash
-# download the cni-plugin-0.3.0
-$ wget https://github.com/containernetworking/plugins/archive/v0.3.0.tar.gz
+# download the cni-plugin-0.2.0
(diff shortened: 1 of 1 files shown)


docs/guides/edgemesh_test_env_guide.md
```diff
@@ -18,19 +18,20 @@ edged:
 > <https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd>
 
 ## Install CNI plugin 
-+ get cni plugin source code of version (0.1.0, 0.2.0, 0.3.0, 0.3.1, 0.4.0) from the github.com. Compile and install.
++ get cni plugin source code of version (0.2.0) from the github.com. Compile and install.
+> It is recommended to use version 0.2.0, which I have tested to ensure stability and availability.
```
fisherxu: Better to remove `I`, just `which have been tested...`.
BTW, how about the newest version like 0.5, 0.6...? Noticed that 0.2 is published at 2016

arcanique: ok. And because containerd currently only supports versions prior to 0.4.1. Previously we stated that we can support it to 0.4.1. Recently we found that there is a bug #892.

fisherxu: ok, this sound good. Once the bug fixed, let's up to the new version.

arcanique: ok

PR #916, opened 2019-07-16
removed dockertools dependency

sids-b:
**What type of PR is this?**
/kind cleanup

Removed docker tools dependency. We started using dockershim hence no need to use dockertools.


Baoqiang-Zhang:
LGTM

Issue #779, opened 2019-06-24
Unable to get GPU information of edge nodes


todayisgood:
**What happened**:

The edge node is GPU type, but can't get the gpu resource info when run `kubectl get node edge-node -o yaml` , olny get the cpu and memory info:
```
allocatable:
    cpu: "4"
    memory: 19816Mi
    pods: "110"
  capacity:
    cpu: "4"
    memory: 19916Mi
    pods: "110"
```

**What you expected to happen**:

The edgecore can report GPU information to the cloud，and can get the  “nvidia.com/gpu”  info  when  run `kubectl get node edge-node -o yaml` , like this:
```
allocatable:
    cpu: 27866m
    memory: 54320540Ki
    nvidia.com/gpu: "1"
    pods: "253"
  capacity:
    cpu: "28"
    memory: 57547164Ki
    nvidia.com/gpu: "1"
    pods: "253"
```

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
the GPU node info:

```
# nvidia-smi
Mon Jun 24 13:12:09 2019       
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 396.26                 Driver Version: 396.26                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  Tesla P4            On   | 00000000:00:05.0 Off |                    0 |
| N/A   29C    P8     6W /  75W |      0MiB /  7611MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

rohitsardesai83:
@todayisgood , thanks for reporting this. Did you get a chance to look for possible causes ?

rohitsardesai83:
```yaml
edged:
    register-node-namespace: default
    hostname-override: fb4ebb70-2783-42b8-b3ef-63e2fd6d242e
    interface-name: eth0
    edged-memory-capacity-bytes: 7852396000
    node-status-update-frequency: 10 # second
[...]

PR #924, opened 2019-07-18
remove gpu codes

edisonxiang:
**What type of PR is this?**
/kind cleanup

Support for the alpha.kubernetes.io/nvidia-gpu resource, which was deprecated in k8s v1.10, has been removed. The latest k8s version is using the resource exposed by DevicePlugins instead (nvidia.com/gpu). 

https://github.com/kubernetes/kubernetes/pull/61498

Since we will upgrade the k8s version in kubeedge, we need to cleanup these codes including gpu and nvida.

* "k8s.io/kubernetes/pkg/kubelet/gpu",	
* "k8s.io/kubernetes/pkg/kubelet/gpu/nvidia"


rohitsardesai83:
@edisonxiang , please resolve the conflicts. Thanks.

rohitsardesai83:
There is an open issue where this might be needed: https://github.com/kubeedge/kubeedge/issues/779. 
Should this code removal be part of the upgrade process ?

edisonxiang:
@rohitsardesai83 rebased. I think we need to make a test for the way using device plugin to support gpu, and give some documents or some modifications in kubeedge.

edisonxiang:
@rohitsardesai83 yep, it is better to make that gpu test after upgrade:)

kevin-wangzefeng:
/approve
Thanks

PR #814, opened 2019-06-26
Add integration tests for device CRD validation

rohitsardesai83:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**:
Add integration tests for validation of device crds.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #926, opened 2019-07-18
fix random unit tests exception

edisonxiang:
**What type of PR is this?**
/kind bug

When running `make cloud_test`, it appears that the unit tests exception like below:
```
  got = {{ID1 PID1 1563431396481 false} {Source1 Group1 Operation1 Resource1} <nil>}, 
 want = {{ID1 PID1 1563431396480 false} {Source1 Group1 Operation1 Resource1} <nil>}
```
That is caused by the actual timestamp and expect timestamp.


rohitsardesai83:
@edisonxiang , I have fixed these tests in https://github.com/kubeedge/kubeedge/pull/814

edisonxiang:
@rohitsardesai83 well.
Thanks your mentions
I will close this pr.
Let's merge #814 since the ci is influenced.

PR #928, opened 2019-07-18
If system has already docker installed then skip docker installation

naveensriram:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

If docker of any version is already installed in the system then skip docker installation

Code developed considering, version of docker installed on the system will be having minimum 2 dots.

### DIFF
--- a/CHANGELOG-1.1.md
+++ b/CHANGELOG-1.1.md
@@ -17,13 +17,13 @@
 
 This feature enables running applications with persistant data store at edge and KubeEdge to support [basic CSI Volume Lifecycle](https://github.com/container-storage-interface/spec/blob/master/spec.md#volume-lifecycle) and is compatible with Kubernetes and CSI.
 
-**Dynamic Admission control Webhook**
+**Dynamic Admission Control Webhook**
 
-Admission control webhook is an effective way of pre-processing the object configuration for KubeEdge API objects like devicemodels, devices. 
+Admission control webhook is an effective way of validating/mutating the object configuration for KubeEdge API objects like devicemodels, devices. 
 
 **Kubernetes Upgrade**
 
-Upgrade the venderod kubernetes version to v1.15.3.
+Upgrade the venderod kubernetes version to v1.15.3, so users can use the feature of new version on the cloud and on the edge side.
 
 **KubeEdge local setup scripts**
 
@@ -37,7 +37,7 @@ A bash script that can start a KubeEdge cluster in a VM with cloudcore, edgecore
 
 ### Other notable changes
 
-- Add New Feature: support CRI Interface docking docker. ([#829](https://github.com/kubeedge/kubeedge/pull/829), [@arcanique](https://github.com/arcanique))
+- Add New Feature: support dockershim in edged. ([#829](https://github.com/kubeedge/kubeedge/pull/829), [@arcanique](https://github.com/arcanique))
 
 - Fix edge_core cannot connect to edgecontroller after disconnecting once ([#870](https://github.com/kubeedge/kubeedge/pull/870), [@shouhong](https://github.com/shouhong))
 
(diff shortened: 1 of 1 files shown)


subpathdev:
I like the previous version because the edge_core can not run with every version of docker. (kubeadm v1.15 needed docker version 1.13 and higher)
I think you can do something with regex to check if a minimum version is installed.

naveensriram:
> I like the previous version because the edge_core can not run with every version of docker. (kubeadm v1.15 needed docker version 1.13 and higher)
> I think you can do something with regex to check if a minimum version is installed.

Referring to https://github.com/kubernetes/kubernetes/pull/77051 & https://kubernetes.io/docs/setup/release/notes/?origin_team=TDZ5TGXQW#cluster-lifecycle I also thought of introducing such a check for minimum version of docker required set to 1.13.1. 
But on referring to https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#external-dependencies I see docker version 18.03 is not been mentioned under valid versions and also in 1.15 release notes "external dependencies" section is missing which should cover the supported docker version section. In such case is it good to introduce minimum docker version check ? Thoughts @sids-b @rohitsardesai83 @subpathdev

subpathdev:
can you please rebase to the master branch and fix ut?

subpathdev:
> > I like the previous version because the edge_core can not run with every version of docker. (kubeadm v1.15 needed docker version 1.13 and higher)
> > I think you can do something with regex to check if a minimum version is installed.
> 
> Referring to [kubernetes/kubernetes#77051](https://github.com/kubernetes/kubernetes/pull/77051) & https://kubernetes.io/docs/setup/release/notes/?origin_team=TDZ5TGXQW#cluster-lifecycle I also thought of introducing such a check for minimum version of docker required set to 1.13.1.
> But on referring to https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#external-dependencies I see docker version 18.03 is not been mentioned under valid versions and also in 1.15 release notes "external dependencies" section is missing which should cover the supported docker version section. In such case is it good to introduce minimum docker version check ? Thoughts @sids-b @rohitsardesai83 @subpathdev @fisherxu

Yes I think we should set a minimum version. I think there will not be a problem with non valid versions because the most people will use their package managing tool (like apt or yum) to install the versions. A non valid version should not be installed via the package managing tool. On the other hand docker should not force to install a non valid version. they should say: install the stable version (if you know what can go wrong you can use the unstable version but this could break your programs / system)

If you want to set the minimum version of docker to 1.13, do you have tested kubeedge with these docker version?

naveensriram:
> > > I like the previous version because the edge_core can not run with every version of docker. (kubeadm v1.15 needed docker version 1.13 and higher)
> > > I think you can do something with regex to check if a minimum version is installed.
> > 
> > 
> > Referring to [kubernetes/kubernetes#77051](https://github.com/kubernetes/kubernetes/pull/77051) & https://kubernetes.io/docs/setup/release/notes/?origin_team=TDZ5TGXQW#cluster-lifecycle I also thought of introducing such a check for minimum version of docker required set to 1.13.1.
> > But on referring to https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#external-dependencies I see docker version 18.03 is not been mentioned under valid versions and also in 1.15 release notes "external dependencies" section is missing which should cover the supported docker version section. In such case is it good to introduce minimum docker version check ? Thoughts @sids-b @rohitsardesai83 @subpathdev @fisherxu
> 
> Yes I think we should set a minimum version. I think there will not be a problem with non valid versions because the most people will use their package managing tool (like apt or yum) to install the versions. A non valid version should not be installed via the package managing tool. On the other hand docker should not force to install a non valid version. they should say: install the stable version (if you know what can go wrong you can use the unstable version but this could break your programs / system)
> 
> If you want to set the minimum version of docker to 1.13, do you have tested kubeedge with these docker version?

Thanks @subpathdev for providing insights on this, will validate with docker 1.13 and make add that check.

sids-b:
@naveensriram , can you please resolve the review comments ??

PR #932, opened 2019-07-19
Fix a crash error of keadm

### LINKED ISSUE #901: keadm init crash
**What happened**:
A crash happend when execute "keadm init" on not supportted systems(eg:debian). 

I know now only support ubuntu.  But it should not crash directly. It should exit  with some errors.

Execute "keadm init" on systems other than ubuntu and centos.

The crash happens because the function GetOSInterface() returns nil on not supported systems and its callers does not check it's return value.

**Environment**:
- KubeEdge version:
- Hardware configuration:
debian 10 (buster)
- Kernel (e.g. `uname -a`):
- Others:

luguanglong:
Keadm crash because of a null pointer error when execute "keadm init" on unsupport systems(like debian and so on).
This change is to fix this error.

**What type of PR is this?**
> /kind bug

Without this PR, keadm crash because of a null pointer error when execute "keadm init" on unsupport systems(like debian and so on).

Fixes #901 

### DIFF
--- a/keadm/app/cmd/util/common.go
+++ b/keadm/app/cmd/util/common.go
@@ -184,7 +184,7 @@ func GetOSInterface() types.OSTypeInstaller {
 	case CentOSType:
 		return &CentOS{}
 	default:
-		panic("unsupport os-release")
+		panic("This OS version is currently un-supported by keadm")
 	}
 }
 
--- a/keadm/app/cmd/util/common.go
+++ b/keadm/app/cmd/util/common.go
@@ -184,8 +184,8 @@ func GetOSInterface() types.OSTypeInstaller {
 	case CentOSType:
 		return &CentOS{}
 	default:
+		panic("unsupport os-release")
 	}
-	return nil
 }
 
 //IsKubeEdgeController identifies if the node is having edge controller and k8s api-server already running.


keadm/app/cmd/util/common.go
```diff
@@ -184,8 +184,8 @@ func GetOSInterface() types.OSTypeInstaller {
 	case CentOSType:
 		return &CentOS{}
 	default:
+		panic("unsupport os-release")
```
naveensriram: @luguanglong , Thanks for the fix, can you re-phrase it to sound better something like "This OS version is currently un-supported by keadm"

luguanglong: @naveensriram OK

luguanglong:
@naveensriram 
Already re-phrase, you can approve.

sids-b:
@luguanglong , can you please squash the commits ?

luguanglong:
@sids-b 
Why?

sids-b:
@luguanglong , there is only 1 logical change in this PR, so it will be better to have only 1 commits so as to keep the commit history clean :)

edisonxiang:
@luguanglong perhaps you can find some information here about `git squash`

https://www.cnblogs.com/edisonxiang/p/6951787.html

luguanglong:
@sids-b @edisonxiang

luguanglong:
@sids-b @edisonxiang OK, I will handle as soon as possible

luguanglong:
@sids-b @edisonxiang 
I will close this request, and open a new one.
In company network environment, I cannot push codes to github by git cmd.

edisonxiang:
Thanks @luguanglong

PR #703, opened 2019-06-13
Adding E2E for bluetooth mapper

lidiyag:
**What type of PR is this?**
 /kind test

**What this PR does / why we need it**:
This PR adds E2E for mapper

**Which issue(s) this PR fixes**:

XREF #597 

**Special notes for your reviewer**:
This PR is in work in progress state

PR #935, opened 2019-07-19
Adding wait group for schedules created at runtime

### LINKED ISSUE #934: Container gets killed in bluetooth mapper when multiple schedules are given runtime
**What happened**:
While giving multiple runtime updates to scheduler in bluetooth mapper , the container gets killed before performing all schedules.

Mapper is expected to execute any number of schedules given runtime.

Run Bluetooth Mapper
Provide multiple runtime schedule updates

sujithsimon22:
**What type of PR is this?**
 /kind bug

This PR contains change to fix the issue of bluetooth maper exiting before completion of all schedules created at runtime

Fixes #934 

### DIFF
--- a/device/bluetooth_mapper/controller/controller.go
+++ b/device/bluetooth_mapper/controller/controller.go
@@ -139,6 +139,7 @@ func (c *Config) handleScheduleCreateMessage(client MQTT.Client, message MQTT.Me
 			glog.Infof("New Schedule: %v", newSchedule)
 		}
 		configuration.Config.Scheduler = c.Scheduler
+		helper.ControllerWg.Add(1)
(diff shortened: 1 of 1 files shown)


sujithsimon22:
cc @sids-b @rohitsardesai83  PTAL

sids-b:
We need to have some UT or e2e test cases to ensure further changes in the mapper code doesn't break these changes

sujithsimon22:
> We need to have some UT or e2e test cases to ensure further changes in the mapper code doesn't break these changes

I agree, We have a PR for e2e tests for bluetooth mapper #703. Once this is PR is ready we can use it to ensure that further changes don't break these changes

PR #936, opened 2019-07-20
bugfix: deviceTwin membership msg unmarshal failed.

### LINKED ISSUE #890: Unmarshal update request body failed, Please check the request
When I enable the devicetwin,   the error messege from edge " Unmarshal update request body failed, Please check the request"

kexun:
**What type of PR is this?**
> /kind bug

Fixes #890 

### DIFF
--- a/edge/pkg/devicetwin/dtcontroller.go
+++ b/edge/pkg/devicetwin/dtcontroller.go
@@ -300,42 +300,32 @@ func classifyMsg(message *dttype.DTMessage) bool {
 		log.LOGGER.Infof("Classify the msg to action %s", action)
 		return true
 	} else if (strings.Compare(msgSource, "edgemgr") == 0) || (strings.Compare(msgSource, "devicecontroller") == 0) {
-		if strings.Contains(message.Msg.Router.Resource, "membership/detail") {
-			message.Action = dtcommon.MemDetailResult
+		switch message.Msg.Content.(type) {
+		case []byte:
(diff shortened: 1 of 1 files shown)


edge/pkg/devicetwin/dtcontroller.go
```diff
@@ -300,42 +300,32 @@ func classifyMsg(message *dttype.DTMessage) bool {
 		log.LOGGER.Infof("Classify the msg to action %s", action)
 		return true
 	} else if (strings.Compare(msgSource, "edgemgr") == 0) || (strings.Compare(msgSource, "devicecontroller") == 0) {
-		if strings.Contains(message.Msg.Router.Resource, "membership/detail") {
-			message.Action = dtcommon.MemDetailResult
+		switch message.Msg.Content.(type) {
```
subpathdev: can you use an if else here?
With if else I think the readability would be increased.

sujithsimon22: +1

sujithsimon22: We could maybe use an assertion and if the assertion fails we can marshal the message ??

sids-b: I think it is better to keep the code style constant everywhere. We have a switch case in edged module

sujithsimon22: fair enough

Issue #806, opened 2019-06-25
Upgrade to Kubernetes 1.15


rohitsardesai83:
**What would you like to be added**:
Currently kubeedge vendors in kubernetes 1.10.9 . 
Gopkg.toml : 

```yaml
[[constraint]]
  name = "k8s.io/api"
  version = "kubernetes-1.10.9"

[[constraint]]
  name = "k8s.io/apiserver"
  version = "kubernetes-1.10.9"

[[constraint]]
  name = "k8s.io/apimachinery"
  version = "kubernetes-1.10.9"
```

Need to upgrade to latest Kubernetes release ( 1.15 )

**Why is this needed**:
Align edged with the latest architectural changes that Kubernetes has gone through from 1.10.9 to 1.15, leverage support of CSI for PV/PVC  , have a clean separation between vendored code and support added in edged to avoid overlapping of code in kubernetes kubelet and edged.

kevin-wangzefeng:
Please list all the relevant tasks for upgrading K8s v1.15 here
/cc @edisonxiang, @fisherxu

PR #939, opened 2019-07-22
Upgrade to Kubernetes v1.15 and Use gomod

### LINKED ISSUE #748: Change to use gomod to manage the dependencies
Use gomod to manage the dependies and remove the vendor from the project

**Why is this needed**:
To manage the dependiences more efficient

edisonxiang:
**What type of PR is this?**
/kind feature

* Upgrade to Kubernetes v1.15 [@edisonxiang]
* Change to use gomod to manage the dependencies [@subpathdev]

Fixes #748 #806

Since these two features are related, so @subpathdev is working together with me.
Thanks @subpathdev for the great work.


fisherxu:
Split to two PR maybe more clear, can change vender to gomod first, @edisonxiang @subpathdev what do you think? :)

edisonxiang:
Thanks @fisherxu we will. Perhaps we can upgrade the k8s version firstly since go mod will depend on that.

fisherxu:
> Perhaps we can upgrade the k8s version firstly since go mod will depend on that

Upgrade the k8s version firstly is also fine, and curious about if we don't upgrade the k8s version, this will block to change gomod?

edisonxiang:
> > Perhaps we can upgrade the k8s version firstly since go mod will depend on that
> 
> Upgrade the k8s version firstly is also fine, and curious about if we don't upgrade the k8s version, this will block to change gomod?

yep. Currently @subpathdev has already finish go mod work based on the upgraded k8s version v1.15.0

Issue #725, opened 2019-06-19
Reliable message delivery from cloud to edge


sids-b:
**What would you like to be added**:
Create a mechanism to make sure each message is delivered reliably to edge.

**Why is this needed**:
In offline scenarios, when there is an update in pods or device, the message is not send to edge and it is lost. So the edgeNode is not aware of the changes.

sids-b:
cc @kadisi @shouhong @chendave @subpathdev

kadisi:
please assign to me

rohitsardesai83:
Let's come up with some kind of design first as to how can we achieve this.

kadisi:
@rohitsardesai83 ok

kadisi:
> In offline scenarios, when there is an update in pods or device, the message is not send to edge and it is lost. So the edgeNode is not aware of the changes.

@rohitsardesai83 @sids-b 

take websocket as an example , we need to deal with  two things.  

one is write message error , we need to judge whether this error is ErrCloseSent.
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/handler/eventhandler.goL255
```
		err = eh.hubIoWrite(hi, info.NodeID, &msg)
		if err != nil {
			bhLog.LOGGER.Errorf("write error, connection for node %s will be closed, affected event %s, reason %s",
				info.NodeID, dumpEventMetadata(event), err.Error())
			stop <- hubioWriteFail
			return
		}
```

add the following judgement:
```
if err == websocket.ErrCloseSent {

}
```
if we found this is `ErrCloseSent`, we need cache this msg. so this is the second thing which we need deal with.

we need change the field `channelPool` in `ChannelEventQueue` struct.  `channelPool` is a map of channel,  we should change it to a map of queue.

the queue struct like this:   refer to `k8s.io/client-go/util/workqueue/queue.go`
```
type EventQueue struct {
	queue []*Event
	lock  sync.RWMutex
}

// Get the first element of the queue, but do not pop it
// You must call Done with item when you have finished processing it.
func (this *EventQueue) Get() *Event {
	return nil
}

// Pop the first element of the queue
func (this *EventQueue) Done() {
}
// add new element to the queue
func (this *EventQueue) Add(e *Event) {
}
```
[...]

PR #950, opened 2019-07-23
Reliable message delivery design proposal

rohitsardesai83:
**What type of PR is this?**
/kind design

Enable reliable message delivery from cloud to edge.

xref #725 


kadisi:
hi @rohitsardesai83,  

even though  any of the cloud components crash, in-flight events will be lost. Messages stored by the Event Queue in-memory would be destroyed. but when edgecontroller restart, event queue will get all event, because we use kubernetes List and watch mechanism .We can get the current full amount of data from kubernetes when edgecontroller restart.

In order to reduce invalid data transmission, should we be concerned about the problem of data duplication ? for example.

there are tow event which are same pod update in event queue .  But at this time they have not been sent to edge node, we just need send the last event.

another example.

there are tow event which are same pod in event queue, the first event is update pod, the second is delete pod, But at this time they have not been sent to edge node , we just need send the second event.

rohitsardesai83:
> hi @rohitsardesai83,
> 
> even though any of the cloud components crash, in-flight events will be lost. Messages stored by the Event Queue in-memory would be destroyed. but when edgecontroller restart, event queue will get all event, because we use kubernetes List and watch mechanism .We can get the current full amount of data from kubernetes when edgecontroller restart.
> 
> In order to reduce invalid data transmission, should we be concerned about the problem of data duplication ? for example.
> 
> there are two event which are same pod update in event queue . But at this time they have not been sent to edge node, we just need send the last event.
> 
> another example.
> 
> there are two event which are same pod in event queue, the first event is update pod, the second is delete pod, But at this time they have not been sent to edge node , we just need send the second event.

Yes, we need to think about how to handle duplicate events. 
Kubernetes did something similar : https://github.com/kubernetes/kubernetes/issues/4073.

docs/proposals/reliable-message-delivery.md
```diff
@@ -0,0 +1,194 @@
...
+
+There is no reliable mechanism to exchange messages between the cloud and the edge nodes.
+The current approach uses a fire-and-forget approach where messages are sent to edge nodes and there is no way to confirm if
+those messages were actually received by the edge node. Unstable networks can result in frequent disconnection of edge nodes and
+this can result in loss of messages sent to edge nodes which can’t be temporarily reached. This proposal addresses this problem
+thus ensuring reliable message delivery.
+
+### Goals
+
+- Provide a reliable message delivery mechanism between cloud and edge.
```
sids-b: from cloud to edge ?

sids-b:
ping @qizha @kevin-wangzefeng @fisherxu for review

fisherxu:
Dumplicate with #1343 , close this first.
/close

PR #952, opened 2019-07-23
Travis stages

### LINKED ISSUE #810: split ci in different stages
It could be useful to split the CI in different stages.
- build all parts
- unit tests
- e2e test
- cross build
....

**Why is this needed**:
- increase clearity of the ci tests

toughs?

subpathdev:
**What type of PR is this?**
/kind test
/kind feature

split the travis build process in multiple stages and adding more checks
increase code quality / contributors could easy understand on which point the build process breaks

Fixes #810 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,26 +33,29 @@ jobs:
     - stage: "dep check"
       script: make depcheck
       name: " Check if imports, Gopkg.toml, and Gopkg.lock are in sync"
-    - stage: "Test Jobs"                # naming the Tests stage
+    - stage: "Test Build"                # naming the Tests stage
       script: make 
       name: "builds cloud and edge components"            # names the first Tests stage job
+    - script: make edge_cross_build
+      name: "cross build edge"     # names the fifth Tests stage job
+    - script: make bluetoothdevice
+      name: "build bluetooth_device"
+    - stage: "Test lint"
+      script: make cloud_lint
+      name: "cloud_lint"     # names the eighth Tests stage job
+    - script: make bluetoothdevice_lint
+      name: "bluetooth_device_lint"
+    - script: make keadm_lint
+      name: "keadm_lint"
     - script: make edge_verify
       name: "verify edge"     # names the second Tests stage job
-    - script: make edge_test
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,26 +34,41 @@ jobs:
...
+      name: "bluetooth_device_lint"
+    - script: make keadm_lint
+      name: "keadm_lint"
     - script: make edge_verify
       name: "verify edge"     # names the second Tests stage job
-    - script: make edge_test
+    - stage: "unit tests"
+      script: make edge_test
       name: "unit test edge"     # names the third Tests stage job
     - script: make edge_integration_test
```
rohitsardesai83: we also have some integration tests in the cloud for device crd validation. Can we include those?
These tests need a K8S cluster and cloud component up and running.

.travis.yml
```diff
@@ -33,26 +34,41 @@ jobs:
...
+    - stage: "Test Build"                # naming the Tests stage
       script: make 
       name: "builds cloud and edge components"            # names the first Tests stage job
+    - script: make edge_cross_build
+      name: "cross build edge"     # names the fifth Tests stage job
+    - script: make bluetoothdevice
+      name: "build bluetooth_device"
+    - script: make edge_small_build
+      name: "build small edge version"
+    - stage: "Test lint"
```
sids-b: lets move the lint stage before the build stage ?

.travis.yml
subpathdev: Yes we can include these tests. But I think there is an issue on k8s side. Compare to the message in the main makefile:
`This has been commented temporarily since there is an issue of CI using same master for all PRs, which is causing failures when run parallely`

.travis.yml
subpathdev: You are right, lint fails more often than the build in the travis tests

.travis.yml
kevin-wangzefeng: This is not blocking, I'd suggest we discuss it in the tracking issue and get this accepted.

sids-b:
Thanks @subpathdev 
/approve

Issue #897, opened 2019-07-12
Support Ingress on edge


qizha:
**What would you like to be added**:
Add ingress support on edge

**Why is this needed**:
User want to deploy services on edge which may include multiple instances with a Nginx in front of them so that they can leverage a powerful machine

qizha:
EdgeMesh can also act as ingress controller. Also, traefic can be another choice

Issue #943, opened 2019-07-22
Support multiple namespaces on edgenode


sids-b:
**What would you like to be added/modified**:
Currently edge_core can only bring up pods in one namespace which is specified in https://github.com/kubeedge/kubeedge/blob/master/edge/conf/edge.yaml#L32
This namespace is used while pulling secrets/configmaps 
for eg: https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edged/edged.go#L1034

Node should support multiple namespaces as node is not a namespaced resource

**Why is this needed**:
1) To support deploying pods from different namespaces
2) To support multi-tenancy

sids-b:
cc @kevin-wangzefeng @qizha @rohitsardesai83 @fisherxu @subpathdev @kadisi

sids-b:
/cc @kevin-wangzefeng @rohitsardesai83 @qizha  @fisherxu @subpathdev @kadisi @subpathdev 

Do you guys think this will be useful ??

PR #953, opened 2019-07-24
proposal for using ingress at kubeedge

sujithsimon22:
**What type of PR is this?**
 /kind design

Provides mechanism to provide ingress support for Kubeedge using Nginx

xref #897 


docs/proposals/ingress.md
```diff
@@ -0,0 +1,91 @@
...
+Using ingress the user can bring all their services under a single umbrella and manage them easily. The ingress support should continue to work even in offline scenarios i.e. even in the cases where 
+connectivity between edge and cloud is lost/unreliable. 
+
+
+## Motivation:
+
+The cloud native and microservice architecture is becoming more and more popular and the edge node is becoming more and more powerful. In these scenarios, users might want to deploy their applications at the edge, 
+which could need to be accessed by external users, through services. Using ingress we would be able to provide name based virtual hosting and load balance traffic.
+
+This design document deals with enabling users to leverage powerful machines at the edge to provide external users
```
sids-b: Can you give more details of what do you mean by external user ?

docs/proposals/ingress.md
```diff
@@ -0,0 +1,91 @@
...
+Traffic routing is controlled by rules defined on the Ingress resource. This proposal uses Node-port service to redirect external users to the ingress service, instead we can also make use of an external load-balancer for this purpose. 
+
+
+## Design Details:
+
+In the current design the communication between external users and the services are handled by the nginx ingress controller. The communication between the services and their respective application pods will be handled by the edge mesh feature of KubeEdge. 
+
+<img src="../images/ingress/ingress.png">
+
+We are making use of nginx controller as our ingress controller to handle ingress resources. The nginx ingress controller runs as a pod (which would be deployed by edge core) in its own namespace (ingress-nginx).
```
sids-b: currently edge_core only support pods from 1 namespace.... if the namespace is ingress-nginx, we may not be able to support pods from multiple namespaces ?

docs/proposals/ingress.md
```diff
@@ -0,0 +1,91 @@
...
+In the current design the communication between external users and the services are handled by the nginx ingress controller. The communication between the services and their respective application pods will be handled by the edge mesh feature of KubeEdge. 
+
+<img src="../images/ingress/ingress.png">
+
+We are making use of nginx controller as our ingress controller to handle ingress resources. The nginx ingress controller runs as a pod (which would be deployed by edge core) in its own namespace (ingress-nginx). 
+The nginx ingress controller pod can be accessed through the nginx service, which also runs in the same namespace as the ingress controller.
+
+
+1. The user first has to deploy the Nginx ingress controller pod and service on the edge through the edge core (edged component to be specific).
+2. Once the pod and service have been created, the user can define the an ingress resource, which contains the rules based on which the ingress controller forwards the requests from the user.
```
sids-b: s/the an/an

sujithsimon22: done, thanks

docs/proposals/ingress.md
sujithsimon22: done, thanks

docs/proposals/ingress.md
sujithsimon22: I think we already have an issue to support multiple pod namespace https://github.com/kubeedge/kubeedge/issues/943. Once that feature is ready i don't think it would be a problem if all resources used by ingress controller resides in a separate namespace. Thoughts ?

sujithsimon22:
> The document is more of a guide on how to use ingress with KubeEdge, i think we can place it in guides folder ?

Have moved it to guides folder as requested

ugvddm:
Why do we watch from k8s-apiserver directly? I don't think it's good to add one more connection with api-server

sids-b:
> Why do we watch from k8s-apiserver directly? I don't think it's good to add one more connection with api-server

+1
cc @qizha @kevin-wangzefeng @rohitsardesai83 @fisherxu @subpathdev thoughts ??

sids-b:
Also one point to note is that the edge node may have connectivity with the node on which edgecontroller is running but not with the node where apiserver is running,

fisherxu:
> Also one point to note is that the edge node may have connectivity with the node on which edgecontroller is running but not with the node where apiserver is running,

Agree, better through edgehub/cloudhub, will be more reliable in edge.

sujithsimon22:
Agreed, I will rework the design and try to utilise the edgehub/cloudhub control flow for the ingress controller as well :)

sujithsimon22:
This design has been moved to a google doc in the KubeEdge google drive: https://docs.google.com/document/d/1n4rp3NUkoo0epIyr_Z8895PBGrp2-6XC2x4h2sHsKos/edit?usp=sharing. Pls feel free to share ur thoughts and comment/edit the document. Another PR can be raised once the final design is finalised.

PR #957, opened 2019-07-24
Add --image-repository  && --pod-network-cidr for kubeadmin init.

jwzl:
-->

**What type of PR is this?**
> /kind feature

 It should be add when kubeadmin init to select the image-repository  and pod-network-cidr.

Fixes #

### DIFF
--- a/keadm/app/cmd/util/common.go
+++ b/keadm/app/cmd/util/common.go
@@ -40,6 +40,9 @@ const (
 
 	KubernetesDownloadURL = "https://apt.kubernetes.io/"
 	KubernetesGPGURL      = "https://packages.cloud.google.com/apt/doc/apt-key.gpg"
+	
+	KubernetesImageRepository = "k8s.gcr.io"
+	KubernetesPodNetworkCidr  = "10.244.0.0/16"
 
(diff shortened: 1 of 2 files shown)


keadm/app/cmd/util/common.go
```diff
@@ -40,6 +40,9 @@ const (
 
 	KubernetesDownloadURL = "https://apt.kubernetes.io/"
 	KubernetesGPGURL      = "https://packages.cloud.google.com/apt/doc/apt-key.gpg"
+	
+	KubernetesImageRepository = "k8s.gcr.io"
+	KubernetesPodNetworkCidr  = "10.244.0.0/16"
```
subpathdev: I think we should use this block 100.64.0.0/10 as default.

subpathdev:
Could you add a configuration parameter to configure the network cidr? (I mean something like this:
`keadm init --pod-network-cidr 192.168.0.0/16 --kubernetes-version 1.15.1`)

keadm/app/cmd/util/common.go
jwzl: Yes, you are right, I will fix this, and recreate PR.

PR #965, opened 2019-07-25
Fix a crash of keadm

### LINKED ISSUE #901: keadm init crash

luguanglong:
Keadm crash because of a null pointer error when execute "keadm init" on unsupport systems(like debian and so on).
This change is to fix this error.

**What type of PR is this?**
>
> /kind bug

Without this PR, keadm crash because of a null pointer error when execute "keadm init" on unsupport systems(like debian and so on).

Fixes #901 

### DIFF
--- a/keadm/app/cmd/util/common.go
+++ b/keadm/app/cmd/util/common.go
@@ -184,8 +184,8 @@ func GetOSInterface() types.OSTypeInstaller {
 	case CentOSType:
 		return &CentOS{}
 	default:
+		panic("This OS version is currently un-supported by keadm")
 	}
-	return nil
 }
 
 //IsKubeEdgeController identifies if the node is having edge controller and k8s api-server already running.


luguanglong:
@sids-b @naveensriram @edisonxiang 
I close previous 932 request, and open this new one for company network reason.

sids-b:
Thanks @luguanglong

fisherxu:
Let's get this merged. @rohitsardesai83 @sids-b

PR #968, opened 2019-07-26
Fix items in edge conf

edisonxiang:
**What type of PR is this?**
/kind test

* add new items in the edge conf to pass the perf test in current version.
* update edge version.

### DIFF
--- a/tests/performance/assets/02-edgeconfigmap.yaml
+++ b/tests/performance/assets/02-edgeconfigmap.yaml
@@ -45,6 +45,7 @@ data:
         register-node-namespace: default
         hostname-override: perf-node-o19egaee6o
         interface-name: eth0
+        edged-memory-capacity-bytes: 7852396000
         node-status-update-frequency: 10 # second
         device-plugin-enabled: false
         gpu-plugin-enabled: false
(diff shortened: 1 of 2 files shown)


tests/performance/assets/02-edgeconfigmap.yaml
```diff
@@ -54,6 +55,14 @@ data:
         docker-address: tcp://localhost:2375
         version: 2.0.0
```
sujithsimon22: @edisonxiang should we maybe update the version field as well to match the current version?

edisonxiang: Thanks. done

kevin-wangzefeng:
/approve
Thanks.

sids-b:
@edisonxiang , can we make the same change in edgesite.yaml as well ?

sids-b:
https://github.com/kubeedge/kubeedge/blob/master/edgesite/conf/edgeSite.yaml this is the link to the file

edisonxiang:
Thanks @sids-b 
We can change edgeSite.yaml in another pr which is related some other modifications about conf.

PR #994, opened 2019-08-06
Reduce time required for edge_small_build

sujithsimon22:
**What type of PR is this?**
 /kind cleanup

This PR reduces the time required for building the edge_core with a smaller binary size

### DIFF
--- a/edge/Makefile
+++ b/edge/Makefile
@@ -58,4 +58,7 @@ armv8:
 .PHONY: small_build
 small_build:
 	go build -ldflags="-w -s -extldflags -static" cmd/edge_core.go
-	upx-ucl --best edge_core
+	# There are options in upx-ucl like --best, which has best compression,
+	# but also has a drawback of taking too long for big files, we are using -9 since it takes less time
+	# Pls use --help for more information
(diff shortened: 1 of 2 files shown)


sujithsimon22:
cc @fisherxu @sids-b @kevin-wangzefeng @rohitsardesai83

edge/Makefile
```diff
@@ -58,4 +58,6 @@ armv8:
 .PHONY: small_build
 small_build:
 	go build -ldflags="-w -s -extldflags -static" cmd/edge_core.go
-	upx-ucl --best edge_core
+	# There are options in upx-ucl like --best, but it has a drawback of taking too long for big files, we are using --9 since it takes less time
```
fisherxu: Need reminder to user that --best has best compression, how about like below? :)
`#` There are options in upx-ucl like --best, which has best compression,
`#` but also has a drawback of taking too long for big files, we are using `-9` since it takes less time

sujithsimon22: Done @fisherxu, thanks :)

fisherxu:
thanks! @sujithsimon22

PR #1003, opened 2019-08-08
Reducing wait time for edgehub unit test

sujithsimon22:
**What type of PR is this?**
 /kind cleanup

Reduces wait time for unit test in edgehub

### DIFF
--- a/edge/pkg/edgehub/config/config_test.go
+++ b/edge/pkg/edgehub/config/config_test.go
@@ -226,7 +226,7 @@ func TestGetConfig(t *testing.T) {
 				t.Error("Unable to generate test YAML file: ", err)
 			}
 			// time to let config be synced again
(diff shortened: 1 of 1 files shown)


sujithsimon22:
cc @fisherxu @kevin-wangzefeng @sids-b @rohitsardesai83

sids-b:
Technically we should not be having sleep here.... We need to find a logical way to deal with the sync problem.

sids-b:
We have sleep's in all test files in edgehub/ directory as well which is not correct way to solve synchronization. Adding sleep for n seconds in a thread ensures that the thread is blocked for n seconds but there is no guarantee that the thread will start executing after n seconds.....

fisherxu:
Do we need the write config file operation here? can read from a exist file? And by the way, when we change to the config API, this test may will be removed :)

PR #1016, opened 2019-08-11
adding verfify vendor script

### LINKED ISSUE #1012: Need add a CI job to check the vender after change to go mod
Add a  CI job to check the vender, need to check:

- go module relate files, like go.mod/go.sum

- the vender files

subpathdev:
**What type of PR is this?**
/kind test
/kind feature

- Check if the vendor directory is up to date
- Check if any patch in the dependencies are present

Fixes #1012 

```release-note
NONE
```


.travis.yml
```diff
@@ -34,6 +34,12 @@ install: true
 
 jobs:
   include:
+    - stage: "Verify vendor"
+      script: bash build/tools/verifyVendor.sh
+      name: "check if the vendor is up to date"
+    - script: bash build/tools/verifyDepVersion.sh
+      name: "check if patches in dependencies are present"
```
fisherxu: Thanks @subpathdev for quick fix! And can we move this to `Makefile` and replace `dep check` in it :)

subpathdev: Yes of course

Makefile
```diff
@@ -124,7 +124,8 @@ edgesiteimage:
 
 .PHONY: depcheck
 depcheck:
```
sids-b: Can we rename to something like vendorCheck ?

build/tools/verifyDepVersion.sh
```diff
@@ -0,0 +1,39 @@
...
+		return
+	fi
+
+	echo 2
+}
+ 
+go get -u=patch
+ret=$(checkModified)
+if [ ${ret} -eq 0 ]; then
+	echo "SUCESS: vendor is up to date"
```
sids-b: nit: s/SUCESS/SUCCESS

build/tools/verifyVendor.sh
```diff
@@ -0,0 +1,39 @@
...
+		return
+	fi
+
+	echo 2
+}
+ 
+go mod vendor
+ret=$(checkModified)
+if [ ${ret} -eq 0 ]; then
+	echo "SUCESS: vendor is up to date"
```
sids-b: nit: s/SUCESS/SUCCESS

build/tools/verifyVendor.sh
```diff
@@ -0,0 +1,39 @@
...
+else
+	echo  "Failed: vendor needs an update; The diff is:"
```
sids-b: s/Failed/FAILED 
to keep the formatting similar to SUCCESS

Applicable to similar places

build/tools/verifyVendor.sh
```diff
@@ -0,0 +1,49 @@
...
+ret=$(checkModified)
+if [ ${ret} -eq 0 ]; then
+	echo "SUCCESS: vendor is up to date"
+else
+	echo  "FAILED: vendor needs an update; The diff is:"
+	git diff
+	exit 1
+fi
+
+go get -u=patch
```
kadisi: why use `go get -u=patch` ?  It will instructs get to update dependencies
to use newer patch releases when available.

Is it necessary ? @subpathdev @sids-b

subpathdev: > It will instructs get to update dependencies to use newer patch releases when available.
Yes you are right

> Is it necessary ?
You have to decide what we want to do. If we want to check if one of our dependencies has been patched then yes it is necessary. If we only check if the vendor has been build after changing the go.mod or go.sum file it is not necessary.

I think we should check if one of our dependencies has been updated. We should include  those patches to not include security issues in our programs. (the minor version will be checked, in an example version string v1.x.y the y is the part who will checked by go get -u=patch.)

build/tools/verifyVendor.sh
```diff
@@ -0,0 +1,49 @@
...
+	modified=$( git status --short 2>/dev/null | grep -e "^.M" | wc -l)
+
+	if [ ${modified} -eq 0 ]; then
+		echo 0
+		return
+	fi
+
+	echo 2
+}
+
```
kadisi: do we need use `go mod tidy` to  check whether adds any missing module requirements and removes unnecessary ones ?

fisherxu: +1

build/tools/verifyVendor.sh
kadisi: @subpathdev ,  that sounds good, but  this will force us to use the latest patch version of the dependency.  This may lead to compatibility issues。 kubernetes also does not use this feature。

build/tools/verifyVendor.sh
subpathdev: do we need this on every upload or should this be done by creating a new release?

build/tools/verifyVendor.sh
subpathdev: I think we should discuss this on an separate pr or?

build/tools/verifyVendor.sh
subpathdev: I have add this

kadisi:
@subpathdev please `git rebase -i ` your commit , then i think /lgtm.  thanks

subpathdev:
I have rebased to 2 commits, one to include the scripts and one to include the changes of go.mod / go.sum and the vendor directory. I didn't like to put all these changes into one commit.

kadisi:
/lgtm

cc @fisherxu @sids-b

sids-b:
Thanks @subpathdev 

/approve

PR #1022, opened 2019-08-13
dtcommon: fix column name error

ethan-daocloud:
Signed-off-by: Guangming Wang <guangming.wang@daocloud.io>

**What type of PR is this?**

/kind bug

i beleive this column name "lastsate" is wrong,  ref from [here](https://github.com/kubeedge/kubeedge/blob/45fb42520b3022ad53132f5140c803f03f8a8122/edge/pkg/devicetwin/dtclient/sqlite.go#L21)

Fixes #

```release-note
NONE
```

### DIFF
--- a/edge/pkg/devicetwin/dtcommon/common.go
+++ b/edge/pkg/devicetwin/dtcommon/common.go
@@ -123,11 +123,6 @@ var (
 	//OperaMembershipResult get memebership
 	OperaMembershipResult = "membership_result"
 
-	// InsertDocument insert document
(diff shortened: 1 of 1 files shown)


sids-b:
/hold

Give me a few minutes... I will check where it is used.

sids-b:
@ethan-daocloud , I think both the InsertDocument and DeleteDocument variables are unused and can be safely deleted.

ethan-daocloud:
> @ethan-daocloud , I think both the InsertDocument and DeleteDocument variables are unused and can be safely deleted.
@sids-b 
ok, then I will do it in this pr.

Issue #1025, opened 2019-08-14
log info is self-contradictory


alien19880726:
**What happened**:
![image](https://user-images.githubusercontent.com/47560998/63003173-3ab26b80-beaa-11e9-97ec-938408fb7887.png)

**What you expected to happen**:there is nothing wrong in the log info

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- KubeEdge version:
- Hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Others:

kadisi:
@alien19880726 , thank you for your issue,  we need continue this loop when we find unsupported operation. https://github.com/kubeedge/kubeedge/blob/45fb42520b3022ad53132f5140c803f03f8a8122/cloud/pkg/edgecontroller/controller/upstream.go#L571

alien19880726:
> @alien19880726 , thank you for your issue, we need continue this loop when we find unsupported operation.
> 
> https://github.com/kubeedge/kubeedge/blob/45fb42520b3022ad53132f5140c803f03f8a8122/cloud/pkg/edgecontroller/controller/upstream.go#L571

Maybe you do not get my words, one line indicates this operation cannot be processed, while another line indicates process successfully in the following logs. I think the second line should be removed.
2019-08-06 19:48:13.574 +08:00 [0;34mINFO[0m controller/upstream.go:571 message: 0130cbc8-c97d-4fc2-8f6d-c988e40dc64a **process failure**, service operation: response unsupported
2019-08-06 19:48:13.574 +08:00 [0;34mINFO[0m controller/upstream.go:573 message: 0130cbc8-c97d-4fc2-8f6d-c988e40dc64a **process successfully**

PR #1026, opened 2019-08-14
bugfix: upstream log info is self-contradictory

kadisi:

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind bug

fix: #1025

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -570,7 +570,6 @@ func (uc *UpstreamController) queryService(stop chan struct{}) {
 			default:
 				log.LOGGER.Infof("message: %s process failure, service operation: %s unsupported", msg.GetID(), msg.GetOperation())
 			}
-			log.LOGGER.Infof("message: %s process successfully", msg.GetID())
 		case <-stop:
 			log.LOGGER.Infof("stop queryService")
 			running = false


cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -566,9 +566,10 @@ func (uc *UpstreamController) queryService(stop chan struct{}) {
 					log.LOGGER.Warnf("message: %s process failure, response failed with error: %s", msg.GetID(), err)
 					continue
 				}
-				log.LOGGER.Warnf("message: %s process successfully", msg.GetID())
+				log.LOGGER.Infof("message: %s process successfully", msg.GetID())
 			default:
 				log.LOGGER.Infof("message: %s process failure, service operation: %s unsupported", msg.GetID(), msg.GetOperation())
+				continue
 			}
 			log.LOGGER.Infof("message: %s process successfully", msg.GetID())
```
sids-b: I think this will always be duplicate of 569 ? Can we remove this one ?

kadisi: ok

anyushun: Remove which one? 569 or 574? I think it is better to delete 569, because 569 is in a case of switch, we may add other cases in the future.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -566,9 +566,9 @@ func (uc *UpstreamController) queryService(stop chan struct{}) {
 					log.LOGGER.Warnf("message: %s process failure, response failed with error: %s", msg.GetID(), err)
 					continue
 				}
-				log.LOGGER.Warnf("message: %s process successfully", msg.GetID())
 			default:
 				log.LOGGER.Infof("message: %s process failure, service operation: %s unsupported", msg.GetID(), msg.GetOperation())
+				continue
```
fisherxu: Maybe the continue don't needed here?

anyushun: If delete this line, will get such log lines:
```
message: xxx process failure, service operation: xxx unsupported
message: xxx process successfully
```
So I think it should not be deleted

cloud/pkg/edgecontroller/controller/upstream.go
sids-b: I would suggest removing 573... for each case we can have a log message at the end which maybe different.

cloud/pkg/edgecontroller/controller/upstream.go
fisherxu: Agree with sid, delete L573. After delete L573, why will get such log lines?
```
message: xxx process failure, service operation: xxx unsupported
message: xxx process successfully
```

anyushun: Sorry, this is a misunderstanding, I mistakenly think that you want to delete continue

cloud/pkg/edgecontroller/controller/upstream.go
anyushun: There should be no difference.

cloud/pkg/edgecontroller/controller/upstream.go
fisherxu: Np :-)

sids-b:
Thanks @kadisi 

/approve

PR #1054, opened 2019-08-20
add image pull progress deadline param

### LINKED ISSUE #1052: ImagePullBackOff error for specific image
**What happened**:
some large size images can not be pulled in edge node
image

image can be pulled in edge node

**Environment**:
- KubeEdge version:
- Hardware configuration:
- Kernel (e.g. `uname -a`):
- Others:

Baoqiang-Zhang:
**What type of PR is this?**
>

> /kind bug

fix image pull backoff error 

Fixes #1052 

```release-note

```

### DIFF
--- a/edge/conf/edge.yaml
+++ b/edge/conf/edge.yaml
@@ -46,6 +46,7 @@ edged:
     remote-image-endpoint: unix:///var/run/dockershim.sock
     runtime-request-timeout: 2
     podsandbox-image: k8s.gcr.io/pause
+    image-pull-progress-deadline: 60 # second
     cgroup-driver: cgroupfs
     node-ip: ""
     cluster-dns: ""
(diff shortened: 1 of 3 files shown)


edge/conf/edge.yaml
```diff
@@ -46,6 +46,7 @@ edged:
     remote-image-endpoint: unix:///var/run/dockershim.sock
     runtime-request-timeout: 2
     podsandbox-image: k8s.gcr.io/pause
+    image-pull-progress-deadline: 60 # second
```
kadisi: does keadm need to be changed as well?

sids-b: I think not required. As the default value is set already.

Also we need to modify keadm handling the configurations from flags at it will be too many flags. Need to think of a simpler way as we cannot expose all configurations using flags.

fisherxu: Yes agree, too many flags are easy to get confused

subpathdev:
/lgtm cancel
please have a look on e2e tests

edisonxiang:
It seems like that the e2e test is easy to throw timeout error:(
I restarted the e2e test.
Let's wait for the result.

sids-b:
Thanks @Baoqiang-Zhang 

/lgtm

fisherxu:
@edisonxiang  ;-)

PR #1075, opened 2019-08-28
Validate device protocol configs based on webhook

### LINKED ISSUE #826: devicecrd: Reject protocol configs for protocols other than OPC UA , Modbus and Bluetooth.
While creating a device instance, protocol configs are optional, but if provided, they can be one of [OPC UA, Modbus, and Bluetooth]. Other protocol configs are not supported.

**Why is this needed**:
Validation of CRDs is needed for ensuring valid metadata is persisted for use by mappers/platform.

chendave:

**What type of PR is this?**
>

/kind bug

Fixes #826

```release-note

```

### DIFF
--- a/cloud/pkg/admissioncontroller/admission.go
+++ b/cloud/pkg/admissioncontroller/admission.go
@@ -2,6 +2,7 @@ package admissioncontroller
 
 import (
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -26,7 +27,9 @@ import (
 
 const (
 	ValidateDeviceModelConfigName  = "validate-devicemodel"
+	ValidateDeviceConfigName       = "validate-device"
 	ValidateDeviceModelWebhookName = "validatedevicemodel.kubeedge.io"
+	ValidateDeviceWebhookName      = "validatedevice.kubeedge.io"
 )
 
 var scheme = runtime.NewScheme()
@@ -92,6 +95,7 @@ func Run(opt *options.AdmissionOptions) {
 		klog.Fatalf("Failed to register the webhook with error: %v", err)
 	}
 
+	http.HandleFunc("/devices", serveDevice)
 	http.HandleFunc("/devicemodels", serveDeviceModel)
 
 	server := &http.Server{
@@ -132,7 +136,7 @@ func configTLS(opt *options.AdmissionOptions, restConfig *restclient.Config) *tl
 	return &tls.Config{}
 }
 
-// Register registers the admission webhook.
+// registerWebhooks registers the admission webhook.
 func (ac *AdmissionController) registerWebhooks(opt *options.AdmissionOptions, cabundle []byte) error {
 	ignorePolicy := admissionregistrationv1beta1.Ignore
 	deviceModelCRDWebhook := admissionregistrationv1beta1.ValidatingWebhookConfiguration{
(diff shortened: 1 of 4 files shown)


chendave:
cc @fisherxu @rohitsardesai83 @subpathdev @sids-b  @kadisi

chendave:
cc @kevin-wangzefeng

fisherxu:
@chendave  Need rebase, we can add a file named admin_device to deal with the device now, how do you think? :)

chendave:
@fisherxu , sure, will do.

kevin-wangzefeng:
We should not reject protocols that is not predefined in-tree, which will prevent people adding their own protocol/mapper implementations.
For protocols already defined in-tree, we can check if all the fields are filled correctly.

chendave:
> We should not reject protocols that is not predefined in-tree, which will prevent people adding their own protocol/mapper implementations.

Agreed, so we might need to update the bug there. @rohitsardesai83 

> For protocols already defined in-tree, we can check if all the fields are filled correctly.

we don't need to do that, this case has been covered by the validation rules defined in the yaml.
so, all we need to do here is make sure only one protocol is defined for each device as what https://github.com/kubeedge/kubeedge/issues/826 said.

cloud/pkg/admissioncontroller/admission.go
```diff
@@ -195,3 +231,57 @@ func registerValidateWebhook(client admissionregistrationv1beta1client.Validatin
...
+		responseAdmissionReview.Response = admit(requestedAdmissionReview)
+	}
+
+	// Return the same UID
+	responseAdmissionReview.Response.UID = requestedAdmissionReview.Request.UID
+	klog.Infof("sending response: %v", responseAdmissionReview.Response)
+
+	respBytes, err := json.Marshal(responseAdmissionReview)
+	if err != nil {
+		klog.Fatalf("cannot marshal to a valid reponse %v", err)
```
rohitsardesai83: reponse  -> response

rohitsardesai83: applies at all such places

rohitsardesai83: what do we return here ?

chendave: thanks, addressed!

rohitsardesai83:
@chendave , can you please fix the lint issue reported by the CI.

cloud/pkg/admissioncontroller/admission.go
```diff
@@ -195,3 +231,58 @@ func registerValidateWebhook(client admissionregistrationv1beta1client.Validatin
...
+// admitFunc is the type we use for all of our validators and mutators
+type admitFunc func(admissionv1beta1.AdmissionReview) *admissionv1beta1.AdmissionResponse
+
+func serve(w http.ResponseWriter, r *http.Request, admit admitFunc) {
+	var body []byte
+	if r.Body != nil {
+		if data, err := ioutil.ReadAll(r.Body); err == nil {
+			body = data
+		}
+	}
```
sids-b: should we return if ``` r.Body == nil ```

cloud/pkg/admissioncontroller/admission.go
```diff
@@ -195,3 +231,58 @@ func registerValidateWebhook(client admissionregistrationv1beta1client.Validatin
...
+
+	// The AdmissionReview that was sent to the webhook
+	requestedAdmissionReview := admissionv1beta1.AdmissionReview{}
+
+	// The AdmissionReview that will be returned
+	responseAdmissionReview := admissionv1beta1.AdmissionReview{}
+
+	deserializer := codecs.UniversalDeserializer()
+	if _, _, err := deserializer.Decode(body, nil, &requestedAdmissionReview); err != nil {
+		klog.Fatalf("decode failed with error: %v", err)
```
sids-b: Just a question. Do we need fatal log ? Fatal log will indicate stopping the service...

cloud/pkg/admissioncontroller/admission.go
```diff
@@ -195,3 +231,58 @@ func registerValidateWebhook(client admissionregistrationv1beta1client.Validatin
...
+	}
+	responseAdmissionReview.Response = admit(requestedAdmissionReview)
+
+	// Return the same UID
+	responseAdmissionReview.Response.UID = requestedAdmissionReview.Request.UID
+	klog.Infof("sending response: %v", responseAdmissionReview.Response)
+
+	respBytes, err := json.Marshal(responseAdmissionReview)
+	if err != nil {
+		klog.Fatalf("cannot marshal to a valid response %v", err)
```
sids-b: nit: please start logs with caps.
Applicable to all similar places

cloud/pkg/admissioncontroller/admit_device.go
```diff
@@ -0,0 +1,68 @@
...
+
+		mdSpec, ok := un.Object["spec"].(map[string]interface{})
+		if ok {
+			mdProto, ok := mdSpec["protocol"].(map[string]interface{})
+			if ok {
+				keys := reflect.ValueOf(mdProto).MapKeys()
+				//device should only have one protocol defined.
+				if len(keys) > 1 {
+					reviewResponse.Allowed = false
+					msg = msg + "multiple protocols found!"
```
sids-b: nit: please add a blank space either before or after the appended message to make it more readable

cloud/pkg/admissioncontroller/admission.go
chendave: I think so, thanks! will address this soon.

cloud/pkg/admissioncontroller/admit_device.go
chendave: Good catch!

sids-b:
Thanks @chendave for promptly addressing all the comments.

/approve

chendave:
let's revisit this later if this is needed, close it for now.

PR #1081, opened 2019-08-29
adjust the reference of udsserver

edisonxiang:
**What type of PR is this?**
/kind cleanup

Adjust the reference of uds server since the uds server is only used to communicate on cloud.
and websocket server & quic server are used to communicate between cloud and edge.

### DIFF
--- a/cloud/pkg/cloudhub/cloudhub.go
+++ b/cloud/pkg/cloudhub/cloudhub.go
@@ -13,6 +13,7 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/util"
 	chconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/config"
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/servers"
+	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/servers/udsserver"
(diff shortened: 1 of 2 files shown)


cloud/pkg/cloudhub/cloudhub.go
```diff
@@ -53,7 +54,7 @@ func (a *cloudHub) Start(c *context.Context) {
 	}
 
 	if util.HubConfig.ProtocolUDS {
-		go servers.StartCloudHub(servers.ProtocolUDS, eventq, c)
+		go udsserver.StartServer(util.HubConfig, c)
```
fisherxu: Could we add some simple comments for the usage of this udsserver？

edisonxiang: Done. Thanks

fisherxu: Thanks! Better not add the websocket and quic server here since we may add other protocols in future
```suggestion
              // The uds server is only used to communicate with csi driver from kubeedge on cloud.
		// It is not used to communicate between cloud and edge.
		go udsserver.StartServer(util.HubConfig, c)
```

edisonxiang: Done. Thanks

PR #1099, opened 2019-09-04
Device value not updated

subpathdev:
**What type of PR is this?**
/kind bug

the device values which are updated by the edge should be synchronized with the device resources in the K8S. (updating reported status)

the first two commits are issues, which are found in the process of identify the error

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -140,9 +140,9 @@ func (q *ChannelEventQueue) Publish(info *model.HubInfo, event *model.Event) err
 	msg := model.EventToMessage(event)
 	switch msg.Router.Source {
 	case model.ResTwin:
-		q.ctx.Send2Group("controller", msg)
+		q.ctx.Send2Group(model.SrcDeviceController, msg)
 	default:
-		q.ctx.Send2Group(model.SrcController, msg)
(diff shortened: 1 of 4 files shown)


sids-b:
I think a problem here is that edgecontroller group should be controller. Currently in edgecontroller/module.go the group function returns "edgecontroller" as module group.

Both devicecontroller and edgecontroller should either belong to same group "controller" or each of them should have a different group "edgecontroller" and "devicecontroller". The "controller" group should not be used only for devicecontroller.

subpathdev:
@sids-b yes you are right this is a problem but can we discuss this in an extra issue and get this merged?

sids-b:
Sure @subpathdev . Can we use sendToModule instead of sendToGroup here to avoid any confusions of module groups ?

subpathdev:
I have renamed the group of the device controllers to mentioned your comment @sids-b. I think this is the best way to handle it. In the beehive module which are used to forward the message to the group of channels is not Send2Module (or SendToModule) implemented.

subpathdev:
@kevin-wangzefeng please check again

sids-b:
Thanks @subpathdev 

/approve

/assign @kevin-wangzefeng

sids-b:
/hold 
for @kevin-wangzefeng re-approval

PR #1100, opened 2019-09-04
fix bug #1097

### LINKED ISSUE #1097: Edgemesh panic
**What happened**:
image

Judge the slice.

**Environment**:
- KubeEdge version:
- Hardware configuration:
- Kernel (e.g. `uname -a`):
- Others:

arcanique:
**What type of PR is this?**
/kind bug
 
fix bug #1097

Fixes #1097 

```release-note

```

### DIFF
--- a/edgemesh/pkg/proxy/proxy.go
+++ b/edgemesh/pkg/proxy/proxy.go
@@ -50,6 +50,7 @@ const (
 	defaultIpPoolSize        = 20
 	defaultTcpClientTimeout  = time.Second * 2
 	defaultTcpReconnectTimes = 3
+	firstPort                = 0
 )
 
 var (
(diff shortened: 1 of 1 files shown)


edgemesh/pkg/proxy/proxy.go
```diff
@@ -544,11 +545,14 @@ func (c *conntrack) processServerProxy() {
 
 //isL4Proxy Determine whether to use L4 proxy
 func IsL4Proxy(svc *v1.Service) bool {
+	if svc.Spec.Ports == nil || len(svc.Spec.Ports) == 0 {
```
fisherxu: Have `len(svc.Spec.Ports) == 0` this one is enough 
```suggestion
	if len(svc.Spec.Ports) == 0 {
```

fisherxu: And switch suggest to:
```suggestion
	switch port.Name {
	case "websocket", "grpc", "https", "tcp":
		return true
	case "http", "udp":
		return false
	default:
		return true
	}
```

fisherxu:
Thanks for quick fix, @arcanique ptal :)

edgemesh/pkg/proxy/proxy.go
arcanique: Thanks for suggestions. please review again

sids-b:
Thanks @arcanique 

/approve

Issue #272, opened 2019-03-20
Support PV, PVC and storageclass


CindyXing:
**What would you like to be added**:
1. Allow customers to store data using PV, PVC and storage class; and implement/deploy stateful app at edge. 
2. Expose ways for customer to config storage CSI plugin

**Why is this needed**:
Often customers need to store data and run stateful app. at edge

PR #1104, opened 2019-09-05
support env in edge pods

edisonxiang:
**What type of PR is this?**
 /kind feature

Currently the pods running at edge can not set the environment variables
by the deployment yaml and inject into the containers,
this pr is to support two kinds of ways to set environment variables.
* Directly setting
```
env:
  - name: CSI_DRIVERNAME
    value: csi-hostpath
```
* Reading from fieldRef
```
env:
  - name: KUBE_NODE_NAME
    valueFrom
      fieldRef:
        fieldPath: spec.nodeName
```

XREF #272


sids-b:
Thanks @edisonxiang 

/approve

PR #1143, opened 2019-09-22
modify  mappers/bluetooth_mapper/data_converter/data_converter.go

dhdljd:
**What type of PR is this?**

>

> /kind bug

fix  the data temperature read from cc2650

Fixes #

```release-note

```

### DIFF
--- a/mappers/bluetooth_mapper/data_converter/data_converter.go
+++ b/mappers/bluetooth_mapper/data_converter/data_converter.go
@@ -19,6 +19,7 @@ package dataconverter
 import (
 	"strconv"
 	"strings"
+	"encoding/hex"
 )
 
 // Bluetooth Protocol Operation type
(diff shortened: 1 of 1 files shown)


mappers/bluetooth_mapper/data_converter/data_converter.go
```diff
@@ -90,9 +91,7 @@ func (operation *ReadOperation) ConvertReadData(data []byte) float64 {
 			initialValue = append(initialValue, data[index])
 		}
 	}
-	for _, value := range initialValue {
-		initialStringValue = initialStringValue + strconv.Itoa(int(value))
-	}
+	initialStringValue = hex.EncodeToString(initialValue)
```
sujithsimon22: Hi @dhdljd ,
Thanks for the change
Did u get a chance to test this change ? Does it produce the same value as the previously expected ?

sujithsimon22: Could u pls test this change....I think the value produced will be different from what was previously returned....It would be helpful if u could verify this....

sujithsimon22: You could take a look use this guide if u would like to test your changes through the e2e env for bluetooth: https://github.com/kubeedge/kubeedge/blob/master/docs/guides/bluetooth_mapper_e2e_guide.md

PR #1145, opened 2019-09-23
Remove coreutils in alpine

osiriswd:
"apk --no-cache add coreutils" may failed when node can't access to Internet.
Using awk '{printf $0}' instead of base64 -w 0.

**What type of PR is this?**
>
> /kind api-change
> 
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Speedup the installation of deploy cloud part into a k8s cluster.

```release-note
NONE
```

### DIFF
--- a/build/cloud/07-deployment.yaml
+++ b/build/cloud/07-deployment.yaml
@@ -25,16 +25,12 @@ spec:
           mountPath: /etc/kubeedge/cloud
           subPath: kubeconfig.yaml
         command:
-        - /bin/sh
-        - -c
-        - |
-          cat | tee /etc/kubeedge/cloud/kubeconfig.yaml <<EOF
           apiVersion: v1
           kind: Config
           clusters:
           - name: kubeedge
             cluster:
-              certificate-authority-data: $(cat /var/run/secrets/kubernetes.io/serviceaccount/ca.crt | base64 | awk '{printf $0}')
+              certificate-authority: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
           users:
           - name: kubeedge
             user:
@@ -48,7 +44,7 @@ spec:
           EOF
       containers:
       - name: cloudcore
-        image: kubeedge/cloudcore:v1.0.0
+        image: kubeedge/cloudcore:v1.1.0
         imagePullPolicy: IfNotPresent
         ports:
         - containerPort: 10000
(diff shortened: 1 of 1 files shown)


build/cloud/07-deployment.yaml
```diff
@@ -28,13 +28,13 @@ spec:
...
         - |
-          apk --no-cache add coreutils && cat | tee /etc/kubeedge/cloud/kubeconfig.yaml <<EOF
+          cat | tee /etc/kubeedge/cloud/kubeconfig.yaml <<EOF
           apiVersion: v1
           kind: Config
           clusters:
           - name: kubeedge
             cluster:
-              certificate-authority-data: $(cat /var/run/secrets/kubernetes.io/serviceaccount/ca.crt | base64 -w 0)
+              certificate-authority-data: $(cat /var/run/secrets/kubernetes.io/serviceaccount/ca.crt | base64 | awk '{printf $0}')
```
johscheuer: Any reason we don't use simply `certificate-authority: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt` ? I mean actually there is no reason to put the ca into the kubeconfig since in all containers the ca will be mounted at ` /var/run/secrets/kubernetes.io/serviceaccount/ca.crt`

osiriswd: Thank @johscheuer ,use simply certificate-authority works fine here. So we don't need base64( or any other advanced functions in coreutils) anymore.

build/cloud/07-deployment.yaml
```diff
@@ -25,16 +25,12 @@ spec:
           mountPath: /etc/kubeedge/cloud
           subPath: kubeconfig.yaml
         command:
-        - /bin/sh
-        - -c
-        - |
-          apk --no-cache add coreutils && cat | tee /etc/kubeedge/cloud/kubeconfig.yaml <<EOF
```
johscheuer: The `sh` command is still required to print the file

fisherxu: I think we can delete the initContainer here, because cloudcore can use `serviceaccount that mount to pod` to connect to apiserver, thought? @johscheuer @osiriswd

fisherxu: No need construct the kubeconfig manually, K8s client-go can use incluster mode. 
cc @kadisi @chendave

osiriswd: Fixed. @johscheuer

fisherxu:
have fixed in #1524 
/close

PR #1176, opened 2019-10-08
update some comments

chendave:
godoc convention is that a function comment needs to start with
the function name.

**What type of PR is this?**

/kind cleanup

```release-note
NONE
```


sids-b:
Thanks @chendave 

/approve

sids-b:
I think we should enable lint for the admission webhook

PR #1178, opened 2019-10-08
change the way to get the HomeDir

NoWhere-zh:
**What type of PR is this?**
> /kind bug

change the way to get the HomeDir.
if the go version is lower than 1.12, using UserHomeDir() may cause compilation errors

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/credentialprovider/config.go
+++ b/vendor/k8s.io/kubernetes/pkg/credentialprovider/config.go
@@ -60,7 +60,7 @@ var (
 	preferredPathLock sync.Mutex
 	preferredPath     = ""
 	workingDirPath    = ""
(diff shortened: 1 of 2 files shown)


vendor/k8s.io/kubernetes/pkg/credentialprovider/config.go
```diff
@@ -60,7 +60,7 @@ var (
 	preferredPathLock sync.Mutex
 	preferredPath     = ""
 	workingDirPath    = ""
-	homeDirPath, _    = os.UserHomeDir()
```
fisherxu: Please don't modify the vendor files directly. If you need to modify them, you need raise PR in their repo first, then upgrade the vendor.

NoWhere-zh: get it. thank u very much.

PR #1184, opened 2019-10-10
Auto register edge node

### LINKED ISSUE #881: Create and register node from edge part
Now when install kubeedge, after install the cloud part, if user want to add a edge node, must create node object by user self. 
Consider create and register node all by edge part, like kubelet done in native k8s.

**Why is this needed**:
User only install edge part, no need create node manually.

### LINKED ISSUE #1045: Keadm join cmd should not has option `k8sserverip`
Now the keadm join cmd has the `k8sserverip` option, which means that need connect Kube-apiserver in edge node, but we have the `cloudcore ip` is enough.

**Why is this needed**:
Remove this option.

/cc @naveensriram @samy2019 @kadisi @subpathdev @sids-b @rohitsardesai83

sids-b:
**What type of PR is this?**
Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:

 /kind feature

Support auto-register edge nodes. We can still allow manual addition of nodes for the administrators to additionally configure nodes by adding labels and other fields.

Fixes #881 

 #1045 will be fixed in next PR

```release-note

```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -384,15 +384,40 @@ func (uc *UpstreamController) updateNodeStatus(stop chan struct{}) {
 			}
 
 			switch msg.GetOperation() {
+			case model.InsertOperation:
+				klog.Infof("Received register node request, creating node")
+				if _, err := uc.createNode(name); err != nil {
+					klog.Errorf("Failed to create node: %s with error: %s", name, err)
+					continue
+				}
+				resMsg := model.NewMessage(msg.GetID())
+				resMsg.Content = "OK"
+				nodeID, err := messagelayer.GetNodeID(msg)
+				if err != nil {
+					klog.Warningf("Message: %s process failure, get node id failed with error: %s", msg.GetID(), err)
+					continue
+				}
+				resource, err := messagelayer.BuildResource(nodeID, namespace, model.ResourceTypeNode, name)
+				if err != nil {
+					klog.Warningf("Message: %s process failure, build message resource failed with error: %s", msg.GetID(), err)
+					continue
+				}
+				resMsg.BuildRouter(constants.EdgeControllerModuleName, constants.GroupResource, resource, model.ResponseOperation)
+				if err = uc.messageLayer.Response(*resMsg); err != nil {
+					klog.Warningf("Message: %s process failure, response failed with error: %s", msg.GetID(), err)
+					continue
+				}
+
+				klog.V(4).Infof("message: %s, update node status successfully, name: %s", msg.GetID(), name)
+
 			case model.UpdateOperation:
 				getNode, err := uc.kubeClient.CoreV1().Nodes().Get(name, metaV1.GetOptions{})
 				if errors.IsNotFound(err) {
-					klog.Infof("Node: %s not found, creating node", name)
-					if getNode, err = uc.createNode(name); err != nil {
-						klog.Errorf("Failed to create node: %s with error: %s", name, err)
-						continue
-					}
-				} else if err != nil {
+					klog.Warningf("message: %s process failure, node %s not found", msg.GetID(), name)
+					continue
+				}
+
+				if err != nil {
 					klog.Warningf("message: %s process failure with error: %s, namespaces: %s name: %s", msg.GetID(), err, namespace, name)
 					continue
 				}
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -333,6 +333,18 @@ func (uc *UpstreamController) updatePodStatus(stop chan struct{}) {
 	}
 }
 
+func (uc *UpstreamController) createNode(name string) (*v1.Node, error) {
+	node := &v1.Node{}
+	labels := make(map[string]string)
+	labels["name"] = name
+	labels["node-role.kubernetes.io/edge"] = ""
+
+	// create node
+	node.Name = name
+	node.Labels = labels
+	return uc.kubeClient.CoreV1().Nodes().Create(node)
+}
+
 func (uc *UpstreamController) updateNodeStatus(stop chan struct{}) {
 	running := true
 	for running {
@@ -375,11 +387,12 @@ func (uc *UpstreamController) updateNodeStatus(stop chan struct{}) {
 			case model.UpdateOperation:
 				getNode, err := uc.kubeClient.CoreV1().Nodes().Get(name, metaV1.GetOptions{})
 				if errors.IsNotFound(err) {
-					klog.Warningf("message: %s process failure, node %s not found", msg.GetID(), name)
-					continue
-				}
-
-				if err != nil {
+					klog.Infof("Node: %s not found, creating node", name)
+					if getNode, err = uc.createNode(name); err != nil {
+						klog.Errorf("Failed to create node: %s with error: %s", name, err)
+						continue
+					}
+				} else if err != nil {
 					klog.Warningf("message: %s process failure with error: %s, namespaces: %s name: %s", msg.GetID(), err, namespace, name)
 					continue
 				}
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -408,7 +408,7 @@ func (uc *UpstreamController) updateNodeStatus(stop chan struct{}) {
 					continue
 				}
 
-				klog.V(4).Infof("message: %s, update node status successfully, name: %s", msg.GetID(), name)
+				klog.V(4).Infof("message: %s, create node successfully, name: %s", msg.GetID(), name)
 
 			case model.UpdateOperation:
 				getNode, err := uc.kubeClient.CoreV1().Nodes().Get(name, metaV1.GetOptions{})
--- a/docs/getting-started/usage.md
+++ b/docs/getting-started/usage.md
@@ -106,14 +106,40 @@ The cert/key will be generated in the `/etc/kubeedge/ca` and `/etc/kubeedge/cert
 ### Run Edge
 
 #### Deploy the Edge node
-We have provided a sample node.json to add a node in kubernetes. Please make sure edge-node is added in kubernetes. Run below steps to add edge-node.
+KubeEdge supports auto registration of edge-nodes but we have provided a sample node.json to manually add a node in kubernetes if required. 
+Run below steps to add edge-node.
 
-+ Modify the `$GOPATH/src/github.com/kubeedge/kubeedge/build/node.json` file and change `metadata.name` to the name of the edge node
-+ Deploy node
++ Copy the `$GOPATH/src/github.com/kubeedge/kubeedge/build/node.json` file and change `metadata.name` to the name of the edge node
+
     ```shell
-    kubectl apply -f $GOPATH/src/github.com/kubeedge/kubeedge/build/node.json
+        mkdir ~/cmd/yaml
+        cp $GOPATH/src/github.com/kubeedge/kubeedge/build/node.json ~/cmd/yaml
     ```
-+ Transfer the certificate file to the edge node
+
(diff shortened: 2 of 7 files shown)


docs/getting-started/usage.md
```diff
@@ -106,14 +106,39 @@ The cert/key will be generated in the `/etc/kubeedge/ca` and `/etc/kubeedge/cert
...
++ Copy the `$GOPATH/src/github.com/kubeedge/kubeedge/build/node.json` file and change `metadata.name` to the name of the edge node
+    
+    ```shell
+        mkdir ~/cmd/yaml
+        cp $GOPATH/src/github.com/kubeedge/kubeedge/build/node.json ~/cmd/yaml
+    ```
+    
++ Make sure role is set to edge for the node. For this a key of the form `"node-role.kubernetes.io/edge"` must be present in `labels` tag of `metadata`.
++ Please ensure to add the label `node-role.kubernetes.io/edge` to the `build/node.json` file.
+
```
kadisi: Why support manual registration since it supports automatic registration?

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -375,11 +387,12 @@ func (uc *UpstreamController) updateNodeStatus(stop chan struct{}) {
 			case model.UpdateOperation:
```
fisherxu: Can we add a create option here? So users can use a flag in edgecore to set whether to manually create node like kubelet.  ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_node_status.go#L370

docs/getting-started/usage.md
sids-b: We can still allow manual addition of nodes for the administrators to additionally configure nodes by adding labels and other fields. By default only role label and name label is added. But if someone wants to add nodes with common labels so that they can be grouped together, it can be useful.

cloud/pkg/edgecontroller/controller/upstream.go
sids-b: I was thinking we keep an option to allow auto registration of nodes at the edgecontroller side. If is it allowed, auto register all nodes that are connecting else reject all. Thoughts ?

kadisi: I think it should be decided by edgecore whether to auto-register instead of cloudcore

fisherxu: Yes, we should do it in edgecore, consistent with kubelet

sids-b: Okay Sure, I will add the insert option in edge_core

sids-b: We already have code in edged. The messages were getting discarded at the cloud side.
https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edged/edged_status.go#L413

sids-b: I guess this change is sufficient or you guys will like me to add something more ?

fisherxu: https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edged/edged_status.go#L413
L413 only register nodestatus in edged, we need a registerNode flag in edgecore. And move `createNode` func to edged. Can refer to the code of kubelet.

tests/e2e/edgesite/edgesite_suite_test.go
```diff
@@ -45,9 +45,8 @@ func TestEdgecoreAppDeployment(t *testing.T) {
 		nodeName = "integration-node-" + utils.GetRandomString(10)
 		nodeSelector = "node-" + utils.GetRandomString(3)
 		Expect(utils.DeploySetup(ctx, nodeName, "edgesite")).Should(BeNil())
-		Expect(utils.StartEdgeSite()).Should(BeNil())
-
 		Expect(utils.RegisterNodeToMaster(nodeName, ctx.Cfg.K8SMasterForKubeEdge+constants.NodeHandler, nodeSelector)).Should(BeNil())
```
fisherxu: Auto register edge node in default, we can remove this line?

cloud/pkg/edgecontroller/controller/upstream.go
sids-b: We already have a function similar to what is there is kubelet for createNode in edged which creates a nodeStatus.
https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edged/edged_status.go#L50

tests/e2e/edgesite/edgesite_suite_test.go
sids-b: It creates node with a node-selector label so that all the pods for the testcase are scheduled on this node. Since we have common setup, it maybe required, once we move to dynamic setup we can remove it.

kadisi:
Suppose there already exist a node (named `node1`) in kubernetes,  and user want to add a new edged node(also named `node1`, Maybe he doesn't know that there is already a node named ` node1`.).   We assume that they set the same projectid 。   What will happens?

sids-b:
> Suppose there already exist a node (named `node1`) in kubernetes, and user want to add a new edged node(also named `node1`, Maybe he doesn't know that there is already a node named ` node1`.). We assume that they set the same projectid 。 What will happens?

Both the nodes will start updating the status of the same node. It happens with kubelet also if 2 machines have same hostname.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -372,6 +384,32 @@ func (uc *UpstreamController) updateNodeStatus(stop chan struct{}) {
...
+					klog.Warningf("Message: %s process failure, build message resource failed with error: %s", msg.GetID(), err)
+					continue
+				}
+				resMsg.BuildRouter(constants.EdgeControllerModuleName, constants.GroupResource, resource, model.ResponseOperation)
+				if err = uc.messageLayer.Response(*resMsg); err != nil {
+					klog.Warningf("Message: %s process failure, response failed with error: %s", msg.GetID(), err)
+					continue
+				}
+
+				klog.V(4).Infof("message: %s, update node status successfully, name: %s", msg.GetID(), name)
```
kevin-wangzefeng:
```suggestion
				klog.V(4).Infof("message: %s, create node successfully, name: %s", msg.GetID(), name)
```

sids-b: Done, Thanks

sids-b:
I like this new feature of github which directly lets you commit the review of reviewers :)

docs/setup/setup.md
```diff
@@ -147,7 +147,8 @@ The cert/key will be generated in the `/etc/kubeedge/ca` and `/etc/kubeedge/cert
 
 
 #### Deploy the edge node
-We have provided a sample node.json to add a node in kubernetes. Please make sure edge-node is added in kubernetes. Run below steps to add edge-node.
+KubeEdge supports auto registration of edge-nodes but we have provided a sample node.json to manually add a node in kubernetes if required. 
+Run below steps to add edge-node.
```
kadisi: We should add  the `register-node: true` configuration  in line287 of setup.md

edge/pkg/edged/edged.go
```diff
@@ -287,6 +288,23 @@ func (e *edged) Start(c *context.Context) {
...
 	e.metaClient = client.New(c)
 
+	autoRegister := config.CONFIG.GetConfigurationByKey("edged.register-node").(bool)
+	if autoRegister {
+		go func() {
+			// keep trying to register with apiserver every 10 seconds
+			for {
+				resource := fmt.Sprintf("%s/%s/%s", e.namespace, model.ResourceTypeNodeStatus, e.nodeName)
+				msg := message.BuildMsg(modules.MetaGroup, "", modules.EdgedModuleName, resource, model.InsertOperation, "")
+				_, err := e.context.SendSync("websocket", *msg, 60*time.Second)
```
kadisi: for `websocket` , Recommended to use constants 。 https://github.com/kubeedge/kubeedge/blob/8819f62b8b4db5f0f83528ad30a1747e9f933afb/edge/pkg/edgehub/module.go#L11

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -372,6 +384,32 @@ func (uc *UpstreamController) updateNodeStatus(stop chan struct{}) {
 			}
 
 			switch msg.GetOperation() {
+			case model.InsertOperation:
+				klog.Infof("Received register node request, creating node")
+				if _, err := uc.createNode(name); err != nil {
```
fisherxu: We should receive the node obj instead of nodename from edge, kubelet has some features(flags like taints/labels) that set to node.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -372,6 +384,32 @@ func (uc *UpstreamController) updateNodeStatus(stop chan struct{}) {
...
+					klog.Errorf("Failed to create node: %s with error: %s", name, err)
+					continue
```
fisherxu: we have retried in edgepart, continue is OK here :)

docs/getting-started/usage.md
```diff
@@ -106,14 +106,40 @@ The cert/key will be generated in the `/etc/kubeedge/ca` and `/etc/kubeedge/cert
 ### Run Edge
 
 #### Deploy the Edge node
-We have provided a sample node.json to add a node in kubernetes. Please make sure edge-node is added in kubernetes. Run below steps to add edge-node.
+KubeEdge supports auto registration of edge-nodes but we have provided a sample node.json to manually add a node in kubernetes if required. 
+Run below steps to add edge-node.
```
fisherxu:
```suggestion
KubeEdge supports auto registration of edge-nodes by default. But we also support manually add edge-nodes in kubernetes if required.
Run below steps to add edge-node manually. 
```

docs/getting-started/usage.md
```diff
@@ -106,14 +106,40 @@ The cert/key will be generated in the `/etc/kubeedge/ca` and `/etc/kubeedge/cert
...
+
++ Copy the `$GOPATH/src/github.com/kubeedge/kubeedge/build/node.json` file and change `metadata.name` to the name of the edge node
+
+    ```shell
+        mkdir ~/cmd/yaml
+        cp $GOPATH/src/github.com/kubeedge/kubeedge/build/node.json ~/cmd/yaml
+    ```
+
++ Make sure role is set to edge for the node. For this a key of the form `"node-role.kubernetes.io/edge"` must be present in `labels` tag of `metadata`.
```
fisherxu: Can delete this line? Add lable is enough. :)

docs/getting-started/usage.md
```diff
@@ -106,14 +106,40 @@ The cert/key will be generated in the `/etc/kubeedge/ca` and `/etc/kubeedge/cert
...
+        "name": "edge-node",
+        "labels": {
+          "name": "edge-node",
+          "node-role.kubernetes.io/edge": ""
+        }
+      }
+    }
+    ```
+    **Note: you need to remember `metadata.name` , because edgecore need it**.
++ If role is not set for the node, the pods, configmaps and secrets created/updated in the cloud cannot be synced with the node they are targeted for.
```
fisherxu: lable is not set for the node. lable and role is different in k8s

cloud/pkg/edgecontroller/controller/upstream.go
kadisi: @fisherxu continue is continue to the next for loop

fisherxu: yes, I mean OK :)

fisherxu:
Any updates?

kuramal:
@sids-b @fisherxu Is there any plan of update?

fisherxu:
/close
Dumplicate with #1401

PR #1190, opened 2019-10-11
fix: immediate return when dns server start err

beautytiger:
**What type of PR is this?**
 /kind bug

Fixes #

```release-note
NONE
```

### DIFF
--- a/edgemesh/pkg/server/dns.go
+++ b/edgemesh/pkg/server/dns.go
@@ -93,10 +93,11 @@ func startDnsServer() {
 		Port: 53,
 	}
 	udpConn, err := net.ListenUDP("udp", laddr)
-	defer udpConn.Close()
 	if err != nil {
 		klog.Errorf("Dns server Start error : %s", err)
+		return
 	}
+	defer udpConn.Close()
 	dnsConn = udpConn
 	for {
 		req := make([]byte, bufSize)


edgemesh/pkg/server/dns.go
```diff
@@ -93,10 +93,11 @@ func startDnsServer() {
 		Port: 53,
 	}
 	udpConn, err := net.ListenUDP("udp", laddr)
-	defer udpConn.Close()
 	if err != nil {
 		klog.Errorf("Dns server Start error : %s", err)
+		return
 	}
+	defer udpConn.Close()
```
sids-b: does this line cause a panic if it is above the if condition ?

beautytiger: No, but the returned conn will be nil, it is no need to close it, like a nil file.
And the following process is also not needed, am I right?

fisherxu:
Oh, just omit my comments, dnsConn is a global variable.
/lgtm

edgemesh/pkg/server/dns.go
fisherxu: Yes, if it is above the if condition, it will cause a panic.
e.g. https://play.golang.org/p/2SpVzVrnGjg

PR #1194, opened 2019-10-15
Enable travis CI on arm64

### LINKED ISSUE #1185: CI available on arm
CI support on arm platform

**Why is this needed**:
current travis CI only support on x86 platform, it would be great to support arm as well, so that we can find the potential issues to run/build kubeedge on arm

chendave:
This patch also did some cleanup.

**What type of PR is this?**

/kind feature

Fixes #1185 

```release-note
travis CI on ARM64 is enabled
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,15 +21,16 @@ before_install:
 before_script:
   - curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | BINARY=golangci-lint sh -s -- -d -b $(go env GOPATH)/bin v1.20.0
   - mkdir -p $HOME/gopath/src/github.com/kubeedge/kubeedge
+  - sudo apt-get install build-essential -y
+  - sudo apt-get install gcc-arm-linux-gnueabi -y
+  - sudo apt-get install upx-ucl -y
+  - sudo apt-get install rsync -y
   - rsync -az ${TRAVIS_BUILD_DIR}/ $HOME/gopath/src/github.com/kubeedge/kubeedge/
   - export TRAVIS_BUILD_DIR=$HOME/gopath/src/github.com/kubeedge/kubeedgee
   - go get github.com/onsi/ginkgo/ginkgo
   - export GO111MODULE=on
   - export GOFLAGS=-mod=vendor
   - cd $HOME/gopath/src/github.com/kubeedge/kubeedge
-  - sudo apt-get install build-essential
-  - sudo apt-get install gcc-arm-linux-gnueabi
-  - sudo apt-get install upx-ucl
 
 install: true
 
@@ -41,7 +42,7 @@ jobs:
 
     - stage: "Test lint"
       script: make cloud_lint
-      name: "cloud_lint"     # names the eighth Tests stage job
+      name: "cloud_lint"
     - script: make bluetoothdevice_lint
       name: "bluetooth_device_lint"
     - script: make keadm_lint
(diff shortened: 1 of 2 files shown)


chendave:
- lint is not included as I think there should no difference for the lint check on different ARCH.
- some of the test is masked cause they are cannot on ARM64 so far, will address them in the following PR.

chendave:
cc @fisherxu  @kadisi

.travis.yml
```diff
@@ -21,70 +21,123 @@ before_install:
...
+      name: "cross build edgeite for armv7"
 
     - stage: "unit tests"
       script: make edge_test
-      name: "unit test edge"     # names the third Tests stage job
+      arch: amd64
+      name: "unit test edge on amd64"
+      #- script: make edge_test
+      #- arch: arm64
+      #- name: "unit test edge on arm64"
```
subpathdev: please remove these comments

subpathdev:
I think we could remove the docker build tests for arm64 or amd64.

In these test we are testing if the dockerfile is correct but these could be done by testing on one architecture.

.travis.yml
```diff
@@ -21,70 +21,123 @@ before_install:
...
+      arch: arm64
+      name: "unit test cloud on arm64"
 
     - stage: "e2e tests"
       script: make e2e_test
-      name: "e2e_test"     # names the seventh Tests stage job
+      arch: amd64
+      name: "e2e_test on amd64"
+      # - script: make e2e_test
+      # - arch: arm64
```
fisherxu: E2e test in arm OS need modify the config like pause container image.

.travis.yml
```diff
@@ -21,70 +21,123 @@ before_install:
...
+      #- script: make edge_integration_test
+      #- arch: arm64
+      #- name: "integration test edge on arm64"
+    - script: make cloud_test
+      arch: amd64
+      name: "unit test cloud on amd64"
     - script: make cloud_test
-      name: "unit test cloud" 
+      arch: arm64
+      name: "unit test cloud on arm64"
```
subpathdev: do we need unit tests on both architectures? I mean the unit test will test if we have made failures on the code and not if it works on this architecture.

.travis.yml
subpathdev: I think this is an important test and we should modify the configuration / test to test both arm64 AND amd64.

fisherxu: Yes, sure, I think @chendave is still testing.

fisherxu: Do we need arm32?

.travis.yml
chendave: this going to be unmasked in the follow on PR along with the fix of the failure.

chendave:
> I think we could remove the docker build tests for arm64 or amd64.
> 
> In these test we are testing if the dockerfile is correct but these could be done by testing on one architecture.

AFAIK, image incompatibility on different ARCH is an important issue we want to capture, compared with X86, docker images are not so well-supported on ARM, we either need to use multi-arch images or should explicitly specify which ARCH the image is sued for. Here is the example: https://github.com/kubeedge/kubeedge/pull/1194/files#diff-354f30a63fb0907d4ad57269548329e3R135

.travis.yml
```diff
@@ -21,70 +21,123 @@ before_install:
...
-    - stage: "Test Build"                # naming the Tests stage
-      script: make 
-      name: "builds cloud and edge components"            # names the first Tests stage job
+    - stage: "Test Build"
+      script: make
+      arch: amd64
+      name: "builds cloud and edge components on amd64"
+    - script: make
+      arch: arm64
+      name: "builds cloud and edge components on arm64"
```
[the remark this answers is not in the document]
chendave: cross-build should not be included since qemu doesn't support on ARM, cause this script passed I guess cross-build is not included in the default target.

.travis.yml
chendave: agreed, but we need make sure all the dependencies needed for the unit test also passed on the ARM, i.e. we need to be able to run those unit test on ARM as well.

.travis.yml
chendave: let's go with arm32 first, I am confident with arm32. :) 
this will going to unmasked when the issue fixed, this PR is the ground work for the feature.

.travis.yml
```diff
@@ -21,70 +21,123 @@ before_install:
...
 jobs:
   include:
     - stage: "Verify vendor"
+      arch: amd64
       script: make verify
-      name: "check if the vendor is up to date"
+      name: "check if the vendor is up to date on amd64"
+    - script: make verify
+      arch: arm64
+      name: "check if the vendor is up to date on arm64"
```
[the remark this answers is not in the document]
chendave: this can be removed from the check list.

sids-b:
THanks @chendave

PR #1209, opened 2019-10-18
use cosntant of modules.enabled for benefit && refactor NewChannelCon…

soolaugust:
…text to singleton

**What type of PR is this?**
>
> /kind feature

1. use constant instead of hard-code string is a good practice.
2. check one small todo item about refactoring NewChannelContext to singleton

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
@@ -21,6 +21,11 @@ const (
 	TickerTimeoutDefault = 20 * time.Millisecond
 )
 
+var (
+	channelContext *ChannelContext
+	onceForChannel sync.Once
+)
+
 // ChannelContext is object for Context channel
 type ChannelContext struct {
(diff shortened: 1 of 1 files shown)


staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
```diff
@@ -33,16 +38,18 @@ type ChannelContext struct {
...
 func NewChannelContext() *ChannelContext {
-	channelMap := make(map[string]chan model.Message)
-	moduleChannels := make(map[string]map[string]chan model.Message)
-	anonChannels := make(map[string]chan model.Message)
-	return &ChannelContext{
-		channels:     channelMap,
-		typeChannels: moduleChannels,
-		anonChannels: anonChannels,
-	}
+	onceForChannel.Do(func() {
```
fisherxu: I noticed several call points for this func(NewChannelContext), and I'm afraid that this function will be used to return different `ChannelContext`. 
If we use the `once`, return the same channelContext, thoughts? @sids-b @kadisi

PR #1222, opened 2019-10-23
make webhook service port configurable

chendave:

**What type of PR is this?**
>

/kind bug

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/admissioncontroller/admission.go
+++ b/cloud/pkg/admissioncontroller/admission.go
@@ -158,6 +158,7 @@ func (ac *AdmissionController) registerWebhooks(opt *options.AdmissionOptions, c
 						Namespace: opt.AdmissionServiceNamespace,
 						Name:      opt.AdmissionServiceName,
 						Path:      strPtr("/devicemodels"),
+						Port:      &opt.Port,
 					},
 					CABundle: cabundle,
 				},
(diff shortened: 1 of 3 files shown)


chendave:
cc @fisherxu

cloud/pkg/admissioncontroller/admission.go
```diff
@@ -158,6 +165,7 @@ func (ac *AdmissionController) registerWebhooks(opt *options.AdmissionOptions, c
 						Namespace: opt.AdmissionServiceNamespace,
 						Name:      opt.AdmissionServiceName,
 						Path:      strPtr("/devicemodels"),
+						Port:      int32Ptr(opt.Port),
```
fisherxu: I think we can change port type to `int32` in opts? @chendave

chendave: thank you for the comments, there is no int32 flags as IntVar. we cannot make it as a int32 options. 

see: https://github.com/kubeedge/kubeedge/blob/master/cloud/cmd/admission/app/options/options.go#L50

fisherxu: How about use `Int32Var` instead?  https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-controller-manager/app/options/replicasetcontroller.go#L36

PR #1228, opened 2019-10-25
Fix local-up script

fisherxu:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/lib/install.sh
+++ b/hack/lib/install.sh
@@ -15,8 +15,8 @@
 # limitations under the License.
 
 # check if kubectl installed
-function check_prerequisites {
-  echo "checking prerequisites"
+function check_kubectl {
+  echo "checking kubectl"
   which kubectl >/dev/null 2>&1
   if [[ $? -ne 0 ]]; then
     echo "kubectl not installed, exiting."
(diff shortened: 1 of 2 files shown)


hack/lib/install.sh
```diff
@@ -33,6 +33,10 @@ function check_kind {
   if [[ $? -ne 0 ]]; then
```
subpathdev: can we use one if-clause here?

subpathdev: the problem is, that this the code in this if clause will not be executed. Because echo will return a success and then you test if echo is executed successfully

fisherxu: This if-clause is used to test whether kind already been installed, see L32

hack/lib/install.sh
```diff
@@ -33,6 +33,10 @@ function check_kind {
...
     echo "installing kind ."
     GO111MODULE="on" go get sigs.k8s.io/kind@v0.5.1
+    if [[ $? -ne 0 ]]; then
```
fisherxu: This if-clause is used to test whether kind is installed successfully :)

hack/lib/install.sh
fisherxu: This if-clause is used to test whether kind already been installed

subpathdev: yes I found my fault

hack/lib/install.sh
subpathdev: yes I found my fault

PR #1258, opened 2019-11-13
UPDATE: Delete redundant code

du2016:
**What type of PR is this?**
UPDATE: Delete redundant code

/kind cleanup

the same as https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/config/config.go#L97,if the  controller protocol type is quic,it's not need to check about ws url


kadisi:
@du2016  please squash your commits.

du2016:
conflict  with other merge pull request,i'll close and recommit

PR #1265, opened 2019-11-15
delete redundant code

du2016:
What type of PR is this?
UPDATE: delete redundant code

/kind cleanup

the same as https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/config/config.go#L97


fisherxu:
@du2016  PR need rebaseed, thanks :)

fisherxu:
/lgtm 
Thanks.

Issue #1117, opened 2019-09-11
Container-Runtime and K8S should not be in the management scope of keadm


sids-b:
**What happened**:
Currently keadm installs of Container-Runtime (Docker) and Kubernetes

**What you expected to happen**:
keadm should only be responsible for management of kubeedge components. The cloud and edge part. We should provide link of guides to setup other prerequisites like docker, containerd, cni, k8s, etc.

sids-b:
Thoughts @kubeedge/maintainers @fisherxu @edisonxiang @chendave @kadisi @subpathdev @sujithsimon22 @lidiyag

fisherxu:
Agreed.  In the cloud and edge side, users can install the k8s and runtime in their own way, and keadm only responsible for kubeedge components.

kadisi:
agree

subpathdev:
+1

sujithsimon22:
Agreed, keadm should be primarily responsible for the setup of only KubeEdge components. But should we also think about have an optional mechanism in place to install prerequisites like Docker and Kubernetes ( This could be helpful for first time users/evaluators of KubeEdge, who might be unfamiliar with these technologies and might just be looking for an easy start-up mechanism). Just an alternate line of thought, what are your opinions on this ?

lidiyag:
> Agreed, keadm should be primarily responsible for the setup of only KubeEdge components. But should we also think about have an optional mechanism in place to install prerequisites like Docker and Kubernetes ( This could be helpful for first time users/evaluators of KubeEdge, who might be unfamiliar with these technologies and might just be looking for an easy start-up mechanism). Just an alternate line of thought, what are your opinions on this ?

+1

rzr:
Yes I have the same opinion, I am patching HEAD to remove installer part and rely on static binaries built from source not downloaded from random location, my longer term project is to provide working deb packages that can be installed on (any) debian system (or derived like ubuntu, raspbian etc).

Track:
https://dl.bintray.com/rzr/deb/pool/k/kubeedge/

Relate-to: https://talk.maemo.org/showthread.php?p=1558991#post1558991

bitvijays:
@fisherxu Maybe we can close this?

fisherxu:
Let's close it. 
/close

PR #1266, opened 2019-11-16
keadm: Installer may be disabled if offline mode is desired

rzr:
Configurations steps are still here.

For smooth transition, the curent behavior is preserved.

Change-Id: Icbf96b5e2b2179671bb09d8f7b5152c547038b70
Relate-to: https://github.com/kubeedge/kubeedge/issues/1117
Signed-off-by: Philippe Coval <p.coval@samsung.com>

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Fixes #

```release-note

```

### DIFF
--- a/keadm/app/cmd/util/common.go
+++ b/keadm/app/cmd/util/common.go
@@ -32,6 +32,8 @@ import (
 
 //Constants used by installers
 const (
+	InstallersOnline = true
+
 	UbuntuOSType = "ubuntu"
 	CentOSType   = "centos"
 
--- a/keadm/app/cmd/util/edgecoreinstaller.go
+++ b/keadm/app/cmd/util/edgecoreinstaller.go
@@ -51,9 +51,12 @@ func (ku *KubeEdgeInstTool) InstallTools() error {
 	ku.SetOSInterface(GetOSInterface())
 	ku.SetKubeEdgeVersion(ku.ToolVersion)
 
-	err := ku.InstallKubeEdge()
-	if err != nil {
-		return err
+	var err error
+	if InstallersOnline {
+		err = ku.InstallKubeEdge()
+		if err != nil {
+			return err
+		}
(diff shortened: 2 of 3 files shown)


keadm/app/cmd/util/common.go
```diff
@@ -32,6 +32,8 @@ import (
 
 //Constants used by installers
 const (
+	InstallersOnline = true
```
fisherxu: How to change the mode mow? :)

keadm/app/cmd/util/edgecoreinstaller.go
```diff
@@ -51,9 +51,12 @@ func (ku *KubeEdgeInstTool) InstallTools() error {
...
 
-	err := ku.InstallKubeEdge()
-	if err != nil {
-		return err
+	var err error
+	if InstallersOnline {
+		err := ku.InstallKubeEdge()
+		if err != nil {
+			return err
+		}
```
fisherxu: In cloudcore, If offline mode, we only start binary, don't copy again, why not add https://github.com/kubeedge/kubeedge/pull/1266/files#diff-f7a2c5d5ad972d622b21fe88054b53c3R184-R193 here?

keadm/app/cmd/util/common.go
rzr: patch source ? this is what i am doing with by my debian packaging branch maybe I could add an --offline flag ?

keadm/app/cmd/util/edgecoreinstaller.go
rzr: thx maybe I'll share an other patch

Is the feature still desirable ?

/assign @kevin-wangzefeng

PR #1270, opened 2019-11-18
change kubeconfig to absolute path

du2016:
**What type of PR is this?**

add support '~' for kubeconfig path,now will return error with "no such file or directory".
golang default can not recognized ~ path

/kind bug

### DIFF
--- a/cloud/conf/controller.yaml
+++ b/cloud/conf/controller.yaml
@@ -5,7 +5,7 @@ controller:
     qps: 5
     burst: 10
     node_update_frequency: 10
-    kubeconfig: "~/.kube/config"   #Enter path to kubeconfig file to enable https connection to k8s apiserver, if master and kubeconfig are both set, master will override any value in kubeconfig.
+    kubeconfig: "/root/.kube/config"   #Enter absolute path to kubeconfig file to enable https connection to k8s apiserver, if master and kubeconfig are both set, master will override any value in kubeconfig.
 cloudhub:
   protocol_websocket: true # enable websocket protocol
   port: 10000 # open port for websocket server
@@ -27,4 +27,4 @@ devicecontroller:
     content_type: "application/vnd.kubernetes.protobuf"
     qps: 5
     burst: 10
-    kubeconfig: "~/.kube/config" #Enter path to kubeconfig file to enable https connection to k8s apiserver,if master and kubeconfig are both set, master will override any value in kubeconfig.
+    kubeconfig: "/root/.kube/config" #Enter absolute path to kubeconfig file to enable https connection to k8s apiserver,if master and kubeconfig are both set, master will override any value in kubeconfig.
--- a/docs/setup/setup.md
+++ b/docs/setup/setup.md
@@ -93,7 +93,7 @@ The cert/key will be generated in the `/etc/kubeedge/ca` and `/etc/kubeedge/cert
         qps: 5
         burst: 10
         node_update_frequency: 10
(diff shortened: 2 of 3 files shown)


du2016:
test success on my local

/retest

kadisi:
@du2016 Have you tested kubernetes  whether support ` ~` path about kubeconfig?

kadisi:
@du2016 Have you tested kubernetes  whether support ` ~` path about kubeconfig?

du2016:
@kadisi  k8s client go is not support ~,i have a simple test like this:

```
KUBECONFIG="~/.kube/config" kubectl get pods
W1119 10:30:54.957146   21897 loader.go:223] Config not found: ~/.kube/config
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

if not add "  like KUBECONFIG=~/.kube/config,system will auto analysis to /root/.kube/config

kadisi:
Do we need to consider consistency with k8s or make a comment ？

du2016:
i think this is because not achieve by golang,or we should not use ~ in config system,in kubernetes,we usually use KUBECONFIG equal some path to ~/...  ,then the system will auto analysis to home dir,but in kubeedge used [go-archaius](github.com/go-chassis/go-archaius),~ just is a string,our local not have a dir named ~

cloud/pkg/devicecontroller/config/kube.go
```diff
@@ -33,6 +35,11 @@ func InitKubeConfig() {
 	if kc, err := config.CONFIG.GetValue("devicecontroller.kube.kubeconfig").ToString(); err != nil {
 		klog.Error("Devicecontroller kube config not set")
 	} else {
+		if strings.HasPrefix(kc, "~") {
+			if homeDir, err := os.UserHomeDir(); err == nil {
+				kc = strings.Replace(kc, "~", homeDir, 1)
```
subpathdev: please use '~/' instead of '~' ; it is possible to enter an other user home directory by using '~otherUser'

kadisi:
/hold

If k8s does not support it, we do not need to support it. It is recommended to keep consistent with the behavior of k8s。

xiliangMa:
I have the sane issue  use kubeedge 1.1.0.
```
W1126 10:39:49.056613   29606 kubeclient.go:12] get kube config failed with error: stat ~/.kube/config: no such file or directory
W1126 10:39:49.056636   29606 upstream.go:1155] create kube client failed with error: stat ~/.kube/config: no such file or directory
E1126 10:39:49.056663   29606 controller.go:44] new upstream controller failed with error: stat ~/.kube/config: no such file or directory
` ``

fisherxu:
> If k8s does not support it, we do not need to support it. It is recommended to keep consistent with the behavior of k8s。

So could we change all the kubeconfigs' path to absolute path in all places in kubeedge? @du2016 @kadisi @sids-b

du2016:
@fisherxu i'am agree

fisherxu:
@du2016 Ok, would you like change them in this PR? :)

du2016:
OK，i will change it soon

fisherxu:
@du2016 Thanks! https://github.com/kubeedge/kubeedge/blob/master/keadm/app/cmd/cloud/init.go#L41  Also plz help fix this line :) And could you revert the first commit, I think we can fix the  '~'  issue in k8s upstream's  client-go.

du2016:
@fisherxu, i have think about this,but we can't fix it in upstream,because `~`is only in shell environment,i have test other language,also not support

fisherxu:
@du2016  Yes, I have tried, client-go don't support `~`, BUT kubectl can use the KUBECONFIG env like `~/.kube/config` ;-)

kadisi:
Did kubectl handle it?

PR #1271, opened 2019-11-18
Use the same golang base image to build images 

johscheuer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Use the same golang base image to build images  and update cloudcore version in deployment manifest to latest tag. This makes the builds more consistent since they use the same golang version (if triggered at the same time and no push of a newer golang version happens). Also I updated the deployment manifest to use the latest (git) tag `1.1.0` instead of 1.0.0 which doesn't exits on docker hub.

Fixes #

```release-note
NONE
```

### DIFF
--- a/build/cloud/Dockerfile
+++ b/build/cloud/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.12-alpine3.9 AS builder
+FROM golang:1.12-alpine3.10 AS builder
 
 COPY . /go/src/github.com/kubeedge/kubeedge
 
--- a/build/cloud/Dockerfile
+++ b/build/cloud/Dockerfile
@@ -6,7 +6,7 @@ RUN CGO_ENABLED=0 go build -v -o /usr/local/bin/cloudcore -ldflags="-w -s" \
 github.com/kubeedge/kubeedge/cloud/cmd/cloudcore
 
 
-FROM gcr.io/distroless/static
+FROM alpine:3.9
 
 ENV GOARCHAIUS_CONFIG_PATH /etc/kubeedge/cloud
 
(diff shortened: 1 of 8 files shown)


johscheuer:
I don't know if the change here https://github.com/kubeedge/kubeedge/pull/1271/files#diff-2bdccdce28c91bcca964f9cb02007d84R51 is worth a release note?

build/cloud/Dockerfile
```diff
@@ -1,12 +1,12 @@
...
+FROM golang:1.12-alpine3.9 AS builder
 
 COPY . /go/src/github.com/kubeedge/kubeedge
 
 RUN CGO_ENABLED=0 go build -v -o /usr/local/bin/cloudcore -ldflags="-w -s" \
 github.com/kubeedge/kubeedge/cloud/cmd/cloudcore
 
 
-FROM alpine:3.9
+FROM gcr.io/distroless/static
```
johscheuer: I switched here to [distroless](https://github.com/GoogleContainerTools/distroless) which seems to be a nice fit here. My first tests where successfully. I can also move this change to another PR. I think we can also update other images to use distroless (has a smaller attack surface and requires less updates.

johscheuer:
I reverted the change (distroless).

johscheuer:
Oh I missed this one. The difference is that `1.12` will point to the latest `1.12` golang docker image e.g. `1.12.13` where as `1.12.1` references a fixed golang version (with all the CVE's). Also this means we should rebuilt all images after this merge since many CVE's will be resolved with building the images with a never golang version.

fisherxu:
> Oh I missed this one. The difference is that 1.12 will point to the latest 1.12 golang docker image e.g. 1.12.13 where as 1.12.1 references a fixed golang version (with all the CVE's). Also this means we should rebuilt all images after this merge since many CVE's will be resolved with building the images with a never golang version.

Ok, make sense!
/lgtm

PR #1272, opened 2019-11-18
Update api versions of kubernetes manifests

johscheuer:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Update api versions of kubernetes manifests and also add the labels to all resources.

Fixes #

```release-note
NONE
```

### DIFF
--- a/build/cloud/01-namespace.yaml
+++ b/build/cloud/01-namespace.yaml
@@ -3,4 +3,4 @@ kind: Namespace
 metadata:
   name: kubeedge
   labels:
-    kubeedge: cloudcore
+    k8s-app: kubeedge
--- a/build/cloud/01-namespace.yaml
+++ b/build/cloud/01-namespace.yaml
@@ -2,3 +2,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: kubeedge
+  labels:
+    kubeedge: cloudcore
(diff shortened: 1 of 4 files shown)


build/cloud/08-service.yaml.example
```diff
@@ -7,6 +7,7 @@ metadata:
     k8s-app: kubeedge
     kubeedge: cloudcore
 spec:
+  type: NodePort
```
chendave: why did this change here?

fisherxu: I think we can use `hostNetwork` in deployment

johscheuer: This was only a mistake, I did this change for my local setup.

build/cloud/01-namespace.yaml
```diff
@@ -2,3 +2,5 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: kubeedge
+  labels:
+    kubeedge: cloudcore
```
fisherxu: Why do we need this label? :) And the `kubeedge` ns will be used for all the components for kubeedge cloudside, not only for cloudcore here.

johscheuer: That's right I wanted to add the `k8s-app: kubeedge` label. Thanks for catching this one !

fisherxu:
/lgtm
Thanks!

PR #1281, opened 2019-11-22
dns query from container can not correctly return back when using edgemesh

cwl233:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

I am tring to use edgemesh by following this:
https://kubeedge.readthedocs.io/en/latest/guides/edgemesh_test_env_guide.html
I have deployed 1 client pod, but I can not use the dns service in the client container.
I try to "curl https://www.baidu.com" in the client container, but no reply. I found that edgemesh received the dns query, but it can not reply correctly. 

Fixes #

NONE

```release-note

```

### DIFF
--- a/edgemesh/pkg/server/dns.go
+++ b/edgemesh/pkg/server/dns.go
@@ -108,8 +108,8 @@ func startDnsServer() {
 		if err != nil {
 			continue
 		}
-		for _, q := range que {
-			q.from = from
+		for i, _ := range que {
+			que[i].from = from
 		}
 		rsp := make([]byte, 0)
 		rsp, err = recordHandler(que, req[0:n])


edgemesh/pkg/server/dns.go
```diff
@@ -108,8 +108,8 @@ func startDnsServer() {
 		if err != nil {
 			continue
 		}
-		for _, q := range que {
```
cwl233: q.from = from seems not executed as expected.

edgemesh/pkg/server/dns.go
```diff
@@ -108,8 +108,8 @@ func startDnsServer() {
...
-			q.from = from
```
cwl233: q.from = from seems not executed as expected.

fisherxu: Good catch, lgtm.  e.g. https://play.golang.org/p/EAMSFaPl9wI   Because the elements of `que` are not pointer. 
cc @arcanique  @kadisi

kevin-wangzefeng:
/approve
Thanks!

edgemesh/pkg/server/dns.go
kadisi: perfect， thanks

PR #1284, opened 2019-11-26
Modify the travis script to speed up the ci process

kadisi:

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind test

  Multiple stages in travis cannot run in parallel, But jobs in a stage can run in parallel
  Now travis can only run 3 jobs in parallel , So we just use 2 stages and 3 jobs in each stage
  The first stage is "ci on amd64", run in amd64 platform
  The second stage is "ci on arm64", run in arm64 platform
  And we put the longest-consuming job first

Fixes #

```release-note
Modify the travis script to speed up the ci process
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,102 +36,103 @@ install: true
 
 jobs:
   include:
+    # Multiple stages in travis cannot run in parallel
+    # But jobs in a stage can run in parallel
+    # Now travis can only run 3 jobs in parallel
+    # So we just use 2 stages and 3 jobs in each stage
+    # The first stage is "ci on amd64", run in amd64 platform
+    # The second stage is "ci on arm64", run in arm64 platform
+    # And we put the longest-consuming job first
+    - stage: "CI on amd64"
+      name: "e2e test, edge test, cloud test"
+      arch: amd64
+      script: 
+        # Now e2e_test takes the longest time, so we put it first
+        - make e2e_test 
+         # unit test edge
+        - make edge_test 
+        # unit test cloud
+        - make cloud_test 
+        
+   
+    - name: "Verify Vendor, lint, build, edge integration test, crossbuild"
+      arch: amd64
+      script: 
+        # check if the vendor is up to date
+        - make verify 
+        # cloud_lint
+        - make cloud_lint 
+        # bluetooth_device_lint
+        - make bluetoothdevice_lint 
+        # keadm_lint
+        - make keadm_lint 
+        # edge_lint
+        - make edge_lint 
+        # make cloudcore edgecore edgesite keadm
+        - make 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -36,101 +36,101 @@ install: true
...
-      script: make
+    # Multiple stages in travis cannot run in parallel
+    # But jobs in a stage can run in parallel
+    # Now travis can only run 3 jobs in parallel
+    # So we just use 2 stages and 3 jobs in each stage
+    # The first stage is "ci on amd64", run in amd64 platform
+    # The second stage is "ci on arm64", run in arm64 platform
+    # And we put the longest-consuming job first
+    - stage: "ci on amd64"
+      name: "unit test, e2e test"
```
kevin-wangzefeng:
```suggestion
      name: "e2e test"
```

.travis.yml
```diff
@@ -36,101 +36,101 @@ install: true
...
+        # cross build edge
+        - make edge_cross_build 
+        # cross build edge for armv7
+        - make edge_cross_build_v7 
+        # cross build edgesite
+        - make edgesite_cross_build 
+        # cross build edgeite for armv7
+        - make edgesite_cross_build_v7 
+
+    - name: "build docker image and test"
```
kevin-wangzefeng:
```suggestion
    - name: "build docker image, unit test and integration test" 
```

.travis.yml
```diff
@@ -36,101 +36,101 @@ install: true
...
+        - make edge_cross_build 
+        # cross build edge for armv7
+        - make edge_cross_build_v7 
+        # cross build edgesite
+        - make edgesite_cross_build 
+        # cross build edgeite for armv7
+        - make edgesite_cross_build_v7 
+
+    - name: "build docker image, unit test and integration test" 
       arch: amd64
```
fisherxu: If ut or integration test failed, we will need to rerun the build docker image here? I think we can put all the build and build image in a job, thought？

kadisi: > If ut or integration test failed, we will need to rerun the build docker image here?
yes

> I think we can put all the build and build image in a job, thought？
I did this at the beginning, but this will be turned into 4 jobs, because only 3 jobs can be executed concurrently at a time, the 4th job must wait until the first 3 jobs are completed, so the total time Becomes very long(Virtual machine creation also takes time). In addition, the build time of the image is relatively short, so this does not affect

kadisi:
cc @kevin-wangzefeng @fisherxu

.travis.yml
kadisi: done

PR #1285, opened 2019-11-26
fix wrong dockerfile of bluetooth_mapper

clumsy456:
**What type of PR is this?**
>
/kind bug

Fix the wrong dockerfile codes of bluetooth_mapper. 

Fixes #

```release-note
Fix wrong dockerfile codes of bluetooth_mapper.
```

### DIFF
--- a/build/edge/Dockerfile
+++ b/build/edge/Dockerfile
@@ -27,6 +27,4 @@ VOLUME ["/etc/kubeedge/certs", "/var/lib/edged", "/var/lib/kubeedge", "/var/run/
 COPY --from=builder /usr/local/bin/edgecore /usr/local/bin/edgecore
 COPY --from=builder /go/src/github.com/kubeedge/kubeedge/edge/conf /etc/kubeedge/edge/conf
 
-EXPOSE 1884
-
 ENTRYPOINT ["edgecore"]
--- a/build/edge/Dockerfile
+++ b/build/edge/Dockerfile
@@ -27,4 +27,6 @@ VOLUME ["/etc/kubeedge/certs", "/var/lib/edged", "/var/lib/kubeedge", "/var/run/
 COPY --from=builder /usr/local/bin/edgecore /usr/local/bin/edgecore
 COPY --from=builder /go/src/github.com/kubeedge/kubeedge/edge/conf /etc/kubeedge/edge/conf
 
+EXPOSE 1884
+
 ENTRYPOINT ["edgecore"]
--- a/mappers/bluetooth_mapper/Dockerfile
+++ b/mappers/bluetooth_mapper/Dockerfile
@@ -1,9 +1,9 @@
 FROM ubuntu:16.04
 
-CMD mkdir -p kubeedge
+RUN mkdir -p kubeedge
 
 COPY . kubeedge/
 
 WORKDIR kubeedge
 
-ENTRYPOINT ["/kubeedge/main","-logtostderr=true"]
+ENTRYPOINT ["/kubeedge/main","--logtostderr=true"]


build/edge/Dockerfile
```diff
@@ -27,4 +27,6 @@ VOLUME ["/etc/kubeedge/certs", "/var/lib/edged", "/var/lib/kubeedge", "/var/run/
 COPY --from=builder /usr/local/bin/edgecore /usr/local/bin/edgecore
 COPY --from=builder /go/src/github.com/kubeedge/kubeedge/edge/conf /etc/kubeedge/edge/conf
 
+EXPOSE 1884
```
fisherxu: If we don't expose port here, the device can't access the 1884 in container?

mappers/bluetooth_mapper/Dockerfile
```diff
@@ -1,9 +1,9 @@
...
-CMD mkdir -p kubeedge
+RUN mkdir -p kubeedge
 
 COPY . kubeedge/
 
 WORKDIR kubeedge
 
-ENTRYPOINT ["/kubeedge/main","-logtostderr=true"]
\ No newline at end of file
+ENTRYPOINT ["/kubeedge/main","--logtostderr=true"]
```
fisherxu: Do we need add compile logic here?

build/edge/Dockerfile
clumsy456: Only if we expose the port here, devices can access the mqtt server **in container and in the same container network**. If we want to access the mqtt server on the host, we still need to map the port on the host. Then [build/edge/docker-compose.yaml](https://github.com/kubeedge/kubeedge/blob/master/build/edge/docker-compose.yaml) or [build/edge/kubernetes/04-deployment-edgenode.yaml](https://github.com/kubeedge/kubeedge/blob/master/build/edge/kubernetes/04-deployment-edgenode.yaml) should be modified to map the ports.

mappers/bluetooth_mapper/Dockerfile
clumsy456: As is shown in [mappers/bluetooth_mapper/Makefile](https://github.com/kubeedge/kubeedge/blob/master/mappers/bluetooth_mapper/Makefile), the bluetooth_mapper are compiled on the host.

clumsy456: Another problem. [mappers/bluetooth_mapper/configuration/config.yaml](https://github.com/kubeedge/kubeedge/blob/master/mappers/bluetooth_mapper/configuration/config.yaml) are packed in the docker image, while it is a config file for cc2650-sensortag device. I think the config file should be able to be modified and mounted in the bluetooth_mapper container.

build/edge/Dockerfile
kadisi: if devices and mqtt server are in the same container network ，device can directly access 127.0.0.1: 1884.

kadisi: > Only if we expose the port here, devices can access the mqtt server in container and in the same container network. If we want to access the mqtt server on the host,

do you mean device and mqttserver are in same pod ?

clumsy456: > > Only if we expose the port here, devices can access the mqtt server in container and in the same container network. If we want to access the mqtt server on the host,
> 
> do you mean device and mqttserver are in same pod ?

Yes. Or if the edgecore is started by docker-compose as is shown in [build/edge/docker-compose.yaml](https://github.com/kubeedge/kubeedge/blob/master/build/edge/docker-compose.yaml), the device should be started by the network "kubeedge-bridge".

I know it looks like strange. If we want to deploy device separately, we should map the ports on the host.

Anyway, 1884 port should be exposed :)

fisherxu: > do you mean device and mqttserver are in same pod ?

Why device and mqttserver could run in same pod？mqttserver run in edgecore, device is another process, if we don't expose the 1884 port, can't we access the mqttserver through containerIP:1884?

clumsy456: > > do you mean device and mqttserver are in same pod ?
> 
> Why device and mqttserver could run in same pod？mqttserver run in edgecore, device is another process, if we don't expose the 1884 port, can't we access the mqttserver through containerIP:1884?

Sorry, my fault. I forgot the containerIP method. And I checked the offical docs of EXPOSE carefully, [EXPOSE](https://docs.docker.com/engine/reference/builder/#expose), then I found I misunderstood the meaning of EXPOSE. Sorry for wating your time, thanks!

> The EXPOSE instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published. To actually publish the port when running the container, use the -p flag on docker run to publish and map one or more ports, or the -P flag to publish all exposed ports and map them to high-order ports.

fisherxu: > And I checked the offical docs of EXPOSE carefully, EXPOSE, then I found I misunderstood the meaning of EXPOSE. Sorry for wating your time, thanks!

Np @clumsy456 If you want to run the edgecore as k8s container, suggest to use the `hostNetwork` mode.  And I'd suggest you run edgecore as process instead of container, because it's related the runtime(docker)、storage in host. :)

clumsy456: > > And I checked the offical docs of EXPOSE carefully, EXPOSE, then I found I misunderstood the meaning of EXPOSE. Sorry for wating your time, thanks!
> 
> Np @clumsy456 If you want to run the edgecore as k8s container, suggest to use the `hostNetwork` mode. And I'd suggest you run edgecore as process instead of container, because it's related the runtime(docker)、storage in host. :)

Thank you so much for your recommandation! Actually I tried deployment, docker-compose, and process.

fisherxu: OK, you can delete the 1st commit, let's got 2nd merged, thanks!

clumsy456: I've revert the first commit, and changed the title and release-node in the first comment.

PR #1286, opened 2019-11-27
add a judgement before decoding the message in edgemesh

cwl233:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Sometimes edgemesh could not decode the message received from metamanager, because the message type is json rather than []byte.

There are such decoding operations in several places, only this place has not made type judgments.

Fixes #

```release-note

```

### DIFF
--- a/edgemesh/pkg/proxy/proxy.go
+++ b/edgemesh/pkg/proxy/proxy.go
@@ -290,9 +290,16 @@ func (at *addrTable) getAddrTable(key string) *serviceTable {
 // filterResourceType implement filter. Proxy cares "Service" and "ServiceList" type
 func filterResourceType(msg model.Message) []v1.Service {
 	svcs := make([]v1.Service, 0)
-	content, ok := msg.Content.([]byte)
-	if !ok {
-		return svcs
+	var content []byte
+	var err error
+	switch msg.Content.(type) {
+	case []byte:
+		content = msg.GetContent().([]byte)
+	default:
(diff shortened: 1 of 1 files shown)


edgemesh/pkg/proxy/proxy.go
```diff
@@ -290,9 +290,16 @@ func (at *addrTable) getAddrTable(key string) *serviceTable {
...
+	var content []byte
+	var err error
+	switch msg.Content.(type) {
+	case []byte:
+		content = msg.GetContent().([]byte)
+	default:
+		content, err = json.Marshal(msg.GetContent())
+		if err != nil {
+			klog.Errorf("marshal message to edgemesh failed, err: %v", err)
+		}
```
fisherxu: Will cause any bugs without this change? I've seen this code block(L293-302) many times...  ;-)

fisherxu: Will there be different types if context that send to edgemesh？

fisherxu:
/lgtm
I think we can move L293-302 to a func, which can be called throughout the project.

edgemesh/pkg/proxy/proxy.go
fisherxu: image

fisherxu: image

PR #1291, opened 2019-11-28
Fix cannot set edgecore_image bug

### LINKED ISSUE #1290: edgecore_image cannot be set correctly by run_daemon.sh
**What happened**:
I tried to execute `./run_daemon.sh set edgecore_image="kubeedge/edgecore:v1.1.0"`, but `EDGECOREIMAGE` variable in [.env](https://github.com/kubeedge/kubeedge/blob/master/build/edge/.env) file is not changed.

The `EDGECOREIMAGE` in [.env](https://github.com/kubeedge/kubeedge/blob/master/build/edge/.env) should equals `kubeedge/edgecore:v1.1.0`.

I checked the `docker_set` function in [run_daemon.sh](https://github.com/kubeedge/kubeedge/blob/master/build/edge/run_daemon.sh) file, and I found the expected input may be `$edge_code_image_tag`. However, the following sed statement uses `${edgecore_image}` variable.

So what should be the expected input? The `edge_code_image_tag` variable which is just the tag, or the `edgecore_image` variable which is the full name including `kubeedge/edgecore:`?

**Environment**:
- KubeEdge version(e.g. `cloudcore/edgecore --version`):  v1.1.0
- bash version: GNU bash, version 4.2.46(2)-release

clumsy456:
**What type of PR is this?**
>
/kind bug

When executing `./run_daemon.sh set edgecore_image="kubeedge/edgecore:v1.1.0"`, EDGECOREIMAGE variable in [.env](https://github.com/kubeedge/kubeedge/blob/master/build/edge/.env) file is not changed. This PR fix the bug, and we don't need to change the usage shown in docs.

Fixes #1290 

Another method to fix the bug is using the `$edge_code_image_tag` variable to set the image tag only, but we should change the usage shown in docs.

```release-note
Fix cannot set edgecore_image bug
```

### DIFF
--- a/build/edge/run_daemon.sh
+++ b/build/edge/run_daemon.sh
@@ -88,7 +88,7 @@ docker_set(){
     #
     # Example
     # 
-    #  ./build.sh set \
+    #  ./run_daemon.sh set \
     #    cloudhub=0.0.0.0:10000 \
     #    edgename=edge-node \
     #    edgecore_image="kubeedge/edgecore:latest" \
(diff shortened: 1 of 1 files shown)


fisherxu:
/lgtm
Thanks!

PR #1301, opened 2019-12-03
Fix the UT failure and open the CI test on ARM64

chendave:
1 millisecond is too small to get everything settle down, which might
cause the UT fail randomly on X86 and always fail on ARM64 with LXD.

The change extends the time to 10 milliseconds which makes the UT could be
stably pass on both X86 and ARM64.

**What type of PR is this?**

/kind bug

```release-note
NONE
```

### DIFF
--- a/edge/pkg/devicetwin/devicetwin_test.go
+++ b/edge/pkg/devicetwin/devicetwin_test.go
@@ -76,6 +76,10 @@ func TestStart(t *testing.T) {
 	// fakeModule is mocked implementation of TestModule.
 	var fakeModule *beehive.MockModule
 
+	const delay = 10 * time.Millisecond
+	const maxRetries = 5
+	var retry int = 0
+
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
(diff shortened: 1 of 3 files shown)


chendave:
cc  @fisherxu  @kevin-wangzefeng

fisherxu:
Thanks! But it seems that UT on ARM can't stably passed :(  Could we merge the fix for x86 first? @chendave

kadisi:
> Thanks! But it seems that UT on ARM can't stably passed :( Could we merge the fix for x86 first? 

agree with @fisherxu , we can merge the fix for x86 first.

chendave:
test couple of times, it's looks okay now.

subpathdev:
Can we use preconditions instead of sleeping? (I didn't know if this is possible in go)

edge/pkg/devicetwin/devicetwin_test.go
```diff
@@ -94,7 +98,8 @@ func TestStart(t *testing.T) {
 	ormerMock.EXPECT().QueryTable(gomock.Any()).Return(querySeterMock).Times(1)
 	querySeterMock.EXPECT().All(gomock.Any()).Return(int64(1), nil).Times(1)
 	go dt.Start()
-	time.Sleep(1 * time.Millisecond)
+	time.Sleep(delay)
+	retry++
```
fisherxu: Need `retry++` here?

chendave: it starts from 0,  the sleep here counts, so it will consistent with other occurrence (5 times total).

chendave:
> Can we use preconditions instead of sleeping? (I didn't know if this is possible in go)

I have no idea either :(

edge/pkg/devicetwin/devicetwin_test.go
fisherxu: ok, sounds fine :)

PR #1305, opened 2019-12-04
WIP: Remove project id

fisherxu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

 /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Fixes #

```release-note
NONE
```

### DIFF
--- a/build/edge/kubernetes/03-configmap-edgenodeconf.yaml
+++ b/build/edge/kubernetes/03-configmap-edgenodeconf.yaml
@@ -17,7 +17,7 @@ data:
         session-queue-size: 100 # A size of how many sessions will be handled. default to 100.
     edgehub:
         websocket:
-            url: wss://0.0.0.0:10000/e632aba927ea4ac2b575ec1603d56f10/edgenode1/events
+            url: wss://0.0.0.0:10000/edgenode1/events
             certfile: /etc/kubeedge/certs/edge.crt
             keyfile: /etc/kubeedge/certs/edge.key
             handshake-timeout: 30 #second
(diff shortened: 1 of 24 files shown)


build/edge/kubernetes/03-configmap-edgenodeconf.yaml
```diff
@@ -17,7 +17,7 @@ data:
         session-queue-size: 100 # A size of how many sessions will be handled. default to 100.
     edgehub:
         websocket:
-            url: wss://0.0.0.0:10000/e632aba927ea4ac2b575ec1603d56f10/edgenode1/events
+            url: wss://0.0.0.0:10000/edgenode1/events
```
kadisi: It is recommended to modify `setup.md` to add instructions。

kadisi: if projectid is useless, I recommend  formatting the `url` to `0.0.0.0:10000` .    then we can automatically assemble final url by `websocket.url` and `controller.node-id` or `edged.hostname-override`. But this needs to deal with compatibility issues。

In addition, `controller.node-id` and `edged.hostname-override` have duplicate settings problem.

fisherxu: > In addition, controller.node-id and edged.hostname-override have duplicate settings problem.

Agreed, but nodeName will be used in Edgehub and Edged, so we also need to think where to place the nodeName field. And I think we can rename all the nodeid to nodeName which is more friendly to developer.

kadisi: +1

Issue #1330, opened 2019-12-19
Only the last schedule called when have multiple schedules in the bluetooth_mapper


markwangC:
**What happened**:

Only the last schedule would be called when using the bluetooth_mapper, if there were multiple schedules defined in the conf.yaml.
It is a code bug in the start function in controller pacakge. When using the goroutines on the loop iterator, the goroutines(schedule.ExecuteSchedule()) will only start until after the loop.

func (c *Config) Start() {
	c.initTopicMap()
	helper.MqttConnect(c.Mqtt.Mode, c.Mqtt.InternalServer, c.Mqtt.Server)
	subscribeAllTopics()
	helper.ControllerWg.Add(1)
	device, err := gatt.NewDevice(option.DefaultClientOptions...)
	if err != nil {
		klog.Fatalf("Failed to open device, err: %s\n", err)
		return
	}
	go c.Watcher.Initiate(device, c.Device.Name, c.Device.ID, c.ActionManager.Actions, c.Converter)

	<-watcher.DeviceConnected
	for _, action := range c.ActionManager.Actions {
		if action.PerformImmediately {
			action.PerformOperation(c.Converter.DataRead)
		}
	}

	for _, schedule := range c.Scheduler.Schedules {
		helper.ControllerWg.Add(1)
		go schedule.ExecuteSchedule(c.ActionManager.Actions, c.Converter.DataRead, c.Device.ID)
	}
	helper.ControllerWg.Wait()
}

**What you expected to happen**:

each schedule to be called not the only the last one

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

To declare a new variable inside the loop to fix it.

	for _, schedule := range c.Scheduler.Schedules {
		helper.ControllerWg.Add(1)
		newSchedule = schedule
		go newSchedule.ExecuteSchedule(c.ActionManager.Actions, c.Converter.DataRead, c.Device.ID)
	}
	helper.ControllerWg.Wait()

**Environment**:
- KubeEdge version(e.g. `cloudcore/edgecore --version`):  

**CloudSide Environment**:
- Hardware configuration (e.g. `lscpu`):
- OS (e.g. `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Go version (e.g. `go version`): 
- Others:

**EdgeSide Environment**:
- edgecore version (e.g. `edgecore --version`):
- Hardware configuration (e.g. `lscpu`):
[...]

PR #1331, opened 2019-12-19
PR:bug fix in the bluetooth_mapper controller

markwangC:
Only the last schedule would be called when using the bluetooth_mapper, if there were multiple schedules defined in the conf.yaml.
It is a code bug in the start function in controller pacakge. When using the goroutines on the loop iterator, the goroutines(schedule.ExecuteSchedule()) will only start until after the loop.

<why this change was made>
To declare a new variable inside the loop to fix it.

markWangC, 5128470@qq.com

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

#1330 
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
_If PR is about `failing-tests or flakes`, please post the related issues/tests in a comment and do not use `Fixes`_*
-->
Fixes #

```release-note

```

### DIFF
--- a/mappers/bluetooth_mapper/controller/controller.go
+++ b/mappers/bluetooth_mapper/controller/controller.go
@@ -86,7 +86,8 @@ func (c *Config) Start() {
 
 	for _, schedule := range c.Scheduler.Schedules {
 		helper.ControllerWg.Add(1)
-		go schedule.ExecuteSchedule(c.ActionManager.Actions, c.Converter.DataRead, c.Device.ID)
+		newSchedule := schedule
+		go newSchedule.ExecuteSchedule(c.ActionManager.Actions, c.Converter.DataRead, c.Device.ID)
 	}
 	helper.ControllerWg.Wait()
 }


mappers/bluetooth_mapper/controller/controller.go
```diff
@@ -86,7 +86,8 @@ func (c *Config) Start() {
 
 	for _, schedule := range c.Scheduler.Schedules {
 		helper.ControllerWg.Add(1)
-		go schedule.ExecuteSchedule(c.ActionManager.Actions, c.Converter.DataRead, c.Device.ID)
+		newSchedule := schedule
+		go newSchedule.ExecuteSchedule(c.ActionManager.Actions, c.Converter.DataRead, c.Device.ID)
```
fisherxu: Good catch, and if we don't declare the newSchedule here, will the last schedule be executed in multi goroutine?

fisherxu: @markwangC I write an example, and it seems that not the last iterator has only been executed, PTAL https://play.golang.org/p/jq8jEUyFe0d

PR #1342, opened 2019-12-24
change /usr/bin/bash to /bin/bash in the edgemesh_init image

cwl233:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

I tried to build the edgemesh_init image by myself, the container can not be run because the bash is in the directory /bin/ but not /usr/bin/.

Fixes #

```release-note

```

### DIFF
--- a/edgemesh/tools/initContainer/script/edgemesh-iptables.sh
+++ b/edgemesh/tools/initContainer/script/edgemesh-iptables.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env bash
 
 function usage() {
 	echo 'this is the edgemesh-iptables usage'


edgemesh/tools/initContainer/script/edgemesh-iptables.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/bin/bash
```
fisherxu: `#!/usr/bin/env bash`  seems better, which searches PATH for bash, and bash is not always in /bin, particularly on non-Linux systems. Thought? Xref:  https://stackoverflow.com/questions/21612980/why-is-usr-bin-env-bash-superior-to-bin-bash

cwl233: yes, "#!/usr/bin/env bash" is better

PR #1350, opened 2019-12-30
Bugfix: modify the default kube qps value for edgecontroller

kuramal:
Signed-off-by: Su Xiaolin <linxxnil@126.com>

/kind bug

Set the default burst to 200, which is twice the value of default kube QPS(100)
‘’‘The rate limiter allows bursts of up to 'burst' to exceed the QPS, while still maintaining a smoothed qps rate of 'qps'.‘’‘
When the request suddenly becomes larger than the QPS，set the burst largger than QPS，the rate limiter will allow request exceed the QPS which can reduced request latency

### DIFF
--- a/common/constants/default.go
+++ b/common/constants/default.go
@@ -11,7 +11,7 @@ const (
 	DefaultKubeContentType         = "application/vnd.kubernetes.protobuf"
 	DefaultKubeNamespace           = v1.NamespaceAll
 	DefaultKubeQPS                 = 100.0
-	DefaultKubeBurst               = 10
+	DefaultKubeBurst               = 200
 	DefaultKubeUpdateNodeFrequency = 20
 
 	DefaultUpdatePodStatusWorkers            = 1
--- a/cloud/pkg/edgecontroller/config/config.go
+++ b/cloud/pkg/edgecontroller/config/config.go
@@ -62,9 +62,9 @@ type Configure struct {
 	KubeConfig string
 	// KubeContentType is the content type communicate with edge master(default is "application/vnd.kubernetes.protobuf")
 	KubeContentType string
-	// KubeQPS is the QPS communicate with edge master(default is 1024)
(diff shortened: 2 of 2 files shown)


common/constants/default.go
```diff
@@ -10,7 +10,7 @@ import (
 const (
 	DefaultKubeContentType         = "application/vnd.kubernetes.protobuf"
 	DefaultKubeNamespace           = v1.NamespaceAll
-	DefaultKubeQPS                 = 100.0
+	DefaultKubeQPS                 = 5.0
```
kadisi: could you please explain why it was changed to 5 ?

kuramal: https://github.com/kubernetes/kubernetes/blob/1e7c07ac4ea28587551c1e766b16d1eeb834185f/staging/src/k8s.io/client-go/rest/config.go#L44

kuramal: The qps should less than the burst.
In keeping with [client-go](https://github.com/kubernetes/kubernetes/blob/1e7c07ac4ea28587551c1e766b16d1eeb834185f/staging/src/k8s.io/client-go/rest/config.go#L44).

kevin-wangzefeng:
For a bugfix, you would need to:
1. describe the bug, how to repro, and what kind of effect it causes
2. how the fix is implemented and why

common/constants/default.go
kevin-wangzefeng: EdgeController is an aggregation of shadow management for pods, nodes, configmaps etc. at edge. The default value 5 in k8s client-go is for a single client e.g. kubectl, which is quite different with EdgeController load here.

kevin-wangzefeng: Would recommand increase burst instea.
And it would be even better if you can provide test result comparison between before and after.

kuramal: @kevin-wangzefeng set the burst to 200, which is twice of qps, is than ok?

Issue #1371, opened 2020-01-07
Reliable message delivery tracking issue


fisherxu:
**What would you like to be added/modified**:

As proposal in #1343, we will implement the reliable transmission message mechanism.
This issue will track the tasks in development.

#### Alpha release (target v1.2)
- [x] Add resource version in message struct. #1368
- [x] Add CRD API which used to save the sending status. #1373 
- [x] Implement the message queue and message sending/receiving between cloud and edge. #1416
- [x] Implement the synccontroller to handle the sending failure events. #1385 

#### Beta release (TBD)
- [ ] Add reliable message delivery for device message
- [ ] Add slide window for NodeMessageQueue
- [ ] optimize NodeMessageQueue message order consistency implementation
- [ ] Consider readd the message to the queue when sending failed after 5 times retries

Abhishek9639:
Hey @fisherxu,
I'd like to work on the remaining Beta release tasks for this issue. Can this be assigned to me?

Here's my quick plan for the pending items:

**1. Reliable message delivery for device message**
- Extend the existing reliable sync mechanism (from #1416) to cover device messages
- Ensure cloud ↔ edge delivery guarantees for device-bound messages

**2. Slide window for NodeMessageQueue**
- Implement a sliding window algorithm to control in-flight message count
- Prevent queue overflow while maintaining throughput

**3. Optimize NodeMessageQueue message order consistency**
- Add sequence tracking to ensure messages are processed in-order
- Handle out-of-order arrival gracefully without dropping messages

**4. Re-enqueue on 5 consecutive send failures**
- Add a retry counter per message
- After 5 failed retries, re-add the message to the tail of the queue with a backoff

I'll first go through #1343, #1373, #1416 and #1385 to fully understand the existing reliable sync design before touching anything. 
Happy to share a detailed design doc before implementation if needed.
Thanks

PR #1385, opened 2020-01-13
Add synccontroller for reliable message delivery

fisherxu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

Add the synccontroller to resend the failed persisted objects to edgenode.
1. Compare the  objects that have been persisted to the edge with the objects in K8s  and generate update and delete events to the edge.
2. Compare the objects in K8s with the objects that have been persisted to the edge,
   If the objects fail to persisted to the edge for the first time, it will be recreated here.
   Just focus on the pod, service, endpoint, because if the pod is passed down, and the configmap and secret cannot be found and which will be queried to the cloud.

Ref: #1371 

```release-note
Add synccontroller for reliable message delivery
```

### DIFF
--- a/cloud/pkg/synccontroller/createfailedobject.go
+++ b/cloud/pkg/synccontroller/createfailedobject.go
@@ -39,7 +39,7 @@ func (sctl *SyncController) manageCreateFailedCoreObject() {
 			continue
 		}
 		// Check whether the pod is successfully persisted to edge
-		_, err := sctl.objectSyncLister.ObjectSyncs(pod.Namespace).Get(buildObjectSyncName(pod.Spec.NodeName, string(pod.UID)))
+		_, err := sctl.objectSyncLister.ObjectSyncs(pod.Namespace).Get(BuildObjectSyncName(pod.Spec.NodeName, string(pod.UID)))
 		if err != nil && apierrors.IsNotFound(err) {
 			msg := buildEdgeControllerMessage(pod.Spec.NodeName, pod.Namespace, model.ResourceTypePod, pod.Name, model.InsertOperation, pod)
 			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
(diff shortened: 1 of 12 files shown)


fisherxu:
It's ready to review now :) cc @kevin-wangzefeng @kadisi

cloud/pkg/synccontroller/createfailedobject.go
```diff
@@ -0,0 +1,97 @@
...
+
+	set := labels.Set{edgemgr.NodeRoleKey: edgemgr.NodeRoleValue}
+	selector := labels.SelectorFromSet(set)
+	allEdgeNodes, err := sctl.nodeLister.List(selector)
+	if err != nil {
+		klog.Errorf("Filed to list all the edge nodes: %v", err)
+	}
+
+	for _, pod := range allPods {
+		if isFromEdgeNode(allEdgeNodes, pod.Spec.NodeName) {
```
kadisi:
```
if !isFromEdgeNode(allEdgeNodes, pod.Spec.NodeName) {
  continue
}
```
?

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+// Group of controller
+func (sctl *SyncController) Group() string {
+	return SyncControllerModuleGroup
+}
+
+// Group of controller
+func (sctl *SyncController) Enable() bool {
+	return true
```
kadisi: add `Enable` field for `SyncController  struct` and then use sctl.Enable.

cloud/pkg/synccontroller/createfailedobject.go
fisherxu: Done.

cloud/pkg/synccontroller/synccontroller.go
fisherxu: Done.

PR #1183, opened 2019-10-09
WIP: fix travis issue with ubuntu16.04

chendave:
**What type of PR is this?**

/kind bug
/kind failing-test

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

```release-note
NONE
```

PR #1396, opened 2020-01-14
Fix edgecore cpu usage issue

### LINKED ISSUE #1205: edgecore cpu usage rate will increase by container's num
**What happened**:
edgecore cpu usage rate will increase by container's num。
cpu usage rate over 10% when 6 pods run 
edgecore's cpu usage rate below than kubelet 
create some pod and check node's cpu usage rate

**Environment**:

**CloudSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

**EdgeSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

fisherxu:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

**Before the fix:** 

- when 40 pods running in edge node, the cpu usage of edgecore will reach 40%.

**With this fix:** 

- when 100 pods running in edge node, the cpu usage of edgecore is between 3%~5%.

Fixes #1205

```release-note
Fix edgecore cpu usage issue
```


.travis.yml
```diff
@@ -5,7 +5,7 @@ git:
 
 sudo: required
 
-dist: trusty
```
fisherxu: image
When using ubuntu 14.04(trusty), the matching docker version is 17.9. This error is reported when starting the pod/container, but the status of the Pod is Running.
So the previous e2e runs have problems, now upgrade to 16.04, the pod is running normally.

.travis.yml
```diff
@@ -78,6 +79,7 @@ jobs:
     - stage: "Test on arm64"
       name: "build, smallbuild"
       arch: arm64
+      dist: trusty
```
fisherxu: ditto

.travis.yml
```diff
@@ -37,6 +37,7 @@ jobs:
     - stage: "Test on amd64"
       name: "build, smallbuild, crossbuild"
       arch: amd64
+      dist: trusty
```
fisherxu: There's some crossbuild error with 16.04, so using 14.04 for workaround here.
/cc @kevin-wangzefeng @chendave @kadisi

kevin-wangzefeng: I'm fine to get the fix in with this restriction, please create a follow-up issue to track this.

fisherxu: https://github.com/kubeedge/kubeedge/pull/1183  There is already a WIP PR to solve this issue, but it‘s a little complicated that relate to the vm image of travis.

kevin-wangzefeng: OK

edge/pkg/edged/pleg/generic.go
```diff
@@ -21,484 +21,29 @@ This file is derived from K8S Kubelet code with reduced set of methods
...
-	"k8s.io/kubernetes/pkg/kubelet/prober"
-	"k8s.io/kubernetes/pkg/kubelet/status"
-	kubetypes "k8s.io/kubernetes/pkg/kubelet/types"
-
-	"github.com/kubeedge/kubeedge/edge/pkg/edged/podmanager"
 )
 
 //GenericLifecycle is object for pleg lifecycle
 type GenericLifecycle struct {
 	pleg.GenericPLEG
```
kadisi: delete `GenericLifecycle ` struct , and use `pleg.GenericPLEG` directly?

fisherxu: Good idea, have done.

PR #1401, opened 2020-01-15
edge node auto register

### LINKED ISSUE #881: Create and register node from edge part

kuramal:
**What type of PR is this?**

/kind documentation
/kind feature

support edge node auto register

Fixes #881

```release-note
Edge node auto register
Now the edge node will automatically register by default,  If you want register manually, you should set the config register-node of edged with false
```


docs/setup/setup.md
```diff
@@ -160,44 +160,10 @@ The cert/key will be generated in the `/etc/kubeedge/ca` and `/etc/kubeedge/cert
...
-      }
-    }
-    ```
-    **Note: you need to remember `metadata.name` , because edgecore need it**.
-+ If role is not set for the node, the pods, configmaps and secrets created/updated in the cloud cannot be synced with the node they are targeted for.
-+ Deploy edge node (**you must run on cloud side**)
-
-    ```shell
-    kubectl apply -f ~/cmd/yaml/node.json
-    ```
```
fisherxu: We can make manually creating node as an option, and add a flag `register-node` to edgecore/edged. @kuramal @kadisi

kuramal: Is there any situation where the user must register manually？ @fisherxu

fisherxu: Yes, in some scenarios users need to manually create a node, they need to set some configuration in node object. Xrf: kubelet also has a flag to set it. https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/options/options.go#L382

kadisi: sounds good

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -349,7 +356,44 @@ func (uc *UpstreamController) updateNodeStatus() {
 			}
 
 			switch msg.GetOperation() {
+			case model.InsertOperation:
+				_, err := uc.kubeClient.CoreV1().Nodes().Get(name, metaV1.GetOptions{})
+				if err == nil {
+					klog.Infof("node: %s already existed, do nothing", name)
```
kevin-wangzefeng:
```suggestion
					klog.Infof("node: %s already exists, do nothing", name)
```

kevin-wangzefeng:
Please add release-note description

edge/conf/edge.yaml
```diff
@@ -33,6 +33,7 @@ edged:
     hostname-override: edge-node
     interface-name: eth0
     edged-memory-capacity-bytes: 7852396000
+    register-node: true
```
kadisi: now we are doing kubeedge component new api config feature, i suggest to add `register-node` field to new api struct.
https://github.com/kubeedge/kubeedge/blob/b61e49266715e493bbb5b2dc7593cc8c8289b8cc/pkg/apis/edgecore/v1alpha1/types.go#L103

kadisi: if  this `register-node` field is not configured,  is it automatically registered by default? I suggest to refer to kubelet。

fisherxu: > i suggest to add register-node field to new api struct.

Agreed.

> if this register-node field is not configured, is it automatically registered by default? I suggest to refer to kubelet

Let's set the `register-node` default as `true`.

cloud/pkg/edgecontroller/controller/upstream.go
kuramal: @kevin-wangzefeng , I didn't deal with this suggested change in time before a force-push, so    I had to change it manually.

docs/setup/setup.md
kuramal: @fisherxu ,  I changed the node auto register to optional, PTAL

edge/conf/edge.yaml
kuramal: done, PTAL

kuramal:
@fisherxu @kadisi update with new config api, PTAL

edgesite/conf/edgeSite.yaml
```diff
@@ -25,6 +25,7 @@ edged:
     register-node-namespace: default
     hostname-override: edge-node
     interface-name: eth0
+    register-node: false
```
fisherxu: Does this change still needed?

edge/conf/edge.yaml
fisherxu: Does this change still needed?

docs/setup/deploy-edge-node.md
```diff
@@ -0,0 +1,43 @@
+# Deploy Edge Node Manually
```
fisherxu: How to jump here from setup.md? I'd suggest to keep it in setup.md. For example:

#### Deploy the edge node
Edge node can be registered automatically.
##### Deploy Edge Node Manually
If you want to deploy edge node manually, here is an example.

fisherxu: please help modify it more clear，like: L5-L16, we can just delete it.

edgesite/conf/edgeSite.yaml
fisherxu: we can delete it in another PR, omit this comment.

edge/conf/edge.yaml
fisherxu: omit this comment.

PR #1404, opened 2020-01-16
For the --minconfig subcommand of edgecore and edgesite, hide the database config:  DriverName and AliasName

kuramal:
**What type of PR is this?**
>

/kind cleanup

For the --minconfig subcommand of edgecore and edgesite, hide the database config:  DriverName and AliasName, which are not required for min config

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/apis/edgesite/v1alpha1/default.go
+++ b/pkg/apis/edgesite/v1alpha1/default.go
@@ -141,8 +141,6 @@ func NewMinEdgeSiteConfig() *EdgeSiteConfig {
 			APIVersion: path.Join(GroupName, APIVersion),
 		},
 		DataBase: &edgecoreconfig.DataBase{
-			DriverName: DataBaseDriverName,
-			AliasName:  DataBaseAliasName,
 			DataSource: DataBaseDataSource,
 		},
 		KubeAPIConfig: &cloudcoreconfig.KubeAPIConfig{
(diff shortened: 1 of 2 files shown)


pkg/apis/edgesite/v1alpha1/default.go
```diff
@@ -141,8 +141,6 @@ func NewMinEdgeSiteConfig() *EdgeSiteConfig {
 			APIVersion: path.Join(GroupName, APIVersion),
 		},
 		DataBase: &edgecoreconfig.DataBase{
-			DriverName: DataBaseDriverName,
```
kadisi: We need to do the same for edgecore

pkg/apis/edgesite/v1alpha1/default.go
```diff
@@ -157,6 +155,7 @@ func NewMinEdgeSiteConfig() *EdgeSiteConfig {
 				ClusterDNS:            "",
 				ClusterDomain:         "",
 				RemoteRuntimeEndpoint: constants.DefaultRemoteRuntimeEndpoint,
+				//TODO (@kuramal) auto set the value of PodSandboxImage by the architecture(x86,amd64,arm or arm64)
```
fisherxu:
```suggestion
				//TODO (@kuramal) Automatically set PodSandboxImage according to the architecture.(x86,amd64,arm or arm64)
```
And please move this line above L160

pkg/apis/edgesite/v1alpha1/default.go
```diff
@@ -157,6 +155,7 @@ func NewMinEdgeSiteConfig() *EdgeSiteConfig {
 				ClusterDNS:            "",
 				ClusterDomain:         "",
 				RemoteRuntimeEndpoint: constants.DefaultRemoteRuntimeEndpoint,
+				//TODO (@kuramal) Automatically set PodSandboxImage according to the architecture.(x86,amd64,arm or arm64)
```
fisherxu: move this line above L160

PR #1412, opened 2020-01-17
Create socket address directory if not exist

chendave:
Hit below issue if the directory does not exist
`failed to listen addr: listen unix //var/lib/kubeedge/kubeedge.sock: bind:
no such file or directory`

**What type of PR is this?**
/kind bug

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/servers/udsserver/uds.go
+++ b/cloud/pkg/cloudhub/servers/udsserver/uds.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"k8s.io/klog"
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudhub/servers/udsserver/uds.go
```diff
@@ -62,6 +63,16 @@ func (us *UnixDomainSocket) StartServer() error {
 		}
 	}
 
+	dir, err := filepath.Abs(filepath.Dir(addr))
+	if err != nil {
+		klog.Errorf("invalid dir of addr found: %v", err)
+	}
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
```
kadisi: ref: https://github.com/kubeedge/kubeedge/blob/6df871802fd63f2a0f0ce4b3ae8e2c015206237d/pkg/util/validation/check.go#L6

fisherxu: Does `FileIsExist` fix this issue `failed to listen addr: listen unix //var/lib/kubeedge/kubeedge.sock: bind: no such file or directory`? @kadisi

kadisi: no, I suggest using a FileIsExist() method to determine whether  file or path is exists.

kevin-wangzefeng: If use `kubeedge/pkg/util/validation/check`, should move the dir creation there as well.
Minor issue, either looks fine to me.

PR #1418, opened 2020-01-20
delete ServiceBus.Enable's tag omitempty

kuramal:
Signed-off-by: kuramal <linxxnil@126.com>

**What type of PR is this?**
>
> /kind api-change
 /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Set the describe of ServiceBus.Enable to false.,because default config is set to false at the init func NewDefaultEdgeCoreConfig()
Delete the ServiceBus.Enable's omitempty attribute, it shoud be displayed when not config

Fixes #

```release-note

```

### DIFF
--- a/pkg/apis/edgecore/v1alpha1/types.go
+++ b/pkg/apis/edgecore/v1alpha1/types.go
@@ -290,8 +290,8 @@ type MetaManager struct {
 // ServiceBus indicates the servicebus module config
 type ServiceBus struct {
 	// Enable indicates whether servicebus is enabled, if set to false (for debugging etc.), skip checking other servicebus configs.
-	// default true
-	Enable bool `json:"enable,omitempty"`
+	// default false
+	Enable bool `json:"enable"`
 }
 
 // DeviceTwin indicates the servicebus module config


kadisi:
why need to delete `omitempty ` json tag?

kuramal:
> why need to delete `omitempty ` json tag?

If tag omitempty is set, when show edgecore's default config with command 'edgecore --defaultconfig', the display of servicebus is as follows:
```
  servicebus: {}
```
the display of servicebus without servicebus.enable's tag omitempty:
```
  servicebus:
    enable: false
```

kevin-wangzefeng:
Having omitempty or not would simply affect the boolean fields be shown or not when set default to false.
The essential question is whether show to user that servicebus is disabled by default.
We can actually turn the default value to true after cloud functionality of servicebus complete

kevin-wangzefeng:
> Having omitempty or not would simply affect the boolean fields be shown or not when set default to false.
> The essential question is whether show to user that servicebus is disabled by default.
> We can actually turn the default value to true after cloud functionality of servicebus complete

Pobably make serviceBus explicitly shown false by default, and later on when turning it default to true, we can make it not explicitly shown.

kevin-wangzefeng:
And DBTest is in the same situation, please remove the "omitempty" as well.

PR #1433, opened 2020-02-04
add gateway proposal

### LINKED ISSUE #1432: Add EdgeGateway as the ingress gateway on Edge
Add a EdgeGateway as the ingress gateway on Edge to dispatch request to multiple backend pods on the Edge side.

**Why is this needed**:
Gateway or edge service in a microservice system provide entry point for the entire application so that external applications can visit this application.

In the native kubernetes, [ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) is provided to achieve the purpose. [Gateway](https://istio.io/docs/reference/config/networking/gateway/) is another object designed in istio as another solution. They provide different capabilities with different designs for the cloud-grade gateway/edge service solution.

However, the implementation on Edge is different. We need an implementation on edge to support the Gateway capabilities.

qizha:
**What type of PR is this?**
>
> /kind design

Gateway or edge service in a microservice system provide entry point for the entire application so that external applications can visit this application.
In the native kubernetes, ingress is provided to achieve the purpose. Gateway is another object designed in istio as another solution. They provide different capabilities with different designs for the cloud-grade gateway/edge service solution.
However, the implementation on Edge is different. We need an implementation on edge to support the Gateway capabilities.

Fixes #1432


docs/proposals/gateway.md
```diff
@@ -0,0 +1,234 @@
...
+        exact: /login
+    - uri:
+        exact: /logout
+    - uri:
+        prefix: /api/v1/products
+    route:
+    - destination:
+        host: productpage
+        port:
+          number: 9080
```
fisherxu: If the pods of a service runs in different available areas, can this rule restrict requests to the corresponding zones and it seems that `productpage` is a service here.

docs/proposals/gateway.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+
+
+3. EdgeController get the definition of Gateway and VirtualService by list/watching the K8S master.
```
kevin-wangzefeng: nit: VirtualService is CRD installed by istio. So this would need to be optional in cases istio may not be installed - better implemented as a plugin.

anyushun:
> When we deploy the gateway to the edgenode, we need to expose the edgenode to public, then access to all the pods in same az? :)

no need expose the edgenode to public, as long as it can be accessed by external services

docs/proposals/gateway.md
anyushun: `productpage` is short name, instead of `productpage.default.svc.cluster.local`

fisherxu: the destination of ingress controller is also service, so what's the difference with ingress rule?

qizha: check [VirtualService](https://istio.io/docs/reference/config/networking/virtual-service/), they provide lots of communication governance capabilities, such as fault injection, canary release and so on. With those capabilities, the VirtualService won't be the definition of the request forward only anymore. We are planning to add those communication governance also in the future.

docs/proposals/gateway.md
```diff
@@ -0,0 +1,234 @@
...
+4. Sending the definition to edge and stored in metaManager in EdgeCore.
+5. EdgeGateway get the definition from EventBus via MQTT and apply them without rebooting
```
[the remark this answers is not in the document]
qizha: You are right. I think we can take advantage of the eventbus as a communication channel between EdgeCore and processes running outside of the the EdgeCore

docs/proposals/gateway.md
qizha: right, we should add this feature as a new module in EC

docs/proposals/gateway.md
fisherxu: Sounds good!

kevin-wangzefeng:
No blokcing issue for now, but it would be better if include design details, development plan (maturity-release alignments), etc. We can get this accepted after v1.2 is out.

PR #1473, opened 2020-02-17
Fix the gating issue with ubuntu distro

chendave:
- upgrade ubuntu distro from `xenial` to `cosmic` and make
it a default config

- fix the cross-building issue by adding the mandatory packages.

- make `cross_build` target build arm64 instead of arm32 since
arm64 is used more widely.

**What type of PR is this?**
>
/kind bug

```release-note
`make cross_build` now target on building arm64 instead of arm32, users can use `make edge_cross_build_v7` to build arm32 binaries.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ git:
 
 sudo: required
 
-dist: xenial
+dist: cosmic
 
 services:
   - docker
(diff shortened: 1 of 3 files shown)


chendave:
cc @fisherxu  @kevin-wangzefeng

kevin-wangzefeng:
Revised PR description, added release-note: 
```
`make cross_build` now target on building arm64 instead of arm32, users can use `make edge_cross_build_v7` to build arm32 binaries.
```

PR #1494, opened 2020-02-24
remove useless validation in cloudcore

zhuguihua:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind cleanup

It is not necessary to check the existence of unixSocketAddress,
it will be created in cloud/pkg/cloudhub/servers/udsserver/uds.go#L71.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/apis/cloudcore/v1alpha1/validation/validation.go
+++ b/pkg/apis/cloudcore/v1alpha1/validation/validation.go
@@ -18,6 +18,7 @@ package validation
 
 import (
 	"fmt"
+	"os"
 	"path"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pkg/apis/cloudcore/v1alpha1/validation/validation.go
```diff
@@ -83,12 +82,6 @@ func ValidateModuleCloudHub(c cloudconfig.CloudHub) field.ErrorList {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("address"),
 			c.UnixSocket.Address, "unixSocketAddress must has prefix unix://"))
 	}
-	s := strings.SplitN(c.UnixSocket.Address, "://", 2)
-	if len(s) > 1 && !utilvalidation.FileIsExist(path.Dir(s[1])) {
```
fisherxu: Thanks for the fix!
Could we create the `path.Dir(s[1])` if it doesn't exist? Ref: https://github.com/kubeedge/kubeedge/blob/master/pkg/apis/edgecore/v1alpha1/validation/validation.go#L50 
@zhuguihua

fisherxu:
> it will be created in cloud/pkg/cloudhub/servers/udsserver/uds.go#L71.

We can move this block to validation?

fisherxu:
/lgtm
It will be  better delete L22 and L27-29 in [execute.sh](https://github.com/kubeedge/kubeedge/blob/master/tests/e2e/scripts/execute.sh) for e2e test :)

zhuguihua:
> /lgtm
> It will be better to delete L22 and L27-29 in [execute.sh](https://github.com/kubeedge/kubeedge/blob/master/tests/e2e/scripts/execute.sh) for e2e test :)

Thanks, updated.

PR #1500, opened 2020-02-26
Fix local up script for new makefile

fisherxu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation

/kind test
> /kind failing-test
> /kind feature

TODO: Use local-up script to start cluster in e2e.

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ ifeq ($(HELP),y)
 all: clean
 	@echo "$$ALL_HELP_INFO"
 else
-all: verify-golang clean
+all: verify-golang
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -31,7 +31,7 @@ ifeq ($(HELP),y)
 all: clean
 	@echo "$$ALL_HELP_INFO"
 else
-all: verify-golang clean
```
fisherxu: It's no need to clean here, if I make cloudcore first, then I make edgecore, cloudcore will be deleted. cc @kadisi

PR #1538, opened 2020-03-10
clarifies location of keadm compiled binary

mimizone:
**What type of PR is this?**
/kind documentation

clarifies the location of keadm compiled binary which was not correct.

### DIFF
--- a/docs/setup/kubeedge_install_keadm.md
+++ b/docs/setup/kubeedge_install_keadm.md
@@ -42,7 +42,7 @@ There are currently two ways to get keadm
 
   2.  If you used `go get`, the `keadm` binary is available in `$GOPATH/bin/`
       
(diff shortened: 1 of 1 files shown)


docs/setup/kubeedge_install_keadm.md
```diff
@@ -40,7 +40,9 @@ There are currently two ways to get keadm
       go get github.com/kubeedge/kubeedge/keadm/cmd/keadm
       ```
 
-  2. Binary `keadm` is available in current path. If you are using `go` get the binary is available in `$GOPATH/bin/`
+  2.  If you used `go get`, the `keadm` binary is available in `$GOPATH/bin/`
+      
+      If you compiled from source, the `keadm` binary is in a subfolder `${KUBEEDGE_OUTPUT_SUBPATH}/bin` which defaults to `_output/local/bin` in your kubeedge project directory, or `$GOPATH/src/github.com/kubeedge/kubeedge/_output/local/bin/`
```
fisherxu:
```suggestion
      If you compiled from source, the `keadm` binary is in `$GOPATH/src/github.com/kubeedge/kubeedge/_output/local/bin/`
```

PR #1600, opened 2020-04-20
Implement HA in hot standby mode

### LINKED ISSUE #1444: HA for CloudCore(Make CloudCore stateless)
HA features for CloudCore(Hereinafter referred to as CC)
**Why is this needed**:
In [2019 Q4 Roadmap](https://github.com/kubeedge/kubeedge/blob/master/docs/getting-started/roadmap.md#2019-q4-roadmap),it says that we would support HA for CC，but not so far.Now I want to open a discussion.

**Make CC Stateless**:
Kubeedge is design to serve hundreds of edgenodes,and it is obviously single CC could hardly run well all the time ，especially when it comes to kubeedge system upgrades.When a CC is down, all edgenodes it connects to will not fetch new data any more until CC restarts.Copy is a powerful weapon against the uncertainty.If we add a LoadBalance(LB) between edgecore and cloudcore, once the connection breaks, edgecore can connect to the cloudcore that run well. The problem is whether the new cloudcore can serve the edgecore as well as the old one, whether there any loss of information during reconnection. The latter could be resolved by the [Reliable message delivery](https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/reliable-message-delivery.md).In the proposal, SyncController will periodically compare the saved objects resourceVersion with the objects in K8s, and then trigger the events such as retry and deletion to ensure data consistency.	

But the new cloudcore can not serve the edgecore as well as the old one because the cloudcore is stateful.For example, when it comes to the delivery of configmap, CC need to maintain a LocationCache to record which node the pod (using the configmap) bind to.As shown in the following figure:
图片

If the LocationCache is not synchronized between the old and new CC, Although SyncController trigger the events delivering the configmap, delivery will fail because of unknown of destination.Now the LocationCache is stored in memory and will lost during crash.And I do not know if there is any other data like this. Maybe many.

There are 3 possible ways, and copies are required.

- The data in edgestore is read from the edge whenever connection is established.

    In view of frequent connection between cloud and edge, it is not a good idea.

- Stateful data is still stored in memory, using **raft** or other consistency protocol to Synchronize data between copies and elect a leader to provide services.[#1192](https://github.com/kubeedge/kubeedge/issues/1192)
- **Make CloudCore stateless**.Store more stateful data in etcd as CRD like [Reliable message delivery](https://github.com/fisherxu/kubeedge/blob/reliab-proposal/docs/proposals/reliable-message-delivery.md) do. 

I prefer the third way.I want to hear from your opinions before start to figer out the stateful data in CC and write the reconnection code between different cloudcores.The ideal structure is shown below:
图片

### LINKED ISSUE #1560: HA support of CloudCore

CloudCore is not deployed as a single instance. Ideally it should be HA. So we may do it in two steps:
1. Hot standby
2. Deploy a cluster of CloudCore 

To be implemented for alpha in v1.3

- [x] Add leader election in CloudCore
- [ ] Add healthcheck method to CloudCore for integration with LB, to tell which instance is currently active
- [ ] Add LB for multiple instances

To be implemented for beta
- [ ] Make CloudCore Stateless, ref: https://github.com/kubeedge/kubeedge/issues/1444

**Why is this needed**:

GsssC:
**What type of PR is this?**
Implement leader elections and health checks ，this pr is for peer review but not the final pr.
>
> /kind api-change
> /kind bug
> /kind cleanup
> kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

Fix #1560
Fix #1444
```release-note

```


cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,181 @@
+package leaderelection
+import (
```
fisherxu, line 2: Add empty line above

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	// Start leaderElection until becoming leader, terminate program if leader lost or context.cancel
+	go leaderElector.Run(beehiveContext.GetContext())
+
+	// Monitor system signal and shutdown gracefully and it should be in main gorutine
+	core.GracefulShutdown()
+	return
+}
+// makeLeaderElectionConfig builds a leader election configuration. It will
```
fisherxu, line 85: empty line

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,181 @@
...
+	return &leaderelection.LeaderElectionConfig{
+		Lock:          rl,
+		LeaseDuration: config.LeaseDuration.Duration,
+		RenewDeadline: config.RenewDeadline.Duration,
+		RetryPeriod:   config.RetryPeriod.Duration,
+		WatchDog:      nil,
+		Name:          "cloudcore",
+	}, nil
+}
+// Try to patch PodReadinessGate if program runs in pod
```
fisherxu, line 117: ditto

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -8,6 +8,7 @@ import (
 	"k8s.io/klog"
 
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
+
```
fisherxu, line 11: rm the line here

pkg/apis/componentconfig/cloudcore/v1alpha1/default.go
```diff
@@ -124,7 +126,17 @@ func NewDefaultCloudCoreConfig() *CloudCoreConfig {
 				Enable: true,
 			},
 		},
+		LeaderElection:	&componentbaseconfig.LeaderElectionConfiguration{
+			LeaderElect:		true,
```
fisherxu, line 130: For the first version of ha, we ca set the `LeaderElect` to false. Which means ha is disable default.

GsssC: get it

pkg/apis/componentconfig/cloudcore/v1alpha1/types.go
```diff
@@ -299,7 +302,7 @@ type DeviceControllerLoad struct {
 
 // SyncController indicates the sync controller
 type SyncController struct {
-	// Enable indicates whether devicecontroller is enabled, if set to false (for debugging etc.), skip checking other devicecontroller configs.
+	// Enable indicates whether SyncController is enabled, if set to false (for debugging etc.), skip checking other devicecontroller configs.
```
kevin-wangzefeng, line 305:
```suggestion
```

pkg/apis/componentconfig/cloudcore/v1alpha1/types.go
```diff
@@ -299,7 +302,6 @@ type DeviceControllerLoad struct {
 
 // SyncController indicates the sync controller
 type SyncController struct {
-	// Enable indicates whether devicecontroller is enabled, if set to false (for debugging etc.), skip checking other devicecontroller configs.
```
kevin-wangzefeng, line 302:
```suggestion
	// Enable indicates whether devicecontroller is enabled, if set to false (for debugging etc.), skip checking other devicecontroller configs.
```

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,208 @@
...
+		podutil.UpdatePodCondition(&getPod.Status, &condition)
+		newJSON, err := json.Marshal(getPod)
+		patchBytes, err := strategicpatch.CreateTwoWayMergePatch(originalJSON, newJSON, corev1.Pod{})
+		if err != nil {
+			return fmt.Errorf("failed to create two way merge patch: %v", err)
+		}
+
+		//Try to patch
+		var isPatchSuccess = false
+		for i := 1; i <= 5; i++ {
```
kevin-wangzefeng, line 155: To avoid magic number, you can define a const e.g. `maxPodPatchRetry` in the beginning of the file.

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -0,0 +1,208 @@
...
+			if _, err = cli.CoreV1().Pods(namespace).Patch(podname, types.StrategicMergePatchType, patchBytes, "status"); err != nil {
```
kevin-wangzefeng, line 156: Should retry only when the error is resource update conflict.
This is not a blocking issue, you may fix with a follow up PR.

pkg/apis/componentconfig/cloudcore/v1alpha1/default.go
```diff
@@ -155,5 +167,8 @@ func NewMinCloudCoreConfig() *CloudCoreConfig {
 				},
 			},
 		},
+		LeaderElection:	&componentbaseconfig.LeaderElectionConfiguration{
+			LeaderElect:		true,
```
kevin-wangzefeng, line 171:
```suggestion
			LeaderElect:		false,
```

kevin-wangzefeng:
@GsssC Please address the review commnets in a followup PR.
/approve

PR #1628, opened 2020-04-30
keadm: fix get version issue

### LINKED ISSUE #1629: Failed to get the latest KubeEdge release version
When do not specify kubeedge version, keadm can not get latest version:
```
# ./_output/local/bin/keadm init
Failed to get the latest KubeEdge release version, will use default version 
...
```

daixiang0:

Fixes #1629

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cloud/init.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/init.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -117,7 +118,7 @@ func Add2ToolsList(toolList map[string]types.ToolsInstaller, flagData map[string
 				return err
 			}
 			if len(version) > 0 {
-				kubeVer = version[1:]
+				kubeVer = strings.TrimPrefix(version, "v")
 				latestVersion = version
 				break
 			}
--- a/keadm/cmd/keadm/app/cmd/cloud/init.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/init.go
@@ -116,7 +116,7 @@ func Add2ToolsList(toolList map[string]types.ToolsInstaller, flagData map[string
 			if err != nil {
 				return err
 			}
-			if len(version) != 0 {
+			if len(version) > 0 {
 				kubeVer = version[1:]
 				latestVersion = version
 				break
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/cloud/init.go
```diff
@@ -112,12 +112,13 @@ func Add2ToolsList(toolList map[string]types.ToolsInstaller, flagData map[string
 	if kubeVer == "" {
 		var latestVersion string
 		for i := 0; i < util.RetryTimes; i++ {
```
kevin-wangzefeng, line 114: The for loop is useless here, since any error will make return the function.

keadm/cmd/keadm/app/cmd/cloud/init.go
```diff
@@ -112,12 +112,13 @@ func Add2ToolsList(toolList map[string]types.ToolsInstaller, flagData map[string
...
-			latestVersion, err := util.GetLatestVersion()
+			version, err := util.GetLatestVersion()
 			if err != nil {
 				return err
 			}
-			if len(latestVersion) != 0 {
```
RainbowMango, line 119: Personaly, I like `len(latestVersion) > 0` better than `len(latestVersion) != 0`

keadm/cmd/keadm/app/cmd/cloud/init.go
```diff
@@ -112,12 +112,13 @@ func Add2ToolsList(toolList map[string]types.ToolsInstaller, flagData map[string
 	if kubeVer == "" {
 		var latestVersion string
 		for i := 0; i < util.RetryTimes; i++ {
-			latestVersion, err := util.GetLatestVersion()
```
RainbowMango, line 115: nice catch!

keadm/cmd/keadm/app/cmd/cloud/init.go
daixiang0: I think this for loop is for retry setting, any error make it return is what we want.

keadm/cmd/keadm/app/cmd/cloud/init.go
```diff
@@ -112,12 +112,13 @@ func Add2ToolsList(toolList map[string]types.ToolsInstaller, flagData map[string
...
+			version, err := util.GetLatestVersion()
 			if err != nil {
 				return err
 			}
-			if len(latestVersion) != 0 {
-				kubeVer = latestVersion[1:]
+			if len(version) > 0 {
+				kubeVer = version[1:]
```
kevin-wangzefeng, line 120: use [strings.TrimPrefix()](https://golang.org/pkg/strings/#TrimPrefix) here to improve readability?

```suggestion
				strings.TrimPrefix(kubeVer, "v")
```

daixiang0: It would be better!

PR #1650, opened 2020-05-09
Improve retransmission logic in leaderelection

Poor12:
**What type of PR is this?**
> /kind bug

In the leaderelection of Cloudcore, when calling the client to patch podReadinessGate, add the logic if patching failed due to conflict then retry, otherwise just return failure.

Which issue(s) this PR fixes:

Fixes #

### DIFF
--- a/cloud/pkg/leaderelection/leaderelection.go
+++ b/cloud/pkg/leaderelection/leaderelection.go
@@ -7,7 +7,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"os"
 	"syscall"
-	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -150,20 +149,26 @@ func TryToPatchPodReadinessGate() error {
 			return fmt.Errorf("failed to create two way merge patch: %v", err)
 		}
 
-		//Try to patch
+		var maxRetries = 3
 		var isPatchSuccess = false
-		for i := 1; i <= 5; i++ {
-			if _, err = cli.CoreV1().Pods(namespace).Patch(podname, types.StrategicMergePatchType, patchBytes, "status"); err != nil {
-				klog.Warningf("Error patching podReadinessGate: kubeedge.io/CloudCoreIsLeader to pod %v through apiserver: %v ,try again, times: %d", podname, err, i)
-				time.Sleep(time.Second)
-			} else {
-				klog.Infof("Successfully patching podReadinessGate: kubeedge.io/CloudCoreIsLeader to pod %q through apiserver", podname)
+		for i := 1; i <= maxRetries; i++ {
+			_, err = cli.CoreV1().Pods(namespace).Patch(podname, types.StrategicMergePatchType, patchBytes, "status")
+			if err == nil {
 				isPatchSuccess = true
+				klog.Infof("Successfully patching podReadinessGate: kubeedge.io/CloudCoreIsLeader to pod %q through apiserver", podname)
 				break
 			}
+			if errors.IsConflict(err) {
+				// If the patch failure is due to update conflict, the necessary retransmission is performed
+				if i >= maxRetries {
+					klog.Errorf("updateMaxRetries(%d) has reached, failed to patching podReadinessGate: kubeedge.io/CloudCoreIsLeader because of update conflict", maxRetries)
+				}
+				continue
+			}
+			break
 		}
-		if isPatchSuccess != true {
-			return fmt.Errorf("failed to patch after 5 attempts")
+		if !isPatchSuccess {
+			return err
 		}
 	} else {
 		klog.Infoln("CloudCore is not running in pod")


fisherxu:
>  There is already a complete retransmission mechanism when client-go sending requests. So I delete the for loop.

I think you mean the code in [client-go here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/rest/request.go#L880)? Here it only repeat for the `GET` requests, noyt for all write request. @Poor12

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -152,18 +152,15 @@ func TryToPatchPodReadinessGate() error {
...
-				klog.Warningf("Error patching podReadinessGate: kubeedge.io/CloudCoreIsLeader to pod %v through apiserver: %v ,try again, times: %d", podname, err, i)
-				time.Sleep(time.Second)
-			} else {
-				klog.Infof("Successfully patching podReadinessGate: kubeedge.io/CloudCoreIsLeader to pod %q through apiserver", podname)
-				isPatchSuccess = true
-				break
-			}
+		if _, err = cli.CoreV1().Pods(namespace).Patch(podname, types.StrategicMergePatchType, patchBytes, "status"); err != nil {
+			klog.Warningf("Error patching podReadinessGate: kubeedge.io/CloudCoreIsLeader to pod %v through apiserver: %v", podname, err)
+			time.Sleep(time.Second)
```
kevin-wangzefeng, line 157: And we don't need sleep at here then.
```suggestion

```

Poor12:
> > There is already a complete retransmission mechanism when client-go sending requests. So I delete the for loop.
> 
> I think you mean the code in [client-go here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/rest/request.go#L880)? Here it only repeats for the `GET` requests, noyt for all write requests. @Poor12

Yeah，Is this for loop mainly for the case where get succeeds but write fails due to resource conflicts?Can we target specific errors?

kevin-wangzefeng:
> > > There is already a complete retransmission mechanism when client-go sending requests. So I delete the for loop.
> > 
> > 
> > I think you mean the code in [client-go here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/rest/request.go#L880)? Here it only repeats for the `GET` requests, noyt for all write requests. @Poor12
> 
> Yeah，Is this for loop mainly for the case where get succeeds but write fails due to resource conflicts?Can we target specific errors?

The reason why client-go didn't implement automatic retry for write calls is that in most cases, an additional GET is needed when conflict occurred.

Poor12:
> > > > There is already a complete retransmission mechanism when client-go sending requests. So I delete the for loop.
> > > 
> > > 
> > > I think you mean the code in [client-go here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/rest/request.go#L880)? Here it only repeats for the `GET` requests, noyt for all write requests. @Poor12
> > 
> > 
> > Yeah，Is this for loop mainly for the case where get succeeds but write fails due to resource conflicts?Can we target specific errors?
> 
> The reason why client-go didn't implement automatic retry for write calls is that in most cases, an additional GET is needed when conflict occurred.

yeah，I review the code in k8s. PatchPodStatus, as a method implementing patching,doestn't use loop either.

Poor12:
I review the code in k8s. PatchPodStatus, as a method implementing patching,doestn't use loop either.I still regard it as an unnecessary thing. @fisherxu @kevin-wangzefeng

fisherxu:
> I review the code in k8s. PatchPodStatus, as a method implementing patching,doestn't use loop either.I still regard it as an unnecessary thing. @fisherxu @kevin-wangzefeng

You can see [this func](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/garbagecollector/operations.go#L79) in k8s, if Conflict occured when patching obj, it will retry, and if err is not conflict, it wil return.

And here in TryToPatchPodReadinessGate, I think the retry is reasonable and harmless, also avoiding occasional failure.

Poor12:
I saw the removeFinalizer function.How about add a logic to judge the type of error same as it？ If conflict then retry ，otherwise just return failure. It can quite reduce some useless retries. @fisherxu

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -150,20 +149,28 @@ func TryToPatchPodReadinessGate() error {
...
+				klog.Infof("Successfully patching podReadinessGate: kubeedge.io/CloudCoreIsLeader to pod %q through apiserver", podname)
 				break
 			}
+			if errors.IsConflict(err) {
+				// If the patch failure is due to update conflict, the necessary retransmission is performed
+				if i < maxRetries {
+					klog.Errorf("Error patching podReadinessGate: kubeedge.io/CloudCoreIsLeader because of update conflicts: try again, times: %d", i)
+				} else {
+					klog.Errorf("Error patching podReadinessGate: kubeedge.io/CloudCoreIsLeader because of update conflicts after %d times", i)
+				}
```
fisherxu, line 167:
```suggestion
				if i >= maxRetries  {
					klog.Errorf("updateMaxRetries(%d) has reached, failed to patching podReadinessGate: kubeedge.io/CloudCoreIsLeader because of update conflict", maxRetries)
				}
```

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -150,20 +149,28 @@ func TryToPatchPodReadinessGate() error {
...
+				continue
+			}
+			break
 		}
-		if isPatchSuccess != true {
-			return fmt.Errorf("failed to patch after 5 attempts")
+		if isPatchSuccess == false {
```
fisherxu, line 172: !isPatchSuccess

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -150,20 +149,28 @@ func TryToPatchPodReadinessGate() error {
...
+			return fmt.Errorf("Error patching podReadinessGate: kubeedge.io/CloudCoreIsLeader to pod %v through apiserver: %v", podname, err)
```
fisherxu, line 173: Failed to patch

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -150,20 +149,26 @@ func TryToPatchPodReadinessGate() error {
...
+					klog.Errorf("updateMaxRetries(%d) has reached, failed to patching podReadinessGate: kubeedge.io/CloudCoreIsLeader because of update conflict", maxRetries)
+				}
+				continue
+			}
+			break
 		}
-		if isPatchSuccess != true {
-			return fmt.Errorf("failed to patch after 5 attempts")
+		if !isPatchSuccess {
+			return fmt.Errorf("Failed to patch")
```
fisherxu, line 171: Looks good to me :) How about return err directly here.

return err

Poor12: Fine:) add a log for other failure reason.

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -150,20 +149,27 @@ func TryToPatchPodReadinessGate() error {
...
+					klog.Errorf("updateMaxRetries(%d) has reached, failed to patching podReadinessGate: kubeedge.io/CloudCoreIsLeader because of update conflict", maxRetries)
+				}
+				continue
+			}
+			break
 		}
-		if isPatchSuccess != true {
-			return fmt.Errorf("failed to patch after 5 attempts")
+		if !isPatchSuccess {
+			klog.Errorln("Failed to patch")
```
fisherxu, line 171: Log here can be deleted now

PR #1654, opened 2020-05-10
fix bug for close nil channel

### LINKED ISSUE #1559: close nil channel
**What happened**:

> W0323 11:12:01.405476   24637 messagehandler.go:174] Timeout to receive heart beat from edge node abc for project e632aba927ea4ac2b575ec1603d56f10
> panic: close of nil channel
> 
> goroutine 21532 [running]:
> github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/handler.(*MessageHandle).UnregisterNode(0xc0008c9420, 0x18fd040, 0xc00075a280, 0xc0007eeba0, 0x4)
> 	/root/go/src/github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/handler/messagehandler.go:299 +0x113
> github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/handler.(*MessageHandle).ServeConn(0xc0008c9420, 0x18fd040, 0xc00075a280, 0xc0007eeba0)
> 	/root/go/src/github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/handler/messagehandler.go:271 +0x289
> created by github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/handler.(*MessageHandle).OnRegister
> 	/root/go/src/github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/handler/messagehandler.go:159 +0x224

```
func (mh *MessageHandle) UnregisterNode() {
	mh.nodeLocks.Delete(info.NodeID)
	mh.nodeConns.Delete(info.NodeID)
	close(mh.KeepaliveChannel[info.NodeID])
	delete(mh.KeepaliveChannel, info.NodeID)
```

**Environment**:

**CloudSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

**EdgeSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

qingchen1203:
**What type of PR is this?**
/kind bug

fix bug for close nil channel 

Fixes #1559

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -142,6 +142,9 @@ func (mh *MessageHandle) OnRegister(connection conn.Connection) {
 
 	if _, ok := mh.KeepaliveChannel[nodeID]; !ok {
 		mh.KeepaliveChannel[nodeID] = make(chan struct{}, 1)
+	} else {
+		klog.Warningf("Node %v/%v has not yet exit, OnRegister failed", projectID, nodeID)
+		return
 	}
 
 	io := &hubio.JSONIO{Connection: connection}


cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -142,6 +142,9 @@ func (mh *MessageHandle) OnRegister(connection conn.Connection) {
 
 	if _, ok := mh.KeepaliveChannel[nodeID]; !ok {
 		mh.KeepaliveChannel[nodeID] = make(chan struct{}, 1)
+	} else {
+		klog.Warningf("Node %v/%v has not yet exit, OnRegister failed", projectID, nodeID)
```
fisherxu, line 146: has not yet exit?

qingchen1203: How about this "Node %v/%v is already registered, OnRegister again failed" ?

qingchen1203: According to #1559 , this bug is due to bad network. Call the function OnRegister() in a short time make the channel 2 times by func (mh *MessageHandle) OnRegister(connection conn.Connection) and with same nodeID. Ideally, the life cycle of a channel looks like this: 2 make operation produces  corresponding connection with different nodeID in KeepaliveChannel,  and close the connection without disturbing each other. The key value in map KeepaliveChannel[nodeID] is nodeID. Therefore, at this time, the two make operations have only one channel and two close operations, which will cause the error of closing the nil channel. This make operation should be atomic. 
Therefore, when registering a node use func OnRegister(connection conn.Connection),  we should make sure to exit the previous channel with the same nodeID, otherwise it will not be registered.

qingchen1203: @fisherxu

GsssC: Based on further investigation #1704 #1670, the problem may not be limited to atomization of registerNode and unRegisterNode

PR #1656, opened 2020-05-11
update the doc kubeedge_configure

XJangel:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind documentation

update the configuration steps

Fixes #

```release-note

```

### DIFF
--- a/docs/setup/kubeedge_configure.md
+++ b/docs/setup/kubeedge_configure.md
@@ -26,7 +26,6 @@ mkdir -p /etc/kubeedge/config/
 Either create a minimal configuration with command `~/kubeedge/cloudcore --minconfig`
 
 ```shell
-
 ~/kubeedge/cloudcore --minconfig > /etc/kubeedge/config/cloudcore.yaml
 ```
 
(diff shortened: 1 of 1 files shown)


XJangel:
> Will add the `advertise-address` in this PR? @XJangel

For the deployment of single CloudCore, the default advertise-address is just the IP address of CloudCore. So I didn't mention it here. Maybe we can do it  in the documentation of HA CloudCore.

fisherxu:
> Maybe we can do it in the documentation of HA CloudCore.

The `advertise address` is not only used in ha mode. If edgecore connect to cloudnode through a public(floating) IP, we should set the floating ip in the config file. And the floating Ip can't be  automatically acquired. :)

XJangel:
> > Maybe we can do it in the documentation of HA CloudCore.
> 
> The `advertise address` is not only used in ha mode. If edgecore connect to cloudnode through a public(floating) IP, we should set the floating ip in the config file. And the floating Ip can't be automatically acquired. :)

Get it! I will modify it right now.

fisherxu:
Here you don't need explain ha, as long as you explain all the addresses(like floating ip) exposed to the edge node should be added to `advertise address`.

XJangel:
> Here you don't need explain ha, as long as you explain all the addresses(like floating ip) exposed to the edge node should be added to `advertise address`.

yep~

docs/setup/kubeedge_configure.md
```diff
@@ -74,6 +74,19 @@ In the cloudcore.yaml, modify the below settings.
 2. Before KubeEdge v1.3: check whether the cert files for `modules.cloudhub.tlsCAFile`, `modules.cloudhub.tlsCertFile`,`modules.cloudhub.tlsPrivateKeyFile` exists.
 
     From KubeEdge v1.3: just skip the above check. If you configure the CloudCore certificate manually, you must check if the path of certificate is right. 
+    
+    
+    
+    **Note:** If your KubeEdge version is before the v1.3, then just skip the step 3.
+    
+3. Configure the IP addresses of CloudCore which are exposed to the edge nodes(like floating IP) in the `advertiseAddress`
```
fisherxu, line 82:
```suggestion
3. Configure all the IP addresses of CloudCore which are exposed to the edge nodes(like floating IP) in the `advertiseAddress`, which will be add to SANs in cert of cloudcore. 
```

PR #1668, opened 2020-05-14
fix logs when node disconnected

ls889:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind bug

The logs have a problem that misleading users when viewing logs.
when the edgecore gets disconnected, the log still shows error `write error, connection for node edge-node will be closed......reason node disconnected`, I think "will be closed" is inaccurate.because it has been closed.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -445,8 +445,13 @@ LOOP:
 func (mh *MessageHandle) send(hi hubio.CloudHubIO, info *model.HubInfo, msg *beehiveModel.Message) {
 	err := mh.hubIoWrite(hi, info.NodeID, msg)
 	if err != nil {
-		klog.Errorf("write error, connection for node %s will be closed, affected event %s, reason %s",
-			info.NodeID, dumpMessageMetadata(msg), err.Error())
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -444,6 +444,11 @@ LOOP:
 
 func (mh *MessageHandle) send(hi hubio.CloudHubIO, info *model.HubInfo, msg *beehiveModel.Message) {
 	err := mh.hubIoWrite(hi, info.NodeID, msg)
+	if err.Error() == "node disconnected" {
```
fisherxu, line 447: Has any difference with L42? :)

kevin-wangzefeng: The only difference is "has been closed" and "will bee closed".
I would suggest to updated the patch to avoid code duplication.

ls889: @kevin-wangzefeng  How to update the patch to avoid code duplication? :)

PR #1669, opened 2020-05-14
keadm: add error message for GetLatestVersion()'s exec.Output()

darunshen:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

reason: 
We need more info about errors of exec.Command,but not just one 'exit status number'.
this may cause it to look a little bloated when related error occurs.
And i only changed the GetLatestVersion()'s exec.Output() , so it will be nice if we can provide unified management for exec.Command.

i'm not sure , just adding more log to error returned

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -27,13 +27,13 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/kubeedge/kubeedge/pkg/util"
 	"github.com/spf13/pflag"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"github.com/kubeedge/kubeedge/pkg/util"
 
 	types "github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
 )
@@ -226,7 +226,7 @@ func IsCloudCore() (types.ModuleRunning, error) {
 func GetLatestVersion() (string, error) {
 	//Download the tar from repo
 	versionURL := "curl -k " + latestReleaseVersionURL
-	latestReleaseData, err := util.Command("sh",[]string{"-c",versionURL})
+	latestReleaseData, err := util.Command("sh", []string{"-c", versionURL})
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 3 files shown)


keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -226,9 +226,11 @@ func GetLatestVersion() (string, error) {
 	//Download the tar from repo
 	versionURL := "curl -k " + latestReleaseVersionURL
 	cmd := exec.Command("sh", "-c", versionURL)
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
```
daixiang0, line 230: Could we simpify it as `cmd.Stderr = &bytes.Buffer{}`?

darunshen: if we do,we should also change this line:
`return "", fmt.Errorf("%v:%v", err, stderr.String())`
to
`return "", fmt.Errorf("%v:%v",err,cmd.Stderr.(*bytes.Buffer).String())`

i don't think it's simpified.

daixiang0: Could we call Command from [util pkg](https://github.com/kubeedge/kubeedge/blob/master/pkg/util/util.go#L453-L462)?

darunshen: yes, it's a better method.

so do we need to add more error print to util.Commond() like this:
```
//Command executes command and returns output
func Command(name string, arg []string) (string, error) {
	cmd := exec.Command(name, arg...)
	var stderr bytes.Buffer
	cmd.Stderr = &stderr
	ret, err := cmd.Output()
	if err != nil {
		err = fmt.Errorf("%v:%v", err, stderr.String())
		klog.Errorf("exec command failed: %v", err)
		return string(ret), err
	}
	return strings.Trim(string(ret), "\n"), nil
}
```
?

daixiang0: Looks good, just add blank after ":".

darunshen: ok, like this : `err = fmt.Errorf("%v:\n%v", err, stderr.String())`;
and the travis-ci tells that `[It] E2E_UPDATE_DEVICE_1`  and `[It] E2E_POD_DEPLOYMENT` get failed ,should i care about them?

daixiang0: Sorry, i mean empty space like "%v: %v". 
Some cases maybe failed since little resource, so please rerun test.

darunshen: ok

fisherxu:
@darunshen  Could please squash the commits and rebase the master branch to fix the CI issue?

fisherxu:
Any updates? @darunshen

PR #1693, opened 2020-05-18
.travis.yml: add interation test for arm64

daixiang0:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ jobs:
       script:
         - make
         - make bluetoothdevice
-        - make smallbuild 
+        - make smallbuild
         - make crossbuild
         - make crossbuild GOARM=GOARM7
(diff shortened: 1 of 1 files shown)


daixiang0:
Error  refer to this [link](https://travis-ci.org/github/kubeedge/kubeedge/jobs/689091015#L529-L539).

daixiang0:
The root cause is that travis ci use LXD to launch arm64 container, it is not a VM, can use root premission to do some test, details refer to:
- https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system
- https://linuxcontainers.org/lxc/security/#privileged-containers

I also check circle ci, it only support amd64, see https://circleci.com/docs/2.0/faq/#which-cpu-architectures-does-circleci-support.

daixiang0:
For same reason, can not run dind to test.

PR #1708, opened 2020-05-20
fix using same namespace while syncPodStatus

### LINKED ISSUE #1700: Pod status can not be updated because of namespaces in message is wrong
**What happened**:
I depoyed some pods which have different namespace on edgenode. At a unknown moment, I found my pods status gotten by "kubectl get -oyaml" are aways unready. Logs shows that message sent from edgenode is wrong in namespace.

Pod status update normally

The bug may reproduce by restarting cloudcore some times, or restarting edgecore. I only remember doing restarting before I found the bug.

- cloudcore log
image
- edgecore log
image

**Environment**:
- KubeEdge version(e.g. `cloudcore/edgecore --version`):  v1.3.0

GsssC:
**What type of PR is this?**
 /kind bug

fix merging different namespace podstatus  into same ns message while syncPodStatus

Fixes #1700

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -512,7 +512,7 @@ func (m *metaManager) processSync(message model.Message) {
 }
 
 func (m *metaManager) syncPodStatus() {
-	klog.Infof("start to sync pod status")
+	klog.Infof("start to sync pod status in edge-store to cloud")
 	podStatusRecords, err := dao.QueryAllMeta("type", model.ResourceTypePodStatus)
 	if err != nil {
 		klog.Errorf("list pod status failed: %v", err)
@@ -522,13 +522,9 @@ func (m *metaManager) syncPodStatus() {
 		klog.Infof("list pod status, no record, skip sync")
 		return
 	}
-
-	var namespace string
-	content := make([]interface{}, 0, len(*podStatusRecords))
+	contents := make(map[string][]interface{})
 	for _, v := range *podStatusRecords {
-		if namespace == "" {
-			namespace, _, _, _ = util.ParseResourceEdge(v.Key, model.QueryOperation)
-		}
+		namespaceParsed, _, _, _ := util.ParseResourceEdge(v.Key, model.QueryOperation)
 		podKey := strings.Replace(v.Key, constants.ResourceSep+model.ResourceTypePodStatus+constants.ResourceSep, constants.ResourceSep+model.ResourceTypePod+constants.ResourceSep, 1)
 		podRecord, err := dao.QueryMeta("key", podKey)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


fisherxu:
Will review today, and could you please rebase the master branch to fix the ci first?

GsssC:
> Will review today, and could you please rebase the master branch to fix the ci first?

done

edge/pkg/metamanager/process.go
```diff
@@ -549,12 +545,17 @@ func (m *metaManager) syncPodStatus() {
 			klog.Errorf("unmarshal podstatus[%s] failed, content[%s]: %v", v.Key, v.Value, err)
 			continue
 		}
-		content = append(content, podStatus)
+		if _, ok := contents[namespaceParsed]; !ok {
+			contents[namespaceParsed] = []interface{}{podStatus}
```
fisherxu, line 549: Do we need this line here? Why not append directly.

GsssC: done : )
you r rigint, I thought the argument couldn't be nil.

edge/pkg/metamanager/process.go
```diff
@@ -549,12 +545,13 @@ func (m *metaManager) syncPodStatus() {
 			klog.Errorf("unmarshal podstatus[%s] failed, content[%s]: %v", v.Key, v.Value, err)
 			continue
 		}
-		content = append(content, podStatus)
+		contents[namespaceParsed] = append(contents[namespaceParsed], podStatus)
+	}
+	for namespace, content := range contents {
+		msg := model.NewMessage("").BuildRouter(MetaManagerModuleName, GroupResource, namespace+constants.ResourceSep+model.ResourceTypePodStatus, model.UpdateOperation).FillBody(content)
+		sendToCloud(msg)
+		klog.Infof("sync pod status successful for namespaces %s, %s", namespace, msgDebugInfo(msg))
```
fisherxu, line 553: We can set the log level here to avoid to many logs, `klog.V(3).Infof`.
Ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/status_manager.go#L578

edge/pkg/metamanager/process.go
```diff
@@ -549,12 +545,14 @@ func (m *metaManager) syncPodStatus() {
...
-
-	msg := model.NewMessage("").BuildRouter(MetaManagerModuleName, GroupResource, namespace+constants.ResourceSep+model.ResourceTypePodStatus, model.UpdateOperation).FillBody(content)
-	sendToCloud(msg)
-	klog.Infof("sync pod status successful, %s", msgDebugInfo(msg))
+	for namespace, content := range contents {
+		msg := model.NewMessage("").BuildRouter(MetaManagerModuleName, GroupResource, namespace+constants.ResourceSep+model.ResourceTypePodStatus, model.UpdateOperation).FillBody(content)
+		sendToCloud(msg)
+		klog.V(3).Infof("sync pod status successfully for namespaces %s, %s", namespace, msgDebugInfo(msg))
+	}
+	klog.Infof("successfully sync pod status in edge-store to cloud")
```
fisherxu, line 555: Default logs should focus on the error log, this successfully log will be printed out periodly, I think it's useless and can remove it, WDYT?

GsssC: Yes, it sounds all reasonable

> Default logs should focus on the error log, this successfully log will be printed out periodly, I think it's useless and can remove it, WDYT?

Yes, it sounds all reasonable

PR #1717, opened 2020-05-22
WIP: auto Patch readiness to false after leader lost

GsssC:
**What type of PR is this?**
> /kind bug

For now, CloudCore which losts leader position and restart will not  dePatch the ReadinessGate. It cause two cloudcore pod both have Readiness to true at the same time, not as expected.

Fixes #

```release-note

```


GsssC:
It does not resolve the situation when program calls klog.Fatal() , shutdown immediatly, and can not call dePatch().

GsssC:
Can we try to dePatch Readiness after the very beginning of program and before leaderelection starting?
@kevin-wangzefeng

PR #1739, opened 2020-05-27
Auto detect pause image

daixiang0:


fisherxu:
EdgeCore is also need to fix :)

PR #1775, opened 2020-06-04
fix wrong parse operation when service url is short

XiaoJiangWang:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

fix wrong parse operation when service url is short in edgemesh load balancer chain

Fixes #

```release-note

```

### DIFF
--- a/edgemesh/pkg/plugin/registry/registry.go
+++ b/edgemesh/pkg/plugin/registry/registry.go
@@ -126,8 +126,9 @@ func (esd *EdgeServiceDiscovery) AutoSync() {}
 func (esd *EdgeServiceDiscovery) Close() error { return nil }
 
 // parseServiceURL parses serviceURL to ${service_name}.${namespace}.svc.${cluster}:${port}, keeps with k8s service
-func parseServiceURL(serviceURL string) (name, namespace string, port int, err error) {
-	name, namespace = common.SplitServiceKey(serviceURL)
+func parseServiceURL(serviceURL string) (string, string, int, error) {
+	var port int
+	var err error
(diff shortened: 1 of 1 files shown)


edgemesh/pkg/plugin/registry/registry.go
```diff
@@ -141,7 +140,9 @@ func parseServiceURL(serviceURL string) (name, namespace string, port int, err e
 	} else {
 		klog.Errorf("[EdgeMesh] service url %s invalid", serviceURL)
 		err = fmt.Errorf("service url %s invalid", serviceURL)
+		return
```
fisherxu, line 143:
```suggestion
		return fmt.Errorf("service url %s invalid", serviceURL)
```
And the error in L137 also need return :)

XiaoJiangWang: It is not necessary. In L129, the function definition showed the return values.

fisherxu: We can delete the unneeded return values.

XiaoJiangWang: fixed

fisherxu:
/lgtm
It seems that the email address you used in your commit hasn't been added to your github personal profile. So the contributions will not be counted in github. :)
image

XiaoJiangWang:
> /lgtm
> It seems that the email address you used in your commit hasn't been added to your github personal profile. So the contributions will not be counted in github. :)
> image

fixed. thanks.

PR #1780, opened 2020-06-05
modifiy edgehub metamanager log to help reading

GsssC:
**What type of PR is this?**
log modification
because there are multiple `process.go` such as MetaManger/process.go, EdgeHub/process.go, etc.
It is necessary to distinguish them more intuitivly in log.

### DIFF
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -115,15 +115,15 @@ func (eh *EdgeHub) sendToKeepChannel(message model.Message) error {
 	defer eh.keeperLock.RUnlock()
 	channel, exist := eh.syncKeeper[message.GetParentID()]
 	if !exist {
-		klog.Errorf("failed to get sync keeper channel, messageID:%+v", message)
-		return fmt.Errorf("failed to get sync keeper channel, messageID:%+v", message)
+		klog.Errorf("EdgeHub failed to get respose message sync keeper channel for messageID:%+v", message)
+		return fmt.Errorf("failed to get response message sync keeper channel for messageID:%+v", message)
 	}
 	// send response into synckeep channel
 	select {
 	case channel <- message:
 	default:
-		klog.Errorf("failed to send message to sync keep channel")
-		return fmt.Errorf("failed to send message to sync keep channel")
+		klog.Errorf("EdgeHub failed to send response message to sync keep channel")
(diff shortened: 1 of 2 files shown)


edge/pkg/edgehub/process.go
```diff
@@ -154,15 +154,15 @@ func (eh *EdgeHub) routeToEdge() {
...
 		message, err := eh.chClient.Receive()
 		if err != nil {
-			klog.Errorf("websocket read error: %v", err)
+			klog.Errorf("EdgeHub websocket read error: %v", err)
 			eh.reconnectChan <- struct{}{}
 			return
 		}
 
-		klog.Infof("received msg from cloud-hub:%+v", message)
+		klog.V(2).Infof("EdgeHub received msg from cloud-hub: %+v", message)
```
kadisi, line 162: why use V2?

GsssC: Message here will be logged by MetaManager or other modules as well. For my view, edgehub would better be transparent and only to log error.

daixiang0:
Is there any possible to add prefix to each component log like `[edgehub] xxxx`?

fisherxu:
> Is there any possible to add prefix to each component log like [edgehub] xxxx?
It would be nice and helpful to locate issue.

Good idea :)

edge/pkg/edgehub/process.go
```diff
@@ -172,8 +172,8 @@ func (eh *EdgeHub) sendToCloud(message model.Message) error {
 	err := eh.chClient.Send(message)
 	eh.keeperLock.Unlock()
 	if err != nil {
-		klog.Errorf("failed to send message: %v", err)
-		return fmt.Errorf("failed to send message, error: %v", err)
+		klog.Errorf("EdgeHub client failed to send message to cloud: %v", err)
+		return fmt.Errorf("edgeHub client failed to send message to cloud, error: %v", err)
```
fisherxu, line 176: EdgeHub?

GsssC:
@daixiang0 I can’t agree more. I just add prefix to some important  and periodic log manually.

fisherxu:
> @daixiang0 I can’t agree more. I just add prefix to some important and periodic log manually.

So we can add the prefix for some logs and add log level for some meaningless logs.

PR #1789, opened 2020-06-08
update the docs of deploying v1.3 cloud part

ls889:
Integrated deployment docs of cloudcore

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind test
> /kind failing-test
> /kind feature

 /kind documentation

**What this PR does / why we need it**:
I integrate the documentations of deploying cloud part.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #1792, opened 2020-06-09
hack/lib/lint.sh: skip deleted files

daixiang0:

**What type of PR is this?**

/kind bug

Fixes this [issue](https://github.com/kubeedge/kubeedge/pull/1789#issuecomment-641005314).


luogangyi:
btw, can we add some friendly tips if we faild tail space check？
I found that if we faild tail space check currently, no any meaningful tips will given, it's hard to find why failed

PR #1796, opened 2020-06-09
edged.go: Don't restart when container exited

### LINKED ISSUE #1364: Network unreachable in container
**What happened**:

I deployed my deployment, but kubeedge ran my container first, then the kubeedge/pause:3.1 container.
So my container can not access the network.

This problem does not happen every time.
When it happens, the "docker ps" command show like this:
image

And in this case, I would get "Network unreachable" in my container.
image

**Environment**:

**CloudSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

**EdgeSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

faicker:
see issue #1364.
When sandbox container is started but not ready, PLEG relist will
generate a container die event. Func syncLoopIteration will add the pod to
restart queue. This is unexpected compared to kubelet's op of deleting
container of the pod.
Latest container will be kept in default in kubelet.
This is a lazy way of doing to ignore the die event.
By the way, the restart logic is probmatic because the app container is
not restarted.

Signed-off-by: Faicker Mo <faicker.mo@gmail.com>

/kind bug
Fixes #1364

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -763,12 +763,16 @@ func (e *edged) syncLoopIteration(plegCh <-chan *pleg.PodLifecycleEvent, houseke
 							break
 						}
 					}
-					klog.Errorf("sync loop get event container died, restart pod [%s]", pod.Name)
-					key := types.NamespacedName{
-						Namespace: pod.Namespace,
-						Name:      pod.Name,
+					if pod.Status.Phase == v1.PodPending {
+						klog.Infof("sync loop ignore die event in pod [%s] pending phase", pod.Name)
+					} else {
+						klog.Errorf("sync loop get event container died, restart pod [%s]", pod.Name)
+						key := types.NamespacedName{
+							Namespace: pod.Namespace,
+							Name:      pod.Name,
+						}
+						e.podAdditionQueue.Add(key.String())
 					}
-					e.podAdditionQueue.Add(key.String())
 				} else {
 					klog.Infof("sync loop get event [%s], ignore it now.", plegEvent.Type)
 				}


faicker:
the restart logic has problem. Because the app container is not restarted, only the pause container is restarted.
Restart logic is in the k8s lib.
This pr is done compared to the kubelet logic.

Baoqiang-Zhang:
@faicker thanks for your PR。the container should be restart if pause container changed, as in the code below:

Baoqiang-Zhang:
> the restart logic has problem. Because the app container is not restarted, only the pause container is restarted.
> Restart logic is in the k8s lib.
> This pr is done compared to the kubelet logic.

Pod lifecycle events are also handled in kubelet code.
https://github.com/kubernetes/kubernetes/blob/fdec14cd4c84b74816a23b5e78bbf0a0d8bf7468/pkg/kubelet/kubelet.go#L1947-L1963

faicker:
The handling of die event when sandbox is started is different with kubelet.
In kubelet, it will do cleanUpContainersInPod(kubernetes v1.17.1) https://github.com/kubernetes/kubernetes/blob/d224476cd0730baca2b6e357d144171ed74192d6/pkg/kubelet/kubelet.go#L1961
the sandbox will not be deleted.
In kubedge, it will restart the pod. https://github.com/kubeedge/kubeedge/blob/bcb88483889b22b0c97067730809e3332cbacceb/edge/pkg/edged/edged.go#L772



When restarting the pod, maybe the app container is ready, but the k8s is not aware of it. It's async. 
It will only restart the sandbox(pause) container.

fisherxu:
@faicker https://github.com/kubernetes/kubernetes/blob/d224476cd0730baca2b6e357d144171ed74192d6/pkg/kubelet/kubelet.go#L1952
In kubelet it will also do the pod sync? In kubeedge, add to the queue also just trigger the pod sync? Not very sure :)

faicker:
I0703 13:47:57.465780   24068 generic.go:155] GenericPLEG: 68ee34e1-9793-4578-912b-90dd34a91146/611d042b9f61df710ee90a2a8fdcc6ad08b718ed9b7825750557848bd3d7a549: non-existent -> exited

the kubelet exec path when received this event,
HandlePodSyncs -> dispatchWork -> UpdatePod and last isWorking is true, options is added to lastUndeliveredWorkUpdate,no syncPod.
and then cleanUpContainersInPod, does not clean really.
This event does not restart the pod.

I will work out this case but keep the restart logic.

fisherxu:
> I will work out this case but keep the restart logic.

Thanks! @faicker  And could you join to our slack channel and have more discussion? The Link: https://join.slack.com/t/kubeedge/shared_invite/enQtNjc0MTg2NTg2MTk0LWJmOTBmOGRkZWNhMTVkNGU1ZjkwNDY4MTY4YTAwNDAyMjRkMjdlMjIzYmMxODY1NGZjYzc4MWM5YmIxZjU1ZDI

faicker:
@fisherxu thanks for joining me.

faicker:
updated, with only pod pending state.

fisherxu:
With #1819 got merged, it seems that the issue #1364 cannot be reproduced...

PR #1819, opened 2020-06-15
edged.go: Fix pod sync copied from kubelet

### LINKED ISSUE #1364: Network unreachable in container

faicker:
/kind bug

Fixes #1364 

It happens when two same message from the cloud-hub(because auto resend)

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -239,7 +239,8 @@ type edged struct {
 
 	dockerLegacyService dockershim.DockerLegacyService
 	// Optional, defaults to simple Docker implementation
-	runner kubecontainer.ContainerCommandRunner
+	runner          kubecontainer.ContainerCommandRunner
+	podLastSyncTime map[types.UID]time.Time
 }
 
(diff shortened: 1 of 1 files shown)


fisherxu:
Thanks for your fix! @faicker 
And I noticed another PR raised by you #1796, they all fixed #1364, do we need combine them in one PR?

faicker:
These are two different cases which leads to the same result.
Maybe not good to combine them, but I can combine them If you want.

fisherxu:
Sorry for the delay and thanks for the fixing! @faicker

PR #1813, opened 2020-06-12
update project maintainers

kevin-wangzefeng:
Signed-off-by: Kevin Wang <kevinwzf0126@gmail.com>

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design

/kind documentation
> /kind test
> /kind failing-test
> /kind feature

**What this PR does / why we need it**:

- Add @chendave, @kadisi and @fisherxu as new maintainers.

- Move @CindyXing @m1093782566 @rohitsardesai83 to emeritus.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #1822, opened 2020-06-16
hack/lib/lint.sh: fix lint error when running on master

daixiang0:

**What type of PR is this?**

/kind bug

Fix [merge issue](https://travis-ci.org/github/kubeedge/kubeedge/jobs/698780226#L673-L677).

If set pipefail, when merge, the grep regex always fail and return 1, so here disable pipefail.

Can not disable pipefail outside function, since we would source other scripts.

### DIFF
--- a/hack/lib/lint.sh
+++ b/hack/lib/lint.sh
@@ -23,6 +23,7 @@ set -o pipefail
 kubeedge::lint::check() {
     cd ${KUBEEDGE_ROOT}
     # skip deleted files
+    set +o pipefail
     git diff --cached --name-only --diff-filter=ACRMTU master | grep -Ev "externalversions|fake|vendor" | xargs --no-run-if-empty sed -i 's/[ \t]*$//'
 
     [[ $(git diff --name-only) ]] && {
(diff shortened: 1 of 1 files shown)


hack/lib/lint.sh
```diff
@@ -18,11 +18,11 @@
 
 set -o errexit
 set -o nounset
-set -o pipefail
 
 kubeedge::lint::check() {
     cd ${KUBEEDGE_ROOT}
     # skip deleted files
+    set +o pipefail
```
chendave, line 25: Just curious, why this doesn't prevent the pr https://github.com/kubeedge/kubeedge/pull/1813 from merging?

kadisi, line 25: I recommend that pipefail be enabled by default on line 21  （keep  `set -o pipefail`）

You can cancel pipefail temporarily in this line.  (`set +o pipefail`)

And then finally turn it on at the end of the function. (`set -o pipefail`)

daixiang0: Since it compares with master branch, but after merge, there is no different files, then grep nothing and return 1.

fisherxu: Should  turn it on at the end of the function. (set -o pipefail)

kadisi:
/lgtm

thanks @daixiang0

PR #1849, opened 2020-06-21
Make: verify copyright header

daixiang0:

**What type of PR is this?**

/kind cleanup

- Add missing copyright header
- Add header check


hack/copyright.go
```diff
@@ -0,0 +1,121 @@
...
+
+`)
+
+	// since different year in Copyright, just check this exist or not.
+	licenseFlag = []byte(`http://www.apache.org/licenses/LICENSE-2.0`)
+
+	usage = []byte(`
+Usage:
+  check:
+  	go run copyright.go check
```
kevin-wangzefeng, line 55: Mixed indent with space and tab?

daixiang0: Fix it.

hack/copyright.go
```diff
@@ -0,0 +1,121 @@
...
+		if filepath.Ext(path) != ".proto" && filepath.Ext(path) != ".go" {
+			return nil
+		}
+
+		b, err := ioutil.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		if !strings.Contains(string(b), string(licenseFlag)) {
```
kevin-wangzefeng, line 87: The checker is to ensure if inputed files have copyright header, relying on "http://www.apache.org/licenses/LICENSE-2.0" as keywords would break easily if any none APL license file inputed.

daixiang0: It check our own files, which should always use apache v2 license.

hack/copyright.go
```diff
@@ -0,0 +1,121 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+var (
+	// license compatible for Go and Proto files.
+	license = []byte(`/*
+Copyright 2020 The KubeEdge Authors.
```
kadisi, line 32: use `time.Now().Year()` to get current year ?

hack/copyright.go
kadisi: Shall we consider checking the `sh` file ? @daixiang0 @kevin-wangzefeng @fisherxu

hack/copyright.go
kevin-wangzefeng: Agreed

cloud/cmd/admission/admission.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2020 The KubeEdge Authors.
```
fisherxu, line 2: The year filed here indicates the date when the file was created, which also indicates the history of the project. So it should be consistent with the created date. Most files was added in 2018, some in 2019, others 2020.

hack/lib/lint.sh
```diff
@@ -33,4 +33,7 @@ kubeedge::lint::check() {
     golangci-lint run
     gofmt -l -w staging
     set -o pipefail
+
+    # check license
+    go run hack/copyright.go check
```
fisherxu, line 38: maybe should add to make verify.

cloud/cmd/admission/admission.go
daixiang0: I will use `git log` to get first time of file, since linux do not store create time of file.

hack/copyright.go
daixiang0: Done.

PR #1855, opened 2020-06-22
add garbage collection of reliablesyncs when node unregisters

ls889:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

 /kind bug
 /kind feature

when we use `kubectl delete node xxx` to unregister, the objectsyncs of the deleted node are still existing. There will be some problems if a new edgenode starts with the same name as the deleted node. So, we need garbage collection of these crds.

Fixes #

cc @fisherxu 

```release-note

```

### DIFF
--- a/cloud/pkg/synccontroller/synccontroller.go
+++ b/cloud/pkg/synccontroller/synccontroller.go
@@ -6,6 +6,8 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
@@ -68,6 +70,9 @@ type SyncController struct {
 	deviceLister            devicelister.DeviceLister
 	clusterObjectSyncLister synclister.ClusterObjectSyncLister
 	objectSyncLister        synclister.ObjectSyncLister
+
+	// client
+	crdClient *versioned.Clientset
 }
 
 func newSyncController(enable bool) *SyncController {
@@ -123,7 +128,15 @@ func newSyncController(enable bool) *SyncController {
 		nodeLister:              nodeInformer.Lister(),
 		clusterObjectSyncLister: clusterObjectSyncInformer.Lister(),
 		objectSyncLister:        objectSyncInformer.Lister(),
+
+		crdClient: crdClient,
 	}
+
+	sctl.nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			sctl.deleteObjectSyncs()
+		},
+	})
 
 	return sctl
 }
@@ -177,6 +190,8 @@ func (sctl *SyncController) Start() {
 	}
 
 	go wait.Until(sctl.reconcile, 5*time.Second, beehiveContext.Done())
+
+	sctl.deleteObjectSyncs()
 }
 
 func (sctl *SyncController) reconcile() {
@@ -188,7 +203,7 @@ func (sctl *SyncController) reconcile() {
 
 	allObjectSyncs, err := sctl.objectSyncLister.List(labels.Everything())
 	if err != nil {
-		klog.Errorf("Filed to list all the ObjectSyncs: %v", err)
+		klog.Errorf("Failed to list all the ObjectSyncs: %v", err)
 	}
 	sctl.manageObjectSync(allObjectSyncs)
 
@@ -221,6 +236,37 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
 			klog.Errorf("Unsupported object kind: %v", sync.Spec.ObjectKind)
 		}
 	}
+}
+
+func (sctl *SyncController) deleteObjectSyncs() {
+	syncs, err := sctl.objectSyncLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("Failed to list all the ObjectSyncs: %v", err)
+	}
+	for _, sync := range syncs {
+		nodeName := getNodeName(sync.Name)
+		isGarbage, err := sctl.checkObjectSync(sync)
+		if err != nil {
+			klog.Errorf("failed to check ObjectSync outdated, %s", err)
+		}
+		if isGarbage {
+			klog.Infof("ObjectSync %s will be deleted since node %s has been deleted", sync.Name, nodeName)
+			err = sctl.crdClient.ReliablesyncsV1alpha1().ObjectSyncs(sync.Namespace).Delete(sync.Name, metav1.NewDeleteOptions(0))
+			if err != nil {
+				klog.Errorf("failed to delete objectSync %s for edgenode %s, err: %v", sync.Name, nodeName, err)
+			}
+		}
+	}
+}
+
+// checkObjectSync checks whether objectSync is outdated
+func (sctl *SyncController) checkObjectSync(sync *v1alpha1.ObjectSync) (bool, error) {
+	nodeName := getNodeName(sync.Name)
+	_, err := sctl.nodeLister.Get(nodeName)
+	if errors.IsNotFound(err) {
+		return true, nil
+	}
+	return false, err
 }
 
 // BuildObjectSyncName builds the name of objectSync/clusterObjectSync


ls889:
@fisherxu  how about now? I run a goroutine `DeleteOutdatedObjectSync()` for gc in synccontroller's start. It will delete the resources when the node deleted. but I think it costs more system resources because it starts a new goroutine.

daixiang0:
In this case, in the for loop, no need to select, it would be duplicate, see https://stackoverflow.com/questions/39107003/select-with-single-case-blocks-adding-default-unblocks.

ls889:
@daixiang0 thanks, it passes.

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -82,6 +90,11 @@ func newSyncController(enable bool) *SyncController {
 	kubeSharedInformers := informers.NewSharedInformerFactory(kubeClient, 0)
 	crdFactory := crdinformerfactory.NewSharedInformerFactory(crdClient, 0)
 
+	nodesManager, err := manager.NewNodesManager(kubeClient, v1.NamespaceAll)
+	if err != nil {
+		klog.Warningf("Create nodes manager failed with error: %s", err)
```
fisherxu, line 95: Need to exit here?

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -222,6 +240,59 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+			node, ok := e.Object.(*v1.Node)
+			if !ok {
+				klog.Warningf("Object type: %T unsupported", node)
+				continue
+			}
+			switch e.Type {
+			case watch.Deleted:
+				syncs, err := sctl.objectSyncLister.List(labels.Everything())
+				if err != nil {
+					klog.Errorf("Filed to list all the ObjectSyncs: %v", err)
```
fisherxu, line 259: Failed

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -222,6 +240,59 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+				}
+				for _, sync := range syncs {
+					nodeName := getNodeName(sync.Name)
+					if isGarbage, err := sctl.objectSyncGarbageCollection(nodeName, sync.Namespace, sync.Name); isGarbage {
+						if err != nil {
+							klog.Errorf("objectSync GC failed, err:%v", err)
+						}
+					}
+				}
+			default:
```
fisherxu, line 269: no need here

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -222,6 +240,59 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+			}
+			switch e.Type {
+			case watch.Deleted:
+				syncs, err := sctl.objectSyncLister.List(labels.Everything())
+				if err != nil {
+					klog.Errorf("Filed to list all the ObjectSyncs: %v", err)
+				}
+				for _, sync := range syncs {
+					nodeName := getNodeName(sync.Name)
+					if isGarbage, err := sctl.objectSyncGarbageCollection(nodeName, sync.Namespace, sync.Name); isGarbage {
```
fisherxu, line 263: what does the isGarbage  mean here?

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -222,6 +240,59 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+				klog.Warningf("Object type: %T unsupported", node)
+				continue
+			}
+			switch e.Type {
+			case watch.Deleted:
+				syncs, err := sctl.objectSyncLister.List(labels.Everything())
+				if err != nil {
+					klog.Errorf("Filed to list all the ObjectSyncs: %v", err)
+				}
+				for _, sync := range syncs {
```
fisherxu, line 261: What's the relationship of the sync and node here? Maybe the code needs some refactor.

cloud/pkg/synccontroller/synccontroller.go
ls889: @fisherxu  isGarbage is true means the object is outdated because edgenode doesn't exist now. we need to delete it.

cloud/pkg/synccontroller/synccontroller.go
ls889: @fisherxu syncs refers to syncObject ,the relationship is the node have syncs.  when node unregisters, it's syncs should be deleted

cloud/pkg/synccontroller/synccontroller.go
GsssC: @fisher how about regarding `node deleted event` as a trigger of deleting outdated syncs, not only deleting specific node-related syncs, in consideration of cloudcore crash?

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -222,6 +240,58 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
 		}
 	}
 }
+func (sctl *SyncController) DeleteOutdatedObjectSync() {
+	for {
```
GsssC, line 244: plz add an init-delete when sync controller starts, node may be deleted when cloudcore crashes.

fisherxu: When the cloudcore start, do one round check and then triggered by node delete event?

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -222,6 +240,58 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+		select {
+		case <-beehiveContext.Done():
+			klog.Warning("Stop synccontroller DeleteOutdatedObjectSync loop")
+			return
+		case e := <-sctl.nodeManager.Events():
+			node, ok := e.Object.(*v1.Node)
```
fisherxu, line 250: Is it possible to reuse the nodeinformer in synccontroller to get the node delete event? @ls889 @GsssC

ls889: @fisherxu  I try to reuse it today, but failed. Because nodeinformer does'n expose lw. we can't access the event from informer.

cloud/pkg/synccontroller/synccontroller.go
fisher: @GsssC are you talking to me?

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -223,6 +244,59 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+			}
+			if err != nil {
+				klog.Fatalf("objectSync GC failed, err:%v", err)
+			}
+		}
+	}
+}
+
+// checkObjectSync checks whether objectSync is outdated
+func (sctl *SyncController) checkObjectSync(nodeName string) (isGarbage bool, err error) {
```
daixiang0, line 290: Is there necessary to specify val name when return:
`func (sctl *SyncController) checkObjectSync(nodeName string) (bool, error) {`

daixiang0: Seems the return val err never used, why not remove it?

cloud/pkg/synccontroller/synccontroller.go
daixiang0: It depends on which action we pick when the env is not clean, directly exit 1 or try some clean things to see if it can run as expected. Could we unify this logic?

cloud/pkg/synccontroller/synccontroller.go
ls889: @daixiang0  yes,(bool, error) is better.  the return val err should be used, I will add some lines to use it.

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -223,6 +244,59 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+				klog.Errorf("Fail to delete objectSync %v of EdgeNode %v which is offline", sync.Name, nodeName)
+			}
+		}
+	}
+}
+
+// checkObjectSync checks whether objectSync is outdated
+func (sctl *SyncController) checkObjectSync(sync *v1alpha1.ObjectSync) (bool, error) {
+	nodeName := getNodeName(sync.Name)
+	_, err := sctl.nodeLister.Get(nodeName)
```
ls889, line 293: @daixiang0 I read the underlying implementation of nodeLister.Get, I found err here is always nil even if it maybe has wrong. the err has two value, nil or NotFound. (in line206 /client-go@v0.17.1/tools/cache/store.go)

daixiang0: Let's keep sync with it :)

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -223,6 +238,37 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+		klog.Errorf("Failed to list all the ObjectSyncs: %v", err)
+	}
+	for _, sync := range syncs {
+		nodeName := getNodeName(sync.Name)
+		isGarbage, err := sctl.checkObjectSync(sync)
+		if err != nil {
+			klog.Errorf("Fail to check ObjectSync outdated, %s", err)
+		}
+		if isGarbage {
+			klog.Infof("ObjectSync %v will be deleted because node %v unregisters  ", sync.Name, nodeName)
```
fisherxu, line 253:
```suggestion
			klog.Infof("ObjectSync %s will be deleted since node %s has been deleted", sync.Name, nodeName)
```

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -223,6 +238,37 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+			err = sctl.crdClient.ReliablesyncsV1alpha1().ObjectSyncs(sync.Namespace).Delete(sync.Name, metav1.NewDeleteOptions(0))
+			if err != nil {
+				klog.Errorf("Fail to delete objectSync %v of EdgeNode %v which is offline", sync.Name, nodeName)
```
fisherxu, line 256:
```suggestion
				klog.Errorf("failed to delete objectSync %s for edgenode %s, err: %v", sync.Name, nodename, err)
```

cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -223,6 +238,37 @@ func (sctl *SyncController) manageObjectSync(syncs []*v1alpha1.ObjectSync) {
...
+func (sctl *SyncController) deleteObjectSyncs() {
+	syncs, err := sctl.objectSyncLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("Failed to list all the ObjectSyncs: %v", err)
+	}
+	for _, sync := range syncs {
+		nodeName := getNodeName(sync.Name)
+		isGarbage, err := sctl.checkObjectSync(sync)
+		if err != nil {
+			klog.Errorf("Fail to check ObjectSync outdated, %s", err)
```
fisherxu, line 250: failed

ls889:
@fisherxu done

PR #1846, opened 2020-06-20
EdgeMesh: support container level proxy

daixiang0:
Signed-off-by: Xiang Dai <long0dai@foxmail.com>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Now EdgeMesh only support hostPort, uses should specify `hostPort` in the yaml like:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
  labels:
    app: nginx
spec:
  #replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx
        imagePullPolicy: IfNotPresent
        ports:
        - containerPort: 80
          hostPort: 80
```

If not, EdgeMesh can not do proxy.

**Does this PR introduce a user-facing change?**:

```release-note
EdgeMesh: support container level proxy
```

PR #1856, opened 2020-06-22
proposal: optimize EdgeMesh

daixiang0:

**What type of PR is this?**
/kind design

### DIFF
--- a/docs/proposals/edgemesh-optimization.md
+++ b/docs/proposals/edgemesh-optimization.md
@@ -1,0 +1,298 @@
+---
+title: EdgeMesh Optimization
+authors:
+    - "@daixiang0"
+approvers:
+creation-date: 2020-06-22
+last-updated: 2020-08-18
+status:
+---
+
+# EdgeMesh Optimization
+
+## Background
+
+Now EdgeMesh works as a DNS and proxy server with some limitations:
+
+- Depend on docker0 network interface, cannot work for other CRIs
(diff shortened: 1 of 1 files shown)


docs/proposals/edgemesh-optimization.md
```diff
@@ -0,0 +1,81 @@
...
+
+We want EdgeMesh as lightweight as possible and support more:
+1. Support more CRIs like docker, containerd, cri-o, etc.
+2. Include more DNS resolver support as in kubernetes coreDNS
+3. Include more protocols support as in kubernetes kube-proxy
+4. Include more types support as in kubernetes kube-proxy
+
+### Use Cases
+
+* Any apps in edge side can commuinicate with each other by svc in case offline with cloud side
```
benjaminhuo, line 54: Sorry, I'm a bit confused about this statement, it means:  any apps in the edge can communicate with each other by svc when edge nodes these apps running on are all disconnected with cloud side?

daixiang0: Yep, EdgeMesh provides a DNS cache to make apps running when edge nodes offline. It would be helpful like app a and b need to communicate with each other but the node which app b run has network issue with cloud side.

daixiang0:
There are some changes implemented, low level design may be simple about those:

- DNS server Alpha version: #1861#1840
- proxy server Alpha version: #1846

PR #1861, opened 2020-06-22
EdgeMesh: remove docker0 dependence

daixiang0:

Now EdgeMesh bind docker0 to work as a DNS server, this pr is to remove this dependence.

The proxy part of EdgeMesh still depends on docker0 network interface, which would be optimized after related proposal confirmed.

### DIFF
--- a/edgemesh/pkg/dns/dns.go
+++ b/edgemesh/pkg/dns/dns.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
 	"strings"
 	"time"
 	"unsafe"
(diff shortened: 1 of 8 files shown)


edgemesh/pkg/dns/dns.go
```diff
@@ -77,29 +77,47 @@ var metaClient client.CoreInterface
 // dnsConn saves DNS protocol
 var dnsConn *net.UDPConn
 
+//Init creates network interface
+func Init(addr string) {
+	// create a network interface
+	checkInterface := fmt.Sprintf("brctl show | grep %s 2>/dev/null", ifi)
```
jimmy-wei, line 83: Is it possible to use [netlink ](https://github.com/vishvananda/netlink) instead of system command?

Poorunga:
@daixiang0 This is a great PR, would you like to move this design to the edgemesh new repo: https://github.com/kubeedge/edgemesh

PR #1862, opened 2020-06-22
add clean namespace in keadm reset

ls889:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind cleanup
/kind bug

when we use `keadm reset`, it can automatically clean the kubeedge namespace. It can reduce some problems about certificates when users use keadm reset.
we should use keadm like `keadm reset --kube-config $HOME/.kube/config`.
I will update the docs of keadm after the pr was merged.

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
+++ b/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
@@ -181,6 +181,10 @@ func (cu *KubeCloudInstTool) TearDown() error {
 
 	//Kill cloudcore process
 	cu.KillKubeEdgeBinary(KubeCloudBinaryName)
-
+	// clean kubeedge namespace
+	err := cu.cleanNameSpace("kubeedge", cu.KubeConfig)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
```diff
@@ -181,6 +181,8 @@ func (cu *KubeCloudInstTool) TearDown() error {
 
 	//Kill cloudcore process
 	cu.KillKubeEdgeBinary(KubeCloudBinaryName)
+	// clean kubeedge namespace
+	cu.cleanNameSpace("kubeedge", cu.KubeConfig)
```
fisherxu, line 185: Could we return the error here? And also return in https://github.com/kubeedge/kubeedge/blob/master/keadm/cmd/keadm/app/cmd/reset.go#L88

ls889: @fisherxu  Done, thanks for reminding.

Issue #1068, opened 2019-08-23
make cross_build faild on centos


kadisi:
**What happened**:
when i make cross build edgecore on centos env. it faild:

```
# make cross_build
export GOARCH=arm \
        export GOOS="linux" \
        export GOARM=6 \
        export CGO_ENABLED=1 \
        export CC=arm-linux-gnueabi-gcc; \
        go build cmd/edgecore/edgecore.go
# runtime/cgo
exec: "arm-linux-gnueabi-gcc": executable file not found in $PATH
make: *** [cross_build] Error 2

```

`arm-linux-gnueabi-gcc` only works on ubuntu.  

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- KubeEdge version:
- Hardware configuration:
- OS (e.g. from /etc/os-release): 
```
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

- Kernel (e.g. `uname -a`):
- Others:

kadisi:
cc @fisherxu @subpathdev @sids-b @kevin-wangzefeng

kadisi:
@subpathdev , after i install arm-gcc in centos, and change CC option from `arm-linux-gnueabi-gcc` to `arm-linux-gnu-gcc`, it still output other error:
```
# yum install gcc-arm-linux-gnu -y
# make cross_build
export GOARCH=arm \
        export GOOS="linux" \
        export GOARM=6 \
        export CGO_ENABLED=1 \
        export CC=arm-linux-gnu-gcc ;\
        go build cmd/edgecore/edgecore.go
# runtime/cgo
_cgo_export.c:3:20: fatal error: stdlib.h: No such file or directory
 #include <stdlib.h>
                    ^
compilation terminated.
make: *** [cross_build] Error 2


```

[...]

PR #1868, opened 2020-06-23
Optimize binary build process

daixiang0:

Optimize binary build:
- move cross build into container
- combine crossbuild and smallbuild logic into one, reduce duplicate codes
- support armv8 for more binary

Fixed:  #1068 #1872


.travis.yml
```diff
@@ -39,17 +39,12 @@ jobs:
       arch: amd64
       before_script:
         - sudo apt-get install upx-ucl -y
-        - sudo apt-get install gcc-aarch64-linux-gnu -y
-        - sudo apt-get install libc6-dev-arm64-cross -y
-        - sudo apt-get install gcc-arm-linux-gnueabi -y
-        - sudo apt-get install libc6-dev-armel-cross -y
```
chendave, line 45: why? iirc, this is for crossing build on amd64

daixiang0: Now make build in container, no need to install them at host.

daixiang0: I mentioned it in the pr descrption.

.travis.yml
```diff
@@ -39,17 +39,12 @@ jobs:
...
         - export GOFLAGS=-mod=vendor
       script:
         - make
         - make bluetoothdevice
         - make smallbuild
-        - make crossbuild
-        - make crossbuild GOARM=GOARM7
```
chendave, line 52: crossbuild should not be deleted, see the name of job @Line38

Makefile
```diff
@@ -151,26 +152,27 @@ crossbuild:
...
 	edgesite
 define SMALLBUILD_HELP_INFO
-# small build components.
+# Small build components.
 #
 # Args:
-#   WHAT: Component names to be lint check. support: $(SMALLBUILD_COMPONENTS)
-#         If not specified, "everything" will be small build.
+#   WHAT: binary name to build. support: $(BINARIES)
+#         the build will produce executable files under $(OUT_DIR)/$$ARCH
```
chendave, line 165: $$ARCH or $ARCH?

Makefile
```diff
@@ -151,26 +152,27 @@ crossbuild:
...
 SMALLBUILD_COMPONENTS=edgecore \
 	edgesite
 define SMALLBUILD_HELP_INFO
-# small build components.
+# Small build components.
 #
 # Args:
-#   WHAT: Component names to be lint check. support: $(SMALLBUILD_COMPONENTS)
-#         If not specified, "everything" will be small build.
+#   WHAT: binary name to build. support: $(BINARIES)
```
chendave, line 164: should be careful here, need to make sure smallbuild works with all those binaries, this used to be "edgecore" and  "edgesite" only.

Makefile
```diff
@@ -179,7 +181,7 @@ smallbuild:
 	@echo "$$SMALLBUILD_HELP_INFO"
 else
 smallbuild: clean
-	hack/make-rules/smallbuild.sh $(WHAT)
+	ENABLE_SMALLBUILD=YES ARCH=$(ARCH) hack/make-rules/crossbuild.sh $(WHAT)
```
chendave, line 184: I still see "hack/make-rules/smallbuild.sh" and "hack/make-rules/crossbuild.sh" in place, and is run with different ways.

Makefile
daixiang0: In fact smallbuild is a pack process, it works for all binaries.

daixiang0:
@chendave thanks for reveiw and sorry for make it hard review, also it is hard to split for me since i start refactor from high level, to be honestly, i have spent 3 times spliting them as long as i spent on the first pr.

Line38:
Please remove my mention from this where you specified an error:

With (@)Line38

.travis.yml
daixiang0: Smallbuild havs cover all logics that crossbuild does, why need to re-test them?

Makefile
daixiang0: It is to print "$ARCH" in the help info.

Makefile
daixiang0: Forget to remove "hack/make-rules/smallbuild.sh", now smallbuild reuse crossbuild.
You remind me, it is better for smallbuild to reuse build logic rather than crossbuild.

.travis.yml
fisherxu: Smallbuild is used to compresses the binary(for amd64 and arm) and crossbuild is used to build for arm, they should be different.

daixiang0: Why need to add limit for arch? We can compress the binary for all arches(amd64, arm, arm64).

daixiang0:
Since small build would do cross build then simallize, why still need cross build test?

daixiang0:
ping @fisherxu

PR #1885, opened 2020-07-03
update GetNodeLabels

### LINKED ISSUE #1875: kubeedge1.3.0 edge can not use local pv
storageclass yaml:
image
pv yaml:
image
pvc yaml:
image
pod yaml:
image

this is node informations:  
image

**compute node that head string is node
random string is edgenode

pod  nodeselector points edgenode   It's in Pending all the time**

this is edgecore informations:
 
image

guojianyu:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
>`/kind feature

Fixes #1875

```release-note

```

### DIFF
--- a/edge/pkg/edged/volume_host.go
+++ b/edge/pkg/edged/volume_host.go
@@ -138,8 +138,14 @@ func (evh *edgedVolumeHost) GetConfigMapFunc() func(namespace, name string) (*ap
 func (evh *edgedVolumeHost) GetExec(pluginName string) utilexec.Interface  { return nil }
 func (evh *edgedVolumeHost) GetHostIP() (net.IP, error)                    { return nil, nil }
 func (evh *edgedVolumeHost) GetNodeAllocatable() (api.ResourceList, error) { return nil, nil }
-func (evh *edgedVolumeHost) GetNodeLabels() (map[string]string, error)     { return nil, nil }
-func (evh *edgedVolumeHost) GetNodeName() types.NodeName                   { return types.NodeName(evh.edge.nodeName) }
+func (evh *edgedVolumeHost) GetNodeLabels() (map[string]string, error) {
+	node, err := evh.edge.initialNode()
+	if err != nil {
+		return nil, fmt.Errorf("error retrieving node: %v", err)
+	}
+	return node.Labels, nil
+}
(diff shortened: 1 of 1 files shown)


edge/pkg/edged/volume_host.go
```diff
@@ -138,8 +137,18 @@ func (evh *edgedVolumeHost) GetConfigMapFunc() func(namespace, name string) (*ap
...
+func (evh *edgedVolumeHost) GetNodeLabels() (map[string]string, error) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		klog.Errorf("couldn't determine hostname: %v", err)
+		return nil, err
+	}
+	if len(evh.edge.nodeName) != 0 {
+		hostname = evh.edge.nodeName
+	}
+	return map[string]string{api.LabelHostname: hostname}, nil
```
fisherxu, line 149: thanks for the fixing, how about using `evh.edge.initialNode()`, it will return all the node info, also the labels.

```
node, err := evh.edge.initialNode()
if err != nil {
	return nil, fmt.Errorf("error retrieving node: %v", err)
}
return node.Labels, nil
```

edge/pkg/edged/volume_host.go
```diff
@@ -44,6 +41,8 @@ import (
 	"k8s.io/kubernetes/pkg/volume/util/subpath"
 	utilexec "k8s.io/utils/exec"
 	"k8s.io/utils/mount"
+	"net"
+	"os"
```
fisherxu, line 45: Don't modify the import here :)

edge/pkg/edged/volume_host.go
guojianyu: thanks,get it.Before it was rude.

edge/pkg/edged/volume_host.go
guojianyu: get it.

edge/pkg/edged/volume_host.go
fisherxu: Yes... In future we can get node info from the local cache. And could you please squash the commits?

guojianyu: ok.

PR #1886, opened 2020-07-03
Add containerLogsDir to store legacy symbolic

### LINKED ISSUE #1883: Failed to create legacy symbolic
**What happened**:
edgecore continuely logs `Failed to create legacy symbolic link` Error for unknown reason.
image
stop logging that error and figure out why.
run edgecore

**Environment**:

Poor12:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind bug

Fixes #1883 

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -131,7 +131,9 @@ const (
 	imageGcHighThreshold = "edged.image-gc-high-threshold"
 	syncMsgRespTimeout   = 1 * time.Minute
 	//DefaultRootDir give default directory
-	DefaultRootDir                   = "/var/lib/edged"
+	DefaultRootDir = "/var/lib/edged"
+	// ContainerLogsDir is the location of container logs.
+	ContainerLogsDir                 = "/var/log/containers"
 	workerResyncIntervalJitterFactor = 0.5
 	//EdgeController gives controller name
 	EdgeController = "edgecontroller"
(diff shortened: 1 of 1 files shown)


edge/pkg/edged/edged.go
```diff
@@ -680,7 +682,12 @@ func (e *edged) initializeModules() error {
 		klog.Errorf("Failed to initialNode %v", err)
 		return err
 	}
-
+	// If the container logs directory does not exist, create it.
+	if _, err := os.Stat(ContainerLogsDir); err != nil {
+		if err := e.os.MkdirAll(ContainerLogsDir, 0755); err != nil {
+			klog.Errorf("Failed to create directory %q: %v", ContainerLogsDir, err)
```
daixiang0, line 688: Here %s is enough, see https://gobyexample.com/string-formatting.

fisherxu: Err is not string here, why use %s? I think %v is consistent with k8s

daixiang0: I mean no need to use %q.

Poor12: It seems the only difference is whether the output contains double quotes.k8s use %q, I think keeping the same is no problem.

PR #1887, opened 2020-07-03
Improve metrics connection

Poor12:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

/kind bug

Kubelet uses the chunked method to return Response when processing logs/exec requests, but does not use this method when requesting /stats. Edged uses the ordinary connection in the same way.

/cc @kadisi 

```release-note

```

### DIFF
--- a/cloud/pkg/cloudstream/streamserver.go
+++ b/cloud/pkg/cloudstream/streamserver.go
@@ -153,18 +153,11 @@ func (s *StreamServer) getMetrics(r *restful.Request, w *restful.Response) {
 		return
 	}
 
-	w.Header().Set("Transfer-Encoding", "chunked")
 	w.WriteHeader(http.StatusOK)
-
-	if _, ok := w.ResponseWriter.(http.Flusher); !ok {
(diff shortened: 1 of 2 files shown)


cloud/pkg/cloudstream/streamserver.go
```diff
@@ -153,18 +153,11 @@ func (s *StreamServer) getMetrics(r *restful.Request, w *restful.Response) {
 		return
 	}
 
-	w.Header().Set("Transfer-Encoding", "chunked")
 	w.WriteHeader(http.StatusOK)
```
kadisi, line 156: need statusok？

Poor12: yes，kubelet actually call the writeAsJson function，which passes the status_ok parameter

kadisi:
/lgtm 

thanks @Poor12 

cc @fisherxu @kevin-wangzefeng

PR #1894, opened 2020-07-06
Edged: simplify umount logic

### LINKED ISSUE #1893: Failed to clearing up volume mount points
**What happened**:
```
.. edged_volumes.go:103] Failed to clearing up volume mount points of pod "9d3c2d6c-4089-4b45-bfab-0489c0c82399": read /var/lib/edged/pods/9d3c2d6c-4089-4b45-bfab-0489c0c82399/volumes: is a directory.
```

Clear up successfully.

**Environment**:
- KubeEdge version(e.g. `cloudcore/edgecore --version`):  1.3.1

daixiang0:

Old logic use `listProcMounts` in wrong way so always return error.

In fact, there is umount logic in each volume logic, no need to handle umount again.

**What type of PR is this?**
/kind bug

Fixes #1893

### DIFF
--- a/edge/pkg/edged/edged_volumes.go
+++ b/edge/pkg/edged/edged_volumes.go
@@ -22,9 +22,6 @@ package edged
 
 import (
 	"fmt"
-	"strconv"
-	"strings"
-
 	api "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -34,8 +31,13 @@ import (
 	"k8s.io/kubernetes/pkg/util/removeall"
 	"k8s.io/kubernetes/pkg/volume"
 	volumetypes "k8s.io/kubernetes/pkg/volume/util/types"
-	utilio "k8s.io/utils/io"
-	"k8s.io/utils/mount"
+)
+
+const (
+	procMountsPath = "/proc/mounts"
+	listTryTime    = 3
+	// mount info field length
+	expectedNumFieldsPerLine = 6
 )
 
 // newVolumeMounterFromPlugins attempts to find a plugin by volume spec, pod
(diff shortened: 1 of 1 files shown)


edge/pkg/edged/edged_volumes.go
```diff
@@ -167,8 +172,7 @@ func (e *edged) podVolumesExist(podUID types.UID) bool {
 
 // listProcMounts is come from k8s.io\kubernetes\pkg\util\mount.listProcMounts
 func listProcMounts(mountFilePath string) ([]mount.MountPoint, error) {
```
fisherxu, line 174: Could we import `ListProcMounts` from https://github.com/kubernetes/utils/blob/master/mount/mount_linux.go#L456?

daixiang0: I think those logic is to list all mount point under `/var/lib/edged/xxx/volumes` then umount, the k8s `ListProcMounts` is to list all mount points from `/proc/mounts`, the older logic wrongly use `/var/lib/edged/xxx/volumes` as `ListProcMounts` param and result error logs.

fisherxu: Need to think if we need cleanupMoutPoints, why k8s has no this func.

daixiang0: K8s check a dir is mount point or not based on `ListProcMounts`, the function is named as `CleanupMountPoint`, refer to https://github.com/kubernetes/utils/blob/master/mount/mount_helper_common.go#L26-L41.

daixiang0:
update

daixiang0:
ping @fisherxu

daixiang0:
@fisherxu mind help review?

daixiang0:
ping @fisherxu @kevin-wangzefeng

PR #1922, opened 2020-07-11
Use lock() and unlock() in DeletePod

### LINKED ISSUE #1920: Race condition in podManager
Hello, I'm reading the source of kubeedge. 
I saw the following code in the file "/edge/pkg/edged/podmanager.go" starting from line 107.
```
func (pm *podManager) DeletePod(pod *v1.Pod) {
	pm.lock.RLock()
	defer pm.lock.RUnlock()
	podFullName := container.GetPodFullName(pod)
	delete(pm.podByUID, pod.UID)
	delete(pm.podByFullName, podFullName)
}
```
Should it use `Lock()` and `Unlock()` instead of `RLock()` and `RUnlock()` to avoid race condition of  `podByUID` and `podByFullName` ?

Congrool:
`/kind bug`

**Modification:**
Use `Lock()` and `Unlock()` instead of `RLock()` and `RUnlock()`in the function `DeletePod`.

Fixes #1920 

Use `Lock()` and `Unlock()` to protect write operation on `pm.podByUID` and `pm.podByFullName`.

### DIFF
--- a/edge/pkg/edged/podmanager/pod_manager.go
+++ b/edge/pkg/edged/podmanager/pod_manager.go
@@ -23,7 +23,7 @@ package podmanager
 import (
 	"sync"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/kubernetes/pkg/kubelet/container"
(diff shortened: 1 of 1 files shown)


fisherxu:
/lgtm
/approve

Thanks for the quick fixing! @Congrool

PR #1923, opened 2020-07-11
Added JYLink to adopter

coversb:
Added Jingying Shuzhi Technology Co., Ltd to adopter

**What type of PR is this?**
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespaces from that line:
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
> /kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #1924, opened 2020-07-12
Fix issue 1905

### LINKED ISSUE #1905: Bad naming style
The name of functions, variables is not appropriate.
file: kubeedge/edge/pkg/devicetwin/dtmanager/membership.go
1. Function name needn't use past tense. Not only this function, but also many other function names should be refined.
Eg:
line 164
// Added add device to the edge group
func Added

2. Bad variable name. Here deviceModel -> deviceInstance. For deviceModel and deviceinstance are two definitions in device manager.
deviceExist  -> isDeviceExist.
line 	178:	deviceModel, deviceExist := context.GetDevice(device.ID)

**Why is this needed**:
Good coding style is the base.

dddddai:
**What type of PR is this?**
> /kind cleanup

1.Make variable and function names more reasonable

2.Fix some spelling mistakes

3.Delete useless comments.If the function name describes itself, there is no need to add a comment

Fixes #1905

No

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/device.go
+++ b/edge/pkg/devicetwin/dtmanager/device.go
@@ -59,7 +59,7 @@ func (dw DeviceWorker) Start() {
 
 func initDeviceActionCallBack() {
 	deviceActionCallBack = make(map[string]CallBack)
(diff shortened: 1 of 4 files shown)


GsssC:
This PR conflicts with the former PR #1923, you can add your suggestions to the former.

PR #1925, opened 2020-07-13
fix use certificate from local directory problem

threestoneliu:
**What type of PR is this?**
/kind bug

when use certificate from local dirctory，there have two problem
1. cloudcore won't create cloudcoresecrect
2. tunnel server cert/key file was not decode

Fixes #

```release-note

```


fisherxu:
cc @ls889 @XJangel

ls889:
@threestoneliu @fisherxu Thanks for pointing out and revising. it seems good

threestoneliu:
@ls889 In the cloudcore.yaml configuration file, is the tunnel server's certificate still allowed to be configured?

ls889:
@threestoneliu  yes,  tunnel server's certificate is allowed to be configured. and now we usually configure tunnel certs using the same certs with edgecore's.

PR #1929, opened 2020-07-14
Cloudcore: support helm deploy

### LINKED ISSUE #1076: Add helm charts and quick install yaml for cloud part(cloudcore&admision)

1. Add hem charts replace the yamls under build/cloud

2. Make the yamls under build/cloud and build/admission as one yaml file, generate it from helm chart.

example xref: [script for generate the yaml](https://github.com/volcano-sh/volcano/blob/master/hack/generate-yaml.sh),  [the generated yaml file](https://github.com/volcano-sh/volcano/blob/master/installer/volcano-development.yaml)
 
**Why is this needed**:
Make the install more easier and clear.

daixiang0:
Support use helm to deploy cloudcore.

**What type of PR is this?**
/kind feature

Fixes #1076

### DIFF
--- a/helm/cloudcore/crds/devices_v1alpha1_device.yaml
+++ b/helm/cloudcore/crds/devices_v1alpha1_device.yaml
@@ -1,0 +1,216 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: devices.devices.kubeedge.io
+spec:
+  group: devices.kubeedge.io
+  names:
+    kind: Device
+    plural: devices
+  scope: Namespaced
+  validation:
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          properties:
+            deviceModelRef:
+              description: 'Required: DeviceModelRef is reference to the device model
+                used as a template to create the device instance.'
+              type: object
+            nodeSelector:
+              description: NodeSelector indicates the binding preferences between
+                devices and nodes. Refer to k8s.io/kubernetes/pkg/apis/core NodeSelector
+                for more details
+              type: object
+            protocol:
+              description: 'Required: The protocol configuration used to connect to
+                the device.'
+              properties:
+                bluetooth:
+                  description: Protocol configuration for bluetooth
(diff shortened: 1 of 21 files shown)


helm/cloudcore/templates/configmap.yaml
```diff
@@ -0,0 +1,27 @@
...
+data:
+  cloudcore.yaml: |
+    apiVersion: cloudcore.config.kubeedge.io/v1alpha1
+    kind: CloudCore
+    kubeAPIConfig:
+      kubeConfig: ""
+      master: ""
+    modules:
+      cloudHub:
+        nodeLimit: 10
```
adriangonz, line 17: Should the configmap also include the `modules.cloudHub.https` block to expose the service running in `10002`?

helm/cloudcore/templates/service.yaml
```diff
@@ -0,0 +1,18 @@
...
+kind: Service
+metadata:
+  namespace: kubeedge
+  name: {{ include "cloudcore.name" . }}
+  labels:
+    {{- include "cloudcore.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.service.port }}
```
adriangonz, line 12: Should the service expose both ports `10000` and `10002`?

helm/cloudcore/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+podAnnotations: {}
+
+podSecurityContext: {}
+
+securityContext: {}
+
+service:
+  create: false
+  # type: NodePort
```
adriangonz, line 25: We could set the default `service.type` to `ClusterIP`. This allows people to then choose whether they want to keep it internal or expose it through an ingress layer.

helm/cloudcore/crds/devices_v1alpha1_device.yaml
```diff
@@ -0,0 +1,216 @@
...
+                        type: string
+                    required:
+                      - value
+                    type: object
+                required:
+                  - propertyName
+                type: object
+              type: array
+          type: object
+  version: v1alpha1
```
adriangonz, line 216: Should the template include the `v1alpha2` version of the `Device` and `DeviceModel` CRDs? Although these ones only seem to work with `v1.4.0` of `cloudcore`.

adriangonz (commented):
This looks great @daixiang0! I was just starting to work on creating a Helm chart to do the same thing, so I've added some comments based on my initial exploration. It would be great to hear your thoughts!

helm/cloudcore/values.yaml
Geethree: Agreed here. NodePort can prove difficult to in practice in hosted versions of k8s backed by an elastic pool of nodes.

helm/cloudcore/crds/devices_v1alpha1_device.yaml
Geethree: Also it is important to know that helm.. does not support updating or deleting CRDs..

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

daixiang0: Let's comment it :)

siredmar:
Can anybodt tell me if this is in a useful state where i can at least try it out?

daixiang0:
update

vikborisov:
hey everyone!
Is this PR a live?

daixiang0:
Need more discuss

PR #1934, opened 2020-07-15
Fix the syntax issue of script for building edgemesh image

chendave:
`sh` is not always symlink to `bash` on some system, for example,
it's linked to `dash` on ubuntu18 while `dash` doesn't recognize
the parentheses after `func`.

Here is how the error looks like,
createImg.sh: 5: createImg.sh: Syntax error: "(" unexpected

This PR fix a typo btw.

**What type of PR is this?**
>

/kind bug

Fixes #

```release-note
NONE
```

### DIFF
--- a/edgemesh/tools/initContainer/createImg.sh
+++ b/edgemesh/tools/initContainer/createImg.sh
@@ -3,8 +3,8 @@
 echo 'create edgemesh init Container image'
 
 function usage() {
-	echo "execute 'sh createImg.sh [rpm | deb]' to create docker image"
-	echo "execute 'sh createImg.sh help for use help'"
+	echo "execute 'bash createImg.sh [rpm | deb]' to create docker image"
+	echo "execute 'bash createImg.sh help for use help'"
 }
 
 path="${1}"
@@ -28,6 +28,6 @@ if command -v docker > /dev/null 2>&1 ; then
 	# delete iptables script
 	rm ./edgemesh-iptables.sh
 else
-	echo 'the docker command is no found!!'
+	echo 'the docker command is not found!!'
 	exit 1
 fi


edgemesh/tools/initContainer/createImg.sh
```diff
@@ -3,8 +3,8 @@
 echo 'create edgemesh init Container image'
 
 function usage() {
-	echo "execute 'sh createImg.sh [rpm | deb]' to create docker image"
-	echo "execute 'sh createImg.sh help for use help'"
+	echo "execute 'bash createImg.sh [rpm | deb]' to create docker image"
+	echo "execute 'bash createImg.sh help for use help'"
```
fisherxu, line 7: What's the difference of them? :)

fisherxu: Oh, just read the PR description, small question, does the bash exist on all os?

chendave: verified on ubuntu18 and ubuntu20, not sure on other OS.

chendave: but we cannot rely on the how the symlink is created, it's more safe to run with "bash" explicitly.

daixiang0:
It confused me that we need to run EdgeMesh as a single instance?

chendave:
> It confused me that we need to run EdgeMesh as a single instance?

this is for build only.

daixiang0:
Build what? If do not need to run as a single instance, whole image build is useless.

chendave:
> Build what? If do not need to run as a single instance, whole image build is useless.

anyhow, I think this image "edgemesh_init" should be okay to build
https://docs.kubeedge.io/en/v1.2.1/guides/edgemesh_test_env_guide.html#how-to-test-end-to-end

btw, this is fixing the issue of the hint in the script.

daixiang0:
Latest [doc](https://docs.kubeedge.io/en/latest/contributing/edgemesh_guide.html) does not need it.

We should delete them in master bracne and update them in release branch.

chendave:
that's great, will have a try, thanks!

fisherxu:
/close
The file is no needed now.

PR #1940, opened 2020-07-17
bugfix: Process of bluetooth mapper scheduler seems wrong 

### LINKED ISSUE #1824: Process of bluetooth mapper scheduler seems wrong
code: kubeedge/mapper/bluetooth_mapper/scheduler/scheduler.go: line 81
issue 1:
actionExists will always be true once it enters  line 85; And it will break once the first action in schedule.Actions was deleted(line 83 mismatch).

issue 2:
No process when the scheduler was deleted.
bug1

sailorvii:
The action in one schedule can be more than one. "actionExists" will always be true once it enters line 85,
which is not expected. Morever, if one action cannot be found, other actions should be run.

**What type of PR is this?**
 /kind bug

The process is wrong.

Fixes #1824 

None

### DIFF
--- a/mappers/bluetooth_mapper/scheduler/scheduler.go
+++ b/mappers/bluetooth_mapper/scheduler/scheduler.go
@@ -24,8 +24,8 @@ import (
 
 	"k8s.io/klog"
 
-	"github.com/kubeedge/kubeedge/mappers/bluetooth_mapper/action_manager"
-	"github.com/kubeedge/kubeedge/mappers/bluetooth_mapper/data_converter"
+	actionmanager "github.com/kubeedge/kubeedge/mappers/bluetooth_mapper/action_manager"
+	dataconverter "github.com/kubeedge/kubeedge/mappers/bluetooth_mapper/data_converter"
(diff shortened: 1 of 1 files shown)


kadisi:
looks good.

cc @fisherxu @kevin-wangzefeng

mappers/bluetooth_mapper/scheduler/scheduler.go
```diff
@@ -95,8 +95,8 @@ func (schedule *Schedule) performScheduleOperation(actionManager []actionmanager
 			schedule.Interval = defaultEventFrequency
 		}
 		if !actionExists {
-			klog.Errorf("Action %s does not exist. Exiting from schedule !!!", actionName)
-			break
+			klog.Errorf("Action %s does not exist.", actionName)
+			continue
```
fisherxu, line 99: Thanks for the fixing, and what's the usage of `continue` here? No matter the actionExists  is true or false, the for will continue

sailorvii: Yes, for there's a sleep after each schedule. When there's no action, I thought the sleep is not needed.
BTW, move this schedule.Interval setting out of this for rotation.

PR #1950, opened 2020-07-20
Remove redundant call

### LINKED ISSUE #1947: Optimize the code
1. Should it use `RWMutex` instead of `Mutex` to avoid the contention between readings?
https://github.com/kubeedge/kubeedge/blob/5921a88695113f45cfda304072097caea80dd117/cloud/pkg/cloudstream/tunnelserver.go#L74-L79

2. It calls `Unlock()` twice,which seems redundant.
https://github.com/kubeedge/kubeedge/blob/2ca99daa5cb247858b28957992c288250fb63507/staging/src/github.com/kubeedge/viaduct/pkg/conn/ws.go#L186-L202

dddddai:
Ref #1947


daixiang0:
It cover 2 issue indeed, better to split. And as i commented in issue, need to dive it deeper.

dddddai:
> It cover 2 issue indeed, better to split.

Thank you, I would split it if it should use `RWMutex`.

PR #1965, opened 2020-07-23
Fix: print token with newline

ESWZY:
**What type of PR is this?**

/kind bug

For a more friendly user interface when get token.

Nope.

I just noticed when `keadm gettoken`, the next command line prompt begins with the end of token (which is difficult to get the token).

before

And I added a newline here:

after

```release-note
NONE
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cloud/gettoken.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/gettoken.go
@@ -25,6 +25,7 @@ to get the token.
 `
 )
 
+// NewGettoken gets the token for edge nodes to join the cluster
 func NewGettoken(out io.Writer, init *common.GettokenOptions) *cobra.Command {
 	if init == nil {
 		init = newGettokenOptions()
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/cloud/gettoken.go
```diff
@@ -74,7 +75,8 @@ func queryToken(namespace string, name string, kubeConfigPath string) ([]byte, e
 
 // showToken prints the token
 func showToken(data []byte, out io.Writer) error {
-	_, err := out.Write(data)
+	token := append(data, byte('\n'))
+	_, err := out.Write(token)
```
fisherxu, line 79: thanks for the fixing! @ESWZY 
how about `_, err := fmt.Fprintln(out, data)`?

ESWZY: Sounds great!

ESWZY:
@fisherxu Same effect as above.

PR #1982, opened 2020-07-25
Update vendor to Kubernetes 1.18

### LINKED ISSUE #1933: Update upstream dependency to K8s 1.18
Upstream K8s is 1.18, we need to update upstream dependency.

**Why is this needed**:
We need to update upstream dependency to avoid more issues when upgrade to 1.19 in the future.

dingyin:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test
/kind feature

Update vendor to Kubernetes 1.18

Fixes #1933 

```release-note

```


fisherxu:
Thanks for the PR, looks like the k8s verion is 1.18.0 and not 1.18.5?

zzxgzgz:
> Thanks for the PR, looks like the k8s verion is 1.18.0 and not 1.18.5?

Yes it is, I named my branch incorrectly. Sorry for the confusion.

fisherxu:
> Yes it is, I named my branch incorrectly. Sorry for the confusion.

I think we need to bump to the latest version(1.18.6), and please fix the ci issue.

dingyin:
> > Yes it is, I named my branch incorrectly. Sorry for the confusion.
> 
> I think we need to bump to the latest version(1.18.6), and please fix the ci issue.

Updated to 1.18.6, and fixed test.

cloud/pkg/csidriver/controllerserver.go
```diff
@@ -20,7 +20,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
-
```
fisherxu, line 23: Keep the empty line here :)

cloud/pkg/leaderelection/leaderelection.go
```diff
@@ -1,13 +1,15 @@
...
+	"context"
 	gocontext "context"
 	"encoding/json"
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"os"
 	"syscall"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+
```
fisherxu, line 12: delete the empty line

edge/pkg/edged/edged.go
```diff
@@ -698,6 +699,13 @@ func (e *edged) initializeModules() error {
 		return err
 	}
 
+	// start the CPU manager in the clcm
+	err = e.clcm.StartCPUManager(e.GetActivePods, edgedutil.NewSourcesReady(e.isInitPodReady), e.statusManager, e.runtimeService)
```
fisherxu, line 703: If we don't add `StartCPUManager` here, can the ci test pass? If ci can pass, maybe we should add the `feature` in another PR :)

hack/update-vendor-licenses.sh
```diff
@@ -202,6 +202,13 @@ for PACKAGE in $(go list -m -json all | jq -r .Path | sort -f); do
     echo "${PACKAGE} doesn't exist in ${DEPS_DIR}, skipping" >&2
     continue
   fi
+  if [[ "${PACKAGE}" = "sigs.k8s.io/structured-merge-diff" ]]; then
```
fisherxu, line 205: cc @kevin-wangzefeng @daixiang0

edge/pkg/edged/edged.go
zzxgzgz: If we don't add this function here, the CPU manager won't start, and one of its fields remains nil, which will lead to edgecore crashing. 

That field was not nil in previous k8s versions, but its implementation changed so that the field remains nil until the start function is called.

fisherxu: Ok, and can we move the `StartCPUManager` to Start func instead of `newedged`?

fisherxu:
@dingyin @zzxgzgz Thanks for the pr, and can you show the memory change after bump to 1.18？

zzxgzgz:
> @dingyin @zzxgzgz Thanks for the pr, and can you show the memory change after bump to 1.18？

How do I check it?

fisherxu:
> > @dingyin @zzxgzgz Thanks for the pr, and can you show the memory change after bump to 1.18？
> 
> How do I check it?

Run edgecore without this patch and with this patch, then use `top` to record the result :)

cloud/pkg/csidriver/controllerserver.go
dingyin: updated.

cloud/pkg/leaderelection/leaderelection.go
dingyin: updated.

edge/pkg/edged/edged.go
dingyin: fixed.

dingyin:
> @dingyin @zzxgzgz Thanks for the pr, and can you show the memory change after bump to 1.18？

@fisherxu Memory consumption is up from 0.4% to 0.5% in a 16GB box (which is roughly 65MB to 80MB).

PR #1984, opened 2020-07-25
Keadm: add test when use `keadm gettoken` in cmdline

daixiang0:


fisherxu:
Has any test? Just test on my env, it works fine with `fmt.Fprintln`.

daixiang0:
Update content, only foucs on the case call `keadm gettoken` in cmdline.

PR #1986, opened 2020-07-25
Fix wrong ContainerRuntimeType in display

### LINKED ISSUE #1972: Wrong Container Runtime Type in Display
**What happened**
On cloud side, using `kubectl get node -o wide` will see the container-runtime is `remote://18.6.1` while the runtimeType in edgecore.yaml is `docker`.
image1-2020-07-24-19-01-18
image2-2020-07-24-19-02-19

**What do you want to see**
The container-runtime should be docker as same as it's described in edgecore.yaml.

I think it's something about the code in function `getNodeInfo` which is in `/kubeedge/kubeedge/edge/pkg/edged/edged_status.go`.
At line 264, it simply output "remote://".

```
        nodeInfo.ContainerRuntimeVersion = fmt.Sprintf("remote://%s", runtimeVersion.String())

	nodeInfo.KernelVersion = kernel
	nodeInfo.OperatingSystem = runtime.GOOS
	nodeInfo.Architecture = runtime.GOARCH
	nodeInfo.KubeletVersion = e.version
	nodeInfo.OSImage = prettyName
	//nodeInfo.ContainerRuntimeVersion = fmt.Sprintf("docker://%s", runtimeVersion.String())
```

To ensure my assumption, I used the following line ` nodeInfo.ContainerRuntimeVersion = fmt.Sprintf("docker://%s", runtimeVersion.String())` instead. And rebuild the `edgecore`, ran it on the edge side. It worked.
image1-2020-07-24-19-13-02

However, I think it should not simply use "remote://" or "docker://". It should use  `containerRuntimeName` which is a member of edged and initialized as `RuntimeType`.
edged go - kubeedge-v1 3 1-linux-amd64 - Visual Studio Code  Unsupported  2020_7_24 19_17_23

Congrool:
**What type of PR is this?**
> /kind bug

Fixed the containerRuntimeType displayed on cloudside, which was always `remote` even if what actually ran on edge side is `docker`.  

Fixes #1972 

None

### DIFF
--- a/edge/pkg/edged/edged_status.go
+++ b/edge/pkg/edged/edged_status.go
@@ -261,14 +261,13 @@ func (e *edged) getNodeInfo() (v1.NodeSystemInfo, error) {
 	if err != nil {
 		return nodeInfo, err
 	}
-	nodeInfo.ContainerRuntimeVersion = fmt.Sprintf("remote://%s", runtimeVersion.String())
+	nodeInfo.ContainerRuntimeVersion = fmt.Sprintf("%s://%s", e.containerRuntimeName, runtimeVersion.String())
 
 	nodeInfo.KernelVersion = kernel
 	nodeInfo.OperatingSystem = runtime.GOOS
 	nodeInfo.Architecture = runtime.GOARCH
 	nodeInfo.KubeletVersion = e.version
 	nodeInfo.OSImage = prettyName
-	//nodeInfo.ContainerRuntimeVersion = fmt.Sprintf("docker://%s", runtimeVersion.String())
 
 	return nodeInfo, nil
 }


edge/pkg/edged/edged_status.go
```diff
@@ -261,14 +261,14 @@ func (e *edged) getNodeInfo() (v1.NodeSystemInfo, error) {
 	if err != nil {
 		return nodeInfo, err
 	}
-	nodeInfo.ContainerRuntimeVersion = fmt.Sprintf("remote://%s", runtimeVersion.String())
+	// It's better to call it ContainerRuntimeTypeAndVersion
```
fisherxu, line 264: Thanks for the fixing, and please delete this comment and squash the commits

Congrool: OK, I'll do that.

PR #1988, opened 2020-07-25
Bump up csidriver example image version

daixiang0:

**What type of PR is this?**
/kind bug

Cmd option has changed.

```
$ docker run -it kubeedge/csidriver:v1.1.0 -h | grep endpoint
      --endpoint string           CSI endpoint (default "unix:///csi/csi.sock")
      --kubeedgeendpoint string   kubeedge endpoint (default "unix:///kubeedge/kubeedge.sock")  <===
```

### DIFF
--- a/build/csisamples/hostpath/deploy/csicontroller.yaml
+++ b/build/csisamples/hostpath/deploy/csicontroller.yaml
@@ -126,7 +126,7 @@ spec:
           - mountPath: /csi
             name: csi-socket-dir
         - name: csi-driver
-          image: kubeedge/csidriver:v1.1.0
+          image: kubeedge/csidriver:v1.3.0
           imagePullPolicy: IfNotPresent
           args:
             - "--v=5"


build/csisamples/hostpath/deploy/csicontroller.yaml
```diff
@@ -132,7 +132,7 @@ spec:
             - "--v=5"
             - "--drivername=$(CSI_DRIVERNAME)"
             - "--endpoint=$(CSI_ENDPOINT)"
-            - "--kubeedge-endpoint=$(KUBEEDGE_ENDPOINT)"
+            - "--kubeedgeendpoint=$(KUBEEDGE_ENDPOINT)"
```
fisherxu, line 135: Seems we need to update the image version, ref https://github.com/kubeedge/kubeedge/blob/master/cloud/cmd/csidriver/app/options/options.go#L43

daixiang0: Done.

PR #1998, opened 2020-07-27
Keadm: add prompting for confirmation

### LINKED ISSUE #1992: Enhancement: add confirmation prompt when reset

Now, when we excute `keadm reset`, it will exit directly, without any confirmation prompt. Maybe we need a step for that.

**Why is this needed**:

To avoid a wrong input, or other reason not really want to reset, we need an intermediate step for confirmation. And it is also a best practice when we want to close or exit something.

For anyone who want to fix that, you can add something to `keadm/cmd/keadm/app/cmd/reset.go`.

daixiang0:
**What type of PR is this?**
/kind feature

Fixes #1992

```release-note
Keadm: add prompting for confirmation
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/reset.go
+++ b/keadm/cmd/keadm/app/cmd/reset.go
@@ -17,8 +17,11 @@ limitations under the License.
 package cmd
 
 import (
+	"bufio"
 	"fmt"
 	"io"
+	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"k8s.io/klog"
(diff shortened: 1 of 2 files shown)


keadm/cmd/keadm/app/cmd/reset.go
```diff
@@ -78,6 +81,18 @@ func NewKubeEdgeReset(out io.Writer, reset *types.ResetOptions) *cobra.Command {
...
+				fmt.Println("[reset] WARNING: Changes made to this host by 'keadm init' or 'keadm join' will be reverted.")
+				fmt.Print("[reset] Are you sure you want to proceed? [y/N]: ")
+				s := bufio.NewScanner(os.Stdin)
+				s.Scan()
+				if err := s.Err(); err != nil {
+					return err
+				}
+				if strings.ToLower(s.Text()) != "y" {
+					return fmt.Errorf("aborted reset operation")
+				}
```
fisherxu, line 94: Do we need to have some retry like 3 times here?

daixiang0: Similar action as `kubeadm`, just add one prompting for confirmation.

PR #2000, opened 2020-07-27
support list-watch proxy in edge side 

### LINKED ISSUE #1758: Support list-watch from edgecore for applications on the edge
## Why do we need it
Some applications running on the edge side need to connect to the k8s master through list-watch interface, but on the edge it cannot directly connect to the k8s master. So we need forward the list-watch requests to the k8s master through the cloud-side channel.

threestoneliu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

Support the list-watch request of the side application. When the apiserver is accessible, the list-watch request is forwarded to the cloud, and when it is not accessible, the local cache is used to construct the response  for the list-watch request.

Fixes #1758
Fixes #1907 

```release-note

```


GsssC:
: ) start to take a look

threestoneliu:
The current edgeproxy implementation is to directly forward the request of the side application to the apiserver. The apiserver address will be exposed in the configuration.

I have used client-go to build an informer for pod, service, secret, and confimap for testing, and the test results are basically in line with expectations.

edge/pkg/edgeproxy/checker/checker.go
```diff
@@ -45,8 +45,9 @@ func (h *healthzChecker) loop() {
 	healthzURL := strings.Join([]string{h.url, "healthz"}, "/")
 	intervalTicker := time.NewTicker(h.interval)
 	isHealthy := false
+	retryTimes := 3
```
dingyin, line 48: We'd better make it configurable.

threestoneliu: fixd

PR #2007, opened 2020-07-29
test: fix keadm test case missing issue

### LINKED ISSUE #2006: make test failed:    can't load package: named files must be .go files:xxx
**Which jobs are failing**:
unit test
make test
make test
**Reason for failure**:
when I add a unit test. Has nothing to do with the following path 
```
[root@centos-kubeedge kubeedge]# make test
hack/make-rules/clean.sh
hack/make-rules/test.sh 
running tests cases 
can't load package: named files must be .go files: ./cloud/pkg/cloudhub/common/model
make: *** [test] Error 1
```
I find the the reason in  hack/lib/golang.sh  kubeedge::golang::get_keadm_test_dirs()
when I try  to perform this step, not only get the directory, but also get the go file
```
[root@centos-kubeedge kubeedge]# find -L ./keadm \
    -name '*_test.go' -print | xargs -0n1 dirname | uniq
./keadm/cmd/keadm/app/cmd/debug/collect_test.go
./keadm/cmd/keadm/app/cmd/util
```

daixiang0:

**What type of PR is this?**
 /kind bug

Fixes #2006

### DIFF
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -371,8 +371,9 @@ kubeedge::golang::get_cloud_test_dirs() {
 kubeedge::golang::get_keadm_test_dirs() {
     cd ${KUBEEDGE_ROOT}
     findDirs=$(find -L ./keadm \
-	    -name '*_test.go' -print | xargs -0n1 dirname | uniq)
-    echo "${findDirs}"
+	    -name '*_test.go' -print | xargs -n1 dirname | uniq)
+    dirArray=(${findDirs// /})
+    echo "${dirArray[@]}"
 }
 
 kubeedge::golang::get_edge_test_dirs() {


shenkonghui:
> Signed-off-by: Xiang Dai [long0dai@foxmail.com](mailto:long0dai@foxmail.com)
> 
> **What type of PR is this?**
> /kind bug
> 
> **What this PR does / why we need it**:
> 
> **Which issue(s) this PR fixes**:
> 
> Fixes #2006

I tested it and there are still problems, I guess it should change "-print" to "-print0" like `kubeedge::golang::get_cloud_test_dirs()`

daixiang0:
@shenkonghui in fact this would fix first part of your issue: new added test files not cover.

PR #2011, opened 2020-07-30
Do not send SIGKILL to stop process

daixiang0:

**What type of PR is this?**
/kind bug

From [signal doc](https://golang.org/pkg/os/signal/#hdr-Types_of_signals)
> The signals SIGKILL and SIGSTOP may not be caught by a program, and therefore cannot be affected by this package.

To shutdown gracefully, we should not use SIGKILL any more.

### DIFF
--- a/edge/test/integration/scripts/execute.sh
+++ b/edge/test/integration/scripts/execute.sh
@@ -17,7 +17,7 @@
 TEST_DIR=$(dirname $(dirname "${BASH_SOURCE[0]}"))
 
 function cleanup() {
-  sudo pkill -9 edgecore || true
+  sudo pkill edgecore || true
   while true; do
     sleep 1
     ps faux | grep -q [e]dgecore || break
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/core.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/core.go
@@ -28,7 +28,7 @@ func StartModules() {
 // GracefulShutdown is if it gets the special signals it does modules cleanup
 func GracefulShutdown() {
 	c := make(chan os.Signal)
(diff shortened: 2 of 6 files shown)


edge/test/integration/scripts/execute.sh
```diff
@@ -17,7 +17,7 @@
 TEST_DIR=$(dirname $(dirname "${BASH_SOURCE[0]}"))
 
 function cleanup() {
-  sudo pkill -9 edgecore || true
```
luogangyi, line 20: We cannot catch SIGKILL, but we still can use SIGKILL to force shutdown a process.
So, is this necessary to remove "-9" here?

daixiang0: Since cannot catch SIGKILL, `reset` would do not graceful shutdown, then would do bad impact on next test.

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -28,7 +28,7 @@ func StartModules() {
 // GracefulShutdown is if it gets the special signals it does modules cleanup
 func GracefulShutdown() {
 	c := make(chan os.Signal)
-	signal.Notify(c, syscall.SIGINT, syscall.SIGHUP, syscall.SIGTERM, syscall.SIGKILL,
```
GsssC, line 31: I see someone say SIGKILL can not be caught by process. The original code here about `SIGKILL` is useless?

daixiang0: Yep, just see doc in this [comment](https://github.com/kubeedge/kubeedge/pull/2011#issue-458903971).

PR #2019, opened 2020-08-01
fixed goroutine leak, when ws connection reconnect, but fifo's channel no close.

### LINKED ISSUE #2018: goroutine leak
**What happened**:
The edge side will happen goroutine leak when the edgecore is often disconnected from the cloudhub communication

When the edgecore is disconnected from the cloudhub, and edge side's connection will reconnection, but it not close fifo's channel, and it will cause  goroutine leak

Disconnect a little bit more often, and then watch if the goroutine number is growing all the time.

**Environment**:
   All version

**CloudSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

**EdgeSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

loongify:
…l not close

**What type of PR is this?**
>
> /kind bug

fixed goroutine leak: when the edgecore is disconnected from the cloudhub, and edge side's connection will reconnection, but it not close fifo's channel, and it will cause goroutine leak

Fixes #2018

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/viaduct/pkg/fifo/msgfifo.go
+++ b/staging/src/github.com/kubeedge/viaduct/pkg/fifo/msgfifo.go
@@ -2,6 +2,7 @@ package fifo
 
 import (
 	"fmt"
+	"sync"
 
 	"k8s.io/klog"
 
@@ -10,7 +11,8 @@ import (
 )
 
 type MessageFifo struct {
-	fifo chan model.Message
+	fifo      chan model.Message
+	closeOnce sync.Once
 }
 
 // set the fifo capacity to MessageFiFoSizeMax
@@ -42,4 +44,10 @@ func (f *MessageFifo) Get(msg *model.Message) error {
 		return fmt.Errorf("the fifo is broken")
 	}
 	return nil
+}
+
+func (f *MessageFifo) Close() {
+	f.closeOnce.Do(func() {
+		close(f.fifo)
+	})
 }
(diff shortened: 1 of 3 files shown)


staging/src/github.com/kubeedge/viaduct/pkg/fifo/msgfifo.go
```diff
@@ -43,3 +45,9 @@ func (f *MessageFifo) Get(msg *model.Message) error {
 	}
 	return nil
 }
+
+func (f *MessageFifo) Close() {
+	f.closeOnce.Do(func() {
+		close(f.fifo)
```
fisherxu, line 51: Why do we only close once? Looks like the fifo will be recreated every time reconnect.

loongify: Only same object close once, two different object will be closed again.

fisherxu: Okey :), so for the same object, is it necessary to use once?

loongify: yes, closing a closed channel will panic, so the channel should be closed once

fisherxu: Make sense! And would you like to split the second commit to another pr?

loongify: ok, I will split it to another pr.

PR #2020, opened 2020-08-01
edgecore support not specifying ip address

chenchunxiu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind test
> /kind failing-test

/kind feature

Allow users to use the IP that can communicate with the cloud directly without specifying it

### DIFF
--- a/pkg/apis/componentconfig/edgecore/v1alpha1/validation/validation.go
+++ b/pkg/apis/componentconfig/edgecore/v1alpha1/validation/validation.go
@@ -22,6 +22,7 @@ import (
 	"path"
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/klog"
 
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha1"
 	utilvalidation "github.com/kubeedge/kubeedge/pkg/util/validation"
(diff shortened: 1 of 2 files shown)


pkg/apis/componentconfig/edgecore/v1alpha1/validation/validation.go
```diff
@@ -63,8 +64,7 @@ func ValidateModuleEdged(e v1alpha1.Edged) field.ErrorList {
 	}
 	allErrs := field.ErrorList{}
 	if e.NodeIP == "" {
-		allErrs = append(allErrs, field.Invalid(field.NewPath("NodeIp"), e.NodeIP,
-			"Need sed NodeIP"))
+		klog.Warningf("NodeIP is empty , use default ip which can connect to cloud.")
```
fisherxu, line 67: Thanks for the PR :) In fact, if users haven't specifed the NodeIP, edgecore will set it to local ip by default, so do we need to care it?

chenchunxiu: yes，edgecore will set it to local ip by default. 
so I removed the error "Need sed NodeIP", and fix the cloudstream to let  it get the default local correctly

fisherxu: Sorry for the delay, sounds good, would you like squash the commits?

chenchunxiu: > Sorry for the delay, sounds good, would you like squash the commits?

done. 
 (thanks for the review and sorry for delay)

pkg/apis/componentconfig/edgecore/v1alpha1/validation/validation.go
```diff
@@ -18,6 +18,7 @@ package validation
 
 import (
 	"fmt"
+	"k8s.io/klog"
```
fisherxu, line 21: Please sorts the imports

chenchunxiu: > Please sorts the imports

done.

PR #2024, opened 2020-08-03
Use context to ensure all goroutines will return

### LINKED ISSUE #1974: An improvement when waiting for other goroutines
I see the following code in edgehub which used to wait other goroutines to return.
It simply uses `sleep()`  to wait other goroutines to return.
I don't think it's wise to work in this way, because which goroutine to run is up to the scheduler, and no assumption should be made in advance.
In the function `start()`, it runs 3 goroutines each loop, called `routeToEdge`, `routeToCloud` and `keepalive`.
image

When some of the three goroutines fail to receive or send msg from or to cloudhub, it will unblock and continue. 
```
               // part of routeToCloud
                err = eh.sendToCloud(message)
		if err != nil {
			klog.Errorf("failed to send message to cloud: %v", err)
			eh.reconnectChan <- struct{}{}
			return
		}
```
```
                // part of routeToEdge
		message, err := eh.chClient.Receive()
		if err != nil {
			klog.Errorf("websocket read error: %v", err)
			eh.reconnectChan <- struct{}{}
			return
		}
```
```
		// part of keepalive
		err := eh.sendToCloud(*msg)
		if err != nil {
			klog.Errorf("websocket write error: %v", err)
			eh.reconnectChan <- struct{}{}
			return
		}
		time.Sleep(time.Duration(config.Config.Heartbeat) * time.Second)
``` 
Then it will execute `Uninit()` to close the connection between edgehub and cloudhub. All of the rest goroutines will come into error and return.

The problem is around `clean`.
We can see that it uses `time.Sleep(time.Duration(config.Config.Heartbeat) * time.Second * 2)` to wait for a period before reconnection and wait all of the goroutines to return as well. 

However, there's no ensurance that all goroutines have returned after the execution of sleep. Just think what will happen when `sleep` has done while some goroutines, such as `routeToEdge`, are still blocking at `receive()` without any opportunity to run or being aware of the closure of the connection. In the worst case, it will result in goroutine leak.

I guess it waits for `time.Duration(config.Config.Heartbeat) * time.Second * 2` because `keepalive` will block at most `time.Duration(config.Config.Heartbeat) * time.Second` as in the above code.

It's better to use other mechanism to ensure that all the goroutines have returned, such as a counter.
It seems no problem currently, because the period of Heartbeat is 15s. It's long enough in most cases.

Congrool:
**What type of PR is this?**
> /kind bug

Imporve the mechanism in IPC. Use `context` and `sync.once` rather than simply waiting.
Use `context.Done` to ensure that all goroutines created in this loop will return, no matter whether it is before or after the next loop beginning.
Use `sync.once` with `errorNotifyChan` to notify parent goroutine that errors have occurred in son goroutines, ensuring that the error signal is sent only once for all goroutines created in that loop, thus there's no need to clean the channel. 
 
Fixes #1974 

None

### DIFF
--- a/common/constants/default.go
+++ b/common/constants/default.go
@@ -7,15 +7,11 @@ import (
 )
 
 const (
-	DefaultConfigDir    = "/etc/kubeedge/config/"
-	DefaultCAFile       = "/etc/kubeedge/ca/rootCA.crt"
-	DefaultCAKeyFile    = "/etc/kubeedge/ca/rootCA.key"
-	DefaultCertFile     = "/etc/kubeedge/certs/server.crt"
-	DefaultKeyFile      = "/etc/kubeedge/certs/server.key"
-	DefaultEdgeCertFile = "/etc/kubeedge/certs/edge.crt"
-	DefaultEdgeKeyFile  = "/etc/kubeedge/certs/edge.key"
-	DefaultCADir        = "/etc/kubeedge/ca"
-	DefaultCertDir      = "/etc/kubeedge/certs"
+	DefaultConfigDir = "/etc/kubeedge/config/"
+	DefaultCAFile    = "/etc/kubeedge/ca/rootCA.crt"
+	DefaultCAKeyFile = "/etc/kubeedge/ca/rootCA.key"
+	DefaultCertFile  = "/etc/kubeedge/certs/server.crt"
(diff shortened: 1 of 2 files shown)


daixiang0:
@Congrool please fix CI issue.

PR #2027, opened 2020-08-04
fix/keadm: fix the logic of auto install script for keadm

ttlv:
**What type of PR is this?**
/kind bug

complete keadm
https://github.com/kubeedge/kubeedge/issues/2057
Fixes #

```release-note

```


GsssC:
Hello, :-) plz complete the describtion of this pr

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -356,48 +360,22 @@ func installKubeEdge(componentType types.ComponentType, arch string, version str
...
-	} else {
-		try := 0
-		for ; try < downloadRetryTimes; try++ {
-			//Download the tar from repo
-			dwnldURL := fmt.Sprintf("cd %s && wget -k --no-check-certificate --progress=bar:force %s/v%s/%s",
-				KubeEdgePath, KubeEdgeDownloadURL, version, filename)
-			if _, err := runCommandWithShell(dwnldURL); err != nil {
+		fmt.Println("Expected or Default KubeEdge version", version, "is already downloaded and will checksum for it.")
+		if success, _ := checkSum(filename, checksumFilename, version); !success {
+			cmdStr := fmt.Sprintf("cd %s && rm -f %s", KubeEdgePath, filename)
```
GsssC, line 365: If fail to checkSum, I think it is better to get a user permission `y or yes`  then `remove file and reDownload`.
if no, log a warning and still use unSafe binaray to install kubeedge.

ttlv: I don't think user would like to continue to install  when uses konws the file is broken, but it's just a personal opinion, I have added selectone option for it  :-)

ttlv: ok, It seems that I have misunderstood your description,I have changed the logic of the code and if you have any ideas about the code.Welcome to have a talk with me .thank you :-)

PR #2046, opened 2020-08-06
fixed pod leak: when pod has been scheduled to another node, the pod should be deleted on edge side

loongify:
**What type of PR is this?**
>
> /kind bug

When pod has been scheduled to another node, the pod should be deleted on edge side


cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -1077,13 +1081,14 @@ func (uc *UpstreamController) queryNode() {
 	}
 }
 
-func (uc *UpstreamController) unmarshalPodStatusMessage(msg model.Message) (ns string, podStatuses []edgeapi.PodStatusRequest) {
+func (uc *UpstreamController) unmarshalPodStatusMessage(msg model.Message) (ns, nodeID string, podStatuses []edgeapi.PodStatusRequest) {
```
daixiang0, line 1084: Could rewrite this func with none-named value return?

loongify: Sorry, I don't get it, isn't it more clarify and clean up use named return parameters? Could you explain why you think should be rewriting it please.

fisherxu:
@loongify Thanks for the fixing :) Could you please add more details for the scene, and how to reproduce it? Maybe pod has been force deleted and still leave in one edgenode, then the cloudside schedule a new pod to another edgenode?

loongify:
> @loongify Thanks for the fixing :) Could you please add more details for the scene, and how to reproduce it? Maybe pod has been force deleted in cloud side and still leave in one edgenode, then the cloudside schedule a new pod to another edgenode?

For example, the cloudhub may lose communication with the edge side due to network reasons, and the cloudhub begins to evict the edge side container, but message can't dispatch to edge side due to network reasons and it will rescheduling the pod to other nodes. After a period of time, the network is restored and the connection beteen cloudhub and edgecore is re-established, then a similar situation will occur

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -224,13 +224,17 @@ func (uc *UpstreamController) updatePodStatus() {
 		case msg := <-uc.podStatusChan:
 			klog.Infof("message: %s, operation is: %s, and resource is: %s", msg.GetID(), msg.GetOperation(), msg.GetResource())
 
-			namespace, podStatuses := uc.unmarshalPodStatusMessage(msg)
+			namespace, nodeID, podStatuses := uc.unmarshalPodStatusMessage(msg)
```
fisherxu, line 227: sorry for the delay, sounds good, how about nodename instead of nodeID here?

loongify: fixed

fisherxu: Thanks, could you please squash the commits?

loongify: done

fisherxu: Using `git rebase` command to squash the 3 commits to 1.

PR #2053, opened 2020-08-10
delete unwanted mutex

elementyang:
Generally speaking, only data synchronization is needed, not function synchronization. So we need to remove unnecessary function synchronization.

### DIFF
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -120,9 +120,7 @@ func (eh *EdgeHub) routeToEdge() {
 }
 
 func (eh *EdgeHub) sendToCloud(message model.Message) error {
-	eh.keeperLock.Lock()
 	err := eh.chClient.Send(message)
-	eh.keeperLock.Unlock()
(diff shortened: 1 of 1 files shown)


fisherxu:
@elementyang Thanks for the fixing :) And `sendToCloud` is called by two goroutine `routeToCloud` and `keepalive`, if we have no the mutex here, edgecore will panic occasionally.

elementyang:
@fisherxu Thanks.
I known sendToCloud and eh.chClient.Send(message) are called by routeToCloud and keepalive. But sendToCloud and eh.chClient.Send() are function, and they should exist in every goroutine. So there is no need to synchronize functions, only need to synchronize data.

fisherxu:
With this patch, we need to test whether edgecore will panic in a little long period.

elementyang:
@fisherxu 
Thanks!

elementyang:
Generally speaking, only data synchronization is needed, not function synchronization. So we need to remove unnecessary function synchronization.
Please check it. 
Thanks.

PR #2058, opened 2020-08-12
fix edge message blocked: add timeout logic at SendResp func, should fail fast when system can't handle the message

loongify:
**What type of PR is this?**
> /kind bug

add timeout logic at SendResp func, should fail fast when system can't handle the message


staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
```diff
@@ -150,8 +150,18 @@ func (ctx *ChannelContext) SendResp(message model.Message) {
...
+
 	if channel, exist := ctx.anonChannels[anonName]; exist {
-		channel <- message
+		select {
+		case channel <- message:
+		case <-sendTimer.C:
+			klog.Warning("timeout when SendResp message")
+			return
+		}
+		sendTimer.Stop()
```
fisherxu, line 164: Thanks for the fixing, small question, Here the message just be written to the channel, will it be blocked?

loongify: Channel operations are blocking in nature, when some data is written to the channel, goroutine is blocked until some other goroutine reads it from that channel, so if no one reads it, it will be blocked.

luogangyi: SendSync already has a timer. Do we need another timer here?

Issue #2057, opened 2020-08-11
checksum when download release source


ttlv:
**What happened**:
There is a broken kubeedge-v1.3.1-linux-amd64.tar.gz in /etc/kubeedge,when I use kubeadm to join and got some error.
**What you expected to happen**:
keadm will checksum for download file when I use keadm to install. don't skip even though kubeedge-v1.3.1-linux-amd64.tar.gz is existed.!!
**How to reproduce it (as minimally and precisely as possible)**:
put a broken  kubeedge-v1.3.1-linux-amd64.tar.gz in your path and use keadm to join 
**Anything else we need to know?**:
no
**Environment**:
- KubeEdge version(e.g. `cloudcore/edgecore --version`):  
both are 1.3.1
**CloudSide Environment**:
- Hardware configuration (e.g. `lscpu`):
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    1
Core(s) per socket:    1
Socket(s):             4
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 79
Model name:            Intel(R) Xeon(R) CPU E5-2650 v4 @ 2.20GHz
Stepping:              1
CPU MHz:               2194.917
BogoMIPS:              4389.83
Hypervisor vendor:     VMware
Virtualization type:   full
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              30720K
NUMA node0 CPU(s):     0-3
Flags:                 fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts mmx fxsr sse sse2 ss syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts nopl xtopology tsc_reliable nonstop_tsc aperfmperf eagerfpu pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm 3dnowprefetch epb cat_l3 cdp_l3 fsgsbase smep cqm_llc cqm_occup_llc cqm_mbm_total cqm_mbm_local dtherm ida arat pln pts
- OS (e.g. `cat /etc/os-release`): 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
[...]

PR #2066, opened 2020-08-13
feat/keadm: add selectone options for checksum failed

ttlv:
**What type of PR is this?**
/kind bug
broken kubeedge release file is useless,lt's essential to checksum everytime even though checksum faile but we can add option for use to decide whether continue to install.
<!-- 

Fixes #
https://github.com/kubeedge/kubeedge/issues/2057

when user use keadm to join node,terminal will show a selectone option for user to choose whther contine if checksum fail


fisherxu:
The vendor files should not be modified  :)

PR #2074, opened 2020-08-14
fix/keadm: finish to do about checksum validation of the downloaded file every time

ttlv:
What type of PR is this?
/kind bug
Because of the bad network,when we use keadm tool to join node,It's esential to checksum everytime even though the release file is already existed and have the same name but the release is broken what's more,I think that providing a optional choose for user about whether continue to install is a better way. 
Which issue(s) this PR fixes:
Fixes #
#2057
no
uses will see some options in the terminal and just select one to continue.

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
 
 	types "github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
 )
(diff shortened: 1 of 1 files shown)


daixiang0:
Please use a more meaningful title first.

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -601,3 +592,92 @@ func hasSystemd() bool {
...
+	if try == downloadRetryTimes {
+		return fmt.Errorf("failed to download %s", filename)
+	}
+	return nil
+}
+
+func ask4confirm() (bool, error) {
+	var s string
+
+	fmt.Println("(yes/no): ")
```
daixiang0, line 667: Support y/N like `kubeadm` is enough.

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -601,3 +592,92 @@ func hasSystemd() bool {
...
+	_, err := fmt.Scan(&s)
+	if err != nil {
+		panic(err)
+	}
+
+	s = strings.TrimSpace(s)
+	s = strings.ToLower(s)
```
daixiang0, line 674: combine them.

keadm/cmd/keadm/app/cmd/util/common.go
ttlv: roger that

keadm/cmd/keadm/app/cmd/util/common.go
ttlv: yeah combine is better

daixiang0:
Please update pr title :)

/cc @fisherxu

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -345,49 +346,39 @@ func installKubeEdge(componentType types.ComponentType, arch string, version str
...
-			}
-
-			if desiredChecksum == actualChecksum {
-				break
-			} else {
-				fmt.Printf("Failed to verify the checksum of %s, try to download it again ... \n\n", filename)
-				//Cleanup the downloaded files
-				cmdStr = fmt.Sprintf("cd %s && rm -f %s", KubeEdgePath, filename)
-				_, err := runCommandWithStdout(cmdStr)
+		fmt.Println("Expected or Default KubeEdge version", version, "is already downloaded and will checksum for it.")
```
daixiang0, line 349: klog?

kevin-wangzefeng:
Please improve the PR name, we should not metion issue number PR name.

Please feel free to unhold when PR name fixed
/hold

keadm/cmd/keadm/app/cmd/util/common.go
ttlv: yeah ,I have used klog to replace all fmt

PR #2076, opened 2020-08-17
Optimize release note

daixiang0:

Now we still to look back on all merged pull requests and copy releate notes by hand, it would take some time.

After this pr, we will get a note then we can easily get all unreleased notes.

Also add a task list would be a tip both for contributors and reviewers.


daixiang0:
WDYT? @kevin-wangzefeng @fisherxu @chendave @dingyin  @GsssC

changelog.md
```diff
@@ -0,0 +1,18 @@
+# changelog
+
+All unreleased and notable changes to this project will be documented in this file.
+
+The format is like below:
```
chendave, line 5:
```suggestion
The format looks like below:
```

changelog.md
```diff
@@ -0,0 +1,18 @@
...
+<pr title> ([#<pr id>](<pr link>), [@<your github name>](<your github homepage link>))
```
chendave, line 6: This line is not rendered correctly, see: https://github.com/kubeedge/kubeedge/blob/c11266109adf97e050c208357ad64d6f16445d16/changelog.md

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -32,10 +32,16 @@ Fixes #
...
+If yes, you should add a release note to changelog.md:
+
+Please write follow this format:
+<pr title> ([#<pr id>](<pr link>), [@<your github name>](<your github homepage link>))
+
+For example:
+Add tree to store copy of dependency's license ([#1847](https://github.com/kubeedge/kubeedge/pull/1847), [@kevin-wangzefeng](https://github.com/kevin-wangzefeng))
+
+The pr title should start with an uppercase letter.
 -->
```
chendave, line 44: I am okay with this change, but all those comments are not shown the template,
see https://github.com/kubeedge/kubeedge/blob/c11266109adf97e050c208357ad64d6f16445d16/.github/PULL_REQUEST_TEMPLATE.md

daixiang0: I have updated format.

kevin-wangzefeng:
My concern is that this may result in changelog edited by to many people. What contributors need to care about is actually the release note content, while the PR number, URL and contributor link can be collected automatically.

So my question here is: do we want to automate the process, or we just ask contributors do more work.

daixiang0:
The final goal is automation i think.

Current release note content in the pr is hard to collect, we need to save it in one place to simpfiy collecting or we need to call GITHUB API which is not stable. After saving them in one place, we can automatically generate final release note.

The new line in changelog would be more obvious and easy to review, and it is a small work i think.

On the other hand, it means we ask them to participate in release process indeed.

kevin-wangzefeng:
Prow already has release-note plugin https://github.com/kubernetes/test-infra/blob/master/prow/plugins/releasenote/releasenote.go , we can simply enable it.

daixiang0:
This plugin only adds label, if we want automation, we still need to call GITHUB API to get release note using this label.

kevin-wangzefeng:
Checkout this: https://github.com/kubernetes/sig-release/blob/master/release-engineering/release-notes.md

daixiang0:
Cool!

Seems that we can enable plugin and introduce `krel`, that is enough!

fisherxu:
Let's use prow to collect the release note, issue #2096 has been created to track :)

PR #2093, opened 2020-08-20
Add the build option to make the binaries debuggable

chendave:
By default, the debug info will be stripped for the binaries, `cloudcore` for example, to keep the binary as small as possible.

This behavior can be overwritten by adding the options `GOLDFLAGS` or `GOGCFLAGS` when building the binaries.

As a example, the binary `cloudcore` built with the below command will have DWARF and symbol table kept, the optimizations and inline will also been disabled, this is helpful when you want to use the debugging tools like `delve`.

`make all WHAT=cloudcore GOLDFLAGS="" GOGCFLAGS="-N -l"`

**What type of PR is this?**
>

/kind feature

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,12 @@ define ALL_HELP_INFO
 #   make all
 #   make all HELP=y
 #   make all WHAT=cloudcore
+#   make all WHAT=cloudcore GOLDFLAGS="" GOGCFLAGS="-N -l"
+#     Note: Specify GOLDFLAGS as an empty string for building unstripped binaries, specify GOGCFLAGS
+#     to "-N -l" to disable optimizations and inlining, this will be helpful when you want to
+#     use the debugging tools like delve. When GOLDFLAGS is unspecified, it defaults to "-s -w" which strips
+#     debug information, see https://golang.org/cmd/link for other flags.
+
 endef
 .PHONY: all
 ifeq ($(HELP),y)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -25,6 +25,12 @@ define ALL_HELP_INFO
 #   make all
 #   make all HELP=y
 #   make all WHAT=cloudcore
+#   make all WHAT=cloudcore GOLDFLAGS="" GOGCFLAGS="-N -l"
```
daixiang0, line 28: It would be better to use as env val, like `GOGCFLAGS="-N -l" make all WHAT=cloudcore`.
Besides, empty val should be hide and considered as default.

chendave: > It would be better to use as env val, like GOGCFLAGS="-N -l" make all WHAT=cloudcore.

makefile will read it as env. so either one is okay to me.

>Besides, empty val should be hide and considered as default.

default config should strip the debug info to keep the binary as small as possible, and this is also align with what k8s does. pls see here, set it as empty string will overwrite the  default string.


https://github.com/kubernetes/kubernetes/blob/14a11060a0775ed609f0810898ebdbe737c59441/build/root/Makefile#L81-L85

kevin-wangzefeng:
@chendave could you add more information in the PR description on how contributors could use and benefit from this change.

PR #2108, opened 2020-08-27
edged support update pod status after consume added pod

### LINKED ISSUE #2107: can not get pod metric data after restart edgecore service
**What happened**:
call URL: 127.0.0.1:10350/stats/summary at edge side, the pods metric data is null:
image

pod metric data should be displayed

restart edgecore service

**Environment**:
1.3.0

lvchenggang:
What type of PR is this?
/kind bug

Once the EdgeCore service is restarted,  if not update pods status after pods consumed, we could not acquired pods metric data by call URL: 127.0.0.1:10350/stats/summary

Which issue(s) this PR fixes:
Fixes #2107

None

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -1206,6 +1206,10 @@ func (e *edged) handlePodListFromMetaManager(content []byte) (err error) {
 			return err
 		}
 		e.addPod(&pod)
+		if err = e.updatePodStatus(&pod); err != nil {
+			klog.Errorf("handlePodListFromMetaManager: update pod %s status error", pod.Name)
+			return err
+		}
 	}
 
 	return nil


daixiang0:
I test at local, stop then start edgecore still can get summary, could you share your restart steps?

lvchenggang:
> I test at local, stop then start edgecore still can get summary, could you share your restart steps?

restart command is: systemctl restart edgecore
    ps: the node's metric data after restart still exist after restart; the pod's metric data is null

daixiang0:
It could be better to update all pod status when edgecore start, or update in each add event would cost too much resource.

edge/pkg/edged/edged.go
```diff
@@ -957,6 +957,11 @@ func (e *edged) consumePodAddition(namespacedName *types.NamespacedName) error {
 
 	e.workQueue.Enqueue(pod.UID, utilwait.Jitter(time.Minute, workerResyncIntervalJitterFactor))
 	klog.Infof("consume added pod [%s] successfully\n", podName)
+	// update pod status after consume added pod, so the pod's metric data could be acquired
+	if err := e.updatePodStatus(pod); err != nil {
+		klog.Errorf("consumePodAddition: update pod %s status error", pod.Name)
+		return err
+	}
```
fisherxu, line 964: Thanks for the fixing! @lvchenggang I just tested in my env, really a bug. Now we only update podstatus from PLEG events. For pod from cloud, pleg can catch the runtime event and update pod status. 

But when edgecore restarted, edged will get all pods info from metamanager, Ref: [edge.go#L1032](https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edged/edged.go#L1032), and have no pleg events now, so I think we can update the podstatus at here.

lvchenggang: It is really better to update podstatus in handlePodListFromMetaManager function, and I tried in my env prove it can work

fisherxu:
/lgtm
/approve
Thanks for the fixing!

PR #2120, opened 2020-09-01
fix resource version compare

### LINKED ISSUE #2119: wrong resource version compare in synccontroller
**What happened**:
sync controller not works correctly in some condition when resource version like "1234" and sync version like "321"

sync controller works corrently.

condition like resource version ="1234" and sync version = "321"

GetResourceVersion returns string, we should no compare two string directly!

```
	if !isDeleteMessage(msg) {
		// If the message doesn't exist in the store, then compare it with
		// the version stored in the database
		if !exist {
			resourceNamespace, _ := edgemessagelayer.GetNamespace(*msg)
			resourceUID, err := GetMessageUID(*msg)
			if err != nil {
				klog.Errorf("fail to get message UID for message: %s", msg.Header.ID)
				return
			}

			objectSync, err := q.ObjectSyncController.ObjectSyncLister.ObjectSyncs(resourceNamespace).Get(synccontroller.BuildObjectSyncName(nodeID, resourceUID))
			if err == nil && msg.GetResourceVersion() <= objectSync.ResourceVersion {
				return
			}
		}

		// Check if message is older than already in store, if it is, discard it directly
		if exist {
			msgInStore := item.(*beehiveModel.Message)
			if msg.GetResourceVersion() <= msgInStore.GetResourceVersion() ||
				isDeleteMessage(msgInStore) {
				return
			}
		}
	}
```
**Environment**:
v1.4, 1.3

**CloudSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

**EdgeSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

luogangyi:
**What type of PR is this?**

sync controller not works correctly in some condition when resource version like "1234" and sync version like "321"

Fixes #2119 

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -107,7 +107,7 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
 			}
 
 			objectSync, err := q.ObjectSyncController.ObjectSyncLister.ObjectSyncs(resourceNamespace).Get(synccontroller.BuildObjectSyncName(nodeID, resourceUID))
-			if err == nil && msg.GetResourceVersion() <= objectSync.ResourceVersion {
+			if err == nil && synccontroller.CompareResourceVersion(msg.GetResourceVersion(), objectSync.ResourceVersion) <= 0 {
 				return
 			}
 		}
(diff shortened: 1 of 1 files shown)


fisherxu:
/lgtm
/approve
Looks like your email addr in the commit that don't exist in your github personal setting ;-)

GsssC:
Good and Thx ! we have ever found that configmap at edge sometime did not renewed, I think this pr may solve our problem!

PR #2125, opened 2020-09-03
add metrics api support in streamserver

### LINKED ISSUE #2106: Suggest support /metrics in stream server
Add "/metric" in stream server to support pulling  "/metrics" such as node_exporter and kubelet provided.

**Why is this needed**:
We need a way to monitor edge node and integrate with Promethus.

Currently, we have metric server. But metric server is not designed for external monitoring system.

AFAK, Promethus + node-exporter + kubelet is widely used in K8S. And adding "/metric" in existing streaming server seems not very hard. 

If you also like the idea, I am happy to implement this feature.

luogangyi:
**What type of PR is this?**

We need a way to monitor edge node and integrate with Promethus.

Fixes #2106

```release-note

```

### DIFF
--- a/cloud/pkg/cloudstream/containermetrics_connection.go
+++ b/cloud/pkg/cloudstream/containermetrics_connection.go
@@ -21,11 +21,11 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strings"
 
 	"k8s.io/klog"
 
 	"github.com/emicklei/go-restful"
-	"github.com/kubeedge/kubeedge/common/constants"
 	"github.com/kubeedge/kubeedge/pkg/stream"
 )
 
@@ -74,8 +74,9 @@ func (ms *ContainerMetricsConnection) SendConnection() (stream.EdgedConnection, 
 		URL:    *ms.r.Request.URL,
 		Header: ms.r.Request.Header,
 	}
+	targetPort := strings.Split(ms.r.Request.Host, ":")[1]
 	connector.URL.Scheme = "http"
-	connector.URL.Host = net.JoinHostPort("127.0.0.1", fmt.Sprintf("%v", constants.ServerPort))
+	connector.URL.Host = net.JoinHostPort("127.0.0.1", fmt.Sprintf("%v", targetPort))
 	m, err := connector.CreateConnectMessage()
 	if err != nil {
 		return nil, err
--- a/pkg/stream/edgedmetricsconnection.go
+++ b/pkg/stream/edgedmetricsconnection.go
@@ -62,6 +62,12 @@ func (ms *EdgedMetricsConnection) Serve(tunnel SafeWriteTunneler) error {
 		return err
 	}
 	req.Header = ms.Header
+	// Since current tunnel implementation only support Text message,
+	// we should force Accept-Encoding to identity to avoid any compression.
+	// For example, user may pass accept-encoding: gzip in header.
+	// TODO: luogangyi
+	// When we support binary message, we can remove this setting.
+	req.Header.Set("accept-encoding", "identity")
 	resp, err := client.Do(req)
 	if err != nil {
 		klog.Errorf("request metrics error %v", err)
--- a/cloud/pkg/cloudstream/streamserver.go
+++ b/cloud/pkg/cloudstream/streamserver.go
@@ -80,6 +80,15 @@ func (s *StreamServer) installDebugHandler() {
 	ws.Route(ws.GET("/{namespace}/{podName}/{uid}/{containerName}").
 		To(s.getMetrics))
 	s.container.Add(ws)
+
+	// metrics api is widely used for Promethus
+	ws = new(restful.WebService)
+	ws.Path("/metrics")
+	ws.Route(ws.GET("").
+		To(s.getMetrics))
+	ws.Route(ws.GET("/cadvisor").
+		To(s.getMetrics))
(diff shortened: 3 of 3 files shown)


cloud/pkg/cloudstream/containermetrics_connection.go
```diff
@@ -74,8 +74,9 @@ func (ms *ContainerMetricsConnection) SendConnection() (stream.EdgedConnection,
 		URL:    *ms.r.Request.URL,
 		Header: ms.r.Request.Header,
 	}
+	targetPort := strings.Split(ms.r.Request.Host, ":")[1]
 	connector.URL.Scheme = "http"
-	connector.URL.Host = net.JoinHostPort("127.0.0.1", fmt.Sprintf("%v", constants.ServerPort))
+	connector.URL.Host = net.JoinHostPort("127.0.0.1", fmt.Sprintf("%v", targetPort))
```
daixiang0, line 79: Still need to change edge streamserver config.

pkg/stream/edgedmetricsconnection.go
```diff
@@ -62,6 +62,12 @@ func (ms *EdgedMetricsConnection) Serve(tunnel SafeWriteTunneler) error {
 		return err
 	}
 	req.Header = ms.Header
+	// Since current tunnel implementation only support Text message,
+	// we should force Accept-Encoding to identity to avoid any compression.
+	// For example, user may pass accept-encoding: gzip in header.
+	// TODO: luogangyi
+	// When we support binary message, we can remove this setting.
+	req.Header.Set("accept-encoding", "identity")
```
daixiang0, line 70: User's request header is `ms.Header`, not need to specify again.

daixiang0 (changes requested):

cloud/pkg/cloudstream/containermetrics_connection.go
luogangyi: no need change  edge streamserver config.
For this implementation, we can support multiple ports.

Fox exmaple, if add two iptables rule like
```
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:10350 to:192.168.96.94:10003
DNAT       tcp  --  0.0.0.0/0            0.0.0.0/0            tcp dpt:9100 to:192.168.96.94:10003
```
then,
```
curl  -k  https://edgenode1:10350/metrics // will forward to  edgenode:10350
curl  -k  https://edgenode1:9100/metrics // will forward to  edgenode:9100 (that is node_exporter)
```

pkg/stream/edgedmetricsconnection.go
luogangyi: this is used to change user's header setting, because we are not support compressed body.
You can try
```
curl  -k -H "Accept-Encoding: gzip" https://edgenode1:10350/metrics|gunzip
```
it returns unexpected EOF

so, we have to force Accept-Encoding to identity

cloud/pkg/cloudstream/containermetrics_connection.go
daixiang0: The common process is that prometheus use SVC to connect node_exporter and scrape.

After we support SVC between edge and cloud, it still works.

So is it necessary?

/cc @fisherxu @kevin-wangzefeng

pkg/stream/edgedmetricsconnection.go
daixiang0: I think it is not a common usage, most users use prometheus directly.

Does k8s support it?

cloud/pkg/cloudstream/containermetrics_connection.go
luogangyi: "After we support SVC between edge and cloud, it still works." 
When can we support SVC between edge and cloud? Seems we still more than half a year :(

pkg/stream/edgedmetricsconnection.go
luogangyi: In fact, this problem is prometheus brings.
I find prometheus will add "Accept-Encoding: gzip" in request...

fisherxu:
/lgtm
/approve
@luogangyi Kindly reminder :) Your email addr in commits DON'T exist in your github personal setting. So github can't count your contributions

luogangyi:
> /lgtm
> /approve
> @luogangyi Kindly reminder :) Your email addr in commits DON'T exist in your github personal setting. So github can't count your contributions

Thanks for reminder :) I had add it back (I remove the address due to previous security censor)

pkg/stream/edgedmetricsconnection.go
donggangcj: > In fact, this problem is prometheus brings.
> I find prometheus will add "Accept-Encoding: gzip" in request...

Indeed it is，How to configure when use prometheus directly?

donggangcj: > In fact, this problem is prometheus brings.
> I find prometheus will add "Accept-Encoding: gzip" in request...

Indeed it is，How to configure when use prometheus directly?

OJumpyo: “Indeed it is，How to configure when use prometheus directly?” hi, has this done? i'm a fresh-hand in the prometheus and kubeedge area, i am wondering how to configure using prometheus to monitor edgenode as k8s worker node do. any help would appreciate.

PR #2130, opened 2020-09-04
support 2 device of multiple namespaces bind to same node

### LINKED ISSUE #2129: bug occur when 2 device of multiple namespaces bind to same node
**What happened**:
I create 2 device of multiple namespaces bind to same node, but cloudcore unable to create second-device configmap. What's more, it's unable to handle the first configmap.
Solve the bug.
Create 2 device of multiple namespaces bind to same node
According to the source code, it's a TODO job in cloud/pkg/devicecontroller/controller/downstream.go line183

**Environment**:
- KubeEdge version:  1.4.0release

wuqihui0317:
**What type of PR is this?**
kind bug

It's a TODO job of kubeedge.Now ,it's ubable to handle 2 device of multiple namespaces bind to same node.

Fixes #2129

NONE
```release-note

```

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -172,7 +172,7 @@ func (dc *DownstreamController) syncDevice() {
 
 // addToConfigMap adds device in the configmap
 func (dc *DownstreamController) addToConfigMap(device *v1alpha2.Device) {
-	configMap, ok := dc.configMapManager.ConfigMap.Load(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0])
+	configMap, ok := dc.configMapManager.ConfigMap.Load(device.Namespace + device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0])
 	if !ok {
 		nodeConfigMap := &v1.ConfigMap{}
 		nodeConfigMap.Kind = ConfigMapKind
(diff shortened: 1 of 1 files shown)


daixiang0:
Please simpify pr title, do not reply issue.

wuqihui0317:
ok，I have reedited my pr.





------------------ Original ------------------

wuqihui0317:
> Please simpify pr title, do not reply issue.

OK,I have reedited it.

luogangyi:
This change requires mappers changes accordingly. Please add some notice.

wuqihui0317:
> accordingly

get it.

luogangyi:
Another problem, How does mapper access configmap?
Previously, configmap name is static for each node, aka "device-profile-config-edgenode-sicp".
But in your patch, configmap name becomes dynamic, like "ns1-device-profile-config-edgenode-sicp". Namespaces can by created dynamicly, which means mappers deployment have to add new volume when new ns created. It seems complicated for mappers.

I suggest we still use one configmap, and add ns prefix to device instance name.

wuqihui0317:
> Another problem, How does mapper access configmap?
> Previously, configmap name is static for each node, aka "device-profile-config-edgenode-sicp".
> But in your patch, configmap name becomes dynamic, like "ns1-device-profile-config-edgenode-sicp". Namespaces can by created dynamicly, which means mappers deployment have to add new volume when new ns created. It seems complicated for mappers.
> 
> I suggest we still use one configmap, and add ns prefix to device instance name.
Sorry, there was something else yesterday.
In my view, there is two kinds of configmap. One is K8s's configmap, another is stored in kubeedge‘s object(dc.configMapManager.ConfigMap), which is changed by my code. So, I just changed the background logic, when we use it, we can only find one configmap in the cluster. 

How do you think about it?

well, I don't know where is the mapper's config from, the first or the second, I'll look source code again.

wuqihui0317:
> Another problem, How does mapper access configmap?
> Previously, configmap name is static for each node, aka "device-profile-config-edgenode-sicp".
> But in your patch, configmap name becomes dynamic, like "ns1-device-profile-config-edgenode-sicp". Namespaces can by created dynamicly, which means mappers deployment have to add new volume when new ns created. It seems complicated for mappers.
> 
> I suggest we still use one configmap, and add ns prefix to device instance name.

Sorry , I think I mistake it. It seems to be better to merge all configmaps in the background to store in dc.configMapManager.ConfigMap?
Anotherway, the bug is fatal, it makes this node's configmap unable to add and delete. Unless restart cloudcore.

luogangyi:
> > Another problem, How does mapper access configmap?
> > Previously, configmap name is static for each node, aka "device-profile-config-edgenode-sicp".
> > But in your patch, configmap name becomes dynamic, like "ns1-device-profile-config-edgenode-sicp". Namespaces can by created dynamicly, which means mappers deployment have to add new volume when new ns created. It seems complicated for mappers.
> > I suggest we still use one configmap, and add ns prefix to device instance name.
> 
> Sorry , I think I mistake it. It seems to be better to merge all configmaps in the background to store in dc.configMapManager.ConfigMap?
> On the other hand, the bug is fatal, it makes this node's configmap unable to add and delete. Unless restart cloudcore.

yes, I think it is better to maintain only one configmap for each edge node.
you can try to re-structure the device-profile configmap, maybe add namespace prefix to instance name? or put device instace under namespace ?

PR #2135, opened 2020-09-04
fix/keadm:  checksum validation of the downloaded file every time

ttlv:
What type of PR is this?
/kind bug
Because of the bad network,when we use keadm tool to join node,It's esential to checksum everytime even though the release file is already existed and have the same name but the release is broken what's more,I think that providing a optional choose for user about whether continue to install is a better way. 
Which issue(s) this PR fixes:
Fixes #
#2057
no
uses will see some options in the terminal and just select one to continue.

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -337,7 +337,7 @@ func installKubeEdge(componentType types.ComponentType, arch string, version sem
 	}
 
 	//Check if the same version exists, then skip the download and just checksum for it
-	//and if checksum failed,there will be a option to choose to continue to untar or quit.
+	//and if checksum failed, there will be a option to choose to continue to untar or quit.
 	//checksum available at download URL. So that both can be compared to see if
 	//proper download has happened and then only proceed further.
 	//Currently it is missing and once checksum is in place, checksum check required
@@ -351,12 +351,12 @@ func installKubeEdge(componentType types.ComponentType, arch string, version sem
 		if success, _ := checkSum(filename, checksumFilename, version); !success {
 			fmt.Printf("%v in your path checksum failed and do you want to delete this file and try to download again? \n", filename)
 			for {
-				result, err := ask4confirm()
+				confirm, err := ask4confirm()
 				if err != nil {
 					fmt.Println(err.Error())
 					continue
 				}
-				if result {
+				if confirm {
 					cmdStr := fmt.Sprintf("cd %s && rm -f %s", KubeEdgePath, filename)
 					if _, err := runCommandWithStdout(cmdStr); err != nil {
 						return err
@@ -367,8 +367,8 @@ func installKubeEdge(componentType types.ComponentType, arch string, version sem
 					}
 				} else {
 					klog.Warningf("failed to checksum and will continue to install.")
-					break
 				}
+				break
 			}
 		} else {
 			klog.Infof("Expected or Default KubeEdge version %v is already downloaded and checksum successfully.", version)
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"k8s.io/klog"
 	"os"
 	"os/exec"
 	"regexp"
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -335,9 +336,9 @@ func installKubeEdge(componentType types.ComponentType, arch string, version sem
 		arch = "arm"
 	}
 
-	//Check if the same version exists, then skip the download and just untar and continue
-	//TODO: It is always better to have the checksum validation of the downloaded file
-	//and checksum available at download URL. So that both can be compared to see if
+	//Check if the same version exists, then skip the download and just checksum for it
+	//and if checksum failed,there will be a option to choose to continue to untar or quit.
```
fisherxu, line 340:
```suggestion
	//and if checksum failed, there will be a option to choose to continue to untar or quit.
```
```suggestion
	//and if checksum failed,there will be a option to choose to continue to untar or quit.
```

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -346,49 +347,39 @@ func installKubeEdge(componentType types.ComponentType, arch string, version sem
...
+					fmt.Println(err.Error())
+					continue
+				}
+				if result {
+					cmdStr := fmt.Sprintf("cd %s && rm -f %s", KubeEdgePath, filename)
+					if _, err := runCommandWithStdout(cmdStr); err != nil {
+						return err
+					}
+					klog.Infof("%v have been deleted and will try to download again", filename)
+					if err := retryDownload(filename, checksumFilename, version); err != nil {
```
fisherxu, line 365: If retryDownload successfully, how to exit the for loop here?

fisherxu (commented):
Thanks for the enhancement, just some nits

keadm/cmd/keadm/app/cmd/util/common.go
ttlv: yeah,you are right.I have ignored this when retry download successfully, I have fix this logic error in next commit.Please have a look。Thanks♪(･ω･)ﾉ

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -595,3 +586,90 @@ func hasSystemd() bool {
...
+			break
+		}
+	}
+	if try == downloadRetryTimes {
+		return fmt.Errorf("failed to download %s", filename)
+	}
+	return nil
+}
+
+func ask4confirm() (bool, error) {
```
fisherxu, line 658: We'd better use `askForConfirm` name style :)

keadm/cmd/keadm/app/cmd/util/common.go
fisherxu: Looks good, thanks :)

keadm/cmd/keadm/app/cmd/util/common.go
fisherxu, line 658:
```suggestion
func askForConfirm() (bool, error) {
```

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -346,49 +347,39 @@ func installKubeEdge(componentType types.ComponentType, arch string, version sem
...
-			} else {
-				fmt.Printf("Failed to verify the checksum of %s, try to download it again ... \n\n", filename)
-				//Cleanup the downloaded files
-				cmdStr = fmt.Sprintf("cd %s && rm -f %s", KubeEdgePath, filename)
-				_, err := runCommandWithStdout(cmdStr)
+		fmt.Printf("Expected or Default KubeEdge version %v is already downloaded and will checksum for it. \n", version)
+		if success, _ := checkSum(filename, checksumFilename, version); !success {
+			fmt.Printf("%v in your path checksum failed and do you want to delete this file and try to download again? \n", filename)
+			for {
+				confirm, err := ask4confirm()
```
fisherxu, line 354:
```suggestion
				confirm, err := askForConfirm()
```

keadm/cmd/keadm/app/cmd/util/common.go
ttlv: I have fixed this in next commit .Thanks for your good advice :)

keadm/cmd/keadm/app/cmd/util/common.go
ttlv: I have fixed this in next commit .Thanks for your good advice :)

PR #2137, opened 2020-09-05
support 2 device of multiple namespaces bind to same node

### LINKED ISSUE #2129: bug occur when 2 device of multiple namespaces bind to same node

wuqihui0317:
…same node ?

**What type of PR is this?**
/kind bug

It's a TODO job of kubeedge.Now ,it's ubable to handle 2 device of multiple namespaces bind to same node.

Fixes #2129 

When we create 2 device of multiple namespaces bind to same node, KE will create two multiple namespaces configmap。 But now, KE doesn't support it. So cloudcore.log will report error, and this node's configmap will be in trouble.

Although we can store all devices in one namespace, sometimes we need to store a kind of devices in a namespace, like temperatureSensor.

What I do is to merge all device's config to a configmap, stored in "default" Namespace(as we usually use, it's ok to stored in another namespace, like "keConfigmap"). And we can create  devices of different namespaces.

None
```release-note

```

### DIFF
--- a/cloud/pkg/devicecontroller/constants/default.go
+++ b/cloud/pkg/devicecontroller/constants/default.go
@@ -13,6 +13,8 @@ const (
 
 	DefaultMessageLayer = "context"
 
+	ConfigmapNamespace = "default"
+
 	DefaultContextSendModuleName     = CloudHubControllerModuleName
 	DefaultContextReceiveModuleName  = DeviceControllerModuleName
 	DefaultContextResponseModuleName = CloudHubControllerModuleName
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -178,21 +178,20 @@ func (dc *DownstreamController) addToConfigMap(device *v1alpha2.Device) {
 		nodeConfigMap.Kind = ConfigMapKind
 		nodeConfigMap.APIVersion = ConfigMapVersion
 		nodeConfigMap.Name = DeviceProfileConfigPrefix + device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0]
-		nodeConfigMap.Namespace = device.Namespace
+		nodeConfigMap.Namespace = "default"
 		nodeConfigMap.Data = make(map[string]string)
-		// TODO: how to handle 2 device of multiple namespaces bind to same node ?
 		dc.addDeviceProfile(device, nodeConfigMap)
 		// store new config map
 		dc.configMapManager.ConfigMap.Store(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0], nodeConfigMap)
 
-		if _, err := dc.kubeClient.CoreV1().ConfigMaps(device.Namespace).Get(context.Background(), nodeConfigMap.Name, metav1.GetOptions{}); err != nil {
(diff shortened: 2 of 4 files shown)


luogangyi:
lgtm

fisherxu:
@wuqihui0317  Thanks for your fixing! Just small question, KubeEdge now can't create two configmap for two device in different ns?

wuqihui0317:
> @wuqihui0317 Thanks for your fixing! Just small question, KubeEdge now can't create two configmap for two device in different ns?

Yes, you can try to create two device of different ns in a node , the configmap won't work until restart cloudcore.

fisherxu:
> > @wuqihui0317 Thanks for your fixing! Just small question, KubeEdge now can't create two configmap for two device in different ns?
> 
> Yes, you can try to create two device of different ns in a node , the configmap won't work until restart cloudcore.

Okey, when we create all the configmaps in default ns, how does the mapper in other ns use the conmfigmap?

luogangyi:
> > > @wuqihui0317 Thanks for your fixing! Just small question, KubeEdge now can't create two configmap for two device in different ns?
> > 
> > 
> > Yes, you can try to create two device of different ns in a node , the configmap won't work until restart cloudcore.
> 
> Okey, when we create all the configmaps in default ns, how does the mapper in other ns use the conmfigmap?

I don't think we can support running mapper in ns so far. 
All device twin code are not consider multiple ns situation. For example, if we craete two device with same in different ns, device twin module cannot differentiate them.

wuqihui0317:
> > 
> > Okey, when we create all the configmaps in default ns, how does the mapper in other ns use the conmfigmap?
> 
> I don't think we can support running mapper in ns so far.
> All device twin code are not consider multiple ns situation. For example, if we craete two device with same in different ns, device twin module cannot differentiate them.

As my superficial understanding in kubeedge, kubeedge is samplified from k8s. And the part of devicecontroller of ke abandons namespace totally. Because in the devicecontroller/controller/downstream, which is watching the changes of device and devicemodel, device and devicemodel is stored by map(key is the corresponding name) .

`dc.deviceModelManager.DeviceModel.Store(deviceModel.Name, deviceModel)`
`dc.deviceManager.Device.Store(device.Name, device)`
`dc.configMapManager.ConfigMap.Store(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0], nodeConfigMap)`

It means that, we can't create two same name device,even if in different ns, in the master. If not, cloudcore will meet the same problem that we create two device of multiple ns bind to same node.

So, I think kubeedge, at device at least, samplifies k8s' namespace. And why not we put all(including device, devicemodel and relevant deployment) in a specific namespace, like device-system?

fisherxu:
> And the part of devicecontroller of ke abandons namespace totally. 

We should not abandon the namespace, adding the namespace back should be our next stage of work

wuqihui0317:
> 
> We should not abandon the namespace, adding the namespace back should be our next stage of work

Ok, thanks. Do you have any plan to add the namespace back? We need to access devices to our edge-computing clutser. If possible, can we push the progress together?

PR #1781, opened 2020-06-05
[WIP] Support cloudcore run as deployment

daixiang0:
Signed-off-by: Xiang Dai <long0dai@foxmail.com>

Issue #1635, opened 2020-05-06
Deploy cloudcore as Kubernetes deployment in keadm init command


fisherxu:
**What would you like to be added/modified**:
Deploy cloudcore as K8s deployment instead of process, and K8s deployment can better manage the lifecycle of cloudcore.

PR #2139, opened 2020-09-07
Run cloudcore as service

GsssC:
**What type of PR is this?**
/kind feature

support cloudcore run as systemd service, so we can restart cloudcore by `systemctl restart cloudcore`

also fix some closely related bugs, so merge them into one pr:

1. Stop to Re-download `service file` if it exsists.
2. More msg about why failed to install.
3. change `klog.Fatal` to `klog.Error`, I think it is too superficial too log func stack here, no help for debug but make user nervers
https://github.com/kubeedge/kubeedge/blob/7c415600ccd66e6cf4c18999939586e71e8ac59d/keadm/cmd/keadm/keadm.go#L27
shown as below:
image
image

Fixes None-related-issue

None
```release-note

```

### DIFF
--- a/build/tools/cloudcore.service
+++ b/build/tools/cloudcore.service
@@ -3,7 +3,7 @@ Description=cloudcore.service
 
 [Service]
 Type=simple
-ExecStart=/etc/kubeedge/cloudcore
+ExecStart=cloudcore
 
 [Install]
 WantedBy=multi-user.target
--- a/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
+++ b/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
@@ -167,10 +167,10 @@ func (cu *KubeCloudInstTool) RunCloudCore() error {
 	cmd.Cmd.Env = os.Environ()
 	err = cmd.ExecuteCmdShowOutput()
 	if err!=nil{
+		if errout := cmd.GetStdErr(); errout != "" {
+			fmt.Printf("failed to run cloudcore by {%s}, stderr: %s\n",command,errout)
+		}
 		return err
-	}
-	if errout := cmd.GetStdErr(); errout != "" {
-		return fmt.Errorf("failed to run cloudcore by {%s} : %s",command,errout)
 	}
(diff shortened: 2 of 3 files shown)


fisherxu:
Why not run cloudcore as container? @GsssC  ref: #1635  #1781

GsssC:
@fisherxu Yes, run as container also matters. Not compare `run as system service` with `run as container`, see it as a upgrade of `run as binary` ->**nohup cloudcore 2 &> 1 1 > cloudcore.log &**, I think it is make sense.
: )

fisherxu:
Let's run it as deployment

GsssC:
@fisherxu So will we still remain the way of `run as binary` ? If no, `run as system service` make no sense.

fisherxu:
Need to review and merge #1781

luogangyi:
+ 1
From user's prospective, running cloudcore as systemd service is useful.
Especially in scenery that user do not have full control of K8S. Running cloudcore need some iptables setting, it's hard to do in third-party K8S service. In that situation, runing cloudcore in a standalone VM is better.

fisherxu:
> Running cloudcore need some iptables setting, it's hard to do in third-party K8S service. 

CloudCore just watch objects from apiserver, for setting iptable rules, we can just use hostnetwork. Within container, cloudcore can also be isolated with other k8s control plane component :)

luogangyi:
> > Running cloudcore need some iptables setting, it's hard to do in third-party K8S service.
> 
> CloudCore just watch objects from apiserver, for setting iptable rules, we can just use hostnetwork. Within container, cloudcore can also be isolated with other k8s control plane component :)

Cloudcore also proxy kube-api's and metric server's api request in logs and metrics, that requires iptabels config. Hostnetwork does not solve all problem.( Sure we can also config iptables in pod when we use NET_RAW and NET_ADMIN flag)

I think running cloudcore in containers is easy to achieve high availability while running in systemd is easy to configure and test. They are not conflict.

luogangyi:
related to #2142

fisherxu:
/hold
Need more discussion.

PR #2164, opened 2020-09-16
Add question issue template

daixiang0:

Now users always mark any question as bug, it would mislead anyone who want to fix bugs.

### DIFF
--- a/.github/ISSUE_TEMPLATE/question.md
+++ b/.github/ISSUE_TEMPLATE/question.md
@@ -1,0 +1,41 @@
+---
+name: Question
+about: Question relating to KubeEdge.
+labels: question
+
+---
+
+<!-- Please use this template while providing as much info as possible. Thanks!-->
+**What happened and what you expected to happen**:
+
+**How to reproduce it (as minimally and precisely as possible)**:
+
+**Anything else we need to know?**:
+
+**Environment**:
+<details><summary>Kubernets Version</summary>
+
+```console
+$ kubectl --version
+# paste output here
+```
+
+</details>
+
+<details><summary>CloudCore Version</summary>
+
+```console
+$ cloudcore --version
+# paste output here
+```
+
+</details>
+
+<details><summary>EdgeCore Version</summary>
+
+```
+$ edgecore --version
+# paste output here
+```
+
+</details>


.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,117 @@
...
+```
+</details>
+
+<details><summary>Kernel </summary>
+```console
+$ uname -a
+# paste output here
+```
+</details>
+
```
GsssC, line 101: repetition
```suggestion

```

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,117 @@
...
```
GsssC, line 62: repetition
```suggestion
```

GsssC:
look good to me, I also think we need an kind of issue to submit question!
👍

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,103 @@
...
+**What you expected to happen**:
+
+**How to reproduce it (as minimally and precisely as possible)**:
+
+**Anything else we need to know?**:
+
+**Environment**:
+<details><summary>Kubernets Version</summary>
+```console
```
fisherxu, line 19: Looks like the ` ```console` doesn't work within the `<details>`? 
image

daixiang0: Fix it, in fact it needs extra empty lines.

fisherxu:
Looks fine :)  But the template file seems a bit long for users to write.. And when we want to see the version of components, we need to click them one by one?

daixiang0:
How about only saying os arch and kernel version for each component, and put the version of components outside, then we can easily to get version info.

kevin-wangzefeng:
Do we really want to receive questions on github?
Or probably just for a period of time until we find a better place.
I'm just afraid of messing up the issues list.

kevin-wangzefeng:
And we do need to add some guide to avoid issues created with "kind/bug" but are actually questions.

daixiang0:
In k8s, the question template is just a entry to tip user that they should go to other places to ask questions.

For now, we still need to split bug and question, the first step is to add a question template.

Then, we can learn k8s, only keep question template as a entry.

PR #2172, opened 2020-09-18
metamanager: fix send sync msg with async SendToCloud

### LINKED ISSUE #2171: odd timeout wait for sync response while report response received at the same time
**What happened**:
Got an odd timeout log message on, both "timeout" and "succeed" for a single parent message.
1. timeout
> W0918 18:47:46.222848   10380 process.go:149] timeout to receive response for message: {Header:{ID:e5025b39-e0e3-4741-bee3-94b56205d8db ParentID: Timestamp:1600426051210 ResourceVersion: Sync:true} Router:{Source:edged Group:meta Operation:update Resource:kubeedge/podstatus/csi-hostpath-edge-fbmh7} Content:{UID:7336cfda-1c99-4cb2-b585-7a3c4f1aafe7 Name:csi-hostpath-edge-fbmh7 Status:{Phase:Running Conditions:[{Type:Initialized Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason: Message:} {Type:Ready Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2020-09-18 18:47:23 +0800 CST Reason: Message:} {Type:PodScheduled Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason: Message:}] Message: Reason: NominatedNodeName: HostIP:192.168.238.169 PodIP:192.168.238.169 PodIPs:[] StartTime:2020-09-18 18:47:22 +0800 CST InitContainerStatuses:[] ContainerStatuses:[{Name:csi-hostpath-driver State:{Waiting:nil Running:&ContainerStateRunning{StartedAt:2020-09-18 18:47:22 +0800 CST,} Terminated:nil} LastTerminationState:{Waiting:nil Running:nil Terminated:nil} Ready:true RestartCount:0 Image:quay.io/k8scsi/hostpathplugin:v1.3.0 ImageID:docker-pullable://quay.io/k8scsi/hostpathplugin@sha256:5871ca68efb924e53f5c7a64ae4e267de1d6834051e66e75b9ed40a79d24f456 ContainerID:docker://15a7a155977bb18f777e0f683424f3d17619ec189a7515ce6ffa1f477120ec62 Started:0xc000f5e7b9} {Name:node-driver-registrar State:{Waiting:nil Running:&ContainerStateRunning{StartedAt:2020-09-18 18:47:22 +0800 CST,} Terminated:nil} LastTerminationState:{Waiting:nil Running:nil Terminated:nil} Ready:true RestartCount:0 Image:quay.io/k8scsi/csi-node-driver-registrar:v1.1.0 ImageID:docker-pullable://quay.io/k8scsi/csi-node-driver-registrar@sha256:13daf82fb99e951a4bff8ae5fc7c17c3a8fe7130be6400990d8f6076c32d4599 ContainerID:docker://6e567c2e7bd38d934ef0225267d96e3047c3b86c30246dbfeaa7897244fb4122 Started:0xc000f5e7ba}] QOSClass:BestEffort EphemeralContainerStatuses:[]}}}

2. succeed
> I0918 18:47:31.221282   10380 process.go:669] get a message {Header:{ID:e5025b39-e0e3-4741-bee3-94b56205d8db ParentID: Timestamp:1600426051210 ResourceVersion: Sync:true} Router:{Source:edged Group:meta Operation:update Resource:kubeedge/podstatus/csi-hostpath-edge-fbmh7} Content:{UID:7336cfda-1c99-4cb2-b585-7a3c4f1aafe7 Name:csi-hostpath-edge-fbmh7 Status:{Phase:Running Conditions:[{Type:Initialized Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason: Message:} {Type:Ready Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2020-09-18 18:47:23 +0800 CST Reason: Message:} {Type:PodScheduled Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason: Message:}] Message: Reason: NominatedNodeName: HostIP:192.168.238.169 PodIP:192.168.238.169 PodIPs:[] StartTime:2020-09-18 18:47:22 +0800 CST InitContainerStatuses:[] ContainerStatuses:[{Name:csi-hostpath-driver State:{Waiting:nil Running:&ContainerStateRunning{StartedAt:2020-09-18 18:47:22 +0800 CST,} Terminated:nil} LastTerminationState:{Waiting:nil Running:nil Terminated:nil} Ready:true RestartCount:0 Image:quay.io/k8scsi/hostpathplugin:v1.3.0 ImageID:docker-pullable://quay.io/k8scsi/hostpathplugin@sha256:5871ca68efb924e53f5c7a64ae4e267de1d6834051e66e75b9ed40a79d24f456 ContainerID:docker://15a7a155977bb18f777e0f683424f3d17619ec189a7515ce6ffa1f477120ec62 Started:0xc000f5e7b9} {Name:node-driver-registrar State:{Waiting:nil Running:&ContainerStateRunning{StartedAt:2020-09-18 18:47:22 +0800 CST,} Terminated:nil} LastTerminationState:{Waiting:nil Running:nil Terminated:nil} Ready:true RestartCount:0 Image:quay.io/k8scsi/csi-node-driver-registrar:v1.1.0 ImageID:docker-pullable://quay.io/k8scsi/csi-node-driver-registrar@sha256:13daf82fb99e951a4bff8ae5fc7c17c3a8fe7130be6400990d8f6076c32d4599 ContainerID:docker://6e567c2e7bd38d934ef0225267d96e3047c3b86c30246dbfeaa7897244fb4122 Started:0xc000f5e7ba}] QOSClass:BestEffort EphemeralContainerStatuses:[]}}}
> I0918 18:47:31.222357   10380 metaclient.go:121] send sync message kubeedge/podstatus/csi-hostpath-edge-fbmh7 successed and response: {{7100fe8b-9762-4a28-b244-bd8152e4e6e8 e5025b39-e0e3-4741-bee3-94b56205d8db 1600426051222  false} {edged meta response kubeedge/podstatus/csi-hostpath-edge-fbmh7} OK}I0918 18:47:31.222395   10380 status_manager.go:118] Status for pod csi-hostpath-edge-fbmh7 updated successfully: {Phase:Running Conditions:[{Type:Initialized Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason: Message:} {Type:Ready Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2020-09-18 18:47:23 +0800 CST Reason: Message:} {Type:PodScheduled Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:0001-01-01 00:00:00 +0000 UTC Reason: Message:}] Message: Reason: NominatedNodeName: HostIP:192.168.238.169 PodIP:192.168.238.169 PodIPs:[] StartTime:2020-09-18 18:47:22 +0800 CST InitContainerStatuses:[] ContainerStatuses:[{Name:csi-hostpath-driver State:{Waiting:nil Running:&ContainerStateRunning{StartedAt:2020-09-18 18:47:22 +0800 CST,} Terminated:nil} LastTerminationState:{Waiting:nil Running:nil Terminated:nil} Ready:true RestartCount:0 Image:quay.io/k8scsi/hostpathplugin:v1.3.0 ImageID:docker-pullable://quay.io/k8scsi/hostpathplugin@sha256:5871ca68efb924e53f5c7a64ae4e267de1d6834051e66e75b9ed40a79d24f456 ContainerID:docker://15a7a155977bb18f777e0f683424f3d17619ec189a7515ce6ffa1f477120ec62 Started:0xc0006fa0b1} {Name:node-driver-registrar State:{Waiting:nil Running:&ContainerStateRunning{StartedAt:2020-09-18 18:47:22 +0800 CST,} Terminated:nil} LastTerminationState:{Waiting:nil Running:nil Terminated:nil} Ready:true RestartCount:0 Image:quay.io/k8scsi/csi-node-driver-registrar:v1.1.0 ImageID:docker-pullable://quay.io/k8scsi/csi-node-driver-registrar@sha256:13daf82fb99e951a4bff8ae5fc7c17c3a8fe7130be6400990d8f6076c32d4599 ContainerID:docker://6e567c2e7bd38d934ef0225267d96e3047c3b86c30246dbfeaa7897244fb4122 Started:0xc0006fa0b2}] QOSClass:BestEffort EphemeralContainerStatuses:[]}

If we see "succeed", we shall not see the "timeout"

0. Suppose you have a running KubeEdge cluster.
1. Turn edgecore log level to at least 2 (to see the "succeed" msg)
2. Deploy a sinle pod and check for edgecore log to wait for the "timeout" warning msg

**Environment**:
KubeEdge v1.4.0-beta.0.128+54c38dbbda787e-dirty

**CloudSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

**EdgeSide Environment**:
- Kernel (e.g. `uname -a`):
- Others:

guihecheng:
The original msg come from edged to metamanager may be sync,
but later we use SendToCloud, which is async, to send msg to cloud,
then edgehub will wait for response and timeout on some type of
resource update operation, e.g. podstatus update which will have
no response from cloud.

Signed-off-by: Gui Hecheng <guihecheng@cmiot.chinamobile.com>

**What type of PR is this?**
/kind bug

Fixes #2171

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -66,7 +66,10 @@ func sendToEdgeMesh(message *model.Message, sync bool) {
 }
 
 func sendToCloud(message *model.Message) {
+	oldSync := message.Header.Sync
+	message.Header.Sync = false
 	beehiveContext.SendToGroup(string(metaManagerConfig.Config.ContextSendGroup), *message)
+	message.Header.Sync = oldSync
 }
 
 // Resource format: <namespace>/<restype>[/resid]


GsssC (commented):
The reason why edgehub still timeout to get sync-resp is edgecontroller or cloudhub (both in cloud) never send a sync-resp to edge, although it get a sync-message.

In `sentToCloud()`, would better use both `beehiveContext.SendToGroup` and `beehiveContext.SendToGroupSync` according to `message.Header.sync`

and add some code about sync-resp at edgecontroller:
https://github.com/kubeedge/kubeedge/blob/c85edd583511bd6a2876da62c69d471307d7040c/cloud/pkg/edgecontroller/controller/upstream.go#L304

daixiang0:
Besides, try add test to cover those cases.

guihecheng:
@GsssC yes, actually I've tried to add a response for podstatus in upstream controller once, but later found it not so good.
First, we come to the place where the msg is sent by metamanager:
https://github.com/kubeedge/kubeedge/blob/c85edd583511bd6a2876da62c69d471307d7040c/edge/pkg/metamanager/process.go#L318-L321
I think here it intends to use the request message directly as the response to send back to edged but not wait for the sync-resp.
And before that, it sends an `async` msg to cloud just to trigger a podstatus update.
So actually here it wants an `async` behavior, but the msg send by edged actually has `sync` field set to true, this cause edgehub to wait for the response which no one actually is interested in.

Second, if we add a response to podstatus in upstream, then edged will receive a lot of podstatus responses which are somewhat meaningless, because metamanager will send a meta sync msg every minute.
https://github.com/kubeedge/kubeedge/blob/c85edd583511bd6a2876da62c69d471307d7040c/edge/pkg/metamanager/metamanager.go#L70-L73

Third, the edged now is not intended to process a response msg of type ResourceTypePodStatus, and this line will be hit:
https://github.com/kubeedge/kubeedge/blob/c85edd583511bd6a2876da62c69d471307d7040c/edge/pkg/edged/edged.go#L1081-L1083

Finally, I came out with this fix, it may seems to be not so clean, but it intends to make edgehub skip wait for the msgs which are not really sync at all.

guihecheng:
@daixiang0 ok, I'll try that if possible.

GsssC:
@guihecheng I get it.
First, make sure that all calls of `sendToCloud` now are async.

Second, go through all msgs passed in `sendToCloud` are got from other modules, not generated, so `msg.Header.Sync` here means sync between other moudles and MetaManager, not between MetaManager and cloud.

Third, if users have to sentToCloudSync, it will be like as below:
https://github.com/kubeedge/kubeedge/blob/c85edd583511bd6a2876da62c69d471307d7040c/edge/pkg/metamanager/process.go#L450-L459

It is better to make a new async msg when pass in `sentToCloud`, because, strictly speaking, these are two msgs, one to MetaManager, one to cloud, although the content is same.

But I think modification now is also a good way for workaround.
/lgtm
/cc fisherxu

guihecheng:
ping @fisherxu

PR #2173, opened 2020-09-18
docs/troubleshooting.md: update release version

daixiang0:

**What type of PR is this?**

/kind documentation


fisherxu:
As an example, do we need always update to new version?

daixiang0:
@fisherxu for version part, i think no, while since we upgrade device crds, it is not easy for new users to change it.

PR #2175, opened 2020-09-18
Add sync/async check when use beehive msg framework

GsssC:
**What type of PR is this?**

/kind improvment

Sync or aync is important for the propagation chain of messages, as #2172 shows, we developer is easy to wrong use send/sendsync, cause propagation chain of messages wrong.

Add check to normalize .

part of fixes #2172

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
@@ -57,6 +57,10 @@ func (ctx *ChannelContext) Cleanup(module string) {
 
 // Send send msg to a module. Todo: do not stuck
 func (ctx *ChannelContext) Send(module string, message model.Message) {
+	// check if msg is sync
+	if message.Header.Sync {
+		klog.Warningf("Get a sync-msg when use beehive.Send(),msg.Header:{%v},msg.Route:{%v}",message.Header,message.Router)
+	}
 	// avoid exception because of channel colsing
 	// TODO: need reconstruction
 	defer func() {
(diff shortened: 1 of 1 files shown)


staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
```diff
@@ -57,6 +57,10 @@ func (ctx *ChannelContext) Cleanup(module string) {
 
 // Send send msg to a module. Todo: do not stuck
 func (ctx *ChannelContext) Send(module string, message model.Message) {
+	// check if msg is sync
+	if message.Header.Sync {
+		klog.Warningf("Get a sync-msg when use beehive.Send(),msg.Header:{%v},msg.Route:{%v}",message.Header,message.Router)
```
daixiang0, line 62: In fact it would be useful for debug, i prefer combine them into one debug level log.

GsssC: yes it is, and how to? :)

daixiang0: By mistake, i mean change to debug level instead.

GsssC: but there is no `klog.Debug`

GsssC: or klog.V(4).Infof()? as https://www.cnblogs.com/iiiiher/p/7879817.html

PR #2194, opened 2020-09-25
CI: use `install: skip` instead of `install: true`

daixiang0:

**What type of PR is this?**

/kind test

Using `install: true` does not skip install step but runs true command. To actually skip install step one should use `install: skip`. See https://docs.travis-ci.com/user/job-lifecycle/#skipping-the-installation-phase and travis-ci/docs-travis-ci-com#2422 (comment)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ before_install:
         travis_terminate 0
       fi
 
-install: true
+install: skip
 
 jobs:
   include:


GsssC (commented):
/lgtm 

`install: skip` truly skip `install step`. 

If yaml like this:
```
install: 
    - true
    - command1
    - command2
```
Seems `install: true` does not skip, does it?

daixiang0:
Yep.

PR #2207, opened 2020-09-28
Add restart options for cloudcore.service and edgecore.service

YaozhongZhang:
**What type of PR is this?**

/kind feature

Add restart options for cloudcore.service and edgecore.service, so that they can restart automatically on unexpected exit.

```release-note

```

### DIFF
--- a/build/tools/cloudcore.service
+++ b/build/tools/cloudcore.service
@@ -4,6 +4,8 @@ Description=cloudcore.service
 [Service]
 Type=simple
 ExecStart=/etc/kubeedge/cloudcore
+Restart=always
+RestartSec=10
 
 [Install]
 WantedBy=multi-user.target
--- a/build/tools/edgecore.service
+++ b/build/tools/edgecore.service
@@ -4,6 +4,8 @@ Description=edgecore.service
 [Service]
 Type=simple
 ExecStart=/etc/kubeedge/edgecore
+Restart=always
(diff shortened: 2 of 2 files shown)


daixiang0:
lgtm 
ping @fisherxu

build/tools/cloudcore.service
```diff
@@ -4,6 +4,8 @@ Description=cloudcore.service
 [Service]
 Type=simple
 ExecStart=/etc/kubeedge/cloudcore
+Restart=always
```
chendave, line 7: Is that meaning service will always try when the exception cannot get recovered?

YaozhongZhang: This "always" means that the service will be restarted regardless of whether the exit code is 0 or not.

YaozhongZhang: I don't think it's a problem. And kubelet dose the same.

> [Unit]
> Description=kubelet: The Kubernetes Node Agent
> Documentation=https://kubernetes.io/docs/
> 
> [Service]
> ExecStart=/usr/bin/kubelet
> Restart=always
> StartLimitInterval=0
> RestartSec=10
> KillMode=process
> 
> [Install]
> WantedBy=multi-user.target

kevin-wangzefeng: I'm fine with this

PR #2208, opened 2020-09-28
support specify edgecore file path by os.env

chenchunxiu:
**What type of PR is this?**
/kind feature

Allow users to use the os.env to config the edgecore file path


fisherxu:
Thanks for the fixing and why not specify it through `--config` option?

chenchunxiu:
--config is better. so close it.

PR #2210, opened 2020-09-29
support list-watch proxy in edge side

### LINKED ISSUE #1758: Support list-watch from edgecore for applications on the edge

threestoneliu:
**What type of PR is this?**

Support the list-watch request of the side application. When the apiserver is accessible, the list-watch request is forwarded to the cloud, and when it is not accessible, the local cache is used to construct the response for the list-watch request.

Fixes #1758 
Fixes #1907 

```release-note

```


kevin-wangzefeng:
How about lwproxy comparing to the name edgeproxy? Just wondering if people would think edgeproxy is something similar to kubeproxy.

threestoneliu:
ok，i wil rename this module.

PR #2218, opened 2020-09-30
cleanup dup codes

guihecheng:
Signed-off-by: Gui Hecheng <guihecheng@cmiot.chinamobile.com>

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged_status.go
+++ b/edge/pkg/edged/edged_status.go
@@ -423,12 +423,9 @@ func (e *edged) syncNodeStatus() {
 			klog.Errorf("Register node failed: %v", err)
 			return
 		}
-		if err := e.updateNodeStatus(); err != nil {
-			klog.Errorf("Unable to update node status: %v", err)
-		}
-	} else {
-		if err := e.updateNodeStatus(); err != nil {
-			klog.Errorf("Unable to update node status: %v", err)
-		}
+	}
+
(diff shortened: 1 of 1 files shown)


edge/pkg/edged/edged_status.go
```diff
@@ -423,12 +423,9 @@ func (e *edged) syncNodeStatus() {
...
-			klog.Errorf("Unable to update node status: %v", err)
-		}
-	} else {
-		if err := e.updateNodeStatus(); err != nil {
-			klog.Errorf("Unable to update node status: %v", err)
-		}
+	}
+
+	if err := e.updateNodeStatus(); err != nil {
+		klog.Errorf("Unable to update node status: %v", err)
```
fisherxu, line 429: @guihecheng Thanks for the fixing! Actually this is not the dup codes, when we first registerNode, then update the nodestatus immediately, or the node will keep not ready for 10s.
And do you have other ways to refactor it?

guihecheng: @fisherxu thanks for the explanation for the intent of the code block, but this cleanup will not change the function logic, since the "update block" is simply "merged" and moved outside the else branch. Here I just want to simplify the code a bit :)
Did I miss anything?

fisherxu: Ohh, just have a double check you are right :)

PR #2235, opened 2020-10-13
Keadm: use latest version in help info

daixiang0:

The help info should better use latest release version.

Do not call `curl` to get since it should be updated in release process.

Fixes #

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cloud/init.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/init.go
@@ -40,7 +40,7 @@ keadm init
 
 - This command will download and install the default version of KubeEdge cloud component
 
-keadm init --kubeedge-version=1.2.1  --kube-config=/root/.kube/config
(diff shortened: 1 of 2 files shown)


pkg/version/version.go
```diff
@@ -40,3 +40,7 @@ func Get() apimachineryversion.Info {
 		Platform:     fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
 	}
 }
+
+func GetLatestVersion() string {
+	return "1.4.0"
```
kevin-wangzefeng, line 45: You can get from https://kubeedge.io/latestversion

daixiang0: This can work with offline case, and update all version info in release process is a common way.

fisherxu: We have a default version in `keadm/cmd/keadm/app/cmd/common/constant.go`

daixiang0: Let's use it :)

PR #2248, opened 2020-10-14
Echo the build error message to stderr instead of stdout

llhuii:
**What type of PR is this?**

For example, **SHOW** the error message `can not find binary: abc` in the terminal when `make all WHAT=abc` (just to **trigger the build script error**):
```
bash hack/verify-golang.sh
go detail version: go version go1.14.7 linux/amd64
go version: 1.14.7
hack/make-rules/build.sh abc 
can not find binary: abc
Makefile:41: recipe for target 'all' failed
make: *** [all] Error 1
```

Fixes #

```release-note

```

### DIFF
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -135,7 +135,7 @@ function start_edgecore {
 
 function check_control_plane_ready {
   echo "wait the control-plane ready..."
-  kubectl wait --for=condition=Ready node/test-control-plane --timeout=60s
+  kubectl wait --for=condition=Ready node/${CLUSTER_NAME}-control-plane --timeout=60s
 }
 
 # Check if all processes are still running. Prints a warning once each time
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -17,6 +17,7 @@
 KUBEEDGE_ROOT=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/..
 ENABLE_DAEMON=${ENABLE_DAEMON:-false}
 LOG_DIR=${LOG_DIR:-"/tmp"}
+TIMEOUT=${TIMEOUT:-60}s
 
 if [[ "${CLUSTER_NAME}x" == "x" ]];then
     CLUSTER_NAME="test"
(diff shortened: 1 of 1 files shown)


daixiang0:
`make HELP=y` is enough.

llhuii:
> `make HELP=y` is enough.

It's another command.

I want to say that the build scripts should echo the error message in the terminal when it wants to.

But now the message would be discard by the shell assignment command like `read -r ldflags <<< "$(kubeedge::version::ldflags)"` even when `kubeedge::version::ldflags` errors.

daixiang0:
So the desribtion of pr is wrong since your point was incorrect **WHAT**.

llhuii:
> So the desribtion of pr is wrong since your point was incorrect **WHAT** while HELP would explain it.
Ok, it is misleading, just to an example to trigger the build error.

fisherxu:
Just small question, is any issue without this patch? :)

llhuii:
> Just small question, is any issue without this patch? :)

Can't find one. I guess no one notice this.

fisherxu:
> > Just small question, is any issue without this patch? :)
> 
> Can't find one. I guess no one notice this.

So let's keep as it is now :)

llhuii:
> > > Just small question, is any issue without this patch? :)
> > 
> > 
> > Can't find one. I guess no one notice this.
> 
> So let's keep as it is now :)

I thought you wanted me to check any open github issues related to this one.

Yes, it has something wrong with me:
- These `echo` commands would be useless without the patch since it would be discard. I think the authors of this code doesn't means that.
- For the view of Error message, I think it should be echoed to stderr either.

llhuii:
PS: These `echo`s can be wrapped into bash functions, like [k8s do this](https://github.com/kubernetes/kubernetes/blob/master/hack/lib/logging.sh).

llhuii:
From this shell guide https://google.github.io/styleguide/shellguide.html#stdout-vs-stderr:

image

llhuii:
FYI, I submited this to k8s project and it merged. https://github.com/kubernetes/kubernetes/pull/97999.

llhuii:
sorry some wrong: my origin commit was overwritten by mistake.

fisherxu:
> sorry something wrong: my origin commit was overwritten by mistake.

Not good choice to use master branch :)

llhuii:
> > sorry something wrong: my origin commit was overwritten by mistake.
> 
> Not good choice to use master branch :)

Learned😄

PR #2258, opened 2020-10-16
Fix bash array usage in hack/lib/golang.sh

llhuii:
**What type of PR is this?**

Fix bash array usage in `kubeedge::check::env` of 'hack/lib/golang.sh'.
The build command below
```sh
export GOPATH=
make all WHAT=edgecore
```
will output the misleading messages:
```
bash hack/verify-golang.sh
go detail version: go version go1.14.7 linux/amd64
go version: 1.14.7
hack/make-rules/build.sh keadm 
Error: GOPATH
Error: environment
Error: value
Error: not
Error: set
Makefile:41: recipe for target 'all' failed
make: *** [all] Error 1
```
which intends to be 
```sh
bash hack/verify-golang.shgo detail version: go version go1.14.7 linux/amd64
go version: 1.14.7
hack/make-rules/build.sh keadm 
Error: GOPATH environment value not set
Makefile:41: recipe for target 'all' failedmake: *** [all] Error 1
```

Fixes #

```release-note

```

### DIFF
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -140,7 +140,7 @@ kubeedge::check::env() {
   # check lenth of errors
   if [[ ${#errors[@]} -ne 0 ]] ; then
     local error
-    for error in ${errors[@]}; do
+    for error in "${errors[@]}"; do
       echo "Error: "$error
     done
     exit 1


daixiang0:
If GOPATH is empty, your go environment is totally wrong!

llhuii:
> If GOPATH is empty, your go environment is totally wrong!

Yes, you are right. But the point here is that error message shouble be `Error: GOPATH environment value not set
 ` instead of
```
Error: GOPATH
Error: environment
Error: value
Error: not
Error: set
```

fisherxu:
> > If GOPATH is empty, your go environment is totally wrong!
> 
> Yes, you are right. But the point here is that error message shouble be `Error: GOPATH environment value not set ` instead of
> 
> ```
> Error: GOPATH
> Error: environment
> Error: value
> Error: not
> Error: set
> ```

Fort the error message issue, it makes sense.

/lgtm
/approve

PR #2261, opened 2020-10-16
Remove useless sync script

daixiang0:

**What type of PR is this?**

 /kind cleanup

Now we do not need sync from sub repo, remove it.


fisherxu:
Let's keep the script there.
/close

PR #2283, opened 2020-10-28
add gateway proposal

### LINKED ISSUE #1432: Add EdgeGateway as the ingress gateway on Edge

FengFees:
**What type of PR is this?**

/kind design

Gateway or edge service in a microservice system provide entry point for the entire application so that external applications can visit this application.
In the native kubernetes, ingress is provided to achieve the purpose. Gateway is another object designed in istio as another solution. They provide different capabilities with different designs for the cloud-grade gateway/edge service solution.
However, the implementation on Edge is different. We need an implementation on edge to support the Gateway capabilities.

Fixes #1432


daixiang0:
@FengFees Thanks for contributions!

Please split pr and focus this pr on proposal, after proposal confirmed, let's code :)

PR #2289, opened 2020-10-31
[WIP] test: add arm64 arch support

daixiang0:

**What type of PR is this?**

/kind feature
/kind test

/kind failing-test

From [travis blog](https://blog.travis-ci.com/2020-09-11-arm-on-aws), only `virt: vm` tag can not enable arm64 feature, need `group` tag too.

Also from [CI log](https://travis-ci.org/github/kubeedge/kubeedge/jobs/736281853#L470-L494), it stills run on amd64 machine.

`kind` can not run well at arm64, so use `minikube` instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,11 @@ go:
   - 1.14
 
 env:
-  - HOME=/home/travis
+  global:
+    - HOME=/home/travis
+    - MINIKUBE_IN_STYLE=false
+    - CHANGE_MINIKUBE_NONE_USER=true
+    - MINIKUBE_WANTUPDATENOTIFICATION=false
 
 cache:
   directories:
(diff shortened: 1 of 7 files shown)


daixiang0:
Now miniube latest support k8s version is 1.19.2 which is used as default.

daixiang0:
Check [doc](https://docs.travis-ci.com/user/multi-cpu-architectures/#multi-cpu-availaibility), we need to switch to `com`.

/cc @kevin-wangzefeng please refer to [doc](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#migrating-a-repository ) to migrate from `org` to `com`.

daixiang0:
Now only e2e on arm64 vm fails, it needs more time to dig into root cause, so implement it in other pr.

kevin-wangzefeng:
According to the latest travis-ci billing model, we will need credits after moved to travis-ci.com
Maybe we can just move x86 CI jobs to other platform, and keep all the travis-ci credits for arm env validations.

Ref: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

daixiang0:
switch to GITHUB, abort it.

PR #2302, opened 2020-11-05
modify one log of eventbus

muxuelan:
**What type of PR is this?**
/kind documentation

For version 1.4，one log of the function onSubscribe is unreasonable.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/eventbus/mqtt/client.go
+++ b/edge/pkg/eventbus/mqtt/client.go
@@ -94,7 +94,7 @@ func OnSubMessageReceived(client MQTT.Client, message MQTT.Message) {
 	klog.Infof("OnSubMessageReceived receive msg from topic: %s", message.Topic())
 	// for "$hw/events/device/+/twin/+", "$hw/events/node/+/membership/get", send to twin
 	// for other, send to hub
-	// for topic, no need to base64 topic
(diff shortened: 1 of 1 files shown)


edge/pkg/eventbus/mqtt/client.go
```diff
@@ -94,7 +94,7 @@ func OnSubMessageReceived(client MQTT.Client, message MQTT.Message) {
 	klog.Infof("OnSubMessageReceived receive msg from topic: %s", message.Topic())
 	// for "$hw/events/device/+/twin/+", "$hw/events/node/+/membership/get", send to twin
 	// for other, send to hub
-	// for topic, no need to base64 topic
+	// for "SYS/dis/upload_records", no need to base64 topic
```
daixiang0, line 97: I think it is still a topic name. Maybe better to comment in doc, and try to add unit test fort this :)

muxuelan: Actually，all topis are 64-encoded，except for "SYS/dis/upload_records".I think  the sentence "for topic"range too much and is incorrect here.

daixiang0: OK.

PR #2308, opened 2020-11-09
Fix some spelling errors

muxuelan:
**What type of PR is this?**
/kind lgtm

**What this PR does / why we need it**:
Fix some spelling errors.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #2309, opened 2020-11-09
Prevent invalid messages from affecting other messages operated

muxuelan:
**What type of PR is this?**
/kind feature

Once a topic is invalid, it will stop the traversal, causing other topics in SubTopics to be unsubscribed.

Fixes #

```release-note

```


kevin-wangzefeng:
Thanks @muxuelan , it looks there's some git author information the github can't recognize. Can you double check that you are using right `git config user.name` and `git config user.email`?

daixiang0:
Some changes are the same with #2308, please check it.

muxuelan:
> Some changes are the same with #2308, please check it.

Thanks, I will check and re-push.

edge/pkg/eventbus/mqtt/client.go
```diff
@@ -83,7 +83,7 @@ func onSubConnect(client MQTT.Client) {
 		token := client.Subscribe(t, 1, OnSubMessageReceived)
 		if rs, err := util.CheckClientToken(token); !rs {
 			klog.Errorf("edge-hub-cli subscribe topic: %s, %v", t, err)
-			return
```
fisherxu, line 86: If Subscribe failed here. do we need to panic directly?

muxuelan: Well，I use an empty topic to test, actually It's just  an invalid topic not a panic, I think the invalid topics such as an empty topic should not affect other valid topics. And  I want to know in which case subscribe will  failed due to panic because I haven't seen it during my study.

fisherxu: The `SubTopics` will be used in kubeedge, so we need to ensure that every top must be subscribed successfully.

PR #2310, opened 2020-11-10
remove some unused code of edge part

lvchenggang:
What type of PR is this?
/kind cleanup

Remove some unused code of edge part, which could make the project code looks neater.

NONE

### DIFF
--- a/edge/pkg/devicetwin/dtcommon/util.go
+++ b/edge/pkg/devicetwin/dtcommon/util.go
@@ -7,11 +7,6 @@ import (
 	"strings"
 )
 
-const (
-	//MAXTWINNUM max twin key
-	MAXTWINNUM = 64
-)
-
 //ValidateValue validate value type
 func ValidateValue(valueType string, value string) error {
 	switch valueType {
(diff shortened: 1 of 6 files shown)


edge/pkg/devicetwin/dtmanager/twin.go
```diff
@@ -23,8 +23,6 @@ const (
 	RestDealType = 0
 	//SyncDealType update form cloud sync
 	SyncDealType = 1
-	//DetailDealType detail update from cloud
-	DetailDealType = 2
```
fisherxu, line 27: We should not delete this one, looks like it's continuous with `SyncDealType` and `SyncTwinDeleteDealType `

fisherxu (commented):
thanks for the fixing :)

edge/pkg/devicetwin/dtmanager/twin.go
lvchenggang: Yes, the DetailDealType may be used later in bussiness and ut code. Has been changed back.

PR #2313, opened 2020-11-11
reorganize constant file

lvchenggang:
What type of PR is this?
/kind cleanup

Reorganize the constatnt file in path common/constants/default.go, which makes the file more readable and clearer.

NONE


fisherxu:
/lgtm
/approve
Thanks for the fixing :)

PR #2329, opened 2020-11-16
change the EventId of GetDetailNode struct in deviectwin to uuid

lvchenggang:
What type of PR is this?
/kind feature

Change the EventId of GetDetailNode struct in deviectwin to uuid instead of stable "123", which could make the project code has  stronger maintainability.

NONE

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/communicate.go
+++ b/edge/pkg/devicetwin/dtmanager/communicate.go
@@ -8,6 +8,7 @@ import (
 
 	"k8s.io/klog"
 
+	"github.com/google/uuid"
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
 	"github.com/kubeedge/beehive/pkg/core/model"
 	connect "github.com/kubeedge/kubeedge/edge/pkg/common/cloudconnection"
@@ -123,10 +124,9 @@ func dealConfirm(context *dtcontext.DTContext, resource string, msg interface{})
 }
 
 func detailRequest(context *dtcontext.DTContext, msg interface{}) (interface{}, error) {
-	//todo eventid uuid
 	getDetail := dttype.GetDetailNode{
 		EventType: "group_membership_event",
-		EventID:   "123",
+		EventID:   uuid.New().String(),
 		Operation: "detail",
 		GroupID:   context.NodeName,
 		TimeStamp: time.Now().UnixNano() / 1000000}


edge/pkg/devicetwin/dtmanager/communicate.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dtcommon"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dtcontext"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dttype"
+	uuid "github.com/satori/go.uuid"
```
daixiang0, line 17: It would be good change, while i am concern is which `uuid` lib should we use:
```
$ grep uuid go.mod
	github.com/google/uuid v1.1.1
	github.com/pborman/uuid v1.2.0
	github.com/satori/go.uuid v1.2.0

```

lvchenggang: The import uuid is refer other packages existing in the project. 
If change to the following: github.com/satori/go.uuid/v1.2.0, the IDE will require to renew sync dependencies of github/kubeedge/kubeedge.

daixiang0: Track as #2330

edge/pkg/devicetwin/dtmanager/communicate.go
```diff
@@ -123,10 +124,9 @@ func dealConfirm(context *dtcontext.DTContext, resource string, msg interface{})
 }
 
 func detailRequest(context *dtcontext.DTContext, msg interface{}) (interface{}, error) {
-	//todo eventid uuid
 	getDetail := dttype.GetDetailNode{
 		EventType: "group_membership_event",
-		EventID:   "123",
+		EventID:   uuid.NewV4().String(),
```
daixiang0, line 129: Double-check, now `EventID` is useless?

lvchenggang: the EventID of GetDetailNode struct now will become the content of message, no other use.

edge/pkg/devicetwin/dtmanager/communicate.go
fisherxu: We can use `github.com/google/uuid` consistent with Kubernetes upstream and can switch all uuid ref to   `github.com/google/uuid` in follow-up PR.

lvchenggang: OK, related content has been modified.

PR #2333, opened 2020-11-17
add proposal for "tarballpath"

Rachel-Shao:
Added some descriptions in “keadm-scope”


daixiang0:
The proposal is a plan for coding, since you have implemented it, no need to update it.

PR #2335, opened 2020-11-17
reorganize edge module name into one file

lvchenggang:
What type of PR is this?
/kind feature

Reorganize the constatnts which is some edge module name into one file in the path of edge/pkg/common/modules/names.go. Could makes the project code has stronger maintainability and more readable.

NONE


GsssC:
seems good, wait for ci !
: )

GsssC (approved):

PR #2337, opened 2020-11-17
fix the groupmap while edgehub processing message

muxuelan:
**What type of PR is this?**
/kind feature

1. Before: The function pubConnectInfo of process.go will send same message twice to metamanager; 
    After: Tunction pubConnectInfo just send message to metamanager once.
2. The group name of ServiceBus is user, so in groupMap of process.go, key "user" is more suitable for ServiceBus instead of Eventbus.
3. Gropus of messages from cloud to edge just include resource(MetaGroup) and twin (TwinGroup) now,  so renaming the key "func" and "user"  don't affect edgehub dispatching message.

Fixes #

```release-note

```


daixiang0:
Good catch!

Could add unit test for this?

muxuelan:
> Good catch!
> 
> Could add unit test for this?

I think process_test.go is no problem, needn't to change.

daixiang0:
Test case for twice send :)

muxuelan:
> Test case for twice send :)

You mean add the test for func  pubConnectInfo ?

muxuelan:
> Good catch!
> 
> Could add unit test for this?

Could this be approved?

muxuelan:
> Test case for twice send :)

Could this be approved too as soon as possible？◠‿◠

PR #2340, opened 2020-11-18
Quic client capture error information

lvchenggang:
What type of PR is this?
/kind feature

Quic client in EdgeHub module add capture error information, which may be printed out by caller.

NONE


daixiang0:
lgtm

PR #2341, opened 2020-11-18
fix err information of certutil

muxuelan:
**What type of PR is this?**
/kind feature

If not add the modify, err equals nil here, and this will affect the generation of certificate.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/servers/httpserver/certutil.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/certutil.go
@@ -7,6 +7,7 @@ import (
 	"crypto/rand"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"errors"
 	"fmt"
 	"math"
 	"math/big"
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudhub/servers/httpserver/certutil.go
```diff
@@ -100,10 +101,12 @@ func NewCertFromCa(cfg *certutil.Config, caCert *x509.Certificate, serverKey cry
 	}
 	if len(cfg.CommonName) == 0 {
 		fmt.Println("must specify a CommonName")
+		err :=errors.New("must specify a CommonName")
 		return nil, err
```
fisherxu, line 105:
```suggestion
		return nil, errors.New("must specify a CommonName")
```

cloud/pkg/cloudhub/servers/httpserver/certutil.go
```diff
@@ -100,10 +101,12 @@ func NewCertFromCa(cfg *certutil.Config, caCert *x509.Certificate, serverKey cry
...
 	}
 	if len(cfg.Usages) == 0 {
 		fmt.Println("must specify at least one ExtKeyUsage")
+		err :=errors.New("must specify at least one ExtKeyUsage")
 		return nil, err
```
fisherxu, line 110:
```suggestion
		return nil, errors.New("must specify at least one ExtKeyUsage")
```

cloud/pkg/cloudhub/servers/httpserver/certutil.go
fisherxu: fmt print can be deleted here.

fisherxu (commented):
Please also squash the commits.

cloud/pkg/cloudhub/servers/httpserver/certutil.go
muxuelan: Okay ,I will ,thanks.

Issue #2026, opened 2020-08-04
EdgeSite is keep failing at "register node" event


pkalincat:
**What happened**:
EdgeSite is keep failing on registration node event. After build EdgeSite from master branch I'am not able to register node in master server. 

```
I0804 11:33:39.013025    7317 edged_status.go:387] Attempting to register node cat-testsite-2
W0804 11:33:39.013197    7317 context_channel.go:295] Failed to get channel, type:websocket
E0804 11:33:39.013221    7317 edged_status.go:393] register node failed, error: bad request module name(websocket)
E0804 11:33:39.013248    7317 edged_status.go:395] response from cloud core: %!s(<nil>)
E0804 11:33:39.013259    7317 edged_status.go:423] Register node failed: bad request module name(websocket)
```

There is also problem with other modules like EdgeMesh:

```
W0804 11:33:39.000456    7317 context_channel.go:295] Failed to get channel, type:edgemesh
W0804 11:33:39.000989    7317 context_channel.go:72] Get bad module name :edgemesh when send message, do nothing
```

EdgeSite documentation page is outdated, there is reference to modify edgeSite.yaml configuration file but that file I can find in version v1.0.0, on master branch there is no such configuration files like edgeSite.yaml, logging.yaml, or modules.yaml

**What you expected to happen**:
EdgeSite will start and register new node in master node. 

**How to reproduce it (as minimally and precisely as possible)**:
- Create k8s cluster.
- Create edge cluster with edgesite, build edgesite from master and run with defaultconfig or minconfig with proper variables inside. 

**Anything else we need to know?**:
Nothing else to know. 

**Environment**:
- KubeEdge version(e.g. `cloudcore/edgecore --version`):  
```
KubeEdge v1.4.0-alpha.0.257+094d66c0d6771b
```

**CloudSide Environment**:
- Hardware configuration (e.g. `lscpu`):
```
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              2
On-line CPU(s) list: 0,1
Thread(s) per core:  1
Core(s) per socket:  1
Socket(s):           2
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               60
Model name:          Intel Core Processor (Haswell, no TSX)
Stepping:            1
CPU MHz:             2095.082
BogoMIPS:            4190.16
Virtualization:      VT-x
Hypervisor vendor:   KVM
[...]

PR #2366, opened 2020-11-24
Fixed bug during node creation in edgesite

antoniomuso:
**What type of PR is this?**
/kind bug

Fixed bug during node creation in edgesite.
```console
register node failed, error: bad request module name(websocket)
```

Fixes [#2026](https://github.com/kubeedge/kubeedge/issues/2026)

```release-note
NONE
```

### DIFF
--- a/edge/pkg/edged/edged_status.go
+++ b/edge/pkg/edged/edged_status.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
+	"github.com/kubeedge/beehive/pkg/core"
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
Thanks for the fixing, @antoniomuso

edge/pkg/edged/edged_status.go
```diff
@@ -24,6 +24,7 @@ package edged
 
 import (
 	"fmt"
+	"github.com/kubeedge/beehive/pkg/core"
```
fisherxu, line 27: Please sorts the import here.

antoniomuso: Ok I sort the import.

fisherxu: Thanks, would you like to squash the commits?

antoniomuso: Done

PR #2369, opened 2020-11-25
cloudcore: support custom namespace

### LINKED ISSUE #2367: customize the installed namespace
customize the installed namespace

**Why is this needed**:
We need to deploy multiple sets of kubeedge, but found that the kubeedeg namespace and secret are fixed.

Hope to be able to customize the namespace in the field below.

```
const (
	NamespaceSystem string = "kubeedge"

...
)
```

shenkonghui:
**What type of PR is this?**
> /kind feature

Fixes #2367

**release-note**
It can generate secret in the namespace where the corresponding cloudcore pod is deployed, not only under kubeedge namesapce.
Determined by the environment variable KUBEEDGE_NAMESPACE

cloucore uses the following configuration.
```
...
  namespace: kubeedge1
...
  env:
  - name: CLOUDCORE_POD_NAMESPACE
    valueFrom:
      fieldRef:
        fieldPath: metadata.namespace
```
The relevant secret key will be automatically generated under the namespace  `kubeedge1`
```
[root@k8s build]# kubectl get secret -n kubeedge1
NAME                  TYPE                                  DATA   AGE
casecret              Opaque                                2      2m8s
cloudcoresecret       Opaque                                2      2m8s
default-token-752b9   kubernetes.io/service-account-token   3      2m8s
tokensecret           Opaque                                1      2m8s
```

### DIFF
--- a/cloud/pkg/cloudhub/servers/httpserver/secretsutil.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/secretsutil.go
@@ -3,6 +3,7 @@ package httpserver
 import (
 	"context"
 	"fmt"
+	"os"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -10,11 +11,10 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/kubeedge/kubeedge/cloud/pkg/edgecontroller/utils"
+	"github.com/kubeedge/kubeedge/common/constants"
 )
 
 const (
-	NamespaceSystem string = "kubeedge"
-
 	TokenSecretName      string = "tokensecret"
 	TokenDataName        string = "tokendata"
 	CaSecretName         string = "casecret"
@@ -24,6 +24,23 @@ const (
 	CloudCoreCertName    string = "cloudcoredata"
 	CloudCoreKeyDataName string = "cloudcorekeydata"
 )
+
+var NamespaceSystem string = constants.KubeEdgeNameSpace
+
+// set namesapce from env CLOUDCORE_POD_NAMESPACE
+func init() {
+	NamespaceSystem = setKubeedgeNamespce(os.Getenv("CLOUDCORE_POD_NAMESPACE"))
+}
+
+func setKubeedgeNamespce(kubeedgeNamespace string) string {
+	if kubeedgeNamespace != "" {
+		NamespaceSystem = kubeedgeNamespace
+	} else {
+		kubeedgeNamespace = constants.KubeEdgeNameSpace
+	}
+
+	return kubeedgeNamespace
+}
 
 func GetSecret(secretName string, ns string) (*corev1.Secret, error) {
 	cli, err := utils.KubeClient()
--- a/cloud/pkg/cloudhub/servers/httpserver/secretsutil_test.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/secretsutil_test.go
@@ -1,0 +1,29 @@
+package httpserver
+
+import "testing"
+
+func TestSetNamespace(t *testing.T) {
+	tests := []struct {
+		name              string
+		kubeedgeNamespace string
+		want              string
+	}{
+		{
+			name:              "null",
+			kubeedgeNamespace: "",
+			want:              "kubeedge",
+		},
+		{
+			name:              "edge-system",
+			kubeedgeNamespace: "edge-system",
+			want:              "edge-system",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := setKubeedgeNamespce(tt.kubeedgeNamespace); got != tt.want {
+				t.Errorf("setKubeedgeNamespce() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
--- a/tests/e2e/cloudcore/cloudcore_suite_test.go
+++ b/tests/e2e/cloudcore/cloudcore_suite_test.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2019 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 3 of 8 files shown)


daixiang0:
@shenkonghui please update unit test and e2e test since they do not use the custom namespace.

GsssC (commented):
https://github.com/kubeedge/kubeedge/blob/65b5f8bb40f181cd3cbaaad61c94b34ceb26dac9/cloud/pkg/leaderelection/leaderelection.go#L137

PTAL, how about modifying together

shenkonghui:
> https://github.com/kubeedge/kubeedge/blob/65b5f8bb40f181cd3cbaaad61c94b34ceb26dac9/cloud/pkg/leaderelection/leaderelection.go#L137
> 
> PTAL, how about modifying together



> https://github.com/kubeedge/kubeedge/blob/65b5f8bb40f181cd3cbaaad61c94b34ceb26dac9/cloud/pkg/leaderelection/leaderelection.go#L137
> 
> PTAL, how about modifying together

already change env to "CLOUDCORE_POD_NAMESPACE"

cloud/pkg/cloudhub/servers/httpserver/secretsutil.go
```diff
@@ -31,13 +30,17 @@ var NamespaceSystem string = "kubeedge"
...
+	NamespaceSystem = setKubeedgeNamespce(os.Getenv("CLOUDCORE_POD_NAMESPACE"))
+}
+
+func setKubeedgeNamespce(kubeedgeNamespace string) string {
 	if kubeedgeNamespace != "" {
-		klog.Infof("found env `CLOUDCORE_POD_NAMESPACE`, set namespce %v", kubeedgeNamespace)
 		NamespaceSystem = kubeedgeNamespace
-		return
+	} else {
+		kubeedgeNamespace = "kubeedge"
```
daixiang0, line 40: Use constant in common/constants/default.go.

cloud/pkg/cloudhub/servers/httpserver/secretsutil_test.go
```diff
@@ -0,0 +1,36 @@
+package httpserver
+
+import "testing"
+
+func Test_setKubeedgeNamespce(t *testing.T) {
```
daixiang0, line 5: Rename `TestSetNamespace`

cloud/pkg/cloudhub/servers/httpserver/secretsutil.go
```diff
@@ -27,6 +27,23 @@ const (
...
 )
 
+var NamespaceSystem string = constants.KubeEdgeNameSpace
+
+// set namesapce from env CLOUDCORE_POD_NAMESPACE
+func init() {
+	NamespaceSystem = setKubeedgeNamespce(os.Getenv("CLOUDCORE_POD_NAMESPACE"))
+}
+
+func setKubeedgeNamespce(kubeedgeNamespace string) string {
```
daixiang0, line 37: In which case, `os.Getenv("CLOUDCORE_POD_NAMESPACE")` return empty?

cloud/pkg/cloudhub/servers/httpserver/secretsutil_test.go
```diff
@@ -0,0 +1,36 @@
+package httpserver
+
+import "testing"
+
+func TestSetNamespace(t *testing.T) {
+	type args struct {
```
daixiang0, line 6: No need to use struct here.

tests/e2e/cloudcore/cloudcore_suite_test.go
```diff
@@ -0,0 +1,48 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cloudcore
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
```
daixiang0, line 22: Do not use dot import.

tests/e2e/cloudcore/cloudcore_test.go
```diff
@@ -0,0 +1,75 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cloudcore
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
```
daixiang0, line 22: Ditto.

tests/e2e/cloudcore/cloudcore_suite_test.go
shenkonghui: I see the case and other codes on the official website do this.
https://github.com/onsi/ginkgo/blob/master/example/books/books_suite_test.go
https://github.com/kubeedge/kubeedge/blob/master/tests/e2e/keadm/keadm_suite_test.go

daixiang0: Better to change it :)

cloud/pkg/cloudhub/servers/httpserver/secretsutil.go
shenkonghui: If this environment variable is not set, it should should return empty

daixiang0: Maybe put it as flag is more helpful, now we do not have configs based on env.

shenkonghui: > Maybe put it as flag is more helpful, now we do not have configs based on env.

The environment variables CLOUDCORE_POD_NAME and CLOUDCORE_POD_NAMESPACE have been used here to set the ReadinessGate.

>
https://github.com/kubeedge/kubeedge/blob/65b5f8bb40f181cd3cbaaad61c94b34ceb26dac9/cloud/pkg/leaderelection/leaderelection.go#L135
>

daixiang0: Ohh, it is to detect running as pod or not, not a config I think. We can discuss it at the following meeting up.

threestoneliu:
@shenkonghui `keadm gettoken` may also need to support custom namespace.

PR #2387, opened 2020-12-02
fix msg in nodestore compare error

threestoneliu:
**What type of PR is this?**
/kind bug

The managePod method  in synccontroller  will check pod is exist in kuberntes, if pod not found ,synccontroller will send a delete operation message to cloudhub, but the message's resourceversion is empty.  then channelq will overwrite the message that stored in nodestore . next time if trigger synccontroller.CompareResourceVersion will cause the cloudcore panic. So we can check wheather the message in node store is a delete operation message to avoid this panic.   

Fixes #

```release-note

```


fisherxu:
Make sense, thanks, we can also pick it to release-1.5 @threestoneliu 
/lgtm
/approve

threestoneliu:
done! [2394](https://github.com/kubeedge/kubeedge/pull/2394)

PR #2388, opened 2020-12-03
Keadm: optimize OS detect

daixiang0:
**What type of PR is this?**

/kind feature

Now we use OS release to detect OS then we need to add specific OS name by hand, while we only want to know which package manager the OS uses, so optimize it.

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -40,13 +40,8 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha1"
 )
 
-//Constants used by installers
+// Constants used by installers
 const (
-	UbuntuOSType   = "ubuntu"
-	RaspbianOSType = "raspbian"
-	DebianOSType   = "debian"
-	CentOSType     = "centos"
-
 	KubeEdgeDownloadURL  = "https://github.com/kubeedge/kubeedge/releases/download"
 	EdgeServiceFile      = "edgecore.service"
 	CloudServiceFile     = "cloudcore.service"
@@ -54,7 +49,8 @@ const (
 	KubeEdgePath         = "/etc/kubeedge/"
 	KubeEdgeUsrBinPath   = "/usr/local/bin"
 	KubeEdgeBinaryName   = "edgecore"
-	KubeCloudBinaryName  = "cloudcore"
+
+	KubeCloudBinaryName = "cloudcore"
 
 	KubeEdgeConfigDir        = KubeEdgePath + "config/"
 	KubeEdgeCloudCoreNewYaml = KubeEdgeConfigDir + "cloudcore.yaml"
@@ -67,6 +63,9 @@ const (
 
 	latestReleaseVersionURL = "https://kubeedge.io/latestversion"
 	RetryTimes              = 5
+
+	APT string = "apt"
+	YUM string = "yum"
 )
 
 //AddToolVals gets the value and default values of each flags and collects them in temporary cache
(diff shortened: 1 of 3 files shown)


keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -116,28 +114,33 @@ func (co *Common) SetOSInterface(intf types.OSTypeInstaller) {
...
-		return &CentOS{}
+	switch GetPackageManager() {
+	case APT:
+		return &DebOS{}
+	case YUM:
+		return &RpmOS{}
 	default:
-		fmt.Printf("This OS version is currently un-supported by keadm, %s", GetOSVersion())
-		panic("This OS version is currently un-supported by keadm,")
+		fmt.Println("Failed to detect apt/yum command, exit")
```
fisherxu, line 142:
```suggestion
		fmt.Println("Failed to detect package manager command, exit")
```
for supporting more types?

daixiang0: Maybe in other prs, `apt` and `yum` should cover most use cases.

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -42,21 +42,28 @@ import (
...
+	KubeEdgeConfigModulesYaml    = KubeEdgeConfPath + "/modules.yaml"
+
+	KubeEdgeCloudCertGenPath     = KubeEdgePath + "certgen.sh"
+	KubeEdgeEdgeCertsTarFileName = "certs.tgz"
+	KubeEdgeCloudConfPath        = KubeEdgePath + "kubeedge/cloud/conf"
+	KubeEdgeCloudCoreYaml        = KubeEdgeCloudConfPath + "/controller.yaml"
+	KubeEdgeCloudCoreModulesYaml = KubeEdgeCloudConfPath + "/modules.yaml"
+	KubeCloudBinaryName          = "cloudcore"
+
+	KubeEdgeConfigDir     = KubeEdgePath + "config/"
```
fisherxu, line 66: Delete the unused consts

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -96,28 +95,33 @@ func (co *Common) SetOSInterface(intf types.OSTypeInstaller) {
 	co.OSTypeInstaller = intf
 }
 
-//GetOSVersion gets the OS name
-func GetOSVersion() string {
-	cmd := NewCommand("source /etc/os-release && echo $ID")
+//GetPackageManager get package manager of OS
+func GetPackageManager() string {
+	cmd := NewCommand("command -v apt || command -v yum")
```
gccio, line 100: if `apt` is a alias, for example `alias apt="apt -y"`, `command -v apt` may not end with `apt`. 

```shell
$ docker run --rm -it ubuntu
root@91d3467eecaf:/# command -v apt
/usr/bin/apt
root@91d3467eecaf:/# alias apt="apt -y"
root@91d3467eecaf:/# command -v apt
alias apt='apt -y'
root@91d3467eecaf:/# which apt
/usr/bin/apt
root@91d3467eecaf:/#
```
maybe we could try with `which`

gccio: We can find the package manager under `/usr/bin/`, most package manager is under this path. And add a parameter in keadm, like `--package-manager` or other. User can provide their package manager to keadm manual.

If keadm have this option, we can provide a interface to user, for example `--package-manage=MYSCRIPT`. The script implement some func like `install mqtt`, . we just use `MYSCRIPT installl mqtt` to install mqtt and other things. user need to complete the install logic, install online or offline, all is ok. :)

daixiang0: Refer to https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script, `which` is not a good choice.

An alias does not make whole install cmd like `apt install -y mqtt` failed, besides, we use `sudo apt`, in most production environments, there is little alias in root account. If exists, fail output would raise error to notice users.

We can auto detect package manager cmd, no need to ask users to specify it. Also, install by  package manager still need network in most cases, it is outside this pr indeed.

gccio (commented):
/lgtm

PR #2389, opened 2020-12-03
Get EdgeNode ip before registerModules to fix stream module problem

lvchenggang:
What type of PR is this?
/kind feature

Premise: we set edged.nodeIP="" in edgecore.yaml.

After that when edgestream register to cloudstream, the SessionInternalIP of connection header  is null, so cloudstream will set Request.RemoteAddr as interalIP. Refer to https://github.com/kubeedge/kubeedge/pull/2020.

But, the Request.RemoteAddr is public network ip, it is different with internalIP most of time. So the problem will happen when use stream module.

NONE

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/kubeedge/beehive/pkg/core"
+	"github.com/kubeedge/kubeedge/common/constants"
 	"github.com/kubeedge/kubeedge/edge/cmd/edgecore/app/options"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/dbm"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin"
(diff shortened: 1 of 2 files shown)


edge/cmd/edgecore/app/server.go
```diff
@@ -144,6 +146,20 @@ func environmentCheck() error {
 	return nil
 }
 
+// get edge node local ip
+func GetNodeIp(config *v1alpha1.EdgeCoreConfig) *v1alpha1.EdgeCoreConfig {
```
daixiang0, line 150: Is it need export?

edge/cmd/edgecore/app/server.go
```diff
@@ -144,6 +146,20 @@ func environmentCheck() error {
...
+	if nodeIP := config.Modules.Edged.NodeIP; nodeIP != "" {
```
daixiang0, line 151: if config.Modules.Edged.NodeIP != "" {

daixiang0 (changes requested):

edge/cmd/edgecore/app/server.go
lvchenggang: This method has not been called elsewhere, feels it is no need to export currently.

edge/cmd/edgecore/app/server.go
lvchenggang: OK.

edge/cmd/edgecore/app/server.go
daixiang0: If so, no need this func. Just write as:
```
if config.Modules.Edged.NodeIP == "" {
  ...
  config.Modules.Edged.NodeIP = localIP
}
```

lvchenggang: OK.

PR #2392, opened 2020-12-03
fix message send problem

threestoneliu:
**What type of PR is this?**
/kind bug

MessageWriteLoop has two problem.
1. if we call nodeQueue.Get() method but don't call nodeQueue.Done() will cause  same message  key cannot be add to nodeQueue, so we may have same message could be send to edge.
Source Code url: [nodeQueue.Get](https://github.com/kubeedge/kubeedge/blob/master/vendor/k8s.io/client-go/util/workqueue/queue.go#L156) [nodeQueue.Add](https://github.com/kubeedge/kubeedge/blob/master/vendor/k8s.io/client-go/util/workqueue/queue.go#L121) [nodeQueue.Done](https://github.com/kubeedge/kubeedge/blob/master/vendor/k8s.io/client-go/util/workqueue/queue.go#L171).

2. when we call trimMessage(msg) ,  at the same time we will change  the message stored in nodeStore . if the edge node offline, message send failed, then when edge node is online, message send succsefully, the objectsync will create or update failed, because the message resource has no node info and find an error namespace.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -425,18 +425,20 @@ func (mh *MessageHandle) MessageWriteLoop(info *model.HubInfo, stopServe chan Ex
 		obj, exist, _ := nodeStore.GetByKey(key.(string))
 		if !exist {
 			klog.Errorf("nodeStore for node %s doesn't exist", info.NodeID)
+			nodeQueue.Done(key)
 			continue
 		}
 		msg := obj.(*beehiveModel.Message)
 
 		if !model.IsToEdge(msg) {
 			klog.Infof("skip only to cloud event for node %s, %s, content %s", info.NodeID, dumpMessageMetadata(msg), msg.Content)
+			nodeQueue.Done(key)
 			continue
 		}
 		klog.V(4).Infof("event to send for node %s, %s, content %s", info.NodeID, dumpMessageMetadata(msg), msg.Content)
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -425,29 +425,31 @@ func (mh *MessageHandle) MessageWriteLoop(info *model.HubInfo, stopServe chan Ex
...
+		trimMessage(copyMsg)
 
 		for {
 			conn, ok := mh.nodeConns.Load(info.NodeID)
 			if !ok {
 				time.Sleep(time.Second * 2)
 				continue
 			}
-			err := mh.sendMsg(conn.(hubio.CloudHubIO), info, msg, copyMsg, nodeStore)
+			err := mh.sendMsg(conn.(hubio.CloudHubIO), info, copyMsg, msg, nodeStore)
```
fisherxu, line 449: Also need to modify the parameters name in sendMsg func :)

threestoneliu: ok,  the parameter name in sendMsg func has been renamed.

fisherxu:
/lgtm
/approve
Also please cherry-pick to 1.5-release branch.

threestoneliu:
done! [pr: 2399](https://github.com/kubeedge/kubeedge/pull/2399)

PR #2393, opened 2020-12-03
fix synccontroller manage same name object error

threestoneliu:
**What type of PR is this?**
/kind bug

 Use stateful set to manage pod. new pod  will use the same name but the pod uid is different when the old pod is deleted. when synccontroller to managepod ,we need to distinguish them by uid。Otherwise we may find the pod with the same name will run in two different node, and three different objectsync but the spec.objectName is same.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/synccontroller/objectsync.go
+++ b/cloud/pkg/synccontroller/objectsync.go
@@ -7,6 +7,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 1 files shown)


fisherxu:
Thanks for the fixing! We get pod from podlister(local cache), when new pod in statefulset came in, the old still hadn't been deleted in cache?

threestoneliu:
> Thanks for the fixing! We get pod from podlister(local cache), when new pod in statefulset came in, the old still hadn't been deleted in cache?

The old pod had been deleted in cache. But if the cloudcore was stoped , and we use kubectl to  delete the pod with option --force --grace-period=0 , the new one will be shedule to other node before nodelifecyclecontroller mark all edgenode notReady。 Then we start cloudcore, problem may occur.

fisherxu:
> > Thanks for the fixing! We get pod from podlister(local cache), when new pod in statefulset came in, the old still hadn't been deleted in cache?
> 
> The old pod had been deleted in cache. But if the cloudcore was stoped , and we use kubectl to delete the pod with option --force --grace-period=0 , the new one will be shedule to other node before nodelifecyclecontroller mark all edgenode notReady。 Then we start cloudcore, problem may occur.

Looks good. Thanks.
/lgtm
/approve

PR #2396, opened 2020-12-03
fix the invalid memory address panic of common_test in eventbus

muxuelan:
**What type of PR is this?**
/kind feature

The func TestHubClientInit and TestLoopConnect under "kubeedge/edge/pkg/eventbus/common/util/common_test.go" build failed because of "invalid memory address or nil pointer dereference", so I add the init().

Fixes #

```release-note

```


fisherxu:
Has ant error logs?

muxuelan:
> Has any error logs?
As follows.
image

PR #2404, opened 2020-12-07
cloudcore: Use LocalIP as advertise address in New config like edgecore

zhijianli88:
both cloudcore and edgecore provide --minconfig to print their default
config. However they call different methods to get the pair ip addresss
before which are possible different. In this case `./hack/local-up-kubeedge.sh`
will fail with below errors:
F1207 09:55:14.851562 1549579 certmanager.go:91] Error: failed to get edge certificate from the cloudcore, error: Get "https://a.b.c.d:10002/edge.crt": x509: certificate is valid for u.x.y.z, not a.b.c.d

Signed-off-by: Li Zhijian <lizhijian@cn.fujitsu.com>

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/pkg/apis/componentconfig/cloudcore/v1alpha1/default.go
+++ b/pkg/apis/componentconfig/cloudcore/v1alpha1/default.go
@@ -17,20 +17,25 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"os"
 	"path"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	utilnet "k8s.io/apimachinery/pkg/util/net"
 	componentbaseconfig "k8s.io/component-base/config"
 
 	"github.com/kubeedge/kubeedge/common/constants"
 	metaconfig "github.com/kubeedge/kubeedge/pkg/apis/componentconfig/meta/v1alpha1"
+	"github.com/kubeedge/kubeedge/pkg/util"
 )
 
 // NewDefaultCloudCoreConfig returns a full CloudCoreConfig object
 func NewDefaultCloudCoreConfig() *CloudCoreConfig {
-	advertiseAddress, _ := utilnet.ChooseHostInterface()
+	hostnameOverride, err := os.Hostname()
+	if err != nil {
+		hostnameOverride = constants.DefaultHostnameOverride
+	}
+	advertiseAddress, _ := util.GetLocalIP(hostnameOverride)
 
 	c := &CloudCoreConfig{
 		TypeMeta: metav1.TypeMeta{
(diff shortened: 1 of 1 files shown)


daixiang0:
Do you have multiple network interfaces on your machine? I can not reproduce it at local. Besides, ensure no iptable rules or firewall to forbid them.

zhijianli88:
> Do you have multiple network interfaces on your machine? I can not reproduce it at local. Besides, ensure no iptable rules or firewall to forbid them.

that's true, i have multiple NICs that route to different destionation. any my hostname is bind to 10.160.226.45 in eth0 which is different from the default route deivce wlan0

so previously, cloudcore got 10.124.3.100 and edgecore got 10.160.226.45.
```
$ route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         _gateway        0.0.0.0         UG    600    0        0 wlan0
10.124.0.0      0.0.0.0         255.255.0.0     U     600    0        0 wlan0
10.160.0.0      10.160.226.1    255.255.0.0     UG    100    0        0 eth0
10.160.226.0    0.0.0.0         255.255.255.0   U     100    0        0 eth0
link-local      0.0.0.0         255.255.0.0     U     1000   0        0 virbr0
172.17.0.0      0.0.0.0         255.255.0.0     U     0      0        0 docker0
172.18.0.0      0.0.0.0         255.255.0.0     U     0      0        0 br-f42ed31bb060
172.20.0.0      0.0.0.0         255.255.0.0     U     0      0        0 br-7c16f098416a
192.168.122.0   0.0.0.0         255.255.255.0   U     0      0        0 virbr0
193.160.0.0     10.160.226.1    255.255.0.0     UG    100    0        0 eth0
193.168.0.0     10.160.226.1    255.255.0.0     UG    100    0        0 eth0
```

daixiang0:
`hack/local-up` is only for test, please use `keadm init` and `CHECK_EDGECORE_ENVIRONMENT="false" keadm join` to test, now we do not have multiple NICs machine to test to ensure productive deployment.

zhijianli88:
> `hack/local-up` is only for test, please use `keadm init` and `CHECK_EDGECORE_ENVIRONMENT="false" keadm join` to test, now we do not have multiple NICs machine to test to ensure productive deployment.

got it.

/hold

fisherxu (commented):
For cloudside, we should use the `host's default interface` through `ChooseHostInterface` as the advertise addr. Ref: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/net/interface.go#L432  
 If it can't be accessed from edge side, we should set it to an external addr.

For edgside, the `nodeIP` should be got through `gitlocalIP` as internal IP. The `server addr` should be same as cloud, btw, it's just used as testing, in real env, we should set it to real cloud external addr.

PR #2414, opened 2020-12-09
Fix edgehub synckeeper use unbuffer channel error

### LINKED ISSUE #2411: Why edgehub keepchannel use the channel without buffer?
Edgehub keepchannel init method: [init method](https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/process.go#L44)
I have test the default statement in select, if we use no buffer channel, we will have enter default statement in some times, but if we use the buffer channel, we will only enter default when all the other  case statement doesn't match.
[buffer channel](https://play.golang.org/p/sHn2oucu_e9)
[nobuffer channel](https://play.golang.org/p/QJDH9ZlP41i)

EdgeCore log: 
edgehuberror

**Environment**:
<details><summary>Kubernets Version</summary>

```console
$ kubectl --version
# paste output here
```

</details>

```console
$ cloudcore --version
# paste output here
```

</details>

<details><summary>EdgeCore Version</summary>

```
$ edgecore --version
# paste output here
```

</details>

threestoneliu:
**What type of PR is this?**
/kind bug

EdgeHub Use buffer channel to avoid some error.
EdgeCore log

Fixes #2411

```release-note

```

### DIFF
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -41,7 +41,7 @@ func (eh *EdgeHub) addKeepChannel(msgID string) chan model.Message {
 	eh.keeperLock.Lock()
 	defer eh.keeperLock.Unlock()
 
-	tempChannel := make(chan model.Message)
+	tempChannel := make(chan model.Message, 1)
 	eh.syncKeeper[msgID] = tempChannel
 
 	return tempChannel
--- a/edge/pkg/edgehub/process_test.go
+++ b/edge/pkg/edgehub/process_test.go
@@ -145,22 +145,14 @@ func TestSendToKeepChannel(t *testing.T) {
 			expectedError:       fmt.Errorf("failed to get sync keeper channel, messageID:%+v", *message),
 		},
 		{
-			name: "Negative Test Case without syncKeeper Error ",
-			hub: &EdgeHub{
-				syncKeeper: make(map[string]chan model.Message),
-			},
-			message:             model.NewMessage("test_id"),
-			keepChannelParentID: "test_id",
-			expectedError:       fmt.Errorf("failed to send message to sync keep channel"),
-		},
-		{
 			name: "Send to keep channel with valid input",
 			hub: &EdgeHub{
 				syncKeeper: make(map[string]chan model.Message),
 			},
 			message:             model.NewMessage("test_id"),
 			keepChannelParentID: "test_id",
-			expectedError:       nil},
+			expectedError:       nil,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


fisherxu:
@threestoneliu Good patch, for the unbuffered channel, if the receive goroutine hasn't been ready, the send goroutine will block or entry the default. So change it to buffer channel is fine, but another thought, what's the impact of deleting the `default` section in select?

threestoneliu:
> @threestoneliu Good patch, for the unbuffered channel, if the receive goroutine hasn't been ready, the send goroutine will block or entry the default. So change it to buffer channel is fine, but another thought, what's the impact of deleting the `default` section in select?

Cool! Deleting the `default` section is also a good way to solve this problem. But I think we'd better to discuss a final conclusion in this [issue](https://github.com/kubeedge/kubeedge/issues/2411).

GsssC:
I think the key is to make sure response temp channel are made before calling func sendToKeepChannel
How about below? : ) 
``` go
func (eh *EdgeHub) sendToCloud(message model.Message) error {
	sendErrChan := make(chan error,1)
	syncKeep := func(message model.Message) {
		tempChannel := eh.addKeepChannel(message.GetID())
		sendTimer := time.NewTimer(time.Duration(config.Config.Heartbeat) * time.Second)
		select {
		case response := <-tempChannel:
			sendTimer.Stop()
			beehiveContext.SendResp(response)
			eh.deleteKeepChannel(response.GetParentID())
		case <-sendTimer.C:
			klog.Warningf("timeout to receive response for message: %+v", message)
			eh.deleteKeepChannel(message.GetID())
		case err :=<- sendErrChan:
			klog.Warningf("failed to receive response for message : %v", err)
			eh.deleteKeepChannel(message.GetID())
		}
	}

	if message.IsSync() {
		go syncKeep(message)
	}
	eh.keeperLock.Lock()
	err := eh.chClient.Send(message)
	eh.keeperLock.Unlock()
	if err != nil {
		err = fmt.Errorf("failed to send message, error: %v", err)
		errChan <- err
		klog.Errorf("failed to send message: %v", err)
		return err
	}
	return nil
}
```

threestoneliu:
> I think the key is to make sure response temp channel are made before calling func sendToKeepChannel
> How about below? : )
> 
> ```go
> func (eh *EdgeHub) sendToCloud(message model.Message) error {
> 	sendErrChan := make(chan error,1)
> 	syncKeep := func(message model.Message) {
> 		tempChannel := eh.addKeepChannel(message.GetID())
> 		sendTimer := time.NewTimer(time.Duration(config.Config.Heartbeat) * time.Second)
> 		select {
> 		case response := <-tempChannel:
> 			sendTimer.Stop()
> 			beehiveContext.SendResp(response)
> 			eh.deleteKeepChannel(response.GetParentID())
> 		case <-sendTimer.C:
> 			klog.Warningf("timeout to receive response for message: %+v", message)
> 			eh.deleteKeepChannel(message.GetID())
> 		case err :=<- sendErrChan:
> 			klog.Warningf("failed to receive response for message : %v", err)
> 			eh.deleteKeepChannel(message.GetID())
> 		}
> 	}
> 
> 	if message.IsSync() {
> 		go syncKeep(message)
> 	}
> 	eh.keeperLock.Lock()
> 	err := eh.chClient.Send(message)
> 	eh.keeperLock.Unlock()
> 	if err != nil {
> 		err = fmt.Errorf("failed to send message, error: %v", err)
> 		errChan <- err
> 		klog.Errorf("failed to send message: %v", err)
> 		return err
> 	}
> 	return nil
> }
> ```

Whether syncKeeper goroutine can receive data is not under our control, it is determined by go.I think this modification can delay the occurrence of the problem, but cannot completely solve the problem. @fisherxu What do you think?

threestoneliu:
When we remove the default section, if the cloudcore send message more than one time the edgehub's routeToEdge goroutine may be blocked, so i think use buffer channel is a better choice.

PR #2418, opened 2020-12-11
Keadm debug get serialization optimization

Peeknut:
**What type of PR is this?**
 /kind design

Refactor func ParseMetaToAPIList

Fixes #

### DIFF
--- a/keadm/cmd/keadm/app/cmd/debug/get.go
+++ b/keadm/cmd/keadm/app/cmd/debug/get.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/klog/v2"
 	api "k8s.io/kubernetes/pkg/apis/core"
+	k8s_v1_api "k8s.io/kubernetes/pkg/apis/core/v1"
 	k8sprinters "k8s.io/kubernetes/pkg/printers"
 	printersinternal "k8s.io/kubernetes/pkg/printers/internalversion"
 	"k8s.io/kubernetes/pkg/printers/storage"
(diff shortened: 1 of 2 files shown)


keadm/cmd/keadm/app/cmd/debug/get.go
```diff
@@ -571,73 +572,68 @@ func ParseMetaToAPIList(metas []dao.Meta) (res []runtime.Object, err error) {
...
 		case model.ResourceTypePod:
-			var pod api.Pod
-
+			var pod v1.Pod
+			var apiPod api.Pod
 			if err = json.Unmarshal([]byte(v.Value), &pod); err != nil {
 				return nil, err
 			}
-			if err = json.Unmarshal(metadata, &pod.ObjectMeta); err != nil {
+			if err := k8s_v1_api.Convert_v1_Pod_To_core_Pod(&pod, &apiPod, nil); err != nil {
```
fisherxu, line 583: Why do we need this conversion? For all external client, we should use the versioned api.

Peeknut: This func ParseMetaToAPIList is called by func HumanReadablePrint, the return result of the function is used as the parameter to generate the table.

1）There is no corresponding table printing function for the specific version, so it needs to be converted to unversioned api to use the table printing function.

2）And the version information will eventually be placed in the table.ResourceVersion. :)

fisherxu: Okey and does kubectl use this convert and `unversioned api`?

GsssC: @fisherxu kubectl get resource Table from apiserver directly, which is generated from internal unversioned api resources.

> 为了避免上述各种潜在的局限性，客户端可以请求服务器端返回对象的表格（Table） 表现形式，从而将打印输出的特定细节委托给服务器。 Kubernetes API 实现标准的 HTTP 内容类型（Content Type）协商：为 GET 调用 传入一个值为 application/json;as=Table;g=meta.k8s.io;v=v1beta1 的 Accept 头部即可请求服务器以 Table 的内容类型返回对象

GsssC:
```
GET /api/v1/pods
Accept: application/json;as=Table;g=meta.k8s.io;v=v1beta1
---
200 OK
Content-Type: application/json

{
    "kind": "Table",
    "apiVersion": "meta.k8s.io/v1beta1",
    ...
    "columnDefinitions": [
        ...
    ]
}
```

qingchen1203:
> /cc @qingchen1203 PTAL



> /cc @qingchen1203 PTAL

/lgtm

PR #2421, opened 2020-12-11
Install golangci-lint for local test

fisherxu:
**What type of PR is this?**
/kind test

Install golangci-lint for local test

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/lib/install.sh
+++ b/hack/lib/install.sh
@@ -17,7 +17,7 @@
 # check if kubectl installed
 function check_kubectl {
   echo "checking kubectl"
-  which kubectl >/dev/null 2>&1
+  command -v kubectl >/dev/null 2>&1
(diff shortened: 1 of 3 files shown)


hack/lib/install.sh
```diff
@@ -47,6 +47,24 @@ function check_kind {
   fi
 }
 
+# check if golangci-lint installed
+function check_golangci-lint {
+  echo "checking golangci-lint"
+  which golangci-lint >/dev/null 2>&1
```
daixiang0, line 53: Better to use `command -v`, see https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script

daixiang0 (changes requested):

daixiang0:
depend on #2422

PR #2425, opened 2020-12-11
Fix the #2150 bug

### LINKED ISSUE #2150: The data in the edgecore.db database is scattered and incomplete
**What happened**:
When I use the sqlite3 tool to view the edgecore.db database data, I found that part of the data is scattered and incomplete.
E.g 1：
When getting current edge node information (`type = node`) from the database ， only get “OK”：
`table: meta`   
`key: default/node/k8s-edge`   
`type: node`   
`value: "OK"`
and another field `nodestatus` saves node status information.
The k8s resource object configuration of node cannot be obtained completely, and the `metadata`/`spec` field is missing，and unable to get the k8s resource object configuration of node.
 
There are similar situations in `pod` and `podstatus`: after the values of type=pod are initialized, the pod.status field will not be updated, update only in `podstatus`.

Why do we need so many fields without considering putting them in the same field? For example, all in node and pod.

qclc:
**What type of PR is this?**
/kind bug

Fix#2150 bug. The data in the edgecore.db database is scattered and incomplete

Fixes #2150 

Analysis of the root cause of bug is in the comments in #2150 

NONE

### DIFF
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -97,12 +97,20 @@ func (eh *EdgeHub) dispatch(message model.Message) error {
 		return fmt.Errorf("msg_group not found")
 	}
 
-	isResponse := eh.isSyncResponse(message.GetParentID()) || message.GetOperation() == model.ResponseOperation
-	if !isResponse {
+	//isResponse := eh.isSyncResponse(message.GetParentID()) || message.GetOperation() == model.ResponseOperation
+	//if !isResponse {
+	//	beehiveContext.SendToGroup(md, message)
+	//	return nil
+	//}
+	//return eh.sendToKeepChannel(message)
+
+	isSyncResp := message.GetOperation() == model.ResponseOperation
+	if !isSyncResp {
 		beehiveContext.SendToGroup(md, message)
-		return nil
+	} else {
+		beehiveContext.SendResp(message)
 	}
-	return eh.sendToKeepChannel(message)
+	return nil
 }
 
 func (eh *EdgeHub) routeToEdge() {
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -97,7 +97,7 @@ func (eh *EdgeHub) dispatch(message model.Message) error {
 		return fmt.Errorf("msg_group not found")
 	}
 
-	isResponse := eh.isSyncResponse(message.GetParentID())
+	isResponse := eh.isSyncResponse(message.GetParentID()) || message.GetOperation() == model.ResponseOperation
 	if !isResponse {
 		beehiveContext.SendToGroup(md, message)
 		return nil


daixiang0:
Please add unit test or e2e test for this case.

fisherxu:
Looks good.
/approve

fisherxu:
Can add UT case here. Ref: https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/process_test.go#L90 @qclc

qclc:
> Can add UT case here. Ref: https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/process_test.go#L90 @qclc

qclc:
> Please add unit test or e2e test for this case.

Thanks for your reply, I will try to add the unit test.

qclc:
> Can add UT case here. Ref: https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/process_test.go#L90 @qclc
Thanks for the suggestion, I will

qclc:
I find a problem with the redundant logic in syncKeeper, and here are the problem analysis and optimization recommendations. Specific code improvements are still being made locally.

qclc:
# Problem

EdgeHub module's member **syncKeeper** has the same effect as beehiveContext's member **anonChannels**. Removing syncKeeper and related logic has no effect on Sync sending and Response message receiving between the cloud and edge.

# Problem analysis

The following figure shows the general flow of edge's modules sending Sync messages to the cloud and receiving response messages from the cloud. In the figure, (3,5,6) and (1,8) marked in red have the problem of logic duplication:

Sync+Resp消息流简图en

## Contrast syncKeeper with anonChannels

- **syncKeeper**: It is a member of the EdgeHub module and is used to save Sync message records uploaded to the cloud and to determine whether a message received from the cloud is a response message.

- **anonChannels**: It is a member of beehiveContext(ChannelContext) and is used to keep records of Sync type messages sent by other modules, including Sync messages sent to the cloud

## The process of sending Sync messages to the cloud and receiving Response messages

The logic associated with **syncKeeper** member is as follows:

- `sendToCloud(...)`: It sends messages from the EdgeHub module to the cloud. If it is a Sync message, it saves (SyncMsgID, channel) record in the syncKeeper member and sets a timer (15 seconds) to wait for the Response message.
- `dispatch(...)`: According to the message type received from the cloud, this function dispatches the message to the corresponding module. If it is a Response message, it is put into the corresponding Channel in the syncKeeper member.

The logic associated with **anonChannels** member is as follows:

- `SendSync(...)`：When this function is called to send Sync messages, save (SyncMsgID, channel) record in an anonChannels member and set a timer (default 30s, customizable) to wait for the response message to come in.

- `SendResp(...)`：Its job is to pass the response message to the corresponding channel in the anonChannels member.

The following is the specific flow chart of `sendToCloud(...), dispatch(...), SendSync(...), SendResp(...)` 

Sync+Resp详细函数调用流程en

As you can see, **the logic in the red section of the flowchart above is repeated, and the same effect can be achieved by removing syncKeeper members and related logic**.

# Optimization proposal

From the above analysis, it can be seen that syncKeeper members act similarly to anonChannels members. If syncKeeper member and the corresponding logic are deleted, when a response message is received, the `beehivecontext. SendResp(response)` function is called directly to pass the response message to anonChannels (a member of ChannelContext). 

The following is the modified flow chart of `sendToCloud(...), dispatch(...), SendSync(...), SendResp(...)`. Where `sendToCloud(...)` and `dispatch(...)` functions are modified, `SendSync(...)` and `SendResp(...)` functions are not modified.

2425优化后逻辑图en

qclc:
The latest submission is to skip the syncKeep logic and directly send the response message to the Beehive module for processing, which has passed the test so far:
https://github.com/kubeedge/kubeedge/pull/2425/commits/1bfcf29f136c07872e44509b2627b84208c7b48f

qclc:
The above is my problem analysis, please read the analysis and make suggestions. @fisherxu @daixiang0 @threestoneliu @GsssC

daixiang0:
@qclc looks good, you can go ahead :)

kevin-wangzefeng:
Let me know when you are ready for another round of view

GsssC:
/close
fix by #2614

PR #2438, opened 2020-12-16
modify unit test of process.go in metamanager

muxuelan:
**What type of PR is this?**
/kind test

modify unit test of process.go in metamanager

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -507,7 +507,7 @@ func (m *metaManager) processNodeConnection(message model.Message) {
 	}
 }
 
-func (m *metaManager) processSync(message model.Message) {
+func (m *metaManager) processSync() {
 	m.syncPodStatus()
 }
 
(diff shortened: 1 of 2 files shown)


edge/pkg/metamanager/process.go
```diff
@@ -508,10 +508,10 @@ func (m *metaManager) processNodeConnection(message model.Message) {
 }
 
 func (m *metaManager) processSync(message model.Message) {
-	m.syncPodStatus()
+	m.syncPodStatus(message)
```
fisherxu, line 511: The `message` is no need here?

muxuelan: The message is just used to tell the edge to sync podstates with its operation，it's dispensable in func processSync , I add it for UT that it will send err message to cloud when unmarshal podstatus failed.

fisherxu: Actually the `feedbackError` will return the message to edged, because it will judge the message source, the source is edged, right? Btw, I think we can delete the message parameter directly here.

PR #2446, opened 2020-12-17
Fix bug: keadm doesn't delete file directly when checkSum is failed

XiaoJiangWang:
**What type of PR is this?**

/kind bug

keadm shouldn't delete file directly when checkSum is failed

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -490,9 +490,8 @@ func checkSum(filename, checksumFilename string, version semver.Version, tarball
 	}
 
 	if getDesiredCheckSum.GetStdOut() != getActualCheckSum.GetStdOut() {
-		fmt.Printf("Failed to verify the checksum of %s, try to download it again ... \n\n", filename)
-		//Cleanup the downloaded files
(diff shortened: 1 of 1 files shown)


fisherxu:
Thanks for the fixing, and if checkSum failed, what's the usage of the package?

XiaoJiangWang:
> Thanks for the fixing, and if checkSum failed, what's the usage of the package?

When checkSum failed, keadm will ask confirm.
If user confirms to download again, keadm will delete the package.
But if user refuses to download again, keadm should use this package to install continually.
You can find in https://github.com/kubeedge/kubeedge/blob/release-1.5/keadm/cmd/keadm/app/cmd/util/common.go#L378

Now the package is deleted before keadm get user's confirmation. So when user refuses to download again, keadm will panic later.

fisherxu:
> > Thanks for the fixing, and if checkSum failed, what's the usage of the package?
> 
> When checkSum failed, keadm will ask user's confirmation to delete the package and download again.
> If user confirms to download again, keadm will delete the package.
> But if user refuses to download again, keadm should use this package to install continually.
> You can find in https://github.com/kubeedge/kubeedge/blob/release-1.5/keadm/cmd/keadm/app/cmd/util/common.go#L378
> 
> Now the package is deleted before keadm get user's confirmation. So when user refuses to download again, keadm will panic later.

/lgtm
/approve

PR #2448, opened 2020-12-17
edgehub: merge same resource query msg.

### LINKED ISSUE #2319: Pod uses the same configmap multiple times, causing the first creation speed to be delayed
**What happened**:
The pod uses the same configmap multiple times, causing a delay in the first creation of the pod.
In the case that the configmap information is not sent to metaManager, Edged will send multiple queries for the same configmap to cloudcore. However, cloudcore compares the resourceVersion of the response message, finally only one that can respond successfully, and the other query requests time out, which causes the first creation of the pod to fail and wait for the next creation. Next time, you can read data directly from the metaManager.

**Environment**:
<details><summary>Kubernets Version</summary>

```console
$ kubelet --version
# paste output here
$ kubectl version
# paste output here
```

</details>

```console
$ cloudcore --version
# paste output here
```

</details>

<details><summary>EdgeCore Version</summary>

```console
$ edgecore --version
# paste output here
```

</details>

**CloudSide Environment**:
<details><summary>Hardware configuration</summary>

```console
$ lscpu
# paste output here
```

</details>

<details><summary>OS</summary>

```console
$ cat /etc/os-release
# paste output here
```

</details>

<details><summary>Kernel</summary>

```console
$ uname -a
# paste output here
```

</details>

<details><summary>Go version</summary>

```console
$ go version
# paste output here
```

</details>

<details><summary>Others</summary>
</details>

**EdgeSide Environment**:
<details><summary>Hardware configuration</summary>

```console
$ lscpu
# paste output here
```

</details>

<details><summary>OS</summary>

```console
$ cat /etc/os-release
# paste output here
```

</details>

<details><summary>Kernel</summary>

```console
$ uname -a
# paste output here
```

</details>

<details><summary>Go version</summary>

```console
$ go version
# paste output here
```

</details>

<details><summary>Others</summary>
</details>

gccio:
**What type of PR is this?**
/kind bug

if we create a pod with some volumes, and these volume's origin is a different key in one configmap. edgecore will send multi query request to cloudcore. cloudcore store message in a sync.Map, the `key` is `resource`. these configmap's query request have the same `resource`, so some request message may be lost. and then the edgecore's request will be timeout.

Fixes #2319 

```release-note
NONE
```

### DIFF
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -2,6 +2,7 @@ package edgehub
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


Peeknut:
Is the var `msgStoreLock sync.Mutex` necessary? 

The structure `sync.Map ` contains the lock,  funcs such as `Load()` and `Delete()` include the operations of acquiring and releasing locks.

gccio:
> Is the var `msgStoreLock sync.Mutex` necessary?
> 
> The structure `sync.Map ` contains the lock, funcs such as `Load()` and `Delete()` include the operations of acquiring and releasing locks.

Thanks. I review my code and update it. And now only use one bigger lock. 
I think we need to ensure that no more request be added to the `model.Message` list while older request has been responsed. If not, we may lost the new message. Because I will delete the list from `msgStore` when old reqeusts response completed.

edge/pkg/edgehub/process.go
```diff
@@ -128,10 +130,44 @@ func (eh *EdgeHub) routeToEdge() {
...
+		err             error
+		resource        = message.GetResource()
+		operation       = message.GetOperation()
+		isQuery         = (operation == model.QueryOperation)
+		_, restype, _   = metamanager.ParseResource(resource)
+		isConfigMap     = restype == "configmap"
+		needSendSyncMsg = true
+	)
+
+	if isQuery && isConfigMap {
```
fisherxu, line 149: Thanks for the fixing, except configmap, if query other objects, we may also need to add code here, so do we need to find a general solution for it?

gccio: I remove the flag `isConfigMap`.

PR #2451, opened 2020-12-18
modify log

LittleBoy18:
**What type of PR is this?**
modify log error

fix log level

```release-note

```


daixiang0:
It is not a good choice to use error level, since here we do not handle the error or do other things.

PR #2463, opened 2020-12-25
edgecore: support customize node labels、taints and annotations

### LINKED ISSUE #1599: Expose --node-labels and --register-with-taints in EdgeCore config
expose `--node-labels` and `--register-with-taints` in EdgeCore Config
**Why is this needed**:
`--node-labels` and `--register-with-taints` are very useful configs when setting up a new node. We should support them in EdgeCore Config

gccio:
**What type of PR is this?**

/kind feature

support indicates annotations、labels and taints in edgecore's config file

Fixes #1599 

NONE

```release-note
NONE
```


relaxtheo:
This feature supported? How to use it? In keadm join?

ttlv:
> This feature supported? How to use it? In keadm join?

nope.it's not for when use keadm for joining,just modify the config file of edgecore

PR #2468, opened 2020-12-27
keadm: clean useless file after keadm_e2e test

gccio:
**What type of PR is this?**

/kind feature

1. remove keadm.test and config.json before keadm_e2e.sh exit. keep directory clean.
2. should use `2>&1` to redirect stderr to stdout.

```
NONE
```

```release-note
NONE
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -398,7 +398,7 @@ func killKubeEdgeBinary(proc string) error {
 
 //isKubeEdgeProcessRunning checks if the given process is running or not
 func isKubeEdgeProcessRunning(proc string) (bool, error) {
-	procRunning := fmt.Sprintf("pidof %s 2&>1", proc)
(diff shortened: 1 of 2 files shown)


keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -398,7 +398,7 @@ func killKubeEdgeBinary(proc string) error {
 
 //isKubeEdgeProcessRunning checks if the given process is running or not
 func isKubeEdgeProcessRunning(proc string) (bool, error) {
-	procRunning := fmt.Sprintf("pidof %s 2&>1", proc)
+	procRunning := fmt.Sprintf("pidof %s 2>&1", proc)
```
daixiang0, line 401: Why change it?

fisherxu: Looks like `2>&1` is right?

daixiang0: Oh my mistake.

PR #2469, opened 2020-12-27
keadm: optimize keadm_e2e.sh cleanup function

gccio:
**What type of PR is this?**

/kind bug

Shell use `set -Ee`. So the shell will exit if any subcommand or pipeline returns a non-zero status. 

`sudo pkill cloudcore/edgecore` will return code 1 if the process not running, then the shell may exit unexpected.

```
NONE
```

```release-note
NONE
```


daixiang0:
lgtm

PR #2470, opened 2020-12-28
Keadm: optimize reset detect logic

### LINKED ISSUE #2464: When cloudcore not running，keadm reset failed
**What happened**:
I have wrong certs，so I wanna use `keadm reset` to reset node，but it failed and have some errors when check `edge`

code located at `keadm/cmd/keadm/app/cmd/util/common.go:411` 
image

skip error，just clean what exist

On cloudcore，`pkill cloudcore` and then run `keadm reset`

I think we need update cloudcore to running as `systemd service` or when check command failed，just return `false，nil`

**Environment**:

```console
$ cloudcore --version

KubeEdge v1.5.0
```

daixiang0:

**What type of PR is this?**
 /kind cleanup

To detect which module running, no need to check it run as a service or not, so simplify it.

Fixes #2464

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/reset.go
+++ b/keadm/cmd/keadm/app/cmd/reset.go
@@ -67,7 +67,7 @@ func NewKubeEdgeReset(out io.Writer, reset *common.ResetOptions) *cobra.Command 
 		Long:    resetLongDescription,
 		Example: resetExample,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			whoRunning, err := util.IsCloudCore()
+			whoRunning, err := util.RunningModule()
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


keadm/cmd/keadm/app/cmd/reset.go
```diff
@@ -67,15 +67,16 @@ func NewKubeEdgeReset(out io.Writer, reset *common.ResetOptions) *cobra.Command
...
 			if err != nil {
 				return err
 			}
 			switch whoRunning {
 			case common.KubeEdgeEdgeRunning:
 				IsEdgeNode = true
 			case common.NoneRunning:
-				return fmt.Errorf("None of KubeEdge components are running in this host")
+				fmt.Println("None of KubeEdge components are running in this host, exit")
+				os.Exit(0)
```
fisherxu, line 79: Has any difference with `return fmt.Errorf`?

fisherxu (commented):
Looks good, but how does this PR fixed the issue linked?

keadm/cmd/keadm/app/cmd/reset.go
daixiang0: If return err, `keadm` would print debug stack and help info which becomes confused for users.

Old logic would detect service exists or not, in fact we just consider binary running or not here, so simplify logic would fix the issue.

PR #2471, opened 2020-12-28
Fix bug: the websocket connection timeout setting doesn't take effect

XiaoJiangWang:
**What type of PR is this?**

Set the timeout period before sending and reading data from the long connection

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/edgehub/clients/wsclient/websocket.go
+++ b/edge/pkg/edgehub/clients/wsclient/websocket.go
@@ -105,6 +105,10 @@ func (wsc *WebSocketClient) UnInit() {
 
 //Send sends the message as JSON object through the connection
 func (wsc *WebSocketClient) Send(message model.Message) error {
+	err := wsc.connection.SetWriteDeadline(time.Now().Add(wsc.config.WriteDeadline))
+	if err != nil {
+		return err
+	}
 	return wsc.connection.WriteMessageAsync(&message)
 }
 


daixiang0:
Is it better to set timeout when create connection?

XiaoJiangWang:
> Is it better to set timeout when create connection?

you can find the example in gorilla/websocket.
https://github.com/gorilla/websocket/blob/master/examples/chat/client.go#L91

The write deadline must be setted before every sending process, because the timeout time stamp is calculated from the current time.

daixiang0:
lgtm

edge/pkg/edgehub/clients/wsclient/websocket.go
```diff
@@ -105,13 +105,22 @@ func (wsc *WebSocketClient) UnInit() {
...
+		return err
+	}
 	return wsc.connection.WriteMessageAsync(&message)
 }
 
 //Receive reads the binary message through the connection
 func (wsc *WebSocketClient) Receive() (model.Message, error) {
+	var err error
 	message := model.Message{}
-	err := wsc.connection.ReadMessage(&message)
```
fisherxu, line 114: Looks good, but seems `ReadMessage` hadn't use this Deadline in  Underlying `WSConnection`?

fisherxu: Writing use it here: https://github.com/kubeedge/kubeedge/blob/master/staging/src/github.com/kubeedge/viaduct/pkg/conn/ws.go#L179

XiaoJiangWang: > Looks good, but seems `ReadMessage` hadn't used this Deadline in Underlying `WSConnection`?

Yes. That's my mistake. `ReadMessage` gets messages from messageFifo in `WSConnection`, the read deadline setting also does't take effect in this method.
But CloudHub won't regularly send messages to the edge to make sure the websocket connection is alive. So is it necessary to set a timeout when read message?

XiaoJiangWang: delete the code for setting read message deadline

fisherxu: ok, I'm also not sure, maybe we don't need set read deadline?

PR #2472, opened 2020-12-28
Refactor informer and client used in cloudcore

threestoneliu:
**What type of PR is this?**

 /kind cleanup

Refactor informer and client used in cloudcore.
1. The client which used to access k8s resource was created where they need . So the QPS、Burst Configure item in cloudcore.yaml may not control the speed of all client in cloudcore. 
2. There has many infomers used in cloudcore is dupplicate, such as pod, svc, objectsync and so on, which will  cost twice memory usage and may has some problem to manage those informers. 

Fixes #

```release-note

```


cloud/cmd/cloudcore/app/server.go
```diff
@@ -98,9 +104,9 @@ kubernetes controller which manages devices so that the device metadata/status d
 
 // registerModules register all the modules started in cloudcore
 func registerModules(c *v1alpha1.CloudCoreConfig) {
-	cloudhub.Register(c.Modules.CloudHub, c.KubeAPIConfig)
-	edgecontroller.Register(c.Modules.EdgeController, c.KubeAPIConfig, "", false)
+	cloudhub.Register(c.Modules.CloudHub)
+	edgecontroller.Register(c.Modules.EdgeController, "", false)
```
daixiang0, line 108: Remove apiconfig directly like cloudhub does?

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -77,7 +76,7 @@ func SortInitContainerStatuses(p *v1.Pod, statuses []v1.ContainerStatus) {
 
 // UpstreamController subscribe messages from edge and sync to k8s api server
 type UpstreamController struct {
-	kubeClient   *kubernetes.Clientset
+	kubeClient   client.KubeEdgeClient
```
daixiang0, line 79: Let's use `Get()` of informers in https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/edgecontroller/controller/upstream.go#L495-L534

threestoneliu: done

cloud/pkg/common/informers/informers.go
```diff
@@ -0,0 +1,167 @@
...
+	ConfigMap() cache.SharedIndexInformer
+	Secrets() cache.SharedIndexInformer
+	Service() cache.SharedIndexInformer
+	Endpoints() cache.SharedIndexInformer
+	Node() cache.SharedIndexInformer
+	EdgeNode() cache.SharedIndexInformer
+	ClusterObjectSync() cache.SharedIndexInformer
+	ObjectSync() cache.SharedIndexInformer
+	Device() cache.SharedIndexInformer
+	Start(stopCh <-chan struct{})
```
fisherxu, line 33: If we put all core and crd informers in a global informers, maybe it's not friendly for the expansion and decoupling.

So I think we can use sharedinformer for the core apis and for the crd, just use themselves.

fisherxu (commented):
Thanks for the great changes :)

cloud/pkg/common/informers/informers.go
threestoneliu: done! PTAL :)

daixiang0: please rebase until #2487 merge

threestoneliu: ok

cloud/pkg/common/client/client.go
```diff
@@ -0,0 +1,61 @@
...
+func InitKubeEdgeClient(config *cloudcoreConfig.KubeAPIConfig) {
+	kubeConfig, err := clientcmd.BuildConfigFromFlags(config.Master,
+		config.KubeConfig)
+	if err != nil {
+		klog.Errorf("Failed to build config, err: %v", err)
+		os.Exit(1)
+	}
+	kubeConfig.QPS = float32(config.QPS)
+	kubeConfig.Burst = int(config.Burst)
+	kubeConfig.ContentType = runtime.ContentTypeJSON
```
fisherxu, line 35: The `ContentType` here, kubernetes core api client should use `application/vnd.kubernetes.protobuf`, and crd client should use runtime.ContentTypeJSON, `protobuf` has better performance than `json`, but crd can only use json..
cc @GsssC

cloud/pkg/common/client/client.go
```diff
@@ -0,0 +1,61 @@
...
+			crdClient: crdClient,
+		}
+	})
+}
+
+func GetKubeEdgeClient() KubeEdgeClient {
+	return keClient
+}
+
+type kubeedgeClient struct {
```
fisherxu, line 50:
```suggestion
type kubeEdgeClient struct {
```

cloud/pkg/common/client/client.go
threestoneliu: done!

cloud/pkg/common/client/client.go
threestoneliu: done!

cloud/pkg/synccontroller/createfailedobject.go
```diff
@@ -92,7 +92,7 @@ func (sctl *SyncController) manageCreateFailedDevice() {
 		// Check whether the device is successfully persisted to edge
 		// TODO: refactor the nodeselector of the device
 		nodeName := device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0]
-		_, err := sctl.objectSyncLister.ObjectSyncs(device.Namespace).Get(BuildObjectSyncName(nodeName, string(device.UID)))
+		_, err := sctl.osLister.ObjectSyncs(device.Namespace).Get(BuildObjectSyncName(nodeName, string(device.UID)))
```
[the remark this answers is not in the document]
threestoneliu: done!

cloud/pkg/edgecontroller/manager/const_test.go
```diff
@@ -0,0 +1,26 @@
+package manager
+
+const mockKubeConfigContent = `
```
fisherxu, line 3: we need this const?

cloud/pkg/common/informers/informers.go
```diff
@@ -0,0 +1,105 @@
...
+	})
+	return globalInformers
+}
+
+type informers struct {
+	defaultResync            time.Duration
+	keClient                 client.KubeEdgeClient
+	lock                     sync.Mutex
+	informers                map[string]cache.SharedIndexInformer
+	startedInformers         map[string]bool
```
fisherxu, line 53: We already have the informerfactory, so do we still need these maps?

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -13,8 +13,8 @@ import (
 
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
 	beehiveModel "github.com/kubeedge/beehive/pkg/core/model"
+	reliablesyncsv1alpha1listers "github.com/kubeedge/kubeedge/cloud/pkg/client/listers/reliablesyncs/v1alpha1"
```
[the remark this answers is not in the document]
threestoneliu: done!

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -30,13 +30,15 @@ type ChannelMessageQueue struct {
 	listQueuePool sync.Map
 	listStorePool sync.Map
 
-	ObjectSyncController *hubconfig.ObjectSyncController
+	osLister  reliablesyncsv1alpha1listers.ObjectSyncLister
```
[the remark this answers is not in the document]
threestoneliu: Done!  All varialiables that with short name has been renamed.

cloud/pkg/edgecontroller/manager/const_test.go
threestoneliu: In package manager, unittest need to create an k8s client, so we need a temp kubeconfig file. When I put this string into the test method, command `make verify` will run error, and warn this string is used in many files. But i didn‘t find a file which is suitable to add this string only for unittest, so i put it here. Otherwise i think we should check the unittest code more strictly, because the old unittest in package manager will skip when create kubeclient failed, and not really cover all mehod logic when github  or travis ci runs.

cloud/pkg/common/informers/informers.go
threestoneliu: The custome Informers stored in the informers field also need to run only once. There will have some problem if we run those informers more than one times, so i think we need it.

cloud/pkg/cloudhub/channelq/channelq.go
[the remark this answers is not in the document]
fisherxu: Okey, someone can use the shortName, but someone like os/dev maybe confusing.

cloud/pkg/common/informers/informers.go
fisherxu: For the pod, [`Informer`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/informers/core/v1/pod.go#L84) here will use the [`InformerFor`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/informers/factory.go#L169) to create the sharedinformer, and if the informer existed, it will return directly, so will it run many times?

threestoneliu: done!

cloud/pkg/common/client/client.go
```diff
@@ -0,0 +1,64 @@
+package client
```
fisherxu, line 1: Add the license header here, 2021 year

cloud/pkg/common/informers/informers.go
```diff
@@ -0,0 +1,98 @@
+package informers
```
fisherxu, line 1: ditto

cloud/pkg/common/client/client.go
```diff
@@ -0,0 +1,64 @@
...
+	crdClient *crdClientset.Clientset
+}
+
+func (kec *kubeEdgeClient) DevicesV1alpha2() devicev1alpha2.DevicesV1alpha2Interface {
+	return kec.crdClient.DevicesV1alpha2()
+}
+
+func (kec *kubeEdgeClient) ReliablesyncsV1alpha1() syncv1alpha1.ReliablesyncsV1alpha1Interface {
+	return kec.crdClient.ReliablesyncsV1alpha1()
+}
```
fisherxu, line 64: If we add new crd, developers need to hard here, so could we separate the crd client and core client?

fisherxu (commented):
Overall looks good, just some small questions. :)

cloud/pkg/edgecontroller/manager/const_test.go
```diff
@@ -0,0 +1,26 @@
+package manager
```
fisherxu, line 1: ditto

cloud/pkg/common/client/client.go
threestoneliu: done!

cloud/pkg/common/informers/informers.go
threestoneliu: done!

cloud/pkg/common/client/client.go
threestoneliu: done!

cloud/pkg/edgecontroller/manager/const_test.go
threestoneliu: done!

fisherxu:
/lgtm
/approve

Let's merge the PR first, if any updates needed, just do them in follow-up prs.

PR #2478, opened 2020-12-29
Keadm: fix install issue

daixiang0:

**What type of PR is this?**

 /kind bug

This `return` would skip unzipping tarballs if tarballs do not exist.

This bug is ignored since we download tarballs first to speed up CI.

Fixes #

```release-note

```


daixiang0:
ping @fisherxu

PR #2491, opened 2021-01-08
keadm: when the service file does not exist, the rm file command is not executed

zc2638:
keadm reset:  When the service file does not exist, the rm file command is not executed

**What type of PR is this?**

/kind bug

Fixes #

```release-note
NONE
```


kevin-wangzefeng:
/lgtm
/approve

Thanks

PR #2495, opened 2021-01-12
Optimize hack/local-up-kubeedge.sh

llhuii:
**What type of PR is this?**

/kind bug

Fixes #

```release-note

```

### DIFF
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -135,7 +135,7 @@ function start_edgecore {
 
 function check_control_plane_ready {
   echo "wait the control-plane ready..."
-  kubectl wait --for=condition=Ready node/test-control-plane --timeout=60s
+  kubectl wait --for=condition=Ready node/${CLUSTER_NAME}-control-plane --timeout=60s
 }
 
 # Check if all processes are still running. Prints a warning once each time
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -17,6 +17,7 @@
 KUBEEDGE_ROOT=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/..
 ENABLE_DAEMON=${ENABLE_DAEMON:-false}
 LOG_DIR=${LOG_DIR:-"/tmp"}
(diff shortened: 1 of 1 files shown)


daixiang0:
Please retitle with a more meaningful title, like `hack/local-up-kubeedge.sh: fix xxx`

llhuii:
> Please retitle with a more meaningful title, like `hack/local-up-kubeedge.sh: fix xxx`

There are 3 commits in this pr, don't know how to say `xxx`. 

May I submit them in three separated pull requests?

daixiang0:
"optimize hack/local-up-kubeedge.sh"


lgtm

hack/local-up-kubeedge.sh
```diff
@@ -121,6 +123,7 @@ function start_edgecore {
   sed -i -e "s|token: .*|token: ${token}|g" \
       -e "s|hostnameOverride: .*|hostnameOverride: edge-node|g" \
       -e "s|/etc/|/tmp/etc/|g" \
+      -e "s|/var/lib/kubeedge/|/tmp&|g" \
```
fisherxu, line 126: What's the meaning of `/tmp&`, for files under `/var/lib/kubeedge`, i'd prefer keep them here, not move to /tmp.

llhuii: > What's the meaning of `/tmp&`?

'&' means the whole matched string, in this context i.e. `/var/lib/kubeedge/`. 

>  For files under `/var/lib/kubeedge`, i'd prefer keep them here, not move to /tmp.
1. What's the difference between `/etc` and `/var/lib/kubeedge`? Why one is under `/tmp`, the other is not?
2. https://github.com/kubeedge/kubeedge/blob/29c26a61b3541331ae51917df83cbf6adcd7ea62/hack/local-up-kubeedge.sh#L48 needs to be changed if `/var/lib/kubeedge` keeps unchanged.

llhuii: Better to move all related files into a directory, also I can make it configurable if you wants.

fisherxu: > &' means the whole matched string, in this context i.e. /var/lib/kubeedge/.

Get it, also fine to put under tmp.

PR #2408, opened 2020-12-08
update CustomResourceDefinition api version

daixiang0:
Signed-off-by: Xiang Dai <long0dai@foxmail.com>

Refer to https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#version-deprecation, apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+.
 /kind cleanup

**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #2290, opened 2020-10-31
update pull request template

daixiang0:
Signed-off-by: Xiang Dai <long0dai@foxmail.com>

**What type of PR is this?**

**What this PR does / why we need it**:

After upgrade prow, no need to comment label.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #2502, opened 2021-01-13
Refactor k8s and crd client in devicecontroller

fisherxu:
**What type of PR is this?**
> /kind cleanup

Follow-up pr of #2472 

Fixes #

/cc @threestoneliu 

```release-note
NONE
```


daixiang0:
please see #2290 or labels would not work.

cloud/pkg/devicecontroller/devicecontroller.go
```diff
@@ -45,22 +58,13 @@ func (dc *DeviceController) Enable() bool {
...
-	if err != nil {
-		klog.Fatalf("new upstream controller failed with error: %s", err)
-	}
-
-	if err := downstream.Start(); err != nil {
+	if err := dc.downstream.Start(); err != nil {
 		klog.Fatalf("start downstream failed with error: %s", err)
 	}
 	// wait for downstream controller to start and load deviceModels and devices
 	// TODO think about sync
```
threestoneliu, line 65: `TODO think about sync` means wait for devciceinformer.HasSynced and devciemodelinformer.HasSynced method?

threestoneliu (commented):
Good changes! All looks good to me. only one question

cloud/pkg/devicecontroller/controller/upstream.go
```diff
@@ -159,9 +159,9 @@ func (uc *UpstreamController) updateDeviceStatus() {
 				klog.Errorf("Failed to marshal device status %v", deviceStatus)
 				continue
 			}
-			result := uc.crdClient.Patch(MergePatchType).Namespace(cacheDevice.Namespace).Resource(ResourceTypeDevices).Name(deviceID).Body(body).Do(context.Background())
-			if result.Error() != nil {
-				klog.Errorf("Failed to patch device status %v of device %v in namespace %v", deviceStatus, deviceID, cacheDevice.Namespace)
+			err = uc.crdClient.DevicesV1alpha2().RESTClient().Patch(MergePatchType).Namespace(cacheDevice.Namespace).Resource(ResourceTypeDevices).Name(deviceID).Body(body).Do(context.Background()).Error()
+			if err != nil {
+				klog.Errorf("Failed to patch device status %v of device %v in namespace %v, err: %v", deviceStatus, deviceID, cacheDevice.Namespace, err)
```
fisherxu, line 164: Now we can't use updatestatus to patch the twin info in device status and only can use the restclient, because:

1. Now the device crd yaml don't  set the status subresource, so updatestatus can't work.
2. If we set status subresource in device crd, when we create the device obj, the status section will not be applied in apiserver/etcd, must use updatestatus to add it.  Now we put desired value in twin of status, we need to create obj with status.  In future, maybe **it's need to split the  desired value in twin to spec and leave the reported value in status**.

cc @luogangyi @sailorvii @kevin-wangzefeng

daixiang0: Refer to https://book-v1.book.kubebuilder.io/basics/status_subresource.html, we can upgrade CRD  from v1beta1 to v1 to enable status subresource as #2408

cloud/pkg/devicecontroller/devicecontroller.go
fisherxu: Looks like it's need to wait all deviceModels and devices cached(cache here maybe also can use the informer, instead of define a new one in devicecontroler) in devicecontroller, `HasSynced`  should be a efficient way, we can do series refactor in folowup prs.

fisherxu:
> @threestoneliu: adding LGTM is restricted to approvers and reviewers in OWNERS files.

Not in owner file under devicecontroller ;) we can modify bot setting to allow all members to add lgtm label.

PR #2517, opened 2021-01-18
hack/local-up-kubeedge.sh: ensure cleanup successfully

daixiang0:

**What type of PR is this?**

If iptables rule does not exist, delete would fail and return 1, then the whole script exits. Since we do cleanup at first, ensure it always succeed is necessary.

Since we set `ENABLE_DAEMON=true` it would not catch EXIT signal then do not trigger this issue.

Fixes #

```release-note

```

### DIFF
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -49,7 +49,7 @@ function uninstall_kubeedge {
   rm -rf /tmp/etc/kubeedge /tmp/var/lib/kubeedge
 
   # delete iptables rule
-  sudo iptables -t nat -D PREROUTING -p tcp --dport 10350 -j REDIRECT --to-port 10003
+  sudo iptables -t nat -D PREROUTING -p tcp --dport 10350 -j REDIRECT --to-port 10003 || true
 }
 
 # clean up


hack/local-up-kubeedge.sh
```diff
@@ -49,7 +49,7 @@ function uninstall_kubeedge {
   rm -rf /tmp/etc/kubeedge /tmp/var/lib/kubeedge
 
   # delete iptables rule
-  sudo iptables -t nat -D PREROUTING -p tcp --dport 10350 -j REDIRECT --to-port 10003
+  sudo iptables -t nat -D PREROUTING -p tcp --dport 10350 -j REDIRECT --to-port 10003 || true
```
chendave, line 52:
```
if [[ "${ENABLE_DAEMON}" = false ]]; then
  trap cleanup EXIT
else
...
```

This `cleanup` is called when this shell exit, and then if the iptables here cannot be deleted, then a error will return, this is not the issue what's described in the commit message. 

But I do think logic of the below script is not correct,
```
if [[ "${ENABLE_DAEMON}" = false ]]; then
  trap cleanup EXIT
```

why we need to cleanup after the script exit? this script is for kubeedge installation, right?

daixiang0: We do cleanup first indeed and catch other signals to enable it as a backend process, not a front process.

chendave: maybe I missed something, but I cannot reproduce the issue, leave it open to others.

fisherxu: > why we need to cleanup after the script exit? this script is for kubeedge installation, right?

When set `{ENABLE_DAEMON}` to false(default value), the localup script will stuck in front after setting up the cluster, at this time we can do some debugging work, then stop the script and cleanup the cluster. This is for quick starting and debugging.

If you want to run in  backend, can set `{ENABLE_DAEMON}` to true, the script will exit after setting up the cluster. The cluster need to cleanup by yourself. @chendave

chendave: Good to know, thanks for the explanation.

PR #2529, opened 2021-01-26
edgecore: add nfs localpath support

swartz-k:
Signed-off-by: swartz-k <9215868@gmail.com>

**What type of PR is this?**

/kind feature

Support use PV of localpath and nfs in edgecore

User can use like this
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
......
  volumeClaimTemplates:
  - metadata:
      name: data
    spec:
      accessModes: [ "ReadWriteOnce" ]
      storageClassName: "local-path"
      resources:
        requests:
          storage: 10Gi
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ edgesite/edgesite
 keadm/keadm
 _output
 _tmp
+build/edge/tmp
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -86,6 +86,8 @@ import (
 	"k8s.io/kubernetes/pkg/volume/downwardapi"
 	"k8s.io/kubernetes/pkg/volume/emptydir"
 	"k8s.io/kubernetes/pkg/volume/hostpath"
+	"k8s.io/kubernetes/pkg/volume/local"
+	"k8s.io/kubernetes/pkg/volume/nfs"
 	"k8s.io/kubernetes/pkg/volume/projected"
 	secretvolume "k8s.io/kubernetes/pkg/volume/secret"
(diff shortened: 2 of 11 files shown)


edge/pkg/edged/edged.go
```diff
@@ -27,6 +27,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"k8s.io/kubernetes/pkg/volume/local"
+	"k8s.io/kubernetes/pkg/volume/nfs"
```
fisherxu, line 31: Please sort the imports :)

swartz-k: Resolved and tested `make verify-vendor`

fisherxu:
Looks like the failed ut is not related to this pr, I will have a look

.gitignore
```diff
@@ -16,3 +16,4 @@ edgesite/edgesite
 keadm/keadm
 _output
 _tmp
+build/edge/tmp
```
daixiang0, line 19: which generates it?

fisherxu:
@swartz-k Please rebase the master branch and let's see the ut result.

.gitignore
swartz-k: [Makefile#L278](https://github.com/kubeedge/kubeedge/blob/2f4447c60b767de371a8b773774662711ead3089/Makefile#L278)`edgeimage`

swartz-k:
> @swartz-k Please rebase the master branch and let's see the ut result.

Thanks a lot. It passed.

fisherxu:
/lgtm
/approve
Thanks for the enhancement！

PR #2554, opened 2021-02-03
keadm: support the installation of cloudcore in systemd

zc2638:
**What type of PR is this?**

/kind feature

Fixes #

```release-note

```


daixiang0:
Now it auto check use systemd if exists, why still need it?

zc2638:
> Now it auto check use systemd if exists, why still need it?

I'm not sure if the default process is necessary, so I support it

PR #2555, opened 2021-02-03
fix：modify a key name called protocol in deviceInstance

coldy1992:
There if an error key name which called 'protocol' in deviceInstance and it should be 'protocolName'.This is because the key is used in kubeedge/mappers/modbus-go/configmap/parse.go, and it's name is 'protocol' in that file.This error can cause unit tests to go wrong

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/mappers/modbus-go/configmap/configmap_test.json
+++ b/mappers/modbus-go/configmap/configmap_test.json
@@ -2,7 +2,7 @@
 	"deviceInstances": [{
 		"id": "sensor-tag-instance-01",
 		"name": "sensor-tag-instance-01",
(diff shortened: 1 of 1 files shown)


daixiang0:
Please move it to https://github.com/kubeedge/mappers-go.

fisherxu:
Ref to here: https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/devicecontroller/types/deviceprofile.go#L24

The key should be `protocol`. I will close the pr now, please reopen later if needed.

/close

PR #2572, opened 2021-02-18
Remove unused hard code map in synccontroller

fisherxu:
**What type of PR is this?**
/kind cleanup

Looks like the objectKind had been [converted to objectresources](https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/handler/messagehandler.go#L554), so we don't need the unused hard code map in synccontroller.
Without this map, synccontroller will has no hard code resources. 

Fixes #

```release-note
NONE
```


fisherxu:
cc @GsssC

GsssC (commented):
/lgtm

PR #2578, opened 2021-02-20
[WIP] support "create update patch" at edge and support kubelet access

GsssC:
**What type of PR is this?**
/kind feature
Support "create update patch" at edge and support kubelet access, need more work in testing.


fisherxu:
/close
Has included in #2587.

PR #2604, opened 2021-03-05
Add timeout for github actions

fisherxu:

**What type of PR is this?**
/kind test

Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

Set timeout for all github actions.  Because the default timeout value is 360 mins, it's too long to report the result if the job get stucked. And usually our jobs will not run for more than 20 mins.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -10,6 +10,7 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     name: Verify vendor, licenses, do lint
     env:
       GOPATH: /home/runner/work/${{ github.repository }}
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -10,6 +10,7 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
```
chendave, line 13: Why not use the default value, can you pls explain a little bit here or in the commit message on why we adopt "30" mins?

fisherxu: Well, will add the detail to PR description :) Because the default timeout value is 360 mins, it's too long to report the result if the job get stucked. And usually our jobs will not run for more than 20 mins.

chendave: thanks!

PR #2613, opened 2021-03-09
fix cloudcore crash when nodekeepalivechannel is nil

lvfei103650:
**What type of PR is this?**

/kind bug

fix cloudcore crash when nodekeepalivechannel is nil

Fixes #

NONE

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -166,7 +166,11 @@ func (mh *MessageHandle) OnRegister(connection conn.Connection) {
 // KeepaliveCheckLoop checks whether the edge node is still alive
 func (mh *MessageHandle) KeepaliveCheckLoop(info *model.HubInfo, stopServe chan ExitCode) {
 	keepaliveTicker := time.NewTimer(time.Duration(mh.KeepaliveInterval) * time.Second)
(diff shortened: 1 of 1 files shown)


lvfei103650:
image

**When the network condition is very weak and edgenode keeps re-registering, cloudcore will crash in here**

fisherxu:
Thanks for the fixing, will have a review :) And kindly reminder, looks like github can't identify your email addr in the commit, you need to associate this Email addr to your github personal setting.
image

cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -166,7 +166,11 @@ func (mh *MessageHandle) OnRegister(connection conn.Connection) {
 // KeepaliveCheckLoop checks whether the edge node is still alive
 func (mh *MessageHandle) KeepaliveCheckLoop(info *model.HubInfo, stopServe chan ExitCode) {
 	keepaliveTicker := time.NewTimer(time.Duration(mh.KeepaliveInterval) * time.Second)
-	nodeKeepaliveChannel, _ := mh.KeepaliveChannel.Load(info.NodeID)
+	nodeKeepaliveChannel, ok := mh.KeepaliveChannel.Load(info.NodeID)
+	if !ok {
+		klog.Errorf("fail to load node %s", info.NodeID)
+		return
```
fisherxu, line 172: Looks good, but it should be not correct that just return here, and we can fix it in follow-up PRs.

PR #2625, opened 2021-03-12
keadm: support to download file without wget

zc2638:
**What type of PR is this?**

/kind feature

Sometimes `wget` is not installed on the node

Fixes #

```release-note

```


daixiang0:
Please split commits and put all vendor changes in one commit, it would help review.

zc2638:
> Please split commits and put all vendor changes in one commit, it would help review.

@daixiang0 I took care of it, thanks

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -29,6 +31,7 @@ import (
 	"strings"
 
 	"github.com/blang/semver"
+	"github.com/schollz/progressbar/v3"
```
daixiang0, line 34: Where need it? I see too much pkgs imported, please double check them really necessary

zc2638: In order to have a good user experience, add the progress bar.
If the network does not allow
Can't see the progress, it affects the experience

daixiang0: Useful logs are enough i think.

zc2638: If the network is too bad, the log is stuck all the time.
And we don't know when it will be ready.
It's a better experience

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -478,6 +482,47 @@ func retryDownload(filename, checksumFilename string, version semver.Version, ta
...
+	if targetPath != "" {
+		_, err = os.Stat(targetPath)
+		if err != nil {
+			mkErr := os.MkdirAll(targetPath, os.ModePerm)
+			if mkErr != nil {
+				return mkErr
+			}
+		}
+	}
+	fullPath := filepath.Join(targetPath, name)
```
SmileCaoJ, line 502: if targetPath = "", targetPath  will be default value?

zc2638: if targetPath = "", the path will be the current directory

fisherxu:
Will have a look today later :)

PR #2642, opened 2021-03-17
Fix image gc issue

fisherxu:
**What type of PR is this?**
/kind bug

Use the right statsProvider and change image gc time to 5min, avoiding print too many err logs when gc failed. Ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L172

Fixes #

```release-note
NONE
```


edge/pkg/edged/images/image_gc_manager.go
```diff
@@ -1,37 +0,0 @@
...
-
-	"github.com/kubeedge/kubeedge/pkg/util"
-)
-
-type imageGCManager struct {
-	imageGCPath string
-}
-
-func (i *imageGCManager) ImageFsStats() (*statsapi.FsStats, error) {
-	fs := syscall.Statfs_t{}
```
fisherxu, line 17: Remove the linux bind func.

PR #2645, opened 2021-03-17
beehive/context: fix the unlocking of the ctx.chsLock in the delChannel

Iceber:
**What type of PR is this?**

/kind bug

Fixes #

```release-note
fix the unlocking of the ctx.chsLock in the delChannel
```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
@@ -308,13 +308,12 @@ func (ctx *ChannelContext) addChannel(module string, moduleCh chan model.Message
 func (ctx *ChannelContext) delChannel(module string) {
 	// delete module channel from channels map
 	ctx.chsLock.Lock()
-	_, exist := ctx.channels[module]
-	if !exist {
+	if _, exist := ctx.channels[module]; !exist {
+		ctx.chsLock.Unlock()
 		klog.Warningf("Failed to get channel, module:%s", module)
 		return
 	}
 	delete(ctx.channels, module)
-
 	ctx.chsLock.Unlock()
 
 	// delete module channel from typechannels map


Iceber:
Is it significant to print a warning log when a module does not exist in the `ctx.channels`?

staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
```diff
@@ -308,13 +308,12 @@ func (ctx *ChannelContext) addChannel(module string, moduleCh chan model.Message
 func (ctx *ChannelContext) delChannel(module string) {
 	// delete module channel from channels map
 	ctx.chsLock.Lock()
-	_, exist := ctx.channels[module]
-	if !exist {
+	if _, exist := ctx.channels[module]; !exist {
+		ctx.chsLock.Unlock()
 		klog.Warningf("Failed to get channel, module:%s", module)
```
Iceber, line 313: Is it significant to print a warning log when a module does not exist in the `ctx.channels`?

or change:
```go
ctx.chsLock.Lock()
delete(ctx.channels, module)
ctx.chsLock.Unlock()
```

fisherxu (commented):
How about using defer to unlock?

Iceber:
> How about using defer to unlock?

@fisherxu There will be no subsequent operation on the `ctx.channels`, so there is no need to keep the`ctx.chsLock` locked.
And a lock will be placed on `typeChsLock` later, which may cause deadlocks if defer is used.

fisherxu:
> > How about using defer to unlock?
> 
> @fisherxu There will be no subsequent operation on the `ctx.channels`, so there is no need to keep the`ctx.chsLock` locked.
> And a lock will be placed on `typeChsLock` later, which may cause deadlocks if defer is used.

Oh, sounds good, had ignored the `typeChsLock` below..
/lgtm

PR #2648, opened 2021-03-17
beehive/context: fix the SendResp of the ChannelContext

Iceber:

**What type of PR is this?**

/kind bug

1. In sync mode, the `SendSync` sending a data timeout will close the channel, so we need to add defer to recover the panic :upside_down_face:
https://github.com/kubeedge/kubeedge/blob/b512232fd7f9f576540b912a3b22e51c0aa51f80/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go#L128-L145
2. There is no need to lock `ctx.anonChannels` when sending response.

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
@@ -146,16 +146,27 @@ func (ctx *ChannelContext) SendSync(module string, message model.Message, timeou
 
 // SendResp send resp for this message when using sync mode
 func (ctx *ChannelContext) SendResp(message model.Message) {
+	// avoid exception because of channel closing
+	// TODO: need reconstruction
+	defer func() {
+		if exception := recover(); exception != nil {
+			klog.Warningf("Recover when sendResp message, exception: %+v", exception)
+		}
+	}()
+
 	anonName := getAnonChannelName(message.GetParentID())
 
 	ctx.anonChsLock.RLock()
-	defer ctx.anonChsLock.RUnlock()
-	if channel, exist := ctx.anonChannels[anonName]; exist {
-		channel <- message
+	channel, exist := ctx.anonChannels[anonName]
+	if !exist {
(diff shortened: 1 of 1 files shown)


staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
```diff
@@ -146,16 +146,27 @@ func (ctx *ChannelContext) SendSync(module string, message model.Message, timeou
...
+	if !exist {
+		ctx.anonChsLock.RUnlock()
+		klog.V(4).Infof("Get bad anonName:%s when sendresp message, do nothing", anonName)
 		return
 	}
+	ctx.anonChsLock.RUnlock()
 
-	klog.V(4).Infof("Get bad anonName:%s when sendresp message, do nothing", anonName)
+	channel <- message
+	return
```
fisherxu, line 169: Have more lines than before? ;)

Iceber: The problem is more complex, here is a deadlock situation.
This needs to be analysed in conjunction with the SendSync function.
https://github.com/kubeedge/kubeedge/blob/b512232fd7f9f576540b912a3b22e51c0aa51f80/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go#L128-L145
When the `sendTimer` expires,  the `ctx.anonChsLock` will be locked in the defer function, but the `ctx.annoChsLock` still locked in the `SendResp`.
This is when a deadlock occurs.

Iceber: Of course it is possible to close the `anonChan` before putting a lock on the `ctx.anonChsLock`, which would also solve the deadlock problem, but I don't feel this is very elegant.

It doesn't feel good to put a lock on a channel operation for non-exceptional reasons, as it is a potentially blocking operation

PR #2652, opened 2021-03-18
edge/devicetwin/dtmanager: fix unlocking  the DTContext

Iceber:

**What type of PR is this?**

/kind bug

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/membership.go
+++ b/edge/pkg/devicetwin/dtmanager/membership.go
@@ -177,8 +177,7 @@ func addDevice(context *dtcontext.DTContext, toAdd []dttype.Device, baseMessage 
 	}
 	for _, device := range toAdd {
 		//if device has existed, step out
-		deviceInstance, isDeviceExist := context.GetDevice(device.ID)
-		if isDeviceExist {
+		if _, exist := context.GetDevice(device.ID); exist {
 			if delta {
 				klog.Errorf("Add device %s failed, has existed", device.ID)
 				continue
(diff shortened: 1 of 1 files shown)


edge/pkg/devicetwin/dtmanager/membership.go
```diff
@@ -220,7 +218,10 @@ func addDevice(context *dtcontext.DTContext, toAdd []dttype.Device, baseMessage
 		if err != nil {
 			klog.Errorf("Add device %s failed due to some error ,err: %#v", device.ID, err)
 			context.DeviceList.Delete(device.ID)
-			context.Unlock(device.ID)
+			if delta {
+				context.Unlock(device.ID)
+			}
+			context.DeviceMutex.Delete(device.ID)
```
Iceber, line 224: If the device is removed from the `DeviceList`, then it should also be removed from the `DeviceMutex`.

edge/pkg/devicetwin/dtmanager/membership.go
```diff
@@ -283,22 +279,18 @@ func removeDevice(context *dtcontext.DTContext, toRemove []dttype.Device, baseMe
 		}
 		//todo
 		context.DeviceList.Delete(device.ID)
-		context.DeviceMutex.Delete(device.ID)
 		if delta {
 			context.Unlock(device.ID)
 		}
+		context.DeviceMutex.Delete(device.ID)
```
Iceber, line 285: If the device is removed from the `DeviceMutex` first, then it may result in the context not being unlocked.
https://github.com/kubeedge/kubeedge/blob/b512232fd7f9f576540b912a3b22e51c0aa51f80/edge/pkg/devicetwin/dtcontext/dtcontext.go#L98-L107

PR #2655, opened 2021-03-18
cloudstream: fix defer cleanup

Iceber:

**What type of PR is this?**

/kind bug

`err` is overwritten. 

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudstream/streamserver.go
+++ b/cloud/pkg/cloudstream/streamserver.go
@@ -99,7 +99,6 @@ func (s *StreamServer) installDebugHandler() {
 
 func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response) {
 	var err error
-
 	defer func() {
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudstream/streamserver.go
```diff
@@ -140,7 +139,7 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
 		klog.Infof("Delete %s from %s", logConnection.String(), session.String())
 	}()
 
-	if err := logConnection.Serve(); err != nil {
+	if err = logConnection.Serve(); err != nil {
 		err = fmt.Errorf("apiconnection Serve %s in %s error %v",
```
fisherxu, line 143: If we don't remove `:` above, the err returned maybe nil, right?

Iceber: yes

fisherxu: But the err here has any usage?

Iceber: `WriteHeader` in the defer func
https://github.com/kubeedge/kubeedge/blob/bf51f6d2db9d83559b3f1da79e5f724492ddc433/cloud/pkg/cloudstream/streamserver.go#L100-L109

fisherxu: Oh, lgtm

PR #2659, opened 2021-03-18
Fix container gc issue

fisherxu:
**What type of PR is this?**
/kind bug

GC running containers started by the previous cluster.
If we stop edgecore without deleting pods and reinstall the cluster, the containers will still run in the edgenode.
When the edgecore reinstalled, these old containers should be GCed.

Fixes #

```release-note
GC running containers started by the previous cluster.
```


edge/pkg/edged/edged.go
```diff
@@ -1409,6 +1418,36 @@ func (e *edged) HandlePodCleanups() error {
...
+	//      though the pod was not considered terminated by the apiserver).
+	// These two conditions could be alleviated by checkpointing kubelet.
+	activePods := e.filterOutTerminatedPods(pods)
+
+	desiredPods := make(map[types.UID]sets.Empty)
+	for _, pod := range activePods {
+		desiredPods[pod.UID] = sets.Empty{}
+	}
+
+	runningPods, err := e.runtimeCache.GetPods()
```
fisherxu, line 1439: Here will get containers to pods from host through cri, and then delete the pods/containers edgecore don't know.

PR #2671, opened 2021-03-20
cloud/devicecontroller: fix the warning log for printing object type

Iceber:

**What type of PR is this?**

/kind bug

The type of the value after the assertion is the type of the assertion.
Here we need to print the type of the e.Object.

Fixes #

```release-note

```


Iceber:
close: https://github.com/kubeedge/kubeedge/pull/2670

PR #2672, opened 2021-03-20
cloud/edgecontroller: fix the syncPod in the downstream controller

Iceber:

**What type of PR is this?**

/kind bug

Continue when the pod event type is not supported

Fixes #

```release-note

```


fisherxu:
/lgtm
/approve
Thanks for the effort, and in future(when edge side list-watch GA) the edgecontroller will be replaced by dynamiccontroller, can pay more attention to it. :)

PR #2680, opened 2021-03-22
support default NVIDIA SMI command

zc2638:
**What type of PR is this?**

/kind feature

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged_status.go
+++ b/edge/pkg/edged/edged_status.go
@@ -51,8 +51,8 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/util"
 )
 
-//GPUInfoQueryTool sets information monitoring tool location for GPU
-var GPUInfoQueryTool = "/var/IEF/nvidia/bin/nvidia-smi"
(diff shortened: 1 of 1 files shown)


edge/pkg/edged/edged_status.go
```diff
@@ -51,8 +51,8 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/util"
 )
 
-//GPUInfoQueryTool sets information monitoring tool location for GPU
-var GPUInfoQueryTool = "/var/IEF/nvidia/bin/nvidia-smi"
+//IEFGPUInfoQueryTool sets information monitoring tool location for GPU
+var IEFGPUInfoQueryTool = "/var/IEF/nvidia/bin/nvidia-smi"
```
fisherxu, line 55: We should remove ief related codes here if possible.

zc2638: okey

zc2638: @fisherxu GPU resource computing is expected to be solved as soon as possible

PR #2682, opened 2021-03-22
Fix dynamiccontroller returned error missing

Abirdcfly:

**What type of PR is this?**

/kind bug

The dynamiccontroller need return raw error to caller.Otherwise caller use kubernetes built-in methods like `errors.IsNotFound(err)` will get unexpected result.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/application.go
+++ b/cloud/pkg/dynamiccontroller/application/application.go
@@ -4,11 +4,11 @@ import (
 	"context"
 	"crypto/md5"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"sync"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -85,7 +85,7 @@ type Application struct {
 	Verb     applicationVerb
 	Nodename string
 	Status   applicationStatus
-	Reason   string // why in this status
+	Reason   []byte // why in this status
 	Option   []byte //
 	ReqBody  []byte // better a k8s api instance
 	RespBody []byte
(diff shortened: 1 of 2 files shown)


cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -477,21 +476,21 @@ func (c *Center) Process(msg model.Message) {
...
-			c.Response(app, msg.GetID(), Approved, "", retObj)
+			c.Response(app, msg.GetID(), Approved, nil, retObj)
 		default:
 			return fmt.Errorf("unsupported Application Verb type :%v", app.Verb)
 		}
 		return nil
 	}()
 	if err != nil {
-		c.Response(app, msg.GetID(), Rejected, err.Error(), nil)
+		c.Response(app, msg.GetID(), Rejected, err,nil)
```
fisherxu, line 486:
```suggestion
		c.Response(app, msg.GetID(), Rejected, err, nil)
```

fisherxu:
> /hold There has a bug...test ci should not pass...I guess we need more e2e test for dynamiccontroller😂

Yes, now we still haven't the ci for dynamiccontroller/mateserver.

cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -4,11 +4,10 @@ import (
 	"context"
 	"crypto/md5"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"sync"
 	"time"
-
+
```
Iceber, line 10: please format

Abirdcfly: Got it. Thx

cloud/pkg/dynamiccontroller/application/application.go
Abirdcfly: Got it, Thx

GsssC:
Start to review. Changing application.Reason to adapt api errors is in evaluation.
Originally just want to use application.Reason to briefly describe why app is in application status, and the scope is bigger than returned error from APIServer, e.g. application is rejected by Application Center before calling APIServer(An authentication layer that must to do but have not yet).
: )

PR #2686, opened 2021-03-22
cloudhub/udsserver: add checks before using the unix addr

Iceber:

**What type of PR is this?**

/kind bug
/kind feature

Better handling of the two cases `"UNIX:///tmp/uds.sock"` and `"unix://    "`

Add more checks before using the addr

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/servers/udsserver/uds.go
+++ b/cloud/pkg/cloudhub/servers/udsserver/uds.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudhub/servers/udsserver/uds.go
```diff
@@ -54,7 +57,7 @@ func (us *UnixDomainSocket) StartServer() error {
 		return err
 	}
 	if proto == "unix" {
-		addr = "/" + addr
+		addr = filepath.Join("/" + addr)
 		if err := os.Remove(addr); err != nil && !os.IsNotExist(err) { //nolint: vetshadow
```
Iceber, line 61: This is a very dangerous operation

Iceber: Add more checks before removing

Iceber:
@fisherxu PTAL

cloud/pkg/cloudhub/servers/udsserver/uds.go
```diff
@@ -54,11 +57,12 @@ func (us *UnixDomainSocket) StartServer() error {
...
-		addr = "/" + addr
-		if err := os.Remove(addr); err != nil && !os.IsNotExist(err) { //nolint: vetshadow
-			klog.Errorf("failed to remove addr: %v", err)
+		addr = filepath.Join("/" + addr)
+		if err := checkUnixSocket(addr); err != nil {
+			klog.Errorf("failed to check unix socket addr: %v", err)
 			return err
 		}
+		defer os.Remove(addr)
 	}
```
gy95, line 66: i think don't need this,  function `checkUnixSocket` already remove socket. or we can remove `os.Remove` from `checkUniSocket` and exec `os.Remove` out of `checkUniSocket` and return error to `StartServer`

Iceber: This is to ensure that the created sockets are removed when the unix server exits，although for the time being it does not appear that the `StartServer` will exit gracefully(TODO)

It is not necessary to have a socket when the program is not running.

gy95: got it, thanks

Iceber:
@GssC PTAL, Thanks

PR #2693, opened 2021-03-23
edgesite: optimize code for edgeSiteCommand

TianTianBigWang:

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note

```

### DIFF
--- a/edgesite/cmd/edgesite/app/server.go
+++ b/edgesite/cmd/edgesite/app/server.go
@@ -61,8 +61,7 @@ runs on edge nodes and manages containerized applications.`,
 			gis.Start(beehiveContext.Done())
 			registerModules(config)
 			// start all modules
-			core.StartModules()
-			core.GracefulShutdown()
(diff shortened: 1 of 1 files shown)


edgesite/cmd/edgesite/app/server.go
```diff
@@ -61,8 +61,7 @@ runs on edge nodes and manages containerized applications.`,
 			gis.Start(beehiveContext.Done())
 			registerModules(config)
 			// start all modules
-			core.StartModules()
-			core.GracefulShutdown()
+			core.Run()
```
Iceber, line 64: It might be clearer to split it into two calls? I'm not sure.

zc2638: The `Run` method combines two methods and has full comment

fisherxu: Looks good.

PR #2695, opened 2021-03-23
edgecontroller: fix downstream easier to read

zc2638:
Signed-off-by: zc <ce.zheng@daocloud.io>

**What type of PR is this?**

/kind feature

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/downstream.go
+++ b/cloud/pkg/edgecontroller/controller/downstream.go
@@ -70,14 +70,14 @@ func (dc *DownstreamController) syncPod() {
 			if !dc.lc.IsEdgeNode(pod.Spec.NodeName) {
 				continue
 			}
-			msg := model.NewMessage("")
-			msg.SetResourceVersion(pod.ResourceVersion)
 			resource, err := messagelayer.BuildResource(pod.Spec.NodeName, pod.Namespace, model.ResourceTypePod, pod.Name)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


cloud/pkg/edgecontroller/controller/downstream.go
```diff
@@ -223,85 +223,96 @@ func (dc *DownstreamController) syncEdgeNodes() {
...
+					}
+					nstatus := string(nsc.Status)
+					status, _ := dc.lc.GetNodeStatus(node.ObjectMeta.Name)
+					dc.lc.UpdateEdgeNode(node.ObjectMeta.Name, nstatus)
+					if nsc.Status != v1.ConditionTrue || status == nstatus {
+						continue
+					}
+					// send all services to edge
+					msg := model.NewMessage("").SetRoute(modules.EdgeControllerModuleName, constants.GroupResource)
+
```
gy95, line 237: the following three msgs(`svcMsg`, `podsMsg`, `epsMsg`) all apply to `SetResourceOperation` with the same operation `model.UpdateOperation` ， so it can be applied to msg in advance
Or dont use the common msg,  new three msgs from the beginning.

zc2638: There is no direct method, it can directly use struct field, which is more appropriate？

gy95: yes, you are right

PR #2698, opened 2021-03-23
Remove excess memory allocation

wzshiming:
**What type of PR is this?**
/kind cleanup

There is no need to alloc memory for Msg-data every time in the loop.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/stream/edgedexecconnection.go
+++ b/pkg/stream/edgedexecconnection.go
@@ -93,6 +93,7 @@ func (e *EdgedExecConnection) Serve(tunnel SafeWriteTunneler) error {
 		}
 	}()
 
+	var data [256]byte
 	for {
 		select {
 		case <-stop:
@@ -100,9 +101,7 @@ func (e *EdgedExecConnection) Serve(tunnel SafeWriteTunneler) error {
 			return nil
 		default:
 		}
-		data := make([]byte, 256)
-
-		n, err := con.Read(data)
+		n, err := con.Read(data[:])
 		if err != nil {
 			if err != io.EOF {
 				klog.Errorf("%v failed to write exec data, err:%v", e.String(), err)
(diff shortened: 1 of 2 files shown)


pkg/stream/edgedexecconnection.go
```diff
@@ -93,16 +93,15 @@ func (e *EdgedExecConnection) Serve(tunnel SafeWriteTunneler) error {
...
 		select {
 		case <-stop:
 			klog.V(6).Infof("receive stop single, so stop exec scan ...")
 			return nil
 		default:
 		}
-		data := make([]byte, 256)
-
-		n, err := con.Read(data)
+		n, err := con.Read(data[:])
```
daixiang0, line 104: ?

wzshiming: There is no need to alloc memory for Msg-data every time in the loop. always use `var data [256]byte`

daixiang0: Why [:]?

wzshiming: Reduce small memory allocation and reduce GC time

wzshiming: data is an Array type, not a Slice, need  [:] as to slice.

fisherxu: Thanks for you fixing! :) And I'm not sure whether `con.Read(data[:])`  will refresh the whole byte in `data` every time? Or there wil be dirty byte in `data`. @wzshiming

zc2638: This is reference passing, similar to pointer passing @fisherxu

wzshiming: @fisherxu 
`n, err := con.Read(data[:])`, will write `n` length data in `data[:]`, below use `data[:n]`.

fisherxu: Make sense.

PR #2700, opened 2021-03-23
Fix Invalid if statement

mengjiao-liu:
**What type of PR is this?**
/kind cleanup

 "if/else if" statements should not have the same condition

A chain of `if/else if `statements is evaluated from top to bottom. At most, only one branch will be executed: the first one with a condition that evaluates to `true`.

Therefore, duplicating a condition automatically leads to `dead code`. Usually, this is due to a copy/paste error. At best, it's simply dead code and at worst, it's a bug that is likely to induce further bugs as the code is maintained, and obviously it could lead to unexpected behavior.

Fixes #

```release-note
None
```

### DIFF
--- a/edge/pkg/edged/volume/csi/csi_client.go
+++ b/edge/pkg/edged/volume/csi/csi_client.go
@@ -217,8 +217,6 @@ func (c *csiDriverClient) NodeGetInfo(ctx context.Context) (
 		var getNodeInfoError error
 		if c.nodeV1ClientCreator != nil {
 			nodeID, maxVolumePerNode, accessibleTopology, getNodeInfoError = c.nodeGetInfoV1(ctx)
-		} else if c.nodeV1ClientCreator != nil {
-			nodeID, maxVolumePerNode, accessibleTopology, getNodeInfoError = c.nodeGetInfoV0(ctx)
 		}
 		if nodeID != "" {
 			return true, nil


edge/pkg/edged/volume/csi/csi_client.go
```diff
@@ -217,7 +217,7 @@ func (c *csiDriverClient) NodeGetInfo(ctx context.Context) (
 		var getNodeInfoError error
 		if c.nodeV1ClientCreator != nil {
 			nodeID, maxVolumePerNode, accessibleTopology, getNodeInfoError = c.nodeGetInfoV1(ctx)
-		} else if c.nodeV1ClientCreator != nil {
+		} else {
```
gy95, line 220: `c.nodeGetInfoV1` will assign values to `c.nodeV1ClientCreator`, so the second `if` statement may be different from the first `if`

edge/pkg/edged/volume/csi/csi_client.go
```diff
@@ -217,7 +217,7 @@ func (c *csiDriverClient) NodeGetInfo(ctx context.Context) (
 		var getNodeInfoError error
 		if c.nodeV1ClientCreator != nil {
 			nodeID, maxVolumePerNode, accessibleTopology, getNodeInfoError = c.nodeGetInfoV1(ctx)
-		} else if c.nodeV1ClientCreator != nil {
```
fisherxu, line 220: `} else if c.nodeV0ClientCreator != nil {` ?

fisherxu (commented):
Please fix the DCO iossue.

edge/pkg/edged/volume/csi/csi_client.go
mengjiao-liu: but the struct have no the field `c.nodeV0ClientCreator`
https://github.com/kubeedge/kubeedge/blob/f7c9abcebc5cf6b3825785b9dfffec9977bec55f/edge/pkg/edged/volume/csi/csi_client.go#L109-L114

git commit 56f91bf6b3bf2e613b7ab3af2df93b692b1cf3ff remove this field `nodeV1ClientCreator`

My guess is that PR didn't delete this else if statement when we updated the code in bulk

The V0 version is no longer in use

mengjiao-liu: Maybe it needs to be cleaned up again

fisherxu: Looks Good.

PR #2705, opened 2021-03-24
cloud/dynamiccontroller: add ProcessApplication

Iceber:

**What type of PR is this?**

/kind feature

add ProcessApplication to handle Application

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/application.go
+++ b/cloud/pkg/dynamiccontroller/application/application.go
@@ -369,8 +369,8 @@ func msgToApplication(msg model.Message) (*Application, error) {
 	return app, nil
 }
 
-// TODO: upgrade to parallel process
 // Process translate msg to application , process and send resp to edge
+// TODO: upgrade to parallel process
 func (c *Center) Process(msg model.Message) {
 	app, err := msgToApplication(msg)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -369,7 +369,6 @@ func msgToApplication(msg model.Message) (*Application, error) {
 	return app, nil
 }
 
-// TODO: upgrade to parallel process
```
GsssC, line 372: is this pr support parallelly process application?

GsssC (commented):
overall lgtm
: ) thx for your contribution

cloud/pkg/dynamiccontroller/application/application.go
Iceber: Parallel processing was not resolved, it may have been accidentally deleted.

But I found areas in the `Process` where there might be concurrency issues.

```go
err := c.HandlerCenter.AddListener(app.ToListener(*option))
```
concurrent map read and map write
https://github.com/kubeedge/kubeedge/blob/1678686e9b557ccb0c2c57985954604c13323372/cloud/pkg/dynamiccontroller/application/eventhandler.go#L140-L163

Is this the problem to be solved if parallel processing is to be implemented, or are there some other problems

cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -369,148 +369,151 @@ func msgToApplication(msg model.Message) (*Application, error) {
...
 	if err != nil {
 		c.Response(app, msg.GetID(), Rejected, err.Error(), nil)
 		klog.Errorf("[metaserver/applicationCenter]failed to process Application(%+v), %v", app, err)
+		return
 	}
+	c.Response(app, msg.GetID(), Approved, "", resp)
 	klog.Infof("[metaserver/applicationCenter]successfully to process Application(%+v)", app)
 }
 
+// ProcessApplication processes application
```
GsssC, line 392:
```suggestion
// ProcessApplication processes application by re-translating it to kube-api request with kube client, 
// and then the request are processed by apiserver eventually. 
// Specially if app.verb == watch, it transfer to a listener and register to HandlerCenter, continuously listen kube-api
// change events and push them to edge node, rather then requetes to apiserver directly.
```

cloud/pkg/dynamiccontroller/application/application.go
GsssC: @Iceber Yes, I think `concurrent map read and map write` is one of the problems in parallelly processing application. This only influence **Processing Watch**, and has no relationship to Get and the Write(create delete update  patch). We need think more and could fix it in next pr. 
:)

Iceber:
@GsssC updated

GsssC:
/lgtm
will add `approved` tag if another one lgtm.

GsssC:
Seems no others oppose.
/approve

PR #2711, opened 2021-03-24
Fix the startup process when using Unix protocol on UDS

wzshiming:

**What type of PR is this?**
/kind bug

1. Avoid deleting empty directory that may exist
2. Allow relative paths https://github.com/kubeedge/kubeedge/blob/16ed69ac85b02bfcbcb51487bb4c9641de3fbe08/pkg/apis/componentconfig/cloudcore/v1alpha1/types.go#L145-L153
3. Create directory if need

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/servers/udsserver/uds.go
+++ b/cloud/pkg/cloudhub/servers/udsserver/uds.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


Iceber:
fixing: https://github.com/kubeedge/kubeedge/pull/2686

wzshiming:
oh, not notice

PR #2716, opened 2021-03-24
Fix wrong field used for calculating length

wzshiming:

**What type of PR is this?**
/kind bug

https://github.com/kubeedge/kubeedge/blob/35c10ff454e70e0489f860e892d5bdda19b050b8/edgemesh/pkg/dns/dns.go#L424-L425
https://github.com/kubeedge/kubeedge/blob/35c10ff454e70e0489f860e892d5bdda19b050b8/edgemesh/pkg/dns/dns.go#L431-L432

Fixes #

```release-note
NONE
```


chendave (commented):
Good catch!

/lgtm
/approve

PR #2719, opened 2021-03-24
Avoid possible OOM

wzshiming:

**What type of PR is this?**
/kind bug

Fixes #

```release-note
NONE
```

### DIFF
--- a/staging/src/github.com/kubeedge/viaduct/pkg/packer/reader.go
+++ b/staging/src/github.com/kubeedge/viaduct/pkg/packer/reader.go
@@ -7,6 +7,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	maxPayloadLen = 32 * 1 << 20 // 32Mi
+)
+
 type Reader struct {
 	reader io.Reader
 }
@@ -30,7 +34,7 @@ func (r *Reader) Read() ([]byte, error) {
 	_, err := io.ReadFull(r.reader, headerBuffer)
 	if err != nil {
 		if err != io.EOF {
-			klog.Error("failed to read package header from buffer")
+			klog.Errorf("failed to read package header from buffer: %v", err)
 		}
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


staging/src/github.com/kubeedge/viaduct/pkg/packer/reader.go
```diff
@@ -27,25 +31,43 @@ func (r *Reader) Read() ([]byte, error) {
...
-			klog.Error("failed to read package header from buffer")
+			klog.Error("failed to read package header from buffer", err)
 		}
 		return nil, err
 	}
+	if HeaderSize != n {
+		err := fmt.Errorf("failed to read package header: invalid header, length want %d, but got %d", HeaderSize, n)
+		klog.Error(err)
+		return nil, err
+	}
```
gy95, line 45: `io.ReadFull` will return an error if fewer bytes were read, so i dont think this `if` statement is neccessary when we have already decided whether `err` is nil or not.

staging/src/github.com/kubeedge/viaduct/pkg/packer/reader.go
```diff
@@ -27,25 +31,43 @@ func (r *Reader) Read() ([]byte, error) {
 	}
 
 	headerBuffer := make([]byte, HeaderSize)
-	_, err := io.ReadFull(r.reader, headerBuffer)
+	n, err := io.ReadFull(r.reader, headerBuffer)
 	if err != nil {
 		if err != io.EOF {
-			klog.Error("failed to read package header from buffer")
+			klog.Error("failed to read package header from buffer", err)
 		}
```
gy95, line 38: need `%v` here

wzshiming:
@gy95  Thanks, updated

staging/src/github.com/kubeedge/viaduct/pkg/packer/reader.go
```diff
@@ -7,6 +7,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
+const (
+	maxPayloadLen = 32 * 1 << 20 // 32Mi
```
fisherxu, line 11: What is the basis of this value？

wzshiming: This type of limit in k8s is generally 10MB, but I feel that this scenario may have a configuration file exceeding 10MB, so increase it to 32MB

PR #2720, opened 2021-03-24
pkg/metaserver: optimize key.go

wzshiming:
**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/metaserver/key.go
+++ b/pkg/metaserver/key.go
@@ -30,68 +30,60 @@ func KeyFuncObj(obj runtime.Object) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	var key string
-	key += "/"
-	gvk := obj.GetObjectKind()
-	if gvk.GroupVersionKind().Empty() {
+
+	objKind := obj.GetObjectKind()
+	gvk := objKind.GroupVersionKind()
+	if gvk.Empty() {
 		return "", fmt.Errorf("could not get group/version/kind information in obj")
 	}
-	group := gvk.GroupVersionKind().Group
-	version := gvk.GroupVersionKind().Version
-	resources := util.UnsafeKindToResource(gvk.GroupVersionKind().Kind)
-	namespaces := accessor.GetNamespace()
+	group := gvk.Group
+	version := gvk.Version
+	resource := util.UnsafeKindToResource(gvk.Kind)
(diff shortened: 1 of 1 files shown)


pkg/metaserver/key.go
```diff
@@ -30,30 +30,26 @@ func KeyFuncObj(obj runtime.Object) (string, error) {
...
+
+	objKind := obj.GetObjectKind()
+	gvk := objKind.GroupVersionKind()
+	if gvk.Empty() {
 		return "", fmt.Errorf("could not get group/version/kind information in obj")
 	}
-	group := gvk.GroupVersionKind().Group
-	version := gvk.GroupVersionKind().Version
-	resources := util.UnsafeKindToResource(gvk.GroupVersionKind().Kind)
+	group := gvk.Group
```
daixiang0, line 39: Could we reduce temp val?

daixiang0:
Merge with #2724 please.

pkg/metaserver/key.go
wzshiming: `GroupVersionKind()` is not a zero-cost function in most implementations, I don't think it is necessary to use it multiple times.
https://github.com/kubeedge/kubeedge/blob/f30b2021cded5191aa116f8ff51dbc9dc1f677d3/vendor/k8s.io/apimachinery/pkg/runtime/schema/group_version.go#L296-L305

pkg/metaserver/key.go
```diff
@@ -30,68 +30,60 @@ func KeyFuncObj(obj runtime.Object) (string, error) {
...
+	namespaces := info.Namespace
+	name := info.Name
+	if namespaces == "" {
+		namespaces = v2.NullNamespace
+	}
+	if name == "" {
+		name = v2.NullName
+	}
+
+	key := "/" + group + "/" + version + "/" + resource + "/" + namespaces + "/" + name
```
daixiang0, line 86: Better to use string.Join or fmt.Sprintf

wzshiming: Thanks, updated

GsssC (commented):
/lgtm thx for contribution! : )

fisherxu (commented):
/lgtm
/approve

PR #2755, opened 2021-04-02
fix proxy append space lines to resolv.conf

k-9527:
**What type of PR is this?**
/kind bug

1. for edgemesh proxy.go, when execure 'ensureResolvForHost()' method will append many space lines to resolv.conf file.
image

2. strings.split() will not return nil slice
`resolv := strings.Split(string(bs), "\n")`  // return []string slice with len==1

3. poor judgement when calculate dnsIdx
`strings.Contains(item, config.Config.ListenIP.String())`
line 'nameserver 172.17.0.100' fit requirement but not equal '172.17.0.1',so we should take a deep check with ip.

Fixes #

```release-note

```

### DIFF
--- a/edgemesh/pkg/proxy/proxy.go
+++ b/edgemesh/pkg/proxy/proxy.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


fisherxu:
@k-9527 Now the edgemesh has been moved to https://github.com/kubeedge/edgemesh , please move this patch to edgemesh repo, thanks.

/close

PR #2768, opened 2021-04-12
skip init edged if disable

GsssC:
**What type of PR is this?**
/kind bug
Although user disable edged, it will be created and inited by func newEdged, which cause unnecessary consumption of resources.
None

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -411,6 +411,12 @@ func (e *edged) cgroupRoots() []string {
 
 //newEdged creates new edged object and initialises it
 func newEdged(enable bool) (*edged, error) {
+	// skip init edged if disabled
+	if !enable {
+		return &edged{
+			enable: enable,
+		}, nil
+	}
 	backoff := flowcontrol.NewBackOff(backOffPeriod, MaxContainerBackOff)
 
 	podManager := podmanager.NewPodManager()


edge/pkg/edged/edged.go
```diff
@@ -286,6 +286,10 @@ func (e *edged) GetRequestedContainersInfo(containerName string, options cadviso
 }
 
 func (e *edged) Start() {
+	if e.Enable() == false {
```
Iceber, line 289:
```
if !e.Enabled {
}
```
is this better?

edge/pkg/edged/edged.go
```diff
@@ -411,6 +415,12 @@ func (e *edged) cgroupRoots() []string {
 
 //newEdged creates new edged object and initialises it
 func newEdged(enable bool) (*edged, error) {
+	// skip init edged if disable
+	if enable == false {
```
Iceber, line 419: ditto

edge/pkg/edged/edged.go
GsssC: @Iceber Done, thx, :)

edge/pkg/edged/edged.go
```diff
@@ -411,6 +415,12 @@ func (e *edged) cgroupRoots() []string {
 
 //newEdged creates new edged object and initialises it
 func newEdged(enable bool) (*edged, error) {
+	// skip init edged if disable
```
fisherxu, line 418:
```suggestion
	// skip init edged if disabled
```

edge/pkg/edged/edged.go
```diff
@@ -286,6 +286,10 @@ func (e *edged) GetRequestedContainersInfo(containerName string, options cadviso
 }
 
 func (e *edged) Start() {
+	if !e.Enable() {
+		klog.Error("edged is disabled, failed to start")
+		return
+	}
```
fisherxu, line 292: Same as comment in #2771

Issue #2760, opened 2021-04-07
[Autonomic Kube-API Endpoint] feature Introduction


GsssC:
# Autonomic Kube-API Endpoint
Autonomic Kube-API Endpoint(hereinafter referred to as AKE) is a very critical feature for meeting the requirement of accessing Kube API Instance at edge. It is a HTTP server to serve the Kube API request from kubernetes client (e.g. kubectl kubelet kubeproxy cni-plugin-agent istio). For now, it could serve most of Kube API verb like **Get List Watch Create Delete Update Patch**. [More Detail](https://github.com/GsssC/kubeedge/blob/ake-proposal/docs/proposals/Autonomic%20Kube-API%20Endpoints.md)
![image](https://user-images.githubusercontent.com/29768082/114855474-97e98780-9e18-11eb-8bf0-b4b9e605f947.png)

## **Benifits**
- **Feature development against edge scene would be decoupled with edgecore**
- **Compatible with kubernetes native solutions, like kubelet, istio, etc**
- **No longer need to extend new controllers in cloudcore if edges want new kind of Kube API.** 
## **Feature track**:
- [x] Support Get List Watch Create Delete Update Patch
- [x] Support to access CRDs
- [ ] **Add proposal and document in both Chinese and English(Urgent)**
- [ ] **Add e2e test(Urgent)**
- [ ] Add AKE client(an implementation of [clientset.Interface](https://github.com/kubernetes/kubernetes/blob/c90330d8f4ca9fd980df24044960a4d8bb28a780/staging/src/k8s.io/client-go/kubernetes/clientset.go#L70)) to meet **in-process accessing** requirement
- [ ] Add **lifecycle and Gabbage-Collection for listener and informer**(components to serve watch request) at cloud
- [ ] Support to serve request when Accept and Content-Type is **"application/vnd.kubernetes.protobuf"**
- [ ] Support **identification and authentication** of request.
- [ ] Try [Bbolt](https://github.com/etcd-io/bbolt) as basic storage.

## **Related Works**:
- [ ] Refactor edged to get Kube API by AKE client
- [ ] Try to move edged out of edgecore as a release of lite-kubelet.
- [ ] Refactor device-twin to get device api by AKE client
# User Manual
## Enable AKE feature
To enable AKE feature we need to config cloudcore and edgecore to enable ake-related modules, and would better disable other module.
### **1. Config CloudCore**
Enable **DynamicController**, **SyncController** and **CloudHub** like below:
```yaml
[...]

PR #2770, opened 2021-04-13
support metaserver to resolve meta request like /api

GsssC:
# About This PR
**What type of PR is this?**
/kind feature
Some MetaServer 's clients will send meta request which is generated by a kind of kube-client called [Discovery](https://github.com/kubernetes/kubernetes/blob/93c7b24562d80959f45c308e7412456a410b9b25/staging/src/k8s.io/client-go/discovery/discovery_client.go#L53-L61):
```go
// DiscoveryInterface holds the methods that discover server-supported API groups,
// versions and resources.
type DiscoveryInterface interface {
	RESTClient() restclient.Interface
	ServerGroupsInterface
	ServerResourcesInterface
	ServerVersionInterface
	OpenAPISchemaInterface
}
``` 
And the REST path of this kind of meta request like below:
1. /api
2. /apis
3. /api/v1

Some native kubernetes application will use this kind of client For example:
1. **kube-proxy**,https://github.com/kubernetes/kubernetes/blob/fc4f91f10b0ea36c0be002a3810e36ae6ac5fe54/cmd/kube-proxy/app/server.go#L747
2. **kubectl**, https://github.com/kubernetes/kubernetes/blob/6c96a059dd4a64f1917378b4a0c13899dff31918/staging/src/k8s.io/kubectl/pkg/cmd/util/factory_client_access.go#L71
3. **event-broker**, https://github.com/kubernetes/kubernetes/blob/ade242288346d968ab97f42ffb30e0855e235afc/staging/src/k8s.io/client-go/tools/events/event_broadcaster.go#L344

To Compatible with it, metaserver must to have the abblity to resolve such meta request. **And I think it is a universal demand that an edge application will use Discovery client to check if an crd are supported or registerd in cluster.**

# Appendix: How to connect metaserver with **kubectl**
The kubectl command line tool lets you control Kubernetes clusters. How to use kubectl please refer to [kubernetes offical guidance document](https://kubernetes.io/docs/reference/kubectl/overview/). Now we:
1.  support part of commands like **get, delete, create, describe, -o yaml**,  
2.  but do not support **No-Object** command like **attach, top, exec, proxy,logs and so on**.
## Enable AKE feature
Plz refer to #2760
## Config kubeconfig
Before using kubectl, we need to prepare a kubeconfig which points to metaserver. There are two ways to achieve it:
**1. create a new kubeconfig**
The content of new kubeconfig is like below, name it _ake-kubeconfig.yaml_, and then use `kubectl --kubeconfig=ake-kubeconfig.yaml` every time when access metaserver.
```yaml
apiVersion: v1
kind: Config
clusters:
- cluster:
    server: http://127.0.0.1:10550
  name: ake-test-cluster
contexts:
- context:
    cluster: ake-test-cluster
    user: ""
  name: ake-test-ctx
current-context: ake-test-ctx
preferences: {}
users: null
```
**2. modify default kubeconfig file at _~/.kube/config_ if have.**
Add a new context called _ake-test-ctx_ and change _current-context_ to it. Use `kubectl config view` to show whether the modified kubeconfig is in effect.
```yaml
apiVersion: v1
kind: Config
clusters:
- cluster: #old cluster, don not touch
  ...
- cluster: #add a new cluster
    server: http://127.0.0.1:10550
  name: ake-test-cluster
contexts:
- context: #old context, do not touch
  ...
- context: #add a new context
    cluster: ake-test-cluster
    user: "" 
  name: ake-test-ctx
current-context: ake-test-ctx #change old context to new
preferences: {} #do not touch
users: #do not touch 
  ...
```
## Example
image
image

image

image

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/application.go
+++ b/cloud/pkg/dynamiccontroller/application/application.go
@@ -28,6 +28,7 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/common/message"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub"
 	"github.com/kubeedge/kubeedge/pkg/metaserver"
+	"github.com/kubeedge/kubeedge/pkg/metaserver/metarequset"
 )
 
 // used to set Message.Route
@@ -424,15 +425,16 @@ func (c *Center) ProcessApplication(app *Application) (interface{}, error) {
 		}
 		return nil, nil
 	case Get:
-		var option = new(metav1.GetOptions)
-		if err := app.OptionTo(option); err != nil {
-			return nil, err
+		switch gvr.Group {
+		case "meta":
+			return processMetaGet(gvr, name)
+		default:
+			var option = new(metav1.GetOptions)
+			if err := app.OptionTo(option); err != nil {
+				return nil, err
+			}
+			return c.kubeclient.Resource(gvr).Namespace(ns).Get(context.TODO(), name, *option)
 		}
-		retObj, err := c.kubeclient.Resource(gvr).Namespace(ns).Get(context.TODO(), name, *option)
-		if err != nil {
-			return nil, err
-		}
-		return retObj, nil
 	case Create:
 		var option = new(metav1.CreateOptions)
 		if err := app.OptionTo(option); err != nil {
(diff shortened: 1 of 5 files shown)


GsssC:
@fisherxu @Iceber PTAL : )

pkg/metaserver/metarequest.go
```diff
@@ -0,0 +1,104 @@
...
+			groupIndex := 2
+			return APIGroups, slice[groupIndex], true
+		case 4: // /apis/group/version
+			groupIndex := 2
+			versionIndex := 3
+			gv := schema.GroupVersion{Group: slice[groupIndex], Version: slice[versionIndex]}
+			name := ResourceListNameParser.New(gv)
+			return APIResourceLists, name, true
+		}
+	}
```
Iceber, line 79: I would use this style of code
```go
        switch info.Path {
        case "/api":
                return APIVersions, "cloud-apiversions", true
        case "/apis":
                return APIGroupLists, "cloud-apigrouplist", true
        case "/api/v1":
                gv := schema.GroupVersion{Group: "", Version: "v1"}
                return APIResourceLists, ResourceListNameParser.New(gv), true
        }

        if strings.HasPrefix(info.Path, "/apis") {
                path := strings.TrimRight(info.Path, "/")
                slice := strings.Split(path, "/")
                switch len(slice) {
                case 3: // /apis/group
                        groupIndex := 2
                        return APIGroups, slice[groupIndex], true
                case 4: // /apis/group/version
                        groupIndex, versionIndex := 2, 3
                        gv := schema.GroupVersion{Group: slice[groupIndex], Version: slice[versionIndex]}
                        return APIResourceLists, ResourceListNameParser.New(gv), true
                }
        }
        return schema.GroupVersionResource{}, "", false
```
The function already specifies that the second return value is `name`, so it may not be necessary to declare a new `name` to indicate what the `ResourceListNameParser.New(gv)` means

pkg/metaserver/metarequest.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+// DecorateMetaRequest check if the request is meta request and transform it to
+// get request.
+// meta request is that wants to get resource which is belong to Group meta like:
+// Path				NewGetPath										Want Resources
+// /api				/api/meta/v1/apiversions/cloud-apiversions		metav1.APIVersions
+// /apis			/api/meta/v1/apigroups/cloud-apigroups			metav1.APIGroups
+// /api/v1			/api/meta/v1/apiresourcelists/{SEP}v1			metav1.APIResourcesList in /v1
+// /apis/apps/v1	/api/meta/v1/apiresourcelists/apps{SEP}v1		metav1.APIResourcesList in apps/v1
```
Iceber, line 32: format comment,eg:
```go
// |     Path     |              NewGetPath                   |         Want Resources            |
// |--------------|-------------------------------------------|-----------------------------------|
// |/api          |/api/meta/v1/apiversions/cloud-apiversions |metav1.APIVersions                 |
// |/apis         |/api/meta/v1/apigroups/cloud-apigroups     |metav1.APIGroups                   |
// |/api/v1       |/api/meta/v1/apiresourcelists/{SEP}v1      |metav1.APIResourcesList in /v1     | 
// |/apis/apps/v1 |/api/meta/v1/apiresourcelists/apps{SEP}v1  |metav1.APIResourcesList in apps/v1 |
```

pkg/metaserver/metarequest.go
GsssC: @Iceber Seems Good! Thx

GsssC:
Adding tests

GsssC:
All is ready, added tests.
/cc @fisherxu @Iceber

cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -525,3 +527,54 @@ func (c *Center) Response(app *Application, parentID string, status applicationS
...
+		default:
+			absPath = "/apis/" + gv.Group + "/" + gv.Version
+		}
+		err = client.GetKubeClient().Discovery().RESTClient().Get().AbsPath(absPath).Do(context.TODO()).Into(v)
+		if err != nil {
+			return nil, err
+		}
+		return &metarequset.NamedAPIResourceList{APIResourceList: *v, ObjectMeta: metadata}, nil
+	default:
+		return nil, fmt.Errorf("unsupprot meta get gvr(%+v),name(%v)", gvr.String(), name)
```
Iceber, line 578: There should be no need to use `%+v` if it is a string

Iceber (commented):
overall lgtm

cloud/pkg/dynamiccontroller/application/application.go
GsssC: get it thx.

fisherxu (commented):
We still need some background details or usage case for this feature. :)

cloud/pkg/dynamiccontroller/application/application.go
GsssC: @Iceber done, thx :)

GsssC:
@fisherxu @Iceber I have updated the description of this pr, PTAL : )

PR #2771, opened 2021-04-13
Skip init device controller if disable

GsssC:
**What type of PR is this?**
/kind cleanup
Although user disable device controller, it will be still created and inited, which cause unnecessary consumption of resources.

### DIFF
--- a/cloud/pkg/devicecontroller/devicecontroller.go
+++ b/cloud/pkg/devicecontroller/devicecontroller.go
@@ -21,6 +21,9 @@ type DeviceController struct {
 }
 
 func newDeviceController(enable bool) *DeviceController {
+	if !enable {
+		return &DeviceController{enable: enable}
+	}
 	downstream, err := controller.NewDownstreamController(informers.GetInformersManager().GetCRDInformerFactory())
 	if err != nil {
 		klog.Fatalf("New downstream controller failed with error: %s", err)


cloud/pkg/devicecontroller/devicecontroller.go
```diff
@@ -58,6 +61,10 @@ func (dc *DeviceController) Enable() bool {
 
 // Start controller
 func (dc *DeviceController) Start() {
+	if !dc.Enable() {
+		klog.Error("DeviceController is disabled, failed to start")
+		return
+	}
```
fisherxu, line 67: No need to return here, for the disabled module, which will not be started.

fisherxu (commented):
lgtm

cloud/pkg/devicecontroller/devicecontroller.go
fisherxu: The log also can be deleted, when disabled, the Start will not be called

GsssC: @fisherxu, delete these three lines or just return with no log? 
and I think it is the last insurance if someone start disabled module, or process will panic "nil pointer" error at later lines.

fisherxu: Only start the enabled module. https://github.com/kubeedge/kubeedge/blob/master/staging/src/github.com/kubeedge/beehive/pkg/core/core.go#L18

GsssC: @fisherxu well, get it

GsssC: @fisherxu done

PR #2777, opened 2021-04-15
Two minor changes:

sailorvii:
1. Add debug info for error.
2. Add type validation according to the device model definition.

**What type of PR is this?**
/kind cleanup

The type in edgecore-device twin is not aligned with cloudcore-devicecontroller and device profile. Refine it.

### DIFF
--- a/edge/pkg/devicetwin/dtcommon/util.go
+++ b/edge/pkg/devicetwin/dtcommon/util.go
@@ -2,6 +2,7 @@ package dtcommon
 
 import (
 	"errors"
+	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


sailorvii:
@fisherxu

fisherxu:
cc @gy95

Please fix the ci issue first :)

edge/pkg/devicetwin/dtcommon/util.go
```diff
@@ -13,15 +14,15 @@ func ValidateValue(valueType string, value string) error {
 	case "":
 		valueType = "string"
 		return nil
-	case "string":
+	case "string", "bytes":
```
fisherxu, line 17: Have any consts for these type?

sailorvii: Yes, there should be.
As defined in device crd, the types include "int, string, double, float, boolean, bytes".

fisherxu (commented):
Please fix the DCO issue.

sailorvii:
Move to another pr.

PR #2781, opened 2021-04-15
Stop to create listener when application center serve list request

GsssC:
Signed-off-by: GsssC <914617455@qq.com>
**What type of PR is this?**
/kind bug
Now when application center gets a list request, it will response a list and create listener to continuously push watched event to edge, even if edge don't care about these subsequent changes. It cause a large extra resource consumption, when list request increase.

Fixes #


fisherxu (commented):
/lgtm
/approve

PR #2806, opened 2021-04-22
Fix: The server could not find the requested resource

### LINKED ISSUE #2801: [AKE]The server could not find the requested resource
**What happened**:
MetaServer' local storage response error "the server could not find the requested resource" when there is no resource , rather than a empty list. It is inconsistent with API Server response.
- MetaServer
image
- APIServer
image

Return empty list.
request metaserver an resource that has no instance in cluster.

**Environment**:
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): v1.6.1

Rachel-Shao:
**What type of PR is this?**
/kind bug

When applying for a resource without an instance in cluster, APIServer will return an empty list.
But MetaServer' local storage response error "the server could not find the requested resource" when there is no resource , rather than an empty list.

Fixes #2801

none

none

### DIFF
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go
@@ -148,9 +148,6 @@ func (s *imitator) List(ctx context.Context, key string) (Resp, error) {
 	if err != nil {
 		return Resp{}, err
 	}
-	if len(*results) == 0 {
-		return Resp{}, fmt.Errorf("the server could not find the requested resource")
-	}
 	resp.Kvs = results
 	return resp, nil
 }


edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go
```diff
@@ -148,9 +148,6 @@ func (s *imitator) List(ctx context.Context, key string) (Resp, error) {
 	if err != nil {
 		return Resp{}, err
 	}
-	if len(*results) == 0 {
-		return Resp{}, fmt.Errorf("the server could not find the requested resource")
-	}
```
fisherxu, line 153: This means when results is empty, just return an empty response, right?

fisherxu (commented):
Please add more description for the PR and please use the meaningful commit message.

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go
Rachel-Shao: yes

PR #2808, opened 2021-04-23
Fix: Separate the data of meta and metav2

### LINKED ISSUE #2802: Separate the data of meta and metav2
When MetaManager receives the event pushed from the dynamicController, it only needs to be stored in metav2

**Why is this needed**:
In the existing code, when the MetaManager receives a event, it will store the event in both metaV2 and meta, as shown in the following code(Insert case as an example):
https://github.com/kubeedge/kubeedge/blob/6ef7f0ca34856bfc50583dacb1827a56d1b9b12f/edge/pkg/metamanager/process.go#L134-L206

This will bring some unknown problems. 
For example, Edged will start unnecessary pods according to the changes of meta, so it is better to separate the data of metav2 from the data of meta.

Rachel-Shao:
**What type of PR is this?**
/kind bug

In the existing code, when the MetaManager receives an event, it will store the event in both metaV2 and meta.
This will bring some unknown problems.
For example, Edged will start unnecessary pods according to the changes of meta, so it is better to separate the data of metav2 from the data of meta.

Fixes #2802

image

In MetaManager, we can perform different processing based on `msg.source` to separate the need to store metav2
What we need to consider is what conditions need to be stored in metav2:
- Events from DynamicController
- Events from SyncController (only those events sent from dynamic controller)

Regarding the first part: no explanation needed
Regarding the second part:

Current situation: 
cloudHub will store the k8s resource object into an object called `ObjectSync` for SyncController to perform synchronization operations
SyncController will read `ObjectSync`, compare `ObjectSync.resourceVersion` and `k8sObject.resourceVersion`, if `ObjectSync.resourceVersion` is less than `k8sObject.resourceVersion`, then synchronize, that is, all use EdgeController as `msg.source` to send msg to edge

Solution (the yellow part):
When ObjectSync is created, mark the `msg.source` information, and then SyncController will send msg to the edge with different `msg.source` based on this information
Considering that some resources may be watched by DynamicController and EdgeController at the same time, multiple msgs may be sent at the same time
For device-related msg, SyncController has not processed it temporarily

need test

none


Rachel-Shao:
@GsssC  PTAL

GsssC:
how about delete?
will edgecontroller's delete msg delete whole sync obj?

Rachel-Shao:
> how about delete?
> will edgecontroller's delete msg delete whole sync obj?

@GsssC 
It seems that for delete msg, it will first add to the `nodeStore` and `nodeQueue` uniformly, and then delete objectSync directly without notification?

https://github.com/kubeedge/kubeedge/blob/66c286ef412c543dd7de86e88a5089c0c1b797cd/cloud/pkg/cloudhub/channelq/channelq.go#L89-L135

https://github.com/kubeedge/kubeedge/blob/66c286ef412c543dd7de86e88a5089c0c1b797cd/cloud/pkg/cloudhub/handler/messagehandler.go#L527-L534

https://github.com/kubeedge/kubeedge/blob/66c286ef412c543dd7de86e88a5089c0c1b797cd/cloud/pkg/cloudhub/handler/messagehandler.go#L578-L582

GsssC:
Seem if one obj are deleted in k8s, in which edgecontroller and dynamiccontroller both interest. Two delete messages will be generated, is there any potential problem?

Rachel-Shao:
> Seem if one obj are deleted in k8s, in which edgecontroller and dynamiccontroller both interest. Two delete messages will be generated, is there any potential problem?

@GsssC 
There do have some problems....When the second delete msg arrives, objectSync will be deleted according to the existing logic, but because objectSync has been deleted when processing the last delete msg, an error will be reported. I can add a judgment to solve this error. How do u think?

GsssC (commented):
/lgtm 
@Iceber Could u help to take a look? :)

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -113,15 +113,23 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
...
 			objectSync, err := q.objectSyncLister.ObjectSyncs(resourceNamespace).Get(synccontroller.BuildObjectSyncName(nodeID, resourceUID))
-			if err == nil && objectSync.Status.ObjectResourceVersion != "" && synccontroller.CompareResourceVersion(msg.GetResourceVersion(), objectSync.Status.ObjectResourceVersion) <= 0 {
+			if err == nil && objectSync.Status.ObjectResourceVersion != "" && synccontroller.CompareResourceVersion(msg.GetResourceVersion(), objectSync.Status.ObjectResourceVersion) == 0 {
+				source := strings.Split(objectSync.Labels["source"], ",")
+				if IsExist(source, msg.GetSource()) {
+					return
+				}
+			}
+
+			if err == nil && objectSync.Status.ObjectResourceVersion != "" && synccontroller.CompareResourceVersion(msg.GetResourceVersion(), objectSync.Status.ObjectResourceVersion) < 0 {
```
Iceber, line 123: suggestion:
```go
                        if err == nil && objectSync.Status.ObjectResourceVersion != "" {
                                cmpValue := synccontroller.CompareResourceVersion(msg.GetResourceVersion(), objectSync.Status.ObjectResourceVersion)
                                if cmpValue < 0 {
                                        return
                                }

                                if cmpValue == 0 {
                                        source := strings.Split(objectSync.Labels["source"], ",")
                                        if IsExist(source, msg.GetSource()) {
                                                return
                                        }
                                }
                        }
```

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -134,6 +142,15 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
...
 
+func IsExist(source []string, msgSource string) bool {
+	for _, s := range source {
+		if s == msgSource {
+			return true
+		}
+	}
+	return false
+}
+
```
Iceber, line 153: In many places are used, it is recommended to put under `cloud/pkg/common/util/` or `/pkg/util/`

cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -529,22 +538,36 @@ func (mh *MessageHandle) saveSuccessPoint(msg *beehiveModel.Message, info *model
 				klog.Errorf("failed to delete message %v, err: %v", msg, err)
 				return
 			}
-			mh.deleteSuccessPoint(resourceNamespace, objectSyncName)
+			_, err := mh.crdClient.ReliablesyncsV1alpha1().ObjectSyncs(resourceNamespace).Get(context.Background(), objectSyncName, metav1.GetOptions{})
+			if err != nil && apierrors.IsNotFound(err) {
+				klog.Infof("ObjectSync: %v has already been deleted", objectSyncName)
+				mh.deleteSuccessPoint(resourceNamespace, objectSyncName)
```
Iceber, line 544: This logic seems to be reversed.
```go
_, err := mh.crdClient.ReliablesyncsV1alpha1().ObjectSyncs(resourceNamespace).Get(context.Background(), objectSyncName, metav1.GetOptions{})<br class="Apple-interchange-newline">
if err != nil && apierrors.IsNotFound(err) {
	klog.Infof("ObjectSync: %v has already been deleted", objectSyncName)
        return
}
mh.deleteSuccessPoint(resourceNamespace, objectSyncName)
return
```
But I suggest just adding an error judgment to the `mh.deleteSuccessPoint` instead of getting it once first.
```go
func (mh *MessageHandle) deleteSuccessPoint(resourceNamespace, objectSyncName string) {
        err := mh.crdClient.ReliablesyncsV1alpha1().ObjectSyncs(resourceNamespace).Delete(context.Background(), objectSyncName, *metav1.NewDeleteOptions(0))
        if err != nil && !apierrors.IsNotFound(err) {
                klog.Errorf("Delete Success Point failed with error: %v", err)
        }
}
```

cloud/pkg/synccontroller/objectsync.go
```diff
@@ -78,8 +80,15 @@ func sendEvents(err error, nodeName string, sync *v1alpha1.ObjectSync, resourceT
...
-		msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, obj)
-		beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		source := strings.Split(sync.Labels["source"], ",")
+		if IsExist(source, modules.DynamicControllerModuleName) {
+			msg := buildDynamicControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, obj)
+			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		}
+		if IsExist(source, modules.EdgeControllerModuleName) {
+			msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, obj)
+			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
```
Iceber, line 90: `buildDynamicControllerMessage` and `buildEdgeControllerMessage`  may return nil, which should be handled for robustness purposes

cloud/pkg/synccontroller/objectsync.go
```diff
@@ -91,8 +100,15 @@ func sendEvents(err error, nodeName string, sync *v1alpha1.ObjectSync, resourceT
...
-		beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		source := strings.Split(sync.Labels["source"], ",")
+		if IsExist(source, modules.DynamicControllerModuleName) {
+			msg := buildDynamicControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.UpdateOperation, obj)
+			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		}
+		if IsExist(source, modules.EdgeControllerModuleName) {
+			msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.UpdateOperation, obj)
+			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		}
```
Iceber, line 111: ditto

cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -511,6 +511,15 @@ func (mh *MessageHandle) send(hi hubio.CloudHubIO, info *model.HubInfo, msg *bee
 	return nil
 }
 
+func IsExist(source []string, msgSource string) bool {
+	for _, s := range source {
```
Iceber, line 515: ditto

Rachel-Shao:
@Iceber fixed  ptal:)

GsssC:
/lgtm 
Better use a constant variable to be represent “127.0.0.1:10550”, could do it in next pr :)
Ping @fisherxu

Rachel-Shao:
need review~

Rachel-Shao:
@fisherxu  plz review!

Rachel-Shao:
@zc2638  plz review

Rachel-Shao:
@daixiang0  plz review

Rachel-Shao:
@fisherxu   plz review

Rachel-Shao:
plz wait,  let me reorganize this pr...

Rachel-Shao:
Chinese:
为了方便review，我来对这个pr重新做一个总结：
image
**现状：**
自从AKE特性开启，云到边就可以从DynamiCcontroller下发资源的更新事件，这些事件消息的来源msg.source是DynamicController的，当然除此之外，云边之间还有来自EdgeController和SyncController的msg，这些个msg到达cloudhub之后，cloudhub会依据它的resourceVersion更新对应的storequeue和nodestore。
cloudhub的另一个协程会从nodeStore中取msg，发送给edge，同时开启ack通道，当收到ack消息的时候，取出该msg对应的resourceVersion更新已存于objectsync的resourceVersion，如果不存在则创建新的objectsync进行存储。
syncController会定时检查objectsync里的resourceVersion和k8s里的对象resourceVersion，如果发现k8s的resourceVersion更大，说明该对象有新的更新没有更新下去，则SyncController会进行更新事件的下发来同步。
在以上过程中，不论是EdgeController或者DynamicController的消息，如果没有及时发送到edge端并确认，则会触发SyncController的同步，而从SyncController发出的同步msg的来源目前都是EdgeController。
以上的消息下发到边缘节点之后，MetaManager负责进行处理，目前MetaManager并未区分来自DynamicController和EdgeController的消息，统一处理，即都会在meta和meta_v2中存储一遍，且在满足相关条件下都会通知edged进行操作，来自SyncController的消息由于msg.source设置为EdgeController所以同上处理。

**可能导致的问题：**
- 数据冗余，同一份数据存储了两遍
- 非相关容器的启动：如#3489 中遇到的问题，因为我在#3293 中将DynamicController下发的msg的resType统一成了单数形式，所以导致edged可以处理这些事件（在此pr之前，如果在edge侧通过MetaServer watch pod，DynamicController也会下发pod到这个发送watch请求的边缘节点，但是由于DynamicController下发的msg.source为复数形式不能被edged处理，所以阴差阳错没有导致一些错误），这种情况下，当在边缘侧执行`curl http://127.0.0.1:10550/api/v1/pods?watch=yes`命令的时候，一些本不属于边缘节点X的pod的更新事件也会被下发至边缘节点X处，在MetaManager中被存储至meta和metav2表，然后被发至edged，edged进而进行相关的操作，于是就可能会导致#3489 提出的问题，即一些本不应该出现在边缘节点X上的pod被运行在了边缘节点X上

**解决方案（图中标黄部分）：**
为了解决这个问题，本pr打算将meta和meta_v2里的数据进行分离，一方面需要在MetaManager的处理逻辑里分离这两种情况，另一方面要解决SyncController下发时的msg.source指定问题：
- MetaManager处理部分：依据msg.source分别处理，对于来自EdgeController的消息，存储在meta，并通知edged做处理，对于来自DynamicController的消息，只是存储于meta_v2，这部分数据用于MetaServer的list/get/watch操作
- SyncController处理部分：SyncController在做同步的时候，用于构造msg的信息来自于objectsync，而objectsync在cloudhub处理信息的时候创建，为了在SyncController下发的时候区分这条消息原本的来源，本pr在cloudhub处理下发消息的过程中，增加了标记msg来源这一过程（ObjectSync.ObjectMeta.Labels字段），同一个pod事件，可能会同时被EdgeController和DynamicController监听到并发送至cloudhub，他们对应同一个objectsync对象（以objectSyncName为key），本pr在cloudhub创建/更新objectsync对象的时候，在ObjectSync.ObjectMeta.Labels字段里标记了对应的msg来源信息（“DC”或者“EC”或者both ”DC,EC“），由此一来，在SyncController做同步的时候，依据ObjectSync.ObjectMeta.Labels字段可以以不同的名义去发送同步消息

同时，我在#3496 中也修复了一个小bug，将由SyncController下发的msg的resType由原来的复数形式改成单数形式，因此现在来自EdgeController、DynamicController、SyncController的所有msg的resType都被统一成了小写的单数形式（不是Pod，也不是pods，是pod）

**还可能的优化：**
- 目前，DynamicController只有在边缘侧收到watch的请求之后才会建立相应的watch连接，是否需要默认开启一些资源的watch？比如pod/configmap/secret等，便于提升meta_v2表内数据的实时性，这样如果遇到断网情况，MetaServer向本地数据库（meta_v2）list/get的内容就会是比较新的，且这样可以保证常用资源在meta和meta_v2里的状态是一致的

以上分析可能有理解错或没考虑到的地方，欢迎讨论(*￣︶￣)

Rachel-Shao:
English:
For the convenience of review, let me re-summarize this pr:
image
**status now:**
Since the AKE feature is turned on, the cloud can send resource events from the `DynamicController`. The msg.source of these event messages is `DynamicController`. Of course, there are also msgs from `EdgeController` and `SyncController` between the cloud and edge. After these msgs arrive at `cloudhub`, `cloudhub` will update the corresponding `storequeue` and `nodestore` according to its `resourceVersion`.
Another `cloudhub` coroutine will fetch msg from the `nodeStore`, send it to the edge, and open the ack channel. When receiving the ack message, take out the `resourceVersion` corresponding to the msg to update the `resourceVersion` stored in `objectsync`, or create a new one if it does not exist `Objectsync` for storage.
`SyncController` will periodically check the `resourceVersion` in `objectsync` and the object `resourceVersion` in k8s. If it finds that the `resourceVersion` of k8s is larger, it means that the object has new updates and has not been updated. `SyncController` will issue events to synchronize.
In the above process, whether it is an `EdgeController` or `DynamicController` message, if it is not sent to the edge side and confirmed in time, it will trigger the synchronization of the `SyncController`, and the source of the synchronization msg sent from the `SyncController` is currently `EdgeController`.
After the above messages are delivered to the edge nodes, `MetaManager` is responsible for processing. Currently, `MetaManager` does not distinguish between the messages from `DynamicController` and `EdgeController`, and they are processed uniformly, that is, they will be stored in `meta` and `meta_v2`, and `Edged` will be notified to proceed if relevant conditions are met. Operation, the message from `SyncController` is processed as above because `msg.source` is set to `EdgeController`.

**Possible problems:**
- Data redundancy, the same data is stored twice
- Start of non-related containers: such as the problem encountered in #3489 , because I unified the `resType` of msg issued by `DynamicController` into a singular form in #3293 . As a result, `Edged` can handle these events (Before this pr, if the pod is watched by `MetaServer` on the edge side, the `DynamicController` will also send the pod events to the edge node who sends the watch request, but because the `msg.source` issued by the `DynamicController` is a plural form, it cannot be  processed by `edged`, so some mistakes did not cause). In this case, when the `curl http://127.0.0.1:10550/api/v1/pods?watch=yes` command is executed on the edge side, some scripts events for pods that do not belong to edge node X will also be sent to edge node X, stored in the `meta` and `meta_v2` tables in `MetaManager`, and then sent to the `Edged`, which may lead to the problem raised by #3489 , that some pods that should not appear on edge node X are run on edge node X

**Solution (the part marked in yellow in the figure):**
In order to solve this problem, the pr intends to separate the data in `meta` and `meta_v2`. On the one hand, the two cases need to be separated in the processing logic of `MetaManager`. On the other hand, it is necessary to solve the problem of `msg.source` specification when `SyncController` is issued:
- MetaManager processing part: separate processing according to `msg.source`, for messages from `EdgeController`, store in `meta`, and notify `Edged` for processing, for messages from `DynamicController`, only store in `meta_v2`, this part of data is used in `MetaServer` list/get/watch operate
- `SyncController` processing part: When `SyncController` is doing synchronization, the information used to construct msg comes from `objectsync`, and `objectsync` is created when `cloudhub` processes the information. In order to distinguish the original source of this message when `SyncController` is issued, in the process of processing the sent message in `cloudhub`, this pr marks the source of msg (`ObjectSync.ObjectMeta.Labels` field). The same pod event may be monitored by `EdgeController` and `DynamicController` at the same time and sent to `cloudhub`. They correspond to the same `objectsync` object. (Using `objectSyncName` as the key), when this pr creates/updates an `objectsync` object in `cloudhub`, the corresponding msg source information is marked in the `ObjectSync.ObjectMeta.Labels` field  (can be "DC" or "EC" or both "DC, EC") . As a result, when `SyncController` is synchronizing, synchronization messages can be sent under different names according to the `ObjectSync.ObjectMeta.Labels` field

At the same time, I also fixed a small bug in #3496, changing the resType of msg issued by `SyncController` from the original plural form to the singular form, so now all msg resTypes from `EdgeController`, `DynamicController`, and `SyncController` are unified into lowercase singular form (not `Pod`, not `pods`, but `pod`)

**Also possible optimizations:**
Currently, the `DynamicController` will only establish the corresponding watch connection after receiving the watch request on the edge side. Is it necessary to enable some resource watches by default? For example, `pod/configmap/secret`, etc., to facilitate the improvement of the real-time performance of the data in the `meta_v2` table, so that if the network is disconnected, the content of the `MetaServer` list/get to the local database (`meta_v2`) will be relatively new, and this can ensure that the resource status in `meta` and `meta_v2` is the same.

The above analysis may be misunderstood or not considered, welcome to discuss  :)

Rachel-Shao:
@fisherxu @gy95 @yz271544  cc

Issue #2764, opened 2021-04-10
customize node labels can't be seen when I add labels in edgecorefile and restart edgecore


ttlv:
**What happened**:
 Kubeedge supports customize node labels  after version 1.6.0,I tryed today but it did't work when I restart edgecore,I can't see any customize node labels when I use ```kubectl get node --show-labels``` command  but when I delete edge node and rejoin,the customize node labels which I add into edgecore's yaml can be showed
**How to reproduce it (as minimally and precisely as possible)**:
just add some customize node labels in edgecore config yaml and then restart edgecore bin
**Environment**:
- Kubernetes version : 1.19.7
- KubeEdge version: cloudcore 1.6.1, edgecore 1.6.1

felix0080:
有更多的细节提供吗？

ttlv:
customsized labels  is meanningful when edgenode first join and if we want to add or update the labels for node, we can use kubectl label command to make it rather than use the configyaml of edgecore, priority of the cloud is top1. so I think it's better  to make keadm to support.

fisherxu:
With this patch #2827, if we restart edgecore, maybe the label will still missing?

ttlv:
> With this patch #2827, if we restart edgecore, maybe the label will still missing?

If we use kubectl label command to add new label for edge node and should we synchronize the new label in the edgecore's config file ?

fisherxu:
>  If we use kubectl label command to add new label for edge node in the cloud and should we synchronize the new label in the edgecore's config file ?

No, I think the edgenode should patch the node spec, if labels in config changed, the labels of edgenode should be patched. @ttlv

PR #2827, opened 2021-05-04
feat/keadm: support customsized labels when use join command

ttlv:
**What type of PR is this?**
/kind feature

we want to use keadm to support customsized labels  

Fixes #
https://github.com/kubeedge/kubeedge/issues/2764

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/join.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join.go
@@ -115,8 +115,8 @@ func addJoinOtherFlags(cmd *cobra.Command, joinOptions *types.JoinOptions) {
 	cmd.Flags().StringVar(&joinOptions.TarballPath, types.TarballPath, joinOptions.TarballPath,
 		"Use this key to set the temp directory path for KubeEdge tarball, if not exist, download it")
 
-	cmd.Flags().StringVarP(&joinOptions.Labels, types.Labels, "l", joinOptions.Labels,
-		`use this key to set the customized labels for node. you can input customized labels like {"key1":"value1", "key2":"value2"}`)
+	cmd.Flags().StringSliceVarP(&joinOptions.Labels, types.Labels, "l", joinOptions.Labels,
+		`use this key to set the customized labels for node. you can input customized labels like key1:value1,key2:value2`)
 }
 
 // newJoinOptions returns a struct ready for being used for creating cmd join flags.
--- a/keadm/cmd/keadm/app/cmd/edge/join.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join.go
@@ -114,6 +114,9 @@ func addJoinOtherFlags(cmd *cobra.Command, joinOptions *types.JoinOptions) {
 
 	cmd.Flags().StringVar(&joinOptions.TarballPath, types.TarballPath, joinOptions.TarballPath,
 		"Use this key to set the temp directory path for KubeEdge tarball, if not exist, download it")
+
+	cmd.Flags().StringVarP(&joinOptions.Labels, types.Labels, "l", joinOptions.Labels,
+		`use this key to set the customized labels for node. you can input customized labels like {"key1":"value1", "key2":"value2"}`)
 }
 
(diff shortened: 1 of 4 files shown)


keadm/cmd/keadm/app/cmd/util/edgecoreinstaller.go
```diff
@@ -122,6 +123,19 @@ func (ku *KubeEdgeInstTool) createEdgeConfigFiles() error {
...
+		s1 := strings.Replace(ku.Labels, "{", "", -1)
+		s2 := strings.Replace(s1, "}", "", -1)
+		labels := strings.Split(s2, ",")
+		for _, label := range labels {
+			key := strings.Split(label, ":")[0]
+			value := strings.Split(label, ":")[1]
+			labelsMap[key] = value
+		}
+		edgeCoreConfig.Modules.Edged.Labels = labelsMap
+	}
```
Iceber, line 138: If the value of the label tag is like `{"key1":"value1", "key2":"value2"}`,then you can use the `json.Unmarshal`
```go
        if strings.TrimSpace(ku.Labels) != "" {
                labels := map[string]string{}
                if err := json.Unmarshal([]byte(ku.Labels), &labels); err != nil {
                        return fmt.Errorf("invalid Labels: %s", err)
                }
                edgeCoreConfig.Modules.Edged.Labels = labels
        }
```

But I recommend using `StringSliceVarP` to resolve the label flag

keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -114,6 +114,9 @@ func addJoinOtherFlags(cmd *cobra.Command, joinOptions *types.JoinOptions) {
 
 	cmd.Flags().StringVar(&joinOptions.TarballPath, types.TarballPath, joinOptions.TarballPath,
 		"Use this key to set the temp directory path for KubeEdge tarball, if not exist, download it")
+
+	cmd.Flags().StringVarP(&joinOptions.Labels, types.Labels, "l", joinOptions.Labels,
+		`use this key to set the customized labels for node. you can input customized labels like {"key1":"value1", "key2":"value2"}`)
```
Iceber, line 119: I recommend using StringSliceVarP to resolve the label flag, like `-l key1=value1,key2=value2`.
What do you think?

ttlv: lgtm,thanks for your suggesstion

keadm/cmd/keadm/app/cmd/util/edgecoreinstaller.go
ttlv: thanks for your suggesstion

ttlv:
@fisherxu this pr is ready to get lgtm now :)

keadm/cmd/keadm/app/cmd/util/edgecoreinstaller.go
```diff
@@ -122,6 +123,15 @@ func (ku *KubeEdgeInstTool) createEdgeConfigFiles() error {
 	}
 	edgeCoreConfig.Modules.EdgeStream.TunnelServer = net.JoinHostPort(cloudCoreIP, strconv.Itoa(constants.DefaultTunnelPort))
 
+	if len(ku.Labels) >= 1 {
+		labelsMap := make(map[string]string)
+		for _, label := range ku.Labels {
+			key := strings.Split(label, "=")[0]
+			value := strings.Split(label, "=")[1]
```
fisherxu, line 130: In the flag comment it's `:` to separate the key and value, here is `=`, which one does the user should chose?

ttlv: yeap I have fixed description and use '=' to replace ':'.I'm sorry for that :)

fisherxu: lgtm, please fix the DCO issue :)

ttlv: ok,DCO had been fixed

fisherxu (commented):
/lgtm
/approve

PR #2828, opened 2021-05-04
Add a full version for beehive and viaduct, which provide the ability to use KubeEdge as dependencies of a custom go module.

Sellto:
**What type of PR is this?**
/kind bug

When we try to add Kubeedge as a dependency of a Go project and try to install it we get the following error:
Capture d’écran 2021-05-04 à 10 15 35

So this PR Add a full version for beehive and viaduct, which provide the ability to use KubeEdge as dependencies of a custom go module.

"NONE"

"NONE"

update in go.mod file:
```release-note
	github.com/kubeedge/beehive v0.0.0-20201125122335-cd19bca6e436
	github.com/kubeedge/viaduct v0.0.0-20201130063818-e33931917980
```


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/kubeedge/kubeedge
+module github.com/sellto/kubeedge
```
daixiang0, line 1: not true

go.mod
```diff
@@ -31,8 +31,8 @@ require (
 	github.com/hashicorp/golang-lru v0.5.3
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/karrick/godirwalk v1.10.12 // indirect
-	github.com/kubeedge/beehive v0.0.0
-	github.com/kubeedge/viaduct v0.0.0
+	github.com/kubeedge/beehive v0.0.0-20201125122335-cd19bca6e436
```
kevin-wangzefeng, line 34: We can't indicate beehive and viaduct to a specific verison here, since they are replaced with `./staging/src/github.com/kubeedge/beehive` and `./staging/src/github.com/kubeedge/viaduct` in the `replace ()` section.
You would have to modify the version tag in your own go.mod where you import the kubeedge/kubeedge repo.

kevin-wangzefeng (commented):
Thanks @Sellto, could you share details about how are you planning to use KubeEdge as dependencies?
Maybe importing the whole repo is not the best solution.

go.mod
Sellto: I know it was just for testing someone

go.mod
Sellto: This the way i have chosen

PR #2830, opened 2021-05-07
devicetwin: ValidateValue supports integer check, just like int

bigclouds:
many device-yaml  files still use integer rather than int, so this patch makes it flexible

**What type of PR is this?**

Fixes #

```release-note

```


fisherxu (commented):
@bigclouds Please signoff the commit to fix the DCO ci issue.

PR #2835, opened 2021-05-08
DeviceTwin: optimize device state's judgement code

bigclouds:
1. optimize device state's judgement code
2. enable to pass "disconnected" and "ok" state to cloudcore

Signed-off-by: longguang.yue <yuelg@chinaunicom.cn>

**What type of PR is this?**

Fixes #

/assign rohitsardesai83

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/device.go
+++ b/edge/pkg/devicetwin/dtmanager/device.go
@@ -86,7 +86,12 @@ func dealDeviceStateUpdate(context *dtcontext.DTContext, resource string, msg in
 	if !ok {
 		return nil, nil
 	}
-	if strings.Compare("online", updatedDevice.State) != 0 && strings.Compare("offline", updatedDevice.State) != 0 && strings.Compare("unknown", updatedDevice.State) != 0 {
+
+	// state refers to definition in mappers-go/pkg/common/const.go
+	state := strings.ToLower(updatedDevice.State)
+	switch state {
+	case "online", "offline", "ok", "unknown", "disconnected":
+	default:
 		return nil, nil
 	}
 	lastOnline := time.Now().Format("2006-01-02 15:04:05")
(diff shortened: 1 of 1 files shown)


edge/pkg/devicetwin/dtmanager/device.go
```diff
@@ -86,7 +90,10 @@ func dealDeviceStateUpdate(context *dtcontext.DTContext, resource string, msg in
 	if !ok {
 		return nil, nil
 	}
-	if strings.Compare("online", updatedDevice.State) != 0 && strings.Compare("offline", updatedDevice.State) != 0 && strings.Compare("unknown", updatedDevice.State) != 0 {
+
+	// state refers to definition in mappers-go/pkg/common/const.go
+	state := strings.ToLower(updatedDevice.State)
+	if _, ok := deviceState[state]; !ok {
```
Iceber, line 96: I think it's better to use switch than to use map.
eg.
```go
        switch updatedDevice.State {
        case "online", "offline", "unknow":
        default:
                return nil, nil
        }
```

edge/pkg/devicetwin/dtmanager/device.go
```diff
@@ -113,7 +118,7 @@ func dealDeviceStateUpdate(context *dtcontext.DTContext, resource string, msg in
 		dtcommon.CommModule,
 		context.BuildModelMessage(modules.BusGroup, "", topic, messagepkg.OperationPublish, payload))
 
-	msgResource := "device/" + device.ID + "/state"
+	msgResource := "device/" + device.ID + dtcommon.DeviceETStateUpdateSuffix
```
fisherxu, line 121: This is a bug?

fisherxu: Actually here updating '/state' to '/state/update', is it intend?

ttlv: the topic which used by mapper for updateing state of device is ```"$hw/events/device/+/state/update"``` and I think it's better to use public constant rather than local too

fisherxu (commented):
/lgtm
/approve

PR #2840, opened 2021-05-12
modify some codes in csidriver

gy95:

**What type of PR is this?**
/kind bug

1、when unmarshal resp failed, should return err but not nil
2、use function `FillBody` to assign content
3、optimize `extractMessage`, first check value, then process msg

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #2842, opened 2021-05-12
remove not used function GetContent

gy95:

**What type of PR is this?**
/kind cleanup

remove not-used function `GetContent`, this function is only used in test cases, And it is not correct to convert `interface` to `string` format simply in this way

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #2848, opened 2021-05-17
admission: fix pod toleration replace

zc2638:
**What type of PR is this?**

/kind bug

So that `admission` can handle `Pod Toleration` normally.

Fixes #

```go
patch := []patchMapValue{{
	Op:   "add",
	Path: "/spec/template/spec/tolerations",
	Value: []map[string]string{{
		"key":      corev1.TaintNodeUnreachable,
		"operator": Exists,
		"effect":   NoExecute,
	}},
}}
```
This patch path is a controller resource, such as `Deployment`.  
However, admission deals with `Pod` resource, and the path should be `/spec/tolerations`.

```release-note

```

### DIFF
--- a/cloud/pkg/admissioncontroller/mutate_offlinemigration.go
+++ b/cloud/pkg/admissioncontroller/mutate_offlinemigration.go
@@ -9,15 +9,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const (
-	Exists    = "Exists"
-	NoExecute = "NoExecute"
-)
-
 type patchMapValue struct {
-	Op    string              `json:"op"`
-	Path  string              `json:"path"`
-	Value []map[string]string `json:"value,omitempty"`
+	Op    string        `json:"op"`
+	Path  string        `json:"path"`
+	Value []interface{} `json:"value,omitempty"`
 }
 
 func mutateOfflineMigration(review admissionv1beta1.AdmissionReview) *admissionv1beta1.AdmissionResponse {
@@ -32,7 +27,6 @@ func mutateOfflineMigration(review admissionv1beta1.AdmissionReview) *admissionv
 	}
 
 	payload := generatePatch(pod.Spec.Tolerations)
-
 	if len(payload) == 0 {
 		return &reviewResponse
 	}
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
Is this a bug? Could you please add more description of the PR?

zc2638:
> Is this a bug? Could you please add more description of the PR?

just have a look

cloud/pkg/admissioncontroller/mutate_offlinemigration.go
```diff
@@ -49,28 +44,25 @@ func mutateOfflineMigration(review admissionv1beta1.AdmissionReview) *admissionv
...
-				toleration.TolerationSeconds = nil
-				patch[0].Op = "replace"
-			}
+	currentTolerations := make([]interface{}, 0, len(tolerations)+1)
+	for _, v := range tolerations {
+		if v.Key == corev1.TaintNodeUnreachable {
+			continue
 		}
+		currentTolerations = append(currentTolerations, v)
+	}
```
fisherxu, line 53: If toleration key is `TaintNodeUnreachable` and effect is `NoSchdule`, do we need to keep it here?

fisherxu:
> just have a look

Really bug.

cloud/pkg/admissioncontroller/mutate_offlinemigration.go
zc2638:
```
currentTolerations = append(currentTolerations, corev1.Toleration{
	Key:      corev1.TaintNodeUnreachable,
	Operator: corev1.TolerationOpExists,
	Effect:   corev1.TaintEffectNoExecute,
})
```
@fisherxu  If toleration key is TaintNodeUnreachable, We use custom toleration content.

zc2638: @fisherxu I think the `effect` can be directly covered into empty to make it adapt to all the effects.

fisherxu: Could we update the `Path: "/spec/template/spec/tolerations",` in another pr first? And leave the key and effect for more discussuion. :)

zc2638: There is no need to open another PR. Except for `path`, other logical results are the same as before

fisherxu: The difference is whether care about `NoSchdule`? Now in the patch only add the `NoExecute`.

zc2638: You can refer to the old version, which is also the default `NoExecution`.
More discussuion, I can open another PR.

zc2638:
need fix

cloud/pkg/admissioncontroller/mutate_offlinemigration.go
```diff
@@ -49,28 +43,28 @@ func mutateOfflineMigration(review admissionv1beta1.AdmissionReview) *admissionv
...
-					toleration.Operator == "Exists" && toleration.TolerationSeconds == nil {
-					return nil
-				}
-				toleration.TolerationSeconds = nil
-				patch[0].Op = "replace"
-			}
+	var hasUnreachable bool
+	currentTolerations := make([]interface{}, 0, len(tolerations)+1)
+	for _, v := range tolerations {
+		if v.Key == corev1.TaintNodeUnreachable {
```
fisherxu, line 49:
```suggestion
		if v.Key == corev1.TaintNodeUnreachable && v.Effect == corev1.TaintEffectNoExecute  {
```
How about add `&& v.Effect == corev1.TaintEffectNoExecute`, if v.Effect is noschedule, we still need to add the `NoExecute`, which is related to the eviction.

zc2638: `NoSchedule `: Kubernetes will not schedule the pod onto that node.
`NoExecute `: the pod will be evicted from the node (if it is already running on the node), and will not be scheduled onto the node (if it is not yet running on the node).

I think `NoSchedule` will be used in the future, and it should not be forcibly overwritten here.

fisherxu: If we only campare the `Key`, and if the effect is `NoSchedule `, we still will set hasUnreachable to true, and will not add the toleration with effect `NoExecute`? And if pod tolerate the effect `NoExecute`, it will not be evicted from node. Do I miss anything? ;-)

zc2638: No, what you understand is correct.
Some design issues are involved here, and we can have some discussion.

fisherxu: Okey, let's have another discussion next Monday :)

fisherxu (commented):
/lgtm
/approve

PR #2853, opened 2021-05-20
support cloudcore running in image container mode

gy95:

**What type of PR is this?**
/kind feature

cloudcore should be HA, and we plan to deploy cloudcore in cluster mode, so should support using keadm to run cluster-running cloudcores.

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
+++ b/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
@@ -1,9 +1,18 @@
 package util
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strings"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
 
 	"github.com/kubeedge/kubeedge/common/constants"
 	types "github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
@@ -18,6 +27,8 @@ type KubeCloudInstTool struct {
 	DNSName          string
 	TarballPath      string
 	CloudCoreRunMode string
+	Replicas         int32
+	NodeSelector     map[string]string
 }
 
 // InstallTools downloads KubeEdge for the specified version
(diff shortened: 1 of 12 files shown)


GsssC:
better keep the way of running cloudcore as binary, add a option like `--run-mode` to keadm,so user could make choice

gy95:
> :) Thx for your contribution. And I think:
> better keep the way of running cloudcore as binary, add a option like `--run-mode` to keadm,so user could make choice

ok

keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
```diff
@@ -70,8 +74,21 @@ func (cu *KubeCloudInstTool) InstallTools() error {
 	return nil
 }
 
-//RunCloudCore starts cloudcore process
-func (cu *KubeCloudInstTool) RunCloudCore() error {
+func (cu *KubeCloudInstTool) RunCloudCoreContainer() error {
+	// start cloudcore image container
+	containerCommand := "for resource in $(ls *.yaml); do kubectl create -f $resource; done"
```
GsssC, line 79: Where this yamls come from? :）

gy95: in source codes `build/cloud/01~08*.yaml`, include namspace, cluster role and etc.

GsssC: @gy95 So users need to download these manaully?

gy95: yes, because if we run cloudcore in container, we should first install pre-requisites, and i think we can follow the guide of `Deploy the cloud part into a k8s cluster`, just create all needed k8s resources at once

GsssC: @gy95
I think `keadm init --run-mod=container` should do more, or this option is confusing to user.

Maybe it is time to start to release cloudcore-image and offer a standare and merged-into-one yaml .
/cc @fisherxu WDYT?

gy95: yes, i think we can download the total.yaml from one website and apply it

GsssC: @zc2638 these yaml is used to init a basic cloudcore but not static, users need to do some configuration to these yaml for some high level feature such as `cloud stream`.

zc2638: @GsssC  I have different opinions.
In addition to the service configuration files, they are basically static files, and users do not need to modify them individually.
e.g. CRD resources、Deployment resources、certgen.sh

zc2638: so,you can use `command` generate the default configuration file. e.g. cloudcore.yaml

GsssC: CRD resources is certainly a static resources, but the others is volatile.
Deploying containerized cloudcore better like deploying `flannel`, `calico` and other k8s application.
```sh
curl cloudcore-deploy.yaml
kubectl apply -f cloudcore-deploy.yaml
```
`keadm` just download and apply `cloudcore-deploy.yaml'

GsssC: According to the feedback by the issues and wechat, cert always confues users, even if there is scripts about cert generation. So we change the way of cert generation from manual into auto. `CloudStream cert` also should generate automaticlly by code.

zc2638: It's a lot of trouble.

You should refer to `kubedm` method. We have specified the deployment method through the command, instead of personalized selection of CNI plugins after platform deployment

GsssC: @zc2638 I do not say personalized selection of CNI plugins after platform deployment, **but Deploying containerized cloudcore would better be like deploying flannel, calico and other k8s application**.

gy95: @GsssC @zc2638 @fisherxu 
now, i use api interface to created prerequisite resources like ns, serviceaccount, and etc.(all yamls are in https://github.com/kubeedge/kubeedge/tree/master/build/cloud) when running cloudcore in container mode, 
but i have a question, now config file is mounted by configmap like this
https://github.com/kubeedge/kubeedge/blob/399b56a80fbfe381c6c4bb680bcd0feb35ca1a81/build/cloud/07-deployment.yaml#L48-L51
and configmap is from this 
https://github.com/kubeedge/kubeedge/blob/master/build/cloud/05-configmap.yaml
Could we mount directly using directory `/etc/kubeedge/config/cloudcore.yaml` to provide configuration.

gy95:
After upgrade golang to 1.16 in PR https://github.com/kubeedge/kubeedge/pull/3190, now we could also use package `embed` to access file from the package directory or subdirectories at compile time
ref : https://pkg.go.dev/embed@go1.16.8
But the difficulty is we could only access file in the package directory or subdirectories at compile time, so this usage like below is not allowed.
```
//go:embed ../../../../../../namespace.yaml
var namespace []byte
```
Could we add go source codes to the directory `build/cloud/01-namespace.yaml` to embed file in go codes? and keadm installer import the new go package to create namespace

PR #2854, opened 2021-05-20
add Gabage-Collection for application

Rachel-Shao:
**What type of PR is this?**
/kind feature

In the current version, the number of `application` will increase with the increase of requests, and `application` will be reused, but it will not be recycled after using. For those `application` that are not used for a long time, it is better to recycle them.

none 

test down

none

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/application.go
+++ b/cloud/pkg/dynamiccontroller/application/application.go
@@ -95,6 +95,7 @@ type Application struct {
 
 	count     uint64 // count the number of current citations
 	countLock sync.Mutex
+	tim       time.Time // record the last closing time of application, only make sense when count == 0
 	//TODO: add lock
 }
 
@@ -120,6 +121,7 @@ func newApplication(ctx context.Context, key string, verb applicationVerb, noden
 		cancel:    cancel,
 		count:     0,
 		countLock: sync.Mutex{},
+		tim:       time.Time{},
 	}
 	app.add()
 	return app
@@ -239,10 +241,20 @@ func (a *Application) Close() {
 		return
 	}
 
+	a.tim = time.Now()
 	a.count--
 	if a.count == 0 {
 		a.Status = Completed
 	}
+}
+
+func (a *Application) LastCloseTime() time.Time {
+	a.countLock.Lock()
+	defer a.countLock.Unlock()
+	if a.count == 0 && !a.tim.Equal(time.Time{}) {
+		return a.tim
+	}
+	return time.Time{}
 }
 
 // used for generating application and do apply
@@ -253,7 +265,9 @@ type Agent struct {
 
 // edged config.Config.HostnameOverride
 func NewApplicationAgent(nodeName string) *Agent {
-	return &Agent{nodeName: nodeName}
+	agent := Agent{nodeName: nodeName}
+	go agent.GC()
+	return &agent
 }
 
 func (a *Agent) Generate(ctx context.Context, verb applicationVerb, option interface{}, obj runtime.Object) *Application {
@@ -326,7 +340,16 @@ func (a *Agent) doApply(app *Application) {
 }
 
 func (a *Agent) GC() {
-
+	time.AfterFunc(time.Minute, func() {
+		a.Applications.Range(func(key, value interface{}) bool {
+			app := value.(*Application)
+			lastCloseTime := app.LastCloseTime()
+			if !lastCloseTime.Equal(time.Time{}) && time.Since(lastCloseTime) >= time.Minute*5 {
+				a.Applications.Delete(key)
+			}
+			return true
+		})
+	})
 }
 
 type Center struct {
--- a/cloud/pkg/dynamiccontroller/application/application.go
+++ b/cloud/pkg/dynamiccontroller/application/application.go
@@ -28,6 +28,7 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/pkg/dynamiccontroller/messagelayer"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/message"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub"
+	metaserverconfig "github.com/kubeedge/kubeedge/edge/pkg/metamanager/metaserver/config"
 	"github.com/kubeedge/kubeedge/pkg/metaserver"
 )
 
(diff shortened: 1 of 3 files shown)


cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -95,6 +95,7 @@ type Application struct {
 
 	count     uint64 // count the number of current citations
 	countLock sync.Mutex
+	tim       time.Time // record the increase and decrease time of application instances
```
GsssC, line 98: better `tim // record the last closing time of application, only make sense when count ==0`
and add a func
```go
func(app *Application) LastCloseTime() tme.Time{
    app.countLock.Lock
    defer app.countLock.Unlock
    if app.count ==0 && app.lastCloseTime != nil {
        return app.lastCloseTIme
    }
    return time.Time{}
}

Comment (Rachel-Shao) [reply cloud/pkg/dynamiccontroller/application/application.go]: fixed

File: cloud/pkg/dynamiccontroller/application/application.go [hunk 2]
```diff
@@ -239,12 +241,22 @@ func (a *Application) Close() {
...
 	a.count--
 	if a.count == 0 {
 		a.Status = Completed
 	}
 }
 
+func (a *Application) LastCloseTime() time.Time {
+	a.countLock.Lock()
+	defer a.countLock.Unlock()
+	if a.count == 0 && !a.tim.Equal(time.Time{}) {
```
GsssC, line 254: better a.tim.IsZero()

cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -326,7 +340,16 @@ func (a *Agent) doApply(app *Application) {
 }
 
 func (a *Agent) GC() {
-
+	time.AfterFunc(time.Minute, func() {
+		a.Applications.Range(func(key, value interface{}) bool {
+			app := value.(*Application)
+			lastCloseTime := app.LastCloseTime()
+			if !lastCloseTime.Equal(time.Time{}) && time.Since(lastCloseTime) >= time.Minute*5 {
```
GsssC, line 347: ditto

cloud/pkg/dynamiccontroller/application/application.go
GsssC: and plz check a.tim != nil

Rachel-Shao: what dose "check a.tim != nil" mean?  the default value of a.tim is time.Time{}, a.tim.IsZero() can detect empty situations?

GsssC: 1. For code robustness, better check a.tim !=nil.
2. yes

Rachel-Shao: I have tried it and will report an error `Cannot convert 'nil' to type 'Time' `

GsssC: my error, it is struct but not point.

Rachel-Shao:
@Iceber  PTAL:)

cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -326,7 +340,16 @@ func (a *Agent) doApply(app *Application) {
...
+	time.AfterFunc(time.Minute, func() {
+		a.Applications.Range(func(key, value interface{}) bool {
+			app := value.(*Application)
+			lastCloseTime := app.LastCloseTime()
+			if !lastCloseTime.IsZero() && time.Since(lastCloseTime) >= time.Minute*5 {
+				a.Applications.Delete(key)
+			}
+			return true
+		})
+	})
```
Iceber, line 352: `AfterFunc`  will only be executed once after one minute, not once every minute.
Doesn't seem to fit the semantics of GC.

cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -253,7 +265,9 @@ type Agent struct {
 
 // edged config.Config.HostnameOverride
 func NewApplicationAgent(nodeName string) *Agent {
-	return &Agent{nodeName: nodeName}
+	agent := Agent{nodeName: nodeName}
+	go agent.GC()
+	return &agent
```
Iceber, line 270: Since there is no limit to the number of `NewApplicationAgent` calls, this may lead to `goroutine leak`.
But currently it will only be used once(https://github.com/kubeedge/kubeedge/search?q=NewApplicationAgent)，so  a TODO or comment can be added to remind other users.

If you want to fix this, you can use the global gc to clean up all the `Applications` of the created agent，
or add a shutdown signal to the agent

cloud/pkg/dynamiccontroller/application/application.go
Rachel-Shao: fixed

cloud/pkg/dynamiccontroller/application/application.go
Rachel-Shao: fixed  PTAL

GsssC: @Rachel-Shao seems not fix the problem of `goroutine leak` if the number of `NewApplicationAgent` calls increase. I prefer to that agent is shared, so how about setting a global agent?
```go
var defaultAgent *Agent
var once sync.Once
func NewApplicationAgent() *Agent{
  once.Do(){
     ...
  }
}

Comment (Rachel-Shao) [reply cloud/pkg/dynamiccontroller/application/application.go]: Modified:)

File: cloud/pkg/dynamiccontroller/application/application.go [hunk 6]
```diff
@@ -239,21 +242,40 @@ func (a *Application) Close() {
...
 }
 
+var defaultAgent *Agent
+var once sync.Once
+
 // edged config.Config.HostnameOverride
 func NewApplicationAgent(nodeName string) *Agent {
-	return &Agent{nodeName: nodeName}
+	once.Do(func() {
+		defaultAgent := Agent{nodeName: nodeName}
```
GsssC, line 273: node name only makesense in first time call. So better delete this parameter

Rachel-Shao:
@GsssC  @Iceber  ptal

GsssC:
/lgtm
/approve 
Thx for ur contribution !

PR #2866, opened 2021-05-31
Fix e2e and keadm e2e ci issue

fisherxu:
**What type of PR is this?**
/kind test

Force to use kind 0.9.0 to create the k8s cluster, since we still don't support v1.21 k8s.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -123,7 +123,7 @@ jobs:
       - name: Install dependences
         run: |
           command -v ginkgo || go get github.com/onsi/ginkgo/ginkgo
-          command -v kind || go get sigs.k8s.io/kind@v0.9.0
+          go get sigs.k8s.io/kind@v0.9.0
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -123,7 +123,7 @@ jobs:
       - name: Install dependences
         run: |
           command -v ginkgo || go get github.com/onsi/ginkgo/ginkgo
-          command -v kind || go get sigs.k8s.io/kind@v0.9.0
+          go get sigs.k8s.io/kind@v0.9.0
```
daixiang0, line 126: I think root cause is https://github.com/kubeedge/kubeedge/blob/master/hack/lib/install.sh#L32, since in Github Action, we always install v0.9.

fisherxu: So do we need to install kind in Github Action?

daixiang0: Still need to avoid install failure since network issue.

PR #2876, opened 2021-06-02
modify the value of tunnel port

muxuelan:
**What type of PR is this?**
/kind bug

While tunnelport is found in "record.IPTunnelPort[localIP]"，it will equals nil without any err，causing panic that invalid memory address or nil pointer deference.

Fixes #

```release-note

```

### DIFF
--- a/cloud/cmd/cloudcore/app/server.go
+++ b/cloud/cmd/cloudcore/app/server.go
@@ -155,27 +155,29 @@ func NegotiateTunnelPort() (*int, error) {
 			return nil, err
 		}
 
-		_, found = record.IPTunnelPort[localIP]
-		if !found {
-			port := negotiatePort(record.Port)
-
-			record.IPTunnelPort[localIP] = port
-			record.Port[port] = true
-
-			recordBytes, err := json.Marshal(record)
-			if err != nil {
-				return nil, err
-			}
-
-			tunnelPort.Annotations[modules.TunnelPortRecordAnnotationKey] = string(recordBytes)
-
-			_, err = kubeClient.CoreV1().ConfigMaps(modules.NamespaceSystem).Update(context.TODO(), tunnelPort, metav1.UpdateOptions{})
-			if err != nil {
-				return nil, err
-			}
-
+		port, found := record.IPTunnelPort[localIP]
+		if found {
 			return &port, nil
 		}
+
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
The cloudcore will panic when restart?

cloud/cmd/cloudcore/app/server.go
```diff
@@ -155,9 +156,9 @@ func NegotiateTunnelPort() (*int, error) {
 			return nil, err
 		}
 
-		_, found = record.IPTunnelPort[localIP]
+		port, found = record.IPTunnelPort[localIP]
```
fisherxu, line 159:
```suggestion
		port, found = record.IPTunnelPort[localIP]
		if found {
		       return port,nil
		}
```

cloud/cmd/cloudcore/app/server.go
```diff
@@ -155,9 +156,9 @@ func NegotiateTunnelPort() (*int, error) {
...
 		if !found {
```
fisherxu, line 160: remove the if here.

cloud/cmd/cloudcore/app/server.go
```diff
@@ -206,11 +207,10 @@ func NegotiateTunnelPort() (*int, error) {
 			return nil, err
 		}
 
-		port := constants.ServerPort
-		return &port, nil
+		port = constants.ServerPort
 	}
 
-	return nil, nil
+	return &port, nil
```
fisherxu, line 213:
```suggestion
	return nil, error("failed to negotiate the tunnel port")
```

muxuelan:
> The cloudcore will panic when restart?

Yes，tunnelport will equals nil directly which causes panic.

cloud/cmd/cloudcore/app/server.go
muxuelan: Okay, I will

fisherxu (commented):
/lgtm
/approve

fisherxu:
@muxuelan Thanks for the fixing, could you please cherry-pick this patch to `release-1.7` branch(use hack/cherry_pick_pull.sh script)?

kevin-wangzefeng:
Fixes: https://github.com/kubeedge/kubeedge/issues/2884

PR #2879, opened 2021-06-03
fix msg log print use correct msg

gy95:

**What type of PR is this?**
/kind bug

message send log print should use sent message itself but not message that it is cloned from 

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #2811, opened 2021-04-25
add an unusual case on kind to resource conversion

Rachel-Shao:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
It added an unusual Kind to Resource conversion.
In #2769 , we found that for resources like `Gateway`, the corresponding resource name should be `gateways` instead of `gatewaies`.  Another example is a resource composed of multiple words, such as `ServiceEntry`. 
These are special cases that needs to be handled specially, so I created a crdmap to record the resource-kind relationship of crd.

**Which issue(s) this PR fixes**:
Fixes #2769

**Special notes for your reviewer**:
none

**Does this PR introduce a user-facing change?**:
none

PR #2888, opened 2021-06-08
fix bug: incorrect storage of some crds

muxuelan:
**What type of PR is this?**
/kind bug

Some crds like gateway，its plural form is gateways instead of gatewaies，causing wrong value stored in the database meta_v2

Fixes 
https://github.com/kubeedge/edgemesh/issues/7
https://github.com/kubeedge/edgemesh/issues/8

```release-note

```

### DIFF
--- a/pkg/metaserver/util/util.go
+++ b/pkg/metaserver/util/util.go
@@ -70,7 +70,7 @@ func UnsafeResourceToKind(r string) string {
 }
 
 func UnsafeKindToResource(k string) string {
-	if len(k) == 0 {
+	if len(k) == 0 || len(k) == 1 {
 		return k
 	}
 	unusualKindToResource := map[string]string{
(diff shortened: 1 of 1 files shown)


Rachel-Shao:
lgtm
Because some resource-kind conversion of crd is irregular, your solution looks good, I am also solving this problem #2811 (for some resources concatenated by multiple words, the resource-kind relationship can only be defined in advance), in order to make the resource-kind conversion of crd more secure, welcome to review:)

muxuelan:
> lgtm
> Because some resource-kind conversion of crd is irregular, your solution looks good, I am also solving this problem #2811 (for some resources concatenated by multiple words, the resource-kind relationship can only be defined in advance), in order to make the resource-kind conversion of crd more secure, welcome to review:)

Okay okay， wait for you~

PR #2889, opened 2021-06-08
optimize device status lifecycle

sailorvii:
1. Add device status update function.
2. Add a debug info for error.
3. Add type validation according to the device model definition.

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/build/crds/devices/devices_v1alpha2_device.yaml
+++ b/build/crds/devices/devices_v1alpha2_device.yaml
@@ -396,6 +396,12 @@ spec:
           type: object
         status:
           properties:
+            state:
+              description: Device state.
+              type: string
+            lastOnline:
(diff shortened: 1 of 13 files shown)


daixiang0:
Please use a more meaningful title like "optimize device status lifecycle"

sailorvii:
> Please use a more meaningful title like "optimize device status lifecycle"

Thanks for the advice.

BTW, do you know how to solve the arm64 compile issue? The error is:
/usr/bin/ld.gold: internal error in try_fix_erratum_843419_optimized, at ../../gold/aarch64.cc:2114

I couldn't produce it in my own env and couldn't solve it by the google.

daixiang0:
I am not familiar with it, you can try build in container to reproduce for same case.

PR #2894, opened 2021-06-11
normalize edgecontroller upstream codes

gy95:

**What type of PR is this?**
/kind cleanup

1. use function instead of assign values
2. initialize map to avoid possible panic when `node labels` map is nil

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #2896, opened 2021-06-11
router processing message add exit rule

gy95:

**What type of PR is this?**
/kind design

cloud router module don't have a rule to skip the infinite loop like other modules, so when beehive context is `Done`, exit the loop

Fixes #

```release-note

```


cloud/pkg/router/listener/message.go
```diff
@@ -54,14 +54,22 @@ func (mh *MessageHandler) getHandler(source string, resource string) (Handle, er
...
+		msg, err := beehiveContext.Receive(module)
+		if err != nil {
 			klog.Errorf("get a message, header:%+v router:%+v, err: %v", msg.Header, msg.Router, err)
+			continue
+		}
+		klog.Infof("get a message, header:%+v router:%+v", msg.Header, msg.Router)
+		err = MessageHandlerInstance.HandleMessage(&msg)
+		if err != nil {
+			klog.Errorf("Process msg error: %s.", err.Error())
+			continue
```
fisherxu, line 72: No need `continue` here?

gy95: yes, I just want to emphasize the act of continuing next loop when some errors happen

fisherxu: can delete it.

gy95: done

fisherxu (commented):
/lgtm
/approve

PR #2897, opened 2021-06-11
make synccontroller manageObject more clear

gy95:

**What type of PR is this?**
/kind cleanup

function `manageObject` is not clear and some codes are confusing, so it's not easy to understand. Refactor the function to make it easier to read.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/synccontroller/objectsync.go
+++ b/cloud/pkg/synccontroller/objectsync.go
@@ -7,10 +7,8 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


cloud/pkg/synccontroller/objectsync.go
```diff
@@ -34,38 +32,26 @@ func (sctl *SyncController) manageObject(sync *v1alpha1.ObjectSync) {
...
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			obj := &unstructured.Unstructured{}
-			obj.SetName(sync.Spec.ObjectName)
-			obj.SetUID(types.UID(getObjectUID(sync.Name)))
-			obj.SetNamespace(sync.Namespace)
-		} else {
-			klog.Errorf("Failed to manage pod sync of %s in namespace %s: %v", sync.Name, sync.Namespace, err)
-			return
-		}
```
gy95, line 67: this `if` branch is unnecessary, `err != nil` happenes only when the `syncObjUID != string(object.GetUID())` and the err is always `notfound`, and here define a not-used `Unstructured` object, so we can delete these codes.

fisherxu (commented):
/lgtm
/approve

PR #2898, opened 2021-06-11
handle processing msg errors more gracefully

gy95:

**What type of PR is this?**
/kind cleanup

use the common way to handle errors gracefully: when errors happen just first return

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -671,16 +671,17 @@ func (m *metaManager) runMetaManager() {
 		for {
 			select {
 			case <-beehiveContext.Done():
-				klog.Warning("MetaManager mainloop stop")
+				klog.Warning("MetaManager main loop stop")
 				return
 			default:
 			}
-			if msg, err := beehiveContext.Receive(m.Name()); err == nil {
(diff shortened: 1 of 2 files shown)


edge/pkg/metamanager/process.go
```diff
@@ -671,16 +671,17 @@ func (m *metaManager) runMetaManager() {
...
 			}
-			if msg, err := beehiveContext.Receive(m.Name()); err == nil {
-				klog.V(2).Infof("get a message %+v", msg)
-				m.process(msg)
-			} else {
+			msg, err := beehiveContext.Receive(m.Name())
+			if err != nil {
 				klog.Errorf("get a message %+v: %v", msg, err)
+				continue
 			}
```
fisherxu, line 682:
```suggestion
			if msg, err := beehiveContext.Receive(m.Name()); err != nil {
				klog.Errorf("get a message %+v: %v", msg, err)
				continue
			}
```

gy95: variable `msg` need to be processed when receive succeed, so cannot use the suggested format.

fisherxu (commented):
/lgtm
/approve

PR #2907, opened 2021-06-16
fix incorrect use of TrimLeft or TrimRight

gy95:

**What type of PR is this?**
/kind bug
/kind cleanup

1. According to the official doc, function `strings.TrimLeft` usage is as follows: 
```
// TrimLeft returns a slice of the string s with all leading
// Unicode code points contained in cutset removed.
//
// To remove a prefix, use TrimPrefix instead.
```
And i think here we should use `TrimPrefix` and `TrimSuffix` instead of `TrimLeft` and `TrimRight`

2. remove not-used function `normalizeResource` in `cloud/pkg/router/utils/path_test.go`

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #2450, opened 2020-12-18
Keadm: support offline deployment

daixiang0:
Signed-off-by: Xiang Dai <long0dai@foxmail.com>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Support offline deployment, users only need download release tarballs then deploy in any offline environment.

**Which issue(s) this PR fixes**:

Fixes #2442 #2307 #2117 #2420 #2424

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Keadm: support offline deployment, with `--checksum=false`, download release tarballs then can deploy in any offline environment.
```

PR #2909, opened 2021-06-16
Solved the checksum file download problem when ke was installed offline

Rachel-Shao:
**What type of PR is this?**
/kind feature

When we use keadm to install kubeedge offline, or when the network condition is not very good, keadm will be very slow to download checksum file, which will greatly reduce the efficiency of our deployment of kubeedge, so we need to provide a function that can download checksum file in advance.

Please download the corresponding checksum file in advance in the specified tarballPath (/etc/kubeedge by default)

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -21,6 +21,7 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


daixiang0:
lgtm

@fisherxu is there a plan to support whole offline deploy like #2450 does?

fisherxu (commented):
/lgtm
/approve

PR #2914, opened 2021-06-19
use function to get content as []byte

gy95:

**What type of PR is this?**
/kind cleanup

we can use provided function to get `[]byte` format, so not need to convert to `[]byte` format using other functions

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #2923, opened 2021-06-24
add function to check hardware architecture

xujingjing-cmss:
Signed-off-by: xujingjing <xujingjing@cmss.chinamobile.com>

**What type of PR is this?**
/kind feature

so that the "keadm debug" can check hardware architecture.

Fixes #


fisherxu (commented):
Thanks for the enhancement, but I think we can't list all the archs. And we can support all the Linux distributions.

xujingjing-cmss:
> Thanks for the enhancement, but I think we can't list all the archs. And we can support all the Linux distributions.

Agree. So I made a commit to delete those description about "check arch" which is redundant.

PR #2938, opened 2021-07-01
Fix kind create cluster failed issue

fisherxu:
Signed-off-by: fisherxu <xufei40@huawei.com>

**What type of PR is this?**
/kind test

**What this PR does / why we need it**:
Use a fixed operating system version in github action to avoid ci failed.

Ref: https://github.com/kubeedge/kubeedge/pull/2937/checks?check_run_id=2949649305
https://github.com/kubeedge/kubeedge/pull/2937/checks?check_run_id=2949649382

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2939, opened 2021-07-02
only when toremove devices exist, remove device

gy95:

**What type of PR is this?**
/kind bug

only when `toRemove` slice is not nil `and` length is not zero, we should call `removeDevice`
`toRemove != nil || len(toRemove) != 0` should be `toRemove != nil && len(toRemove) != 0`, the logic should be `and`
and can be simplified to `len(toRemove) != 0` 

Fixes #

```release-note

```


gy95:
fixed by #2938 , waiting this PR merged

fisherxu (commented):
/lgtm
/approve

PR #2971, opened 2021-07-22
delete duplicate module name eventbus and twin

gy95:

**What type of PR is this?**
/kind cleanup

use the variable already defined in common package, instead of duplicately redefine

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/servicebus/servicebus.go
+++ b/edge/pkg/servicebus/servicebus.go
@@ -98,10 +98,10 @@ func processMessage(msg *beehiveModel.Message) {
 		}
 		return
 	}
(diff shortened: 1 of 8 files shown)


edge/pkg/servicebus/servicebus.go
```diff
@@ -98,7 +98,7 @@ func processMessage(msg *model.Message) {
 		}
 		return
 	}
-	content, err := json.Marshal(msg.GetContent())
+	content, err := msg.GetContentData()
 	if err != nil {
 		klog.Errorf("marshall message content failed %v", err)
```
fisherxu, line 103: Modify the error logs.

PR #2973, opened 2021-07-22
move module name definition to common package

gy95:

**What type of PR is this?**
/kind cleanup

move `EdgeHub` and `MetaManager` module definitions to common package, keep the same with other edge modules

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/common/modules/names.go
+++ b/edge/pkg/common/modules/names.go
@@ -1,16 +1,18 @@
 package modules
 
 const (
-	// EdgedModuleName name
+	// EdgedModuleName is Edged module name
 	EdgedModuleName = "edged"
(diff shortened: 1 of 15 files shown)


edge/pkg/common/modules/names.go
```diff
@@ -1,14 +1,18 @@
 package modules
 
 const (
-	// EdgedModuleName name
+	// EdgeHubModuleName is EdgeHub module name
+	EdgeHubModuleName = "websocket"
```
zc2638, line 5: It seems to conflict with #3085

gy95: yes, when conflict, i'll rebase this PR.

PR #2975, opened 2021-07-22
fix RuleStatus update fail

### LINKED ISSUE #2920: The subresource Status of the routing module crd resource Rule failed to update
**What happened**:
When I am testing custom sending messages，I found that the number of times the message was forwarded was always 0，The resource field is Status.SuccessMessages
When I send a message, the Rule.Status.SuccessMessages increases
Follow this [tutorial，](https://kubeedge.io/en/docs/developer/custom_message_deliver/#how-to-deliver-custom-messages)

**Environment**:
- Kubernetes version (use `kubectl version`): 1.19.1
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`):1.7.1

Read the source code to find the cause of this [upstream.go-updateRuleStatus](https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/edgecontroller/controller/upstream.go)
```go
rule, err = uc.crdClient.RulesV1().Rules(namespace).Get(context.Background(), ruleID, metaV1.GetOptions{})
			if err != nil {
				klog.Warningf("message: %s process failure, get rule with error: %s, namespaces: %s name: %s", msg.GetID(), err, namespace, ruleID)
				continue
			}
			content, ok := msg.Content.(routerrule.ExecResult)
			if !ok {
				klog.Warningf("message: %s process failure, get rule content with error: %s, namespaces: %s name: %s", msg.GetID(), err, namespace, ruleID)
				continue
			}
			if content.Status == "SUCCESS" {
				rule.Status.SuccessMessages++
			}
			if content.Status == "FAIL" {
				rule.Status.FailMessages++
				errSlice := make([]string, 0)
				rule.Status.Errors = append(errSlice, content.Error.Detail)
			}
			newStatus := &rulesv1.RuleStatus{}
			body, err := json.Marshal(newStatus)
			if err != nil {
				klog.Warningf("message: %s process failure, content marshal err: %s", msg.GetID(), err)
				continue
			}
			var data []byte = []byte(body)
			_, err = uc.crdClient.RulesV1().Rules(namespace).Patch(context.Background(), ruleID, controller.MergePatchType, data, metaV1.PatchOptions{})
			if err != nil {
				klog.Warningf("message: %s process failure, update ruleStatus failed with error: %s, namespace: %s, name: %s", msg.GetID(), err, namespace, ruleID)
			} else {
				klog.Infof("UpdateRulestatus successfully!")
			}

```
There are two problems here. A newStatus is only initialized. Second, Rule.Status is a sub-resource and needs to be enabled in crd yaml

xujiaja:
**What type of PR is this?**

/kind bug

Created a new structure of RuleStatus and then assign the value of Rule'status.

Fixes #2920 

none

none

```release-note

```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -220,6 +220,11 @@ func (uc *UpstreamController) dispatchMessage() {
 		}
 	}
 }
+
+type RuleStatus struct {
+	Status rulesv1.RuleStatus `json:"status"`
+}
+
 func (uc *UpstreamController) updateRuleStatus() {
 	for {
 		select {
@@ -257,7 +262,8 @@ func (uc *UpstreamController) updateRuleStatus() {
 				errSlice := make([]string, 0)
 				rule.Status.Errors = append(errSlice, content.Error.Detail)
 			}
-			newStatus := &rulesv1.RuleStatus{}
+			newStatus := &RuleStatus{}
+			newStatus.Status = rule.Status
 			body, err := json.Marshal(newStatus)
 			if err != nil {
 				klog.Warningf("message: %s process failure, content marshal err: %s", msg.GetID(), err)
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -83,6 +83,11 @@ func SortInitContainerStatuses(p *v1.Pod, statuses []v1.ContainerStatus) {
 	}
 }
 
+// RuleStatus is structure to patch rule status
+type RuleStatus struct {
+	Status rulesv1.RuleStatus `json:"status"`
+}
+
 // UpstreamController subscribe messages from edge and sync to k8s api server
 type UpstreamController struct {
 	kubeClient   kubernetes.Interface
@@ -220,11 +225,6 @@ func (uc *UpstreamController) dispatchMessage() {
 		}
 	}
 }
-
-type RuleStatus struct {
-	Status rulesv1.RuleStatus `json:"status"`
-}
-
 func (uc *UpstreamController) updateRuleStatus() {
 	for {
 		select {


cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -220,6 +220,11 @@ func (uc *UpstreamController) dispatchMessage() {
 		}
 	}
 }
+
+type RuleStatus struct {
```
fisherxu, line 224: Why do we need this struct?

fisherxu (commented):
please fix the failed DCO ci test(signoff the commit).

cloud/pkg/edgecontroller/controller/upstream.go
xujiaja: k8s patch RuleStatus must specify the object of status, may be t this structure is placed in the wrong place

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -257,7 +262,8 @@ func (uc *UpstreamController) updateRuleStatus() {
 				errSlice := make([]string, 0)
 				rule.Status.Errors = append(errSlice, content.Error.Detail)
 			}
-			newStatus := &rulesv1.RuleStatus{}
+			newStatus := &RuleStatus{}
+			newStatus.Status = rule.Status
```
fisherxu, line 266: If we always patch the empty status, the count will always 0. We should patch the rule's real status.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -257,7 +262,8 @@ func (uc *UpstreamController) updateRuleStatus() {
...
 			body, err := json.Marshal(newStatus)
```
fisherxu, line 267: Thanks for the fixing, and has any test here? Is the status worked as desired? For the patch func, the data should be patch bytes created through olddata and newdata. 

Ref: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/cmd/annotate/annotate.go#L346

xujiaja:
The k8s cluster failed to initialize due to timeout, please re-execute CI

cloud/pkg/edgecontroller/controller/upstream.go
fisherxu, line 266: Could we use update status as below?
```go
if content.Status == "SUCCESS" {
				rule.Status.SuccessMessages++
			}
			if content.Status == "FAIL" {
				rule.Status.FailMessages++
				errSlice := make([]string, 0)
				rule.Status.Errors = append(errSlice, content.Error.Detail)
			}

			_, err = uc.crdClient.RulesV1().Rules(namespace).UpdateStatus(context.Background(), rule, metaV1.UpdateOptions{})
			if err != nil {
				klog.Warningf("message: %s process failure, update ruleStatus failed with error: %s, namespace: %s, name: %s", msg.GetID(), err, namespace, ruleID)
			} else {
				klog.Infof("UpdateRulestatus successfully!")
			}
```

fisherxu:
And please rebase to master branch to fix the failed ci tests :)

PR #2999, opened 2021-08-04
Add iptables manager for tunnel port

fisherxu:
**What type of PR is this?**
/kind feature

In v1.7, we add the negotiate part, which negotiate the tunnel port for cloudcore. In this pr, we add the iptables manager, which will add the iptables rules for all the cloudcore automatically.

Now the iptables manager is coupled in cloudcore, in future, we can move it to an independent process when deployment ready.

The iptables rule as below:

```shell
Chain PREROUTING (policy ACCEPT 38 packets, 2248 bytes)
 pkts bytes target     prot opt in     out     source               destination         
  823 92521 DOCKER     all  --  *      *       0.0.0.0/0            0.0.0.0/0            ADDRTYPE match dst-type LOCAL
 1633 97916 TUNNEL-PORT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubeedge tunnel port */

Chain OUTPUT (policy ACCEPT 91 packets, 6672 bytes)
 pkts bytes target     prot opt in     out     source               destination         
   19  1140 DOCKER     all  --  *      *       0.0.0.0/0           !127.0.0.0/8          ADDRTYPE match dst-type LOCAL
24329 1505K TUNNEL-PORT  all  --  *      *       0.0.0.0/0            0.0.0.0/0            /* kubeedge tunnel port */

Chain TUNNEL-PORT (2 references)
 pkts bytes target     prot opt in     out     source               destination         
    1    60 DNAT       tcp  --  *      *       0.0.0.0/0            0.0.0.0/0            tcp dpt:10351 to:192.168.1.16:10003
```

Ref: https://github.com/kubeedge/kubeedge/issues/1560

```release-note
Add iptables manager for tunnel port
```

### DIFF
--- a/cloud/cmd/cloudcore/app/server.go
+++ b/cloud/cmd/cloudcore/app/server.go
@@ -24,6 +24,7 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub"
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/servers/httpserver"
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudstream"
+	"github.com/kubeedge/kubeedge/cloud/pkg/cloudstream/iptables"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/informers"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/modules"
@@ -87,6 +88,9 @@ kubernetes controller which manages devices so that the device metadata/status d
 
 			registerModules(config)
 
+			// IptablesManager manages tunnel port related iptables rules
+			go iptables.NewIptablesManager(config.Modules.CloudStream).Run()
+
 			// Start all modules
 			core.StartModules()
 			gis.Start(beehiveContext.Done())
@@ -143,7 +147,7 @@ func NegotiateTunnelPort() (*int, error) {
 
 	localIP := getLocalIP()
 
-	var record options.TunnelPortRecord
+	var record iptables.TunnelPortRecord
 	if err == nil {
 		recordStr, found := tunnelPort.Annotations[modules.TunnelPortRecordAnnotationKey]
 		recordBytes := []byte(recordStr)
@@ -181,12 +185,13 @@ func NegotiateTunnelPort() (*int, error) {
 	}
 
 	if apierror.IsNotFound(err) {
-		record := options.TunnelPortRecord{
+		port := negotiatePort(record.Port)
+		record := iptables.TunnelPortRecord{
 			IPTunnelPort: map[string]int{
-				localIP: constants.ServerPort,
+				localIP: port,
 			},
 			Port: map[int]bool{
-				constants.ServerPort: true,
+				port: true,
 			},
 		}
 		recordBytes, err := json.Marshal(record)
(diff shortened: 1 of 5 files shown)


pkg/apis/componentconfig/cloudcore/v1alpha1/validation/validation.go
```diff
@@ -45,18 +43,6 @@ func ValidateCloudCoreConfiguration(c *v1alpha1.CloudCoreConfig) field.ErrorList
 	return allErrs
 }
 
-//ValidateLeaderElectionConfiguration validates part `l` and returns an errorList if it is invalid, the rest will be validated at run time
-func ValidateLeaderElectionConfiguration(l componentbaseconfig.LeaderElectionConfiguration) field.ErrorList {
```
chendave, line 49: not pretty sure why this is removed, doesn't that mean `leaderelelction` is no longer supported?

chendave (commented):
I guess it's challenging to get those func tested.

/lgtm

pkg/apis/componentconfig/cloudcore/v1alpha1/validation/validation.go
fisherxu: Now we will support active-active mode of cloudcore, so leaderelelction is no loanger needed. :)

chendave:
let me retest whole CI check.

fisherxu:
> let me retest whole CI check.

The failed e2e test is because the compatibility of kind and github action env, already have an issue to track it. #3002

chendave:
/approve 
/hold
hold in case you want someone else to review this, or else feel free to lift the label.

sdghchj:
mark

fisherxu:
/hold cancel
Let's merge it first, then I will add some e2e test to test this feature. Thanks for the review from @chendave :)

cloud/cmd/cloudcore/app/server.go
```diff
@@ -217,10 +221,10 @@ func NegotiateTunnelPort() (*int, error) {
 
 func negotiatePort(portRecord map[int]bool) int {
 	for port := constants.ServerPort; ; {
+		port++
```
sdghchj, line 224: Why start with 10351?

sdghchj, line 224: Why it starts with 10351?

sdghchj:
It looks like that rules will be add many times if cloudcore runs in multi-instances mode

sdghchj:
Cloudcore is required to run in host network in order to add rules, isn't it ?

kevin-wangzefeng:
> It looks like that rules will be add many times if cloudcore runs in multi-instances mode

Actually the number of rules will be equal to number of cloudcore instances

sdghchj:
Can it works in k8s deployment?

lvfei103650:
Ask a question, is cloudcore deployed on k8s master node?

fisherxu:
> Ask a question, is cloudcore deployed on k8s master node?

@lvfei103650 Now it's need to run at master node if users need to set iptables through cloudcore, but we need to make it can run on another node.

lvfei103650:
Ask a question, we use nginx for load balancing between cloudcore and edgecore, how can we ensure that port 10000 and port 10004 are loaded to the same cloudcore server?

We use this nginx conf , not useless.
nginx conf：
image

PR #3011, opened 2021-08-09
Supports defining the message buffer length through configuration

zc2638:
**What type of PR is this?**

/kind feature

```
Supports defining the message buffer length through configuration.

In many cases, because the buffer size of the message channel at the edge is only 1024, the message processing stops, resulting in false death.

When the resources are controllable, the message channel length of each module can be customized.
```

Fixes #

```release-note

```


fisherxu (commented):
@zc2638  Thanks for the fixing, and could you please rebase to the master branch to fix the failed ci tests?

fisherxu (commented):
Looks like there's too many size configuration fields, maybe we can use one common config?

zc2638:
No matter what the problem is, it should support custom message buffer length.

Issue #2967, opened 2021-07-21
get secret from metaManager failed, err: timed out waiting for the condition


sdghchj:
**What happened**:
When edgecore runs into the function getSecretsFromMetaManager, it reports error: "get secret from metaManager failed, err: timed out waiting for the condition"

The same issue: #1909 

**What you expected to happen**:
get secrets successfully.

**How to reproduce it (as minimally and precisely as possible)**:
1. Install the edgecore whith a nodeName on an edge node; 
2. Apply a deployment, which includes imagePullSecrets to pull images from a private project in a harbor, to that edge node;
3. Uninstall the edgecore from that edge node; 
4. Install again the edgecore whith the same nodeName on that edge node;
5. Apply a deployment with the same imagePullSecrets  to that edge node, at this step, it fails and report the error.

  </details>
  - edgecore version (e.g. `edgecore --version`): all the versions
  </details>

**Solution**:
I have viewed the source code, and found edgecore first query secrets from local db, if not found, then send the query request to the cloudcore, but the machanism of ObjectSync CRD stopped cloudcore to response the secret again, that is, cloudcore only sync once one same resource, including secret, to one same node.
I think, if the operation of a message is response which indicates edgecore doesn't have a resource in local and request it from cloud, cloudcore has to sync it down to edge.

PR #3014, opened 2021-08-09
sync the response message unconditionally

sdghchj:
if the operation is response, force to sync the resource message, since the edgecore requests it

**What type of PR is this?**
/kind bug

In case that the edgecore with the same node name is re-installed or moved to another edge node, it can't find imagePullSecrets or configMaps in local metamanager and also failed to get response from cloudcore.
If edgecore request a resouce from cloudcore, which means the resource doesn't exist in local metamanager, cloudcore has to sync the response resource down to edge.

#2967

NONE

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -100,9 +100,10 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
 		return
 	}
 
-	item, exist, _ := nodeStore.GetByKey(messageKey)
-
-	if !isDeleteMessage(msg) {
+	//if the operation is delete, force to sync the resource message
+	//if the operation is response, force to sync the resource message, since the edgecore requests it
+	if !isDeleteMessage(msg) && msg.GetOperation() != beehiveModel.ResponseOperation {
+		item, exist, _ := nodeStore.GetByKey(messageKey)
 		// If the message doesn't exist in the store, then compare it with
 		// the version stored in the database
 		if !exist {


fisherxu (commented):
@sdghchj Thanks for the fixing, and could you please rebase to the master branch to fix the failed e2e ci?

sdghchj:
> @sdghchj Thanks for the fixing, and could you please rebase to the master branch to fix the failed e2e ci?

@fisherxu   Rebased.

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -100,9 +100,10 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
 		return
 	}
 
-	item, exist, _ := nodeStore.GetByKey(messageKey)
-
-	if !isDeleteMessage(msg) {
+	//if the operation is delete, force to sync the resource message
+	//if the operation is response, force to sync the resource message, since the edgecore requests it
+	if !isDeleteMessage(msg) && msg.GetOperation() != beehiveModel.ResponseOperation {
```
fisherxu, line 105: Thanks  for the fixing!

Here the `if !isDeleteMessage(msg)` means if the message is not delete message, we will do some check like remove the dumplicate message. And if the message is delete message, we will send it directly.

And the Response Message we will also send to edge at here: https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/channelq/channelq.go#L168

So this is not the root cause of the issue.

sdghchj: Yes, if `exist` is true which means the duplicate message in nodeStore, as you mentioned,  some checks will be done like remove the dumplicate message. But if `exist` is false, I see `objectSync` may prevent the message to be send to edge: https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/channelq/channelq.go#L116.

And as a reponse to query a secret,  the field `content` stores the replied secret object, not string `OK`, so the boolean value of the experession at https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/channelq/channelq.go#L170 will be false which means the message is not list resource.

At last, I have tested my PR,  and it did work.

fisherxu: Yes, I just double check the issue, it's really bug in this scenario, we can have more discussions.

sdghchj: What to do next ?  @fisherxu

fisherxu: If the edgenode is crashed or reinstalled(data in db lost), the pod will also can't recovered, right? I think we need a design or full solution for it?

sdghchj:
> If the edgenode is crashed or reinstalled(data in db lost), the pod will also can't recovered, right? I think we need a design or full solution for it?
@fisherxu  
Actually, If one uninstalled edgecore, he had to be aware of  the effect made by his behavior.
To me, I can accept the fact that I have to re-deploy all the pods manually, but I am disappointed to see the  pods failed to run because of related resources such as secrets、configMaps.

Of course, I have a suggestion that edgecore also query the pods from cloud when starting if no pods found in local db ,and cloud reponse all the pods down.

fisherxu:
> Of course, I have a suggestion that edgecore also query the pods from cloud when starting if no pods found in local db ,and cloud reponse all the pods down.

This patch can fix restarting a new node with same name, and send cm\secret to edge, we can get it merged. :) And for edge recovering from edgenode crashed, we can have further design.

sdghchj:
> > Of course, I have a suggestion that edgecore also query the pods from cloud when starting if no pods found in local db ,and cloud reponse all the pods down.
> 
> This patch can fix restarting a new node with same name, and send cm\secret to edge, we can get it merged. :) And for edge recovering from edgenode crashed, we can have further design.

OK

fisherxu (commented):
/approve

PR #3021, opened 2021-08-11
support k8s v1.21.X

### LINKED ISSUE #3002: KubeEdge Support K8s v1.21

Now when we use version v1.21 of K8s master, the application schduled to the edgenode can't run normally. Maybe need some adaption work in edgecontroller.

cc @gy95 
**Why is this needed**:
Support  K8s v1.21.

gy95:

**What type of PR is this?**
/kind feature

To support k8s v1.21.X

Fixes #3002

when kubeedge running in the k8s v1.21.X environment, pod cannot run successfully. And the edgecore logs error: `GetServiceAccountToken unsupported`, so we should implement the function `func (evh *edgedVolumeHost) GetServiceAccountTokenFunc()` https://github.com/kubeedge/kubeedge/blob/597c2b4bf33f59009390e3d6f6fed3b03226225d/edge/pkg/edged/volume_host.go#L171-L175

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -158,7 +158,7 @@ func isListResource(msg *beehiveModel.Message) bool {
 	if strings.Contains(msgResource, beehiveModel.ResourceTypePodlist) ||
 		strings.Contains(msgResource, "membership") ||
 		strings.Contains(msgResource, "twin/cloud_updated") ||
-		strings.Contains(msgResource, commonconst.ResourceTypeServiceAccount) {
+		strings.Contains(msgResource, beehiveModel.ResourceTypeServiceAccountToken) {
 		return true
 	}
 
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -157,7 +157,8 @@ func isListResource(msg *beehiveModel.Message) bool {
 	msgResource := msg.GetResource()
 	if strings.Contains(msgResource, beehiveModel.ResourceTypePodlist) ||
 		strings.Contains(msgResource, "membership") ||
-		strings.Contains(msgResource, "twin/cloud_updated") {
+		strings.Contains(msgResource, "twin/cloud_updated") ||
+		strings.Contains(msgResource, commonconst.ResourceTypeServiceAccount) {
 		return true
 	}
 
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -86,7 +86,8 @@ func requireRemoteQuery(resType string) bool {
 		resType == constants.ResourceTypePersistentVolume ||
 		resType == constants.ResourceTypePersistentVolumeClaim ||
 		resType == constants.ResourceTypeVolumeAttachment ||
-		resType == model.ResourceTypeNode
+		resType == model.ResourceTypeNode ||
+		resType == model.ResourceTypeServiceAccountToken
 }
 
 func isConnected() bool {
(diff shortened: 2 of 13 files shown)


fisherxu:
Bump the kind in localup script and github action from `0.9.0` to `0.11.0` to get the ci passed.

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -157,7 +157,8 @@ func isListResource(msg *beehiveModel.Message) bool {
 	msgResource := msg.GetResource()
 	if strings.Contains(msgResource, beehiveModel.ResourceTypePodlist) ||
 		strings.Contains(msgResource, "membership") ||
-		strings.Contains(msgResource, "twin/cloud_updated") {
+		strings.Contains(msgResource, "twin/cloud_updated") ||
+		strings.Contains(msgResource, commonconst.ResourceTypeServiceAccount) {
```
fisherxu, line 161: The `token request` is a k8s object, so it should not be the `List` resource here.

edge/pkg/metamanager/process.go
```diff
@@ -492,6 +515,10 @@ func (m *metaManager) process(message model.Message) {
 		constants.CSIOperationTypeControllerPublishVolume,
 		constants.CSIOperationTypeControllerUnpublishVolume:
 		m.processVolume(message)
+	case constants.OperationTypeGetServiceAccount:
```
fisherxu, line 518: Is it possible to reuse the `QueryOperation` here?

gy95: `QueryOperation`  involves the operation of the `database`, and it reduces codes readability if we directly modify in the `QueryOperation` related funtions , So IMO, it's better to just add one more operation to process `serviceaccount token` exclusively

fisherxu: > QueryOperation involves the operation of the database

The result should be stored in the database.

> it's better to just add one more operation to process serviceaccount token exclusively

It's better to reuse the lib func instead of adding the redundant code.

gy95: ok

cloud/pkg/cloudhub/channelq/channelq.go
gy95: but it seems to not be able to be distinguished from other model messages.

fisherxu: Just as a k8s object like pod, cm, etc.

fisherxu: Note: The tokenreview obj's resource version is empty, so it's reasonable to use the `List` here.

edge/pkg/metamanager/process.go
gy95: done

hack/lib/install.sh
```diff
@@ -32,7 +32,7 @@ function check_kind {
   command -v kind >/dev/null 2>&1
   if [[ $? -ne 0 ]]; then
     echo "installing kind ."
-    GO111MODULE="on" go get sigs.k8s.io/kind@v0.9.0
+    GO111MODULE="on" go get sigs.k8s.io/kind@v0.11.0
```
fisherxu, line 35:
```suggestion
    GO111MODULE="on" go get sigs.k8s.io/kind@v0.11.1
```

fisherxu (commented):
LGTM

fisherxu:
Please fix the dco :)

fisherxu (commented):
/lgtm
/approve

fisherxu:
Please cherry pick this pr to release-1.7 using the cherry-pick script.

gy95:
ok

PR #3024, opened 2021-08-12
use UpdateDeviceStatusWorkers as updateDeviceStatus routines

gy95:

**What type of PR is this?**
/kind bug

`config.Config.Buffer.UpdateDeviceStatus` represents channel size, default value is 1024, and it's too big, cannot use this value as `UpdateDeviceStatus` Goroutine numbers
So use the correct value `config.Config.Load.UpdateDeviceStatusWorkers` to open go routines

Fixes #

```release-note

```


fisherxu (commented):
/lgtm

PR #3025, opened 2021-08-12
Handling node/*/membership/detail

subpathdev:
**What type of PR is this?**
/kind bug

A lot of warnings are generated by the devicecontroller/controller/upstream.go where the resource type cannot be identified. Most of these messages are of the resource type node/*/membersip/detail. These resource type will not generate a warning any more.


fisherxu (commented):
Thanks for the fixing, and could you please squash some commits?

fisherxu (commented):
/lgtm
/approve

PR #3028, opened 2021-08-12
simplify the way of getting message []byte format

gy95:

**What type of PR is this?**
/kind cleanup

`model.Message` has provided the function `GetContentData` to get `[]byte` format of message body,
So we can use function instead of implementing function itself.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/client/configmap.go
+++ b/edge/pkg/metamanager/client/configmap.go
@@ -58,15 +58,9 @@ func (c *configMaps) Get(name string) (*api.ConfigMap, error) {
 		return nil, fmt.Errorf("get configmap from metaManager failed, err: %v", err)
 	}
 
-	var content []byte
-	switch msg.Content.(type) {
-	case []byte:
-		content = msg.GetContent().([]byte)
-	default:
-		content, err = json.Marshal(msg.Content)
-		if err != nil {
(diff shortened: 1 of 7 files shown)


edge/pkg/metamanager/client/configmap.go
```diff
@@ -58,15 +58,9 @@ func (c *configMaps) Get(name string) (*api.ConfigMap, error) {
...
-	case []byte:
-		content = msg.GetContent().([]byte)
-	default:
-		content, err = json.Marshal(msg.Content)
-		if err != nil {
-			return nil, fmt.Errorf("marshal message to configmap failed, err: %v", err)
-		}
+	content, err := msg.GetContentData()
+	if err != nil {
+		return nil, fmt.Errorf("marshal message to configmap failed, err: %v", err)
```
Iceber, line 63: For `GetContentData` that return an error containing `marshal message content failed`.
https://github.com/kubeedge/kubeedge/blob/c51143389cdd3061a804b73be21b951c446bb68a/staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go#L142-L152

I recommend avoiding including marshal-related information in an error

gy95: done, modify to `parse message to configmap failed`

fisherxu (commented):
/lgtm

PR #3044, opened 2021-08-19
Feature/edgeSite

blackdobb:
**What type of PR is this?**
/kind feature

This submit for add the validation of hostname due to the problem caused by long name. Meanwhile,I create a new function for the redundant code

Fixes #

```release-note

```

### DIFF
--- a/pkg/apis/componentconfig/edgesite/v1alpha1/default.go
+++ b/pkg/apis/componentconfig/edgesite/v1alpha1/default.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"os"
 	"path"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


pkg/apis/componentconfig/edgesite/v1alpha1/default.go
```diff
@@ -31,10 +32,7 @@ import (
 
 // NewDefaultEdgeSiteConfig returns a full EdgeSiteConfig object
 func NewDefaultEdgeSiteConfig() *EdgeSiteConfig {
-	hostnameOverride, err := os.Hostname()
-	if err != nil {
-		hostnameOverride = constants.DefaultHostnameOverride
-	}
```
fisherxu, line 37: There are many places that get the `hostnameOverride`, could we move `func getHostname() string {` to a common place, then import it in all places? :) And please squash the commits.

blackdobb: well done~

PR #3049, opened 2021-08-23
Fixed modbus config parameters null value invalid

TianTianBigWang:

**What type of PR is this?**

Fixes #

```release-note

```


fisherxu (commented):
Thanks forthe fixing, but where will them be set to empty?

TianTianBigWang:
> Thanks forthe fixing, but where will them be set to empty?

Generally, the parameter does not need to be set to empty, just to avoid the parameter as the default value when the zero value is passed, which leads to the problem that the required parameter is ignored.

sailorvii:
> problem

Just for discussion.  I have some questions:
1. There's no default value to be set. The value is gotten from the device instance CRD yaml file. 
2. Why the offset could not be 0?
3. How the pointer method works to prevent the value to be 0?

Thanks,
Wei

TianTianBigWang:
> > problem
> 
> Just for discussion. I have some questions:
> 
> 1. There's no default value to be set. The value is gotten from the device instance CRD yaml file.
> 2. Why the offset could not be 0?
> 3. How the pointer method works to prevent the value to be 0?
> 
> Thanks,
> Wei

In fact, the offset can be 0, but setting it to 0 will cause me to get the following error: 

`"Device.devices.kubeedge.io \"xxx\" is invalid: spec.propertyVisitors.modbus.offset: Required value"
` 

 so I set the offset as a pointer to allow my 0 value to be passed into the variable.

 I don't need to use pointers to avoid the value 0, I only need 0 to be recognized by the parameter.

sailorvii:
> but setting it to 0 will cause me to get the fo

Oh, sorry for my misunderstanding.

I tried with 0 value but did not hit the same error.
Could you paste your device instance yaml and the whole output error?

Regards,
Wei

TianTianBigWang:
> > but setting it to 0 will cause me to get the fo
> 
> Oh, sorry for my misunderstanding.
> 
> I tried with 0 value but did not hit the same error.
> Could you paste your device instance yaml and the whole output error?
> 
> Regards,
> Wei

kubectl apply the yaml is ok, but when using the client DeviceInterface Create interface will cause the above error

```
apiVersion: devices.kubeedge.io/v1alpha2
kind: DeviceModel
metadata:
  labels:
    app: temperture
  name: xxx
  namespace: default
spec:
  properties:
  - name: humidity
    type:
      string:
        accessMode: ReadWrite
  - name: temperature
    type:
      string:
        accessMode: ReadWrite
---
apiVersion: devices.kubeedge.io/v1alpha2
kind: Device
metadata:
  labels:
    app: temperture
  name: xxx
  namespace: default
spec:
  data: {}
  deviceModelRef:
    name: xxx
  nodeSelector:
    nodeSelectorTerms: null
  propertyVisitors:
  - collectCycle: 1
    modbus:
      isSwap: true
      limit: 1
      offset: 0 # which i want to set
      register: HoldingRegister
    propertyName: humidity
    reportCycle: 1
  - collectCycle: 1
    modbus:
      isSwap: true
      limit: 1
      offset: 1
      register: HoldingRegister
    propertyName: temperature
    reportCycle: 1
  protocol:
    common:
      com:
        baudRate: 9600
        dataBits: 8
        parity: none
        serialPort: /dev/ttyUSB0
        stopBits: 1
    modbus:
      slaveID: 1
status:
  twins:
  - desired:
      metadata:
        timestamp: "1629766468272"
        type: string
      value: ""
    propertyName: humidity
    reported:
      value: ""
  - desired:
      metadata:
        timestamp: "1629766468272"
        type: string
      value: ""
    propertyName: temperature
    reported:
      value: ""

```

zc2638:
It may be that when the interface is called, the `struct` is converted to JSON, and the field is lost due to the `omitempty`attribute.
@TianTianBigWang @sailorvii

TianTianBigWang:
There are two ways to solve this problem:

1. Use pointers for parameters
2. Remove the `omitempty` attribute of json

@sailorvii @zc2638

sailorvii:
> There are two ways to solve this problem:
> 
> 1. Use pointers for parameters
> 2. Remove the `omitempty` attribute of json
> 
> @sailorvii @zc2638

I'm not familiar with the json conversion, but it's unreasonable that the json couldn't analyze the integer 0. Is there any other issue with the input json of the client?

fisherxu:
> It may be that when the interface is called, the `struct` is converted to JSON, and the field is lost due to the `omitempty`attribute.
> @TianTianBigWang @sailorvii

I think this is right. And can `Offset`\`Limit` be optional? Or must be required.

TianTianBigWang:
These two parameters are required and can be `0`, but `0` is the default value of `int64` in go and will be filtered by the json`omitempty` attribute. 

In addition, these two parameters need to be manually filled in by the user and must not be parsed as default values by json.

So it is recommended to use pointer.
@fisherxu @sailorvii

fisherxu:
> These two parameters are required and can be `0`, but `0` is the default value of `int64` in go and will be filtered by the json`omitempty` attribute.
> 
> In addition, these two parameters need to be manually filled in by the user and must not be parsed as default values by json.
> 
> So it is recommended to use pointer.
> @fisherxu @sailorvii

Sounds good.
/approve

PR #3060, opened 2021-08-27
add message handler for edgehub

gy95:

**What type of PR is this?**
/kind feature

introduce message handler to process some specific messages.
When we want to process some particular messages from cloud, we could call `RegisterHandler` function to register message handlers, so when the compliant messages come, we could process it. And this method is a non-intrusive way to add more message processing logics, we don't neet to add more branch cases directly to the blow block `swich case` anymore, and this makes codes more readability and elegant.

https://github.com/kubeedge/kubeedge/blob/9abf3aa6ee4df0a0746acd8507e1c9b18af59eb6/edge/pkg/edgehub/process.go#L47-L59

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/edgehub/common/msghandler/handler.go
+++ b/edge/pkg/edgehub/common/msghandler/handler.go
@@ -1,0 +1,45 @@
+package msghandler
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/kubeedge/beehive/pkg/core/model"
+	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/clients"
+)
+
+// Handler handler different messages
+type Handler interface {
+	Filter(message *model.Message) bool
+	Process(message *model.Message, clientHub clients.Adapter) error
+}
+
+var Handlers []Handler
+var lock sync.RWMutex
+
+func init() {
+	Handlers = make([]Handler, 0)
+}
+
+// RegisterHandler register msg handler.
+func RegisterHandler(handler Handler) {
+	lock.Lock()
+	defer lock.Unlock()
+	Handlers = append(Handlers, handler)
+}
+
+// ProcessHandler return true if handler filtered
+func ProcessHandler(message model.Message, client clients.Adapter) error {
+	lock.RLock()
+	defer lock.RUnlock()
+	for _, handle := range Handlers {
+		if handle.Filter(&message) {
+			err := handle.Process(&message, client)
+			if err != nil {
+				return fmt.Errorf("failed to handle message, message group: %s, error: %+v", message.GetGroup(), err)
+			}
+			return nil
+		}
+	}
+	return fmt.Errorf("failed to handle message, no handler found for the message, message group: %s", message.GetGroup())
+}
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -12,6 +12,7 @@ import (
 	messagepkg "github.com/kubeedge/kubeedge/edge/pkg/common/message"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/modules"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/clients"
+	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/common/msghandler"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/config"
 )
 
@@ -37,11 +38,25 @@ func (eh *EdgeHub) initial() (err error) {
 	return nil
 }
 
-func (eh *EdgeHub) isSyncResponse(msgID string) bool {
+func isSyncResponse(msgID string) bool {
 	return msgID != ""
 }
 
-func (eh *EdgeHub) dispatch(message model.Message) error {
+func init() {
+	handler := &defaultHandler{}
+	msghandler.RegisterHandler(handler)
+}
+
+type defaultHandler struct {
+}
+
+func (*defaultHandler) Filter(message *model.Message) bool {
+	group := message.GetGroup()
+	return group == messagepkg.ResourceGroupName || group == messagepkg.TwinGroupName ||
+		group == messagepkg.FuncGroupName || group == messagepkg.UserGroupName
+}
+
+func (*defaultHandler) Process(message *model.Message, clientHub clients.Adapter) error {
 	group := message.GetGroup()
 	md := ""
 	switch group {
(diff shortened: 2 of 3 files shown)


kevin-wangzefeng:
This is a refectoring work, I would suggest include change on existing code.
Otherwise it's no-op.

gy95:
This message handler is better for processing messages which are only from `cloudcore` and not from inner modules of `edgecore`.

using this handler for devicetwin is not suitable, because meesages that need to be processed by `devicetwin` module, are not only from cloudcore, but also from eventbus etc. If devicetwin also register handlers, will make codes more complex

gy95:
/kindly @fisherxu @kevin-wangzefeng ,PTAL, is this modification ok or not? if ok, i'll squalish the commits.

edge/pkg/edgehub/process.go
```diff
@@ -53,26 +68,34 @@ func (eh *EdgeHub) dispatch(message model.Message) error {
...
 
+func (eh *EdgeHub) dispatch(message model.Message) error {
+	// handler for msg.
+	isProcessed, err := msghandler.ProcessHandler(message, eh.chClient)
+	if isProcessed {
+		return err
+	}
+
+	klog.Errorf("msg not processed due to not-supported, group: %v", message.GetGroup())
+	return fmt.Errorf("msg not processed due to not-supported, group: %v", message.GetGroup())
```
fisherxu, line 96: Will always return error?

edge/pkg/edgehub/common/msghandler/handler.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, handle := range Handlers {
+		if handle.Filter(&message) {
+			err := handle.Process(&message, client)
+			if err != nil {
+				return true, fmt.Errorf("failed to handle message, error: %+v", err)
+			}
+			return true, nil
+		}
+	}
+	return false, nil
```
fisherxu, line 44:
```suggestion
	return true, nil
```
?

edge/pkg/edgehub/common/msghandler/handler.go
```diff
@@ -0,0 +1,45 @@
...
+
+// ProcessHandler return true if handler filtered
+func ProcessHandler(message model.Message, client clients.Adapter) (bool, error) {
+	lock.RLock()
+	defer lock.RUnlock()
+	for _, handle := range Handlers {
+		if handle.Filter(&message) {
+			err := handle.Process(&message, client)
+			if err != nil {
+				return true, fmt.Errorf("failed to handle message, error: %+v", err)
```
fisherxu, line 39:
```suggestion
				return false, fmt.Errorf("failed to handle message, error: %+v", err)
```
?

edge/pkg/edgehub/process.go
```diff
@@ -53,26 +68,34 @@ func (eh *EdgeHub) dispatch(message model.Message) error {
...
+		beehiveContext.SendToGroup(md, *message)
 	}
 	return nil
 }
 
+func (eh *EdgeHub) dispatch(message model.Message) error {
+	// handler for msg.
+	isProcessed, err := msghandler.ProcessHandler(message, eh.chClient)
+	if isProcessed {
+		return err
```
fisherxu, line 92:
```suggestion
		return nil
```
?

edge/pkg/edgehub/common/msghandler/handler.go
```diff
@@ -0,0 +1,45 @@
...
+
+// RegisterHandler register msg handler.
+func RegisterHandler(handler Handler) {
+	lock.Lock()
+	defer lock.Unlock()
+	Handlers = append(Handlers, handler)
+}
+
+// ProcessHandler return true if handler filtered
+func ProcessHandler(message model.Message, client clients.Adapter) (bool, error) {
```
fisherxu, line 32: Do we need the bool here?

edge/pkg/edgehub/common/msghandler/handler.go
```diff
@@ -0,0 +1,45 @@
...
+
+func init() {
+	Handlers = make([]Handler, 0)
+}
+
+// RegisterHandler register msg handler.
+func RegisterHandler(handler Handler) {
+	lock.Lock()
+	defer lock.Unlock()
+	Handlers = append(Handlers, handler)
```
fisherxu, line 28: Will we have many handler in future? @wackxu

edge/pkg/edgehub/process.go
```diff
@@ -37,11 +38,25 @@ func (eh *EdgeHub) initial() (err error) {
...
+	msghandler.RegisterHandler(handler)
+}
+
+type defaultHandler struct {
+}
+
+func (*defaultHandler) Filter(message *model.Message) bool {
+	group := message.GetGroup()
+	return group == messagepkg.ResourceGroupName || group == messagepkg.TwinGroupName ||
+		group == messagepkg.FuncGroupName || group == messagepkg.UserGroupName
```
fisherxu, line 56: Is there some message that filter failed? @wackxu

edge/pkg/edgehub/process.go
gy95: if been processed, we need to return nil if no-error, return error if error happens, so just return `err`, and it contains nil or error circumanstances.

edge/pkg/edgehub/common/msghandler/handler.go
gy95: this bool return value represents for whether message been processed by processed processors

edge/pkg/edgehub/common/msghandler/handler.go
gy95: no, it means not matched processors and don't be processed.

edge/pkg/edgehub/process.go
gy95: all supported messages will be processed by processors. So if the codes run here, it means that it's not supported, so just return error

edge/pkg/edgehub/process.go
fisherxu: If err != nil,  dispatch msg should be failed.

edge/pkg/edgehub/common/msghandler/handler.go
```diff
@@ -0,0 +1,45 @@
...
+		if handle.Filter(&message) {
+			err := handle.Process(&message, client)
+			if err != nil {
+				return false, fmt.Errorf("failed to handle message, error: %+v", err)
+			}
+			return true, nil
+		}
+	}
+	return true, nil
+}
```
fisherxu, line 45:
```suggestion
func ProcessHandler(message model.Message, client clients.Adapter) error {
	lock.RLock()
	defer lock.RUnlock()
	for _, handle := range Handlers {
		if handle.Filter(&message) {
			err := handle.Process(&message, client)
			if err != nil {
				return fmt.Errorf("failed to handle message, message group: %s, error: %+v", message.GetGroup(), err)
			}
			return nil
		}
	}
	return fmt.Errorf("failed to handle message, no handler found for the message, message group: %s", message.GetGroup())
}
```

edge/pkg/edgehub/process.go
```diff
@@ -53,26 +68,34 @@ func (eh *EdgeHub) dispatch(message model.Message) error {
...
+func (eh *EdgeHub) dispatch(message model.Message) error {
+	// handler for msg.
+	isProcessed, err := msghandler.ProcessHandler(message, eh.chClient)
+	if isProcessed {
+		return nil
+	}
+
+	klog.Errorf("msg not processed due to not-supported, group: %v", message.GetGroup())
+	return fmt.Errorf("msg not processed due to not-supported, group: %v", message.GetGroup())
+}
```
fisherxu, line 97:
```suggestion
func (eh *EdgeHub) dispatch(message model.Message) error {
	// handler for msg.
	err := msghandler.ProcessHandler(message, eh.chClient)
	if err!=nil {
		return err
	}
        
        return nil
}
```

fisherxu:
Please squash the commits, thanks.

gy95:
> Please squash the commits, thanks.

done， PTAL

fisherxu (commented):
/lgtm
/approve

PR #3069, opened 2021-08-30
fixed format content in beehive/pkg/core/context/context_channel.go

orangegzx:
**What type of PR is this?**
/kind bug

>The initial letter of the content of the formatted string is not capitalized uniformly

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
@@ -171,7 +171,7 @@ func (ctx *ChannelContext) SendToGroup(moduleType string, message model.Message)
 		select {
 		case ch <- message:
 		default:
-			klog.Warningf("the module %s message channel is full, message: %+v", module, message)
+			klog.Warningf("The module %s message channel is full, message: %+v", module, message)
 			ch <- message
 		}
 	}
@@ -262,11 +262,10 @@ func (ctx *ChannelContext) SendToGroupSync(moduleType string, message model.Mess
 		case <-sendTimer.C:
 			cleanup()
 			if timeoutCounter != 0 {
-				errInfo := fmt.Sprintf("timeout to send message, several %d timeout when send", timeoutCounter)
-				return fmt.Errorf(errInfo)
+				return  fmt.Errorf("timeout to send message, several %d timeout when send", timeoutCounter)
 			}
 			klog.Error("Timeout to sendToGroupsync message")
-			return fmt.Errorf("Timeout to send message")
+			return fmt.Errorf("timeout to send message")
 		}
 	}
 


gy95:
the error fmt commonly starts with a lowercase letter, more readable when print errors to logs. 😄 🥲

staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
```diff
@@ -80,7 +80,7 @@ func (ctx *ChannelContext) Receive(module string) (model.Message, error) {
 	}
 
 	klog.Warningf("Failed to get channel for module:%s when receive message", module)
-	return model.Message{}, fmt.Errorf("failed to get channel for module(%s)", module)
+	return model.Message{}, fmt.Errorf("Failed to get channel for module(%s)", module)
```
zc2638, line 83: The first letter of the `log` output needs to be capitalized, but the `error` not recommended

fisherxu: +1 :)

gy95:
when use `klog.Errorf` or `klog.Infof` etc. to print logs, we can use capitalized string.
But when we use `fmt.Errorf` or error.New to format error info, msg should start with lowercase

staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
```diff
@@ -262,11 +262,11 @@ func (ctx *ChannelContext) SendToGroupSync(moduleType string, message model.Mess
 		case <-sendTimer.C:
 			cleanup()
 			if timeoutCounter != 0 {
-				errInfo := fmt.Sprintf("timeout to send message, several %d timeout when send", timeoutCounter)
+				errInfo := fmt.Sprintf("Timeout to send message, several %d timeout when send", timeoutCounter)
 				return fmt.Errorf(errInfo)
```
zc2638, line 266:
```suggestion
				return fmt.Errorf(Timeout to send message, several %d timeout when send", timeoutCounter)
```
It is recommended to combine the two methods

staging/src/github.com/kubeedge/beehive/pkg/core/context/context_channel.go
```diff
@@ -262,11 +262,10 @@ func (ctx *ChannelContext) SendToGroupSync(moduleType string, message model.Mess
 		case <-sendTimer.C:
 			cleanup()
 			if timeoutCounter != 0 {
-				errInfo := fmt.Sprintf("timeout to send message, several %d timeout when send", timeoutCounter)
-				return fmt.Errorf(errInfo)
+				return  fmt.Errorf("Timeout to send message, several %d timeout when send", timeoutCounter)
```
gy95, line 265: need to fixed from `Timeout` to `timeout`

fisherxu (commented):
/approve
Seems that there's many redundant logs, as first klog print the error, then also return the error, then print it  again.

PR #3071, opened 2021-08-30
support correct node limit

zc2638:
**What type of PR is this?**

/kind feature

> support correct node limit

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -7,7 +7,6 @@ import (
 	"regexp"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -110,8 +109,9 @@ func (mh *MessageHandle) HandleServer(container *mux.MessageContainer, writer mu
 	nodeID := container.Header.Get("node_id")
 	projectID := container.Header.Get("project_id")
 
-	// TODO(iceber): check node limits at registration
-	if atomic.LoadInt32(&mh.NodeNumber) >= mh.NodeLimit {
+	// check node limit
+	keys := informers.GetInformersManager().EdgeNode().GetStore().ListKeys()
```
fisherxu, line 113: Now the nodelimit is only for one cloudcore instance, and for active-active mode, informers will list all the edge nodes. But if we set the limit to whole cluster, and cloudcore maybe can't serve that :) 

But I think it's great to count the nodes from cache store, also need more discussion.

zc2638: This is the node limit of the whole cluster. I will add node connection limit in another pr.
@fisherxu

fisherxu: > This is the node limit of the whole cluster. I will add node connection limit in another pr.
> @fisherxu

We can add node connection limit per cloudcore first, for whole cluster, there should be no specific value yet :)

zc2638: why?

fisherxu: For whole cluster, there can be no limit in theory, it can be 5k, 8k, 10k etc. And also lack test data now.

zc2638: So I have to limit it myself in `admission`?
Maybe `kubernetes` doesn't have this limit now.
But `kubeedge` should need.

PR #3077, opened 2021-08-31
refactor edge metamanager: introduce processors to distinguish messages

### LINKED ISSUE #3070: Refactor edge metamanager
Refactor edgecore metamanager
introduce a structure that contains resourceType and operation like below, introduce a global map value `processors` to contain different processors,

So we can distinguish or dispatch messages by resourceType and Operation to different processors to process messages.  and we can process message like function `func Process(message model.Message)` just as belows.

So when we add modules to process message, we just need to add a new processor to the global map `processors`, but not need to intrusively modify in the function `func (m *metaManager) process(message model.Message)` 
```
var (
	processors = make(map[queryKey]Processor)
)

type queryKey struct {
	resType   string
	operation string
}

type Processor interface {
	Process(message model.Message)
}

func Process(message model.Message) {

	op := message.GetOperation()
	_, resType, _ := parseResource(message.GetResource())
	p := processors[queryKey{resType: resType, operation: op}]
	if resType != "" && p == nil {
		p = processors[queryKey{operation: op}]
	}
	if p != nil {
		p.Process(message)
	} else {
		klog.Errorf("No processor to process message,op:%v,resType:%v", op, resType)
	}
}
```

https://github.com/kubeedge/kubeedge/blob/9abf3aa6ee4df0a0746acd8507e1c9b18af59eb6/edge/pkg/metamanager/process.go#L431-L460
**Why is this needed**:

gy95:

**What type of PR is this?**
/kind feature

introduce the processors, and use operationType and resourceType of message to distinguish, so if we want to add more  function modules to add functions of processing more different msgs, we only need to add a new processor, and don't need to modify the below codes intrusively.
https://github.com/kubeedge/kubeedge/blob/c2f3b2af1d59fdfd39ca6944d292eda9b2d1aa7f/edge/pkg/metamanager/process.go#L431-L460

Fixes #3070

```release-note

```


edge/pkg/common/modules/names.go
```diff
@@ -11,4 +11,6 @@ const (
 	EdgeStreamModuleName = "edgestream"
 	// DeviceTwinModuleName name
 	DeviceTwinModuleName = "twin"
+	// MetaManagerModuleName name
+	MetaManagerModuleName = "metaManager"
```
zc2638, line 15: `EdgeStreamModuleName = "edgestream"`
Uniform case is recommended.

edge/pkg/metamanager/processor/common_processor.go
```diff
@@ -0,0 +1,279 @@
...
+
+type insertProcessor struct{}
+
+type updateProcessor struct{}
+
+type deleteProcessor struct{}
+
+type queryProcessor struct{}
+
+type responseProcessor struct{}
```
zc2638, line 42: It is recommended that the method follow the `struct` to increase readability

gy95: done

edge/pkg/common/modules/names.go
gy95: we can open another PR to fix it, need to modify all "metaManager" related codes, and move this string literal to the common modules just as processor also need to use this module name. If not, compile will be failed due to `import cycle not allowed `

zc2638: okey.

gy95:
ping @fisherxu PTAL, review please. 😃

PR #3080, opened 2021-08-31
goroutine util

qiuax:
What type of PR is this?
 Add goroutine util:

Which issue(s) this PR fixes:

/assign @kevin-wangzefeng

### DIFF
--- a/common/goroutine/goroutine.go
+++ b/common/goroutine/goroutine.go
@@ -12,13 +12,13 @@ func Goroutine(function interface{}, args ...interface{}) {
 
 	f := reflect.ValueOf(function)
 	if f.Kind() != reflect.Func {
-		log.Println("function 参数必须是个 function")
+		log.Println("function parameter must be a function")
 		return
 	}
 
 	go func() {
 		defer func() {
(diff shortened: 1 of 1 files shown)


common/goroutine/goroutine.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func Goroutine(function interface{}, args ...interface{}) {
+	if function == nil {
+		return
+	}
+
+	f := reflect.ValueOf(function)
+	if f.Kind() != reflect.Func {
+		log.Println("function 参数必须是个 function")
```
daixiang0, line 15: please use English here

common/goroutine/goroutine.go
```diff
@@ -0,0 +1,35 @@
...
+		return
+	}
+
+	go func() {
+		defer func() {
+			if pnc := recover(); pnc != nil { // recover 必须在 defer 函数中直接调用才能拦截异常
```
daixiang0, line 21: same here

common/goroutine/goroutine.go
```diff
@@ -0,0 +1,35 @@
+package goroutine
```
daixiang0, line 1: I do not find where to call it, do you miss it?

PR #3084, opened 2021-08-31
unified namespace reference

zc2638:

**What type of PR is this?**

/kind feature

> unified namespace reference

Fixes #

```release-note

```

### DIFF
--- a/common/constants/default.go
+++ b/common/constants/default.go
@@ -14,8 +14,8 @@ const (
 	// NodeName is for the clearer log of cloudcore.
 	NodeName = "NodeName"
 
-	KubeEdge          = "kubeedge"
-	KubeEdgeNameSpace = "kubeedge"
+	SystemName      = "kubeedge"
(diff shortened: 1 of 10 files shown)


common/constants/default.go
```diff
@@ -14,8 +14,8 @@ const (
 	// NodeName is for the clearer log of cloudcore.
 	NodeName = "NodeName"
 
-	KubeEdge          = "kubeedge"
-	KubeEdgeNameSpace = "kubeedge"
+	Name      = "kubeedge"
+	Namespace = Name
```
fisherxu, line 18: `Namespace` and `Name` don’t sound so clearly when developers import...?

zc2638: like `SystemNamespace` ?

zc2638: For example k8s, defined as `NamespaceDefault`

fisherxu (commented):
/lgtm
/approve

PR #3089, opened 2021-08-31
rpminstaller:add support for openEuler

CooperLi:
keadm will install epel-release when edge node want to join cluster,
openEuler does not provide epel-release, it's better to support keadm
join on openEuler ^.^

> openEuler is an innovative platform nurtured by community collaboration. It aims to build a unified and open OS that supports multiple processor architectures, and to advance the hardware/software application ecosystem.

Signed-off-by: CooperLi <a710905118@163.com>

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/rpminstaller.go
+++ b/keadm/cmd/keadm/app/cmd/util/rpminstaller.go
@@ -18,10 +18,15 @@ package util
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/blang/semver"
 
 	types "github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
+)
+
+const (
+	openEulerVendorName = "openEuler"
 )
 
 // RpmOS struct objects shall have information of the tools version to be installed
@@ -51,13 +56,24 @@ func (r *RpmOS) InstallMQTT() error {
 		return nil
 	}
 
-	// install MQTT
-	for _, command := range []string{
+	commands := []string{
 		"yum -y install epel-release",
 		"yum -y install mosquitto",
 		"systemctl start mosquitto",
 		"systemctl enable mosquitto",
-	} {
+	}
+
+	vendorName, err := getOSVendorName()
+	if err != nil {
+		fmt.Printf("Get OS vendor name failed: %v\n", err)
+	}
+	// epel-release package does not included in openEuler
+	if vendorName == openEulerVendorName {
+		commands = commands[1:]
+	}
+
+	// install MQTT
+	for _, command := range commands {
 		cmd := NewCommand(command)
 		if err := cmd.Exec(); err != nil {
 			return err
@@ -115,4 +131,14 @@ func (r *RpmOS) IsKubeEdgeProcessRunning(proc string) (bool, error) {
 // IsProcessRunning checks if the given process is running or not
 func (r *RpmOS) IsProcessRunning(proc string) (bool, error) {
 	return isKubeEdgeProcessRunning(proc)
+}
+
+func getOSVendorName() (string, error) {
+	cmd := NewCommand("cat /etc/os-release | grep -E \"^NAME=\" | awk -F'=' '{print $2}'")
+	if err := cmd.Exec(); err != nil {
+		return "", err
+	}
+	vendor := strings.Trim(cmd.GetStdOut(), "\"")
+
+	return vendor, nil
 }


keadm/cmd/keadm/app/cmd/util/rpminstaller.go
```diff
@@ -51,13 +56,23 @@ func (r *RpmOS) InstallMQTT() error {
...
 		"systemctl enable mosquitto",
-	} {
+	}
+
+	if vendor, err := getOSVendor(); err != nil {
+		return err
+	} else {
+		// epel-release package does not included in openEuler OS
+		if vendor != openEulerOS {
+			commands = append(commands[:1], "yum -y install epel-release")
```
gy95, line 70: It seems that the commands are executed in sequence , so it should be like this `commands = append([]string{"yum -y install epel-release"}, commands...)`

CooperLi:
@gy95 PTAL~

gy95:
Another question, kubeedge components: `cloudcore` and `edgecore` installed method  should also be installed. 
For example, `cloudcore` install should also be influenced by os, now we support three package managers of `os`: apt, yum, pacman,
you can run this command :  `command -v apt || command -v yum || command -v pacman` to find the type of package manager. or refer source codes logic: https://github.com/kubeedge/kubeedge/blob/7e599ed3c01a1ab0d4840ec68894f8c18c124adc/keadm/cmd/keadm/app/cmd/util/common.go#L131-L143

install of cloudcore
https://github.com/kubeedge/kubeedge/blob/7e599ed3c01a1ab0d4840ec68894f8c18c124adc/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go#L32

And i think you can use `keadm` to install kubeedge on `openEuler` to check whether kubeedge works on `openEuler`, I cannot do this test due to not owning a openEuler machine. 😄

CooperLi (commented):
> Another question, kubeedge components: `cloudcore` and `edgecore` installed method should also be installed.
> For example, `cloudcore` install should also be influenced by os, now we support three package managers of `os`: apt, yum, pacman,
> you can run this command : `command -v apt || command -v yum || command -v pacman` to find the type of package manager. or refer source codes logic:
> 
> https://github.com/kubeedge/kubeedge/blob/7e599ed3c01a1ab0d4840ec68894f8c18c124adc/keadm/cmd/keadm/app/cmd/util/common.go#L131-L143
> 
> install of cloudcore
> https://github.com/kubeedge/kubeedge/blob/7e599ed3c01a1ab0d4840ec68894f8c18c124adc/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go#L32
> 
> And i think you can use `keadm` to install kubeedge on `openEuler` to check whether kubeedge works on `openEuler`, I cannot do this test due to not owning a openEuler machine. 😄

actually, the whole test have been done on the openEuler machine, and all works good. OpenEuler also use `yum` as package manager, so once we fix the `epel-release` problem, everything will be good.

```bash
[root@cloud kubeedge]# kubectl get node -A -owide
NAME             STATUS     ROLES                  AGE     VERSION                   INTERNAL-IP    EXTERNAL-IP   OS-IMAGE          KERNEL-VERSION               CONTAINER-RUNTIME
cloud.kubeedge   Ready      control-plane,master   3d17h   v1.20.2                   9.63.252.224   <none>        openEuler 21.09   5.10.0-5.7.0.21.oe1.x86_64   iSulad://2.0.8
edge.kubeedge    Ready   agent,edge             2d22h   v1.19.3-kubeedge-v1.7.2   9.63.252.227   <none>        openEuler 21.09   5.10.0-5.7.0.21.oe1.x86_64   remote://2.0.8
```

keadm/cmd/keadm/app/cmd/util/rpminstaller.go
```diff
@@ -51,13 +56,23 @@ func (r *RpmOS) InstallMQTT() error {
...
+	}
+
+	if vendorName, err := getOSVendorName(); err != nil {
+		return err
+	} else {
+		// epel-release package does not included in openEuler
+		if vendorName != openEulerVendorName {
+			commands = append([]string{"yum -y install epel-release"}, commands...)
+		}
+	}
```
zc2638, line 72:
```suggestion
        commands := []string{
		"yum -y install epel-release",
		"yum -y install mosquitto",
		"systemctl start mosquitto",
		"systemctl enable mosquitto",
	}
	
	vendorName, err := getOSVendorName()
	if err != nil {
		return err
	}
	// epel-release package does not included in openEuler
      if vendorName == openEulerVendorName {
	      commands = commands[1:]
      }
```
I think this coding method will be better.What do you think?

CooperLi: OK~

CooperLi:
@zc2638 @gy95 @alrs PTAL :grinning:

keadm/cmd/keadm/app/cmd/util/rpminstaller.go
```diff
@@ -51,13 +56,24 @@ func (r *RpmOS) InstallMQTT() error {
...
 		"yum -y install mosquitto",
 		"systemctl start mosquitto",
 		"systemctl enable mosquitto",
-	} {
+	}
+
+	vendorName, err := getOSVendorName()
+	if err != nil {
+		return err
+	}
```
zc2638, line 69: Is there no need to throw an error here?
If you can't get it, execute it in the established way.
However, log output is necessary.

gy95: use fmt.Println to print error

zc2638: what's your opinion? @fisherxu

CooperLi: already print error here

CooperLi:
@zc2638 @gy95 @alrs PTAL 😀

PR #3095, opened 2021-09-01
goroutine util

qiuax:
What type of PR is this?
 Add goroutine util:

Which issue(s) this PR fixes:

/assign @kevin-wangzefeng

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/core.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/core.go
@@ -1,6 +1,7 @@
 package core
 
 import (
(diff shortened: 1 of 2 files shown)


staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -20,7 +21,8 @@ func StartModules() {
 		beehiveContext.AddModule(name)
 		// Assemble typeChannels for sendToGroup
 		beehiveContext.AddModuleGroup(name, module.Group())
-		go module.Start()
```
fisherxu, line 23: Thanks for the fixing, and if the module panic, the whole process should exit, so it's not reasonable to recover it here.

PR #3098, opened 2021-09-01
Replaced 'kubeedge/pause' with 'k8s.gcr.io/pause' multi arch image

siredmar:
**What type of PR is this?**
/kind feature

This PR uses official `pause` container image which are already multiarch images. Therefore the need to explicitly define the arch for the pause image wonn't be needed anymore. Examples `kubeedge/pause`, `kubeedge/pause-arm64` and `kubeedge/pause-arm`.

```release-note
NONE
```


subpathdev:
can you resolve the conflict please?

siredmar:
> can you resolve the conflict please?

done

subpathdev (approved):

fisherxu:
/hold 
Let's have more discussion :)

siredmar:
> /hold
> Let's have more discussion :)

You told me in the kubeedge slack channel, that you've experienced difficulties using k8s.gcr.io from all regions. 
In this case i propose some sort of mirror to the original pause image, as this is already multi arch for amd64, arm and arm64. The mirrored image can be put to `kubeedge/pause`. Therefore there is no need to manually maintain the versions (`k8s.gcr.io/pause` is already 3.5) and the edgecore.yaml files can be the same for all archs.

gy95:
you can rebase to solve the conflict, and now this PR contains too much other commit source codes. 
we could simplify this PR that only contains `kubeedge/pause` or `k8s.gcr.io/pause` image related modifications. 😄

chendave:
pls send your second commit in other PR, @fisherxu  do you think we need to update `kubeedge/pause`  to a multi-arch image as well?

fisherxu:
> > /hold
> > Let's have more discussion :)
> 
> You told me in the kubeedge slack channel, that you've experienced difficulties using k8s.gcr.io from all regions.
> In this case i propose some sort of mirror to the original pause image, as this is already multi arch for amd64, arm and arm64. The mirrored image can be put to `kubeedge/pause`. Therefore there is no need to manually maintain the versions (`k8s.gcr.io/pause` is already 3.5) and the edgecore.yaml files can be the same for all archs.

@siredmar  Make sense, let's sync the images to `kubeedge/pause` first, then merge this patch :)

siredmar:
@chendave done. See #3114 
@fisherxu put back in kubeedge/pause

Lets continue the discussion in #3114

PR #3099, opened 2021-09-01
fix help message when white noise issue occurs

gy95:

**What type of PR is this?**
/kind cleanup

When ci lint or `make lint` command checks white noise, the suggestion message should suggest using `git status` or `git diff` to find the reason, but not using `make lint` command. Or, developers or who first encounters whitenoise,  don't know how to fix this problem.

Fixes #

```release-note

```

### DIFF
--- a/hack/lib/lint.sh
+++ b/hack/lib/lint.sh
@@ -61,7 +61,7 @@ kubeedge::lint::check() {
     fi
 
     [[ $(git diff --name-only) ]] && {
-      echo "Some files have white noise issue, please run \`make lint\` to slove this issue"
+      echo "Some files have white noise issue, please run \`make lint\` to solve and \`git status\` to find and fix this issue"
       return 1
     }
     set -o pipefail


hack/lib/lint.sh
```diff
@@ -61,7 +61,7 @@ kubeedge::lint::check() {
     fi
 
     [[ $(git diff --name-only) ]] && {
-      echo "Some files have white noise issue, please run \`make lint\` to slove this issue"
+      echo "Some files have white noise issue, please run \`git status\` to find whitenoise and solve this issue"
```
fisherxu, line 64:
```suggestion
      echo "Some files have white noise issue, please run \`make lint\` to solve and \`git status\` to find and fix this issue"
```

fisherxu: If users find this issue in the ci, they also need to run make lint in local env.

gy95: ok

fisherxu (commented):
/approve

PR #3117, opened 2021-09-03
fix modbus slaveID cannot be 0

TianTianBigWang:
**What type of PR is this?**
/kind bug

Like #3049 , if I want to set slaveId to 0, I will get an error: 
`"Device.devices.kubeedge.io \"xxx\" is invalid: spec.protocol.modbus.slaveID: Required value"`

Fixes #

```release-note

```


daixiang0:
good catch

/lgtm

fisherxu (commented):
/approve

PR #3119, opened 2021-09-03
meta server addr configurable

TianTianBigWang:

**What type of PR is this?**
/kind feature

Fixes #

```release-note

```

### DIFF
--- a/pkg/apis/componentconfig/edgecore/v1alpha1/default.go
+++ b/pkg/apis/componentconfig/edgecore/v1alpha1/default.go
@@ -135,6 +135,7 @@ func NewDefaultEdgeCoreConfig() *EdgeCoreConfig {
 				MetaServer: &MetaServer{
 					Enable: false,
 					Debug:  false,
+					Server: constants.DefaultMetaServerAddr,
 				},
 			},
 			ServiceBus: &ServiceBus{
--- a/pkg/apis/componentconfig/edgecore/v1alpha1/types.go
+++ b/pkg/apis/componentconfig/edgecore/v1alpha1/types.go
@@ -383,8 +383,9 @@ type MetaManager struct {
 }
 
 type MetaServer struct {
-	Enable bool `json:"enable"`
-	Debug  bool `json:"debug"`
+	Enable bool   `json:"enable"`
+	Debug  bool   `json:"debug"`
+	Server string `json:"server"`
 }
 
 // ServiceBus indicates the ServiceBus module config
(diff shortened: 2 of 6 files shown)


gy95:
also need to do some work that we can use `edgecore --defaultconfig > edgecore.yaml` cmd to generate this default config.

gy95:
oh, sorry, ignore the previous comment, variable `Httpaddr` is also used in the test file

pkg/apis/componentconfig/edgecore/v1alpha1/default.go
```diff
@@ -135,6 +135,7 @@ func NewDefaultEdgeCoreConfig() *EdgeCoreConfig {
 				MetaServer: &MetaServer{
 					Enable: false,
 					Debug:  false,
+					Server: "127.0.0.1:10550",
```
zc2638, line 138: You can set it as a constant and reference it elsewhere.

tests/integration/metaserver/access_test.go
```diff
@@ -10,12 +10,12 @@ import (
...
 	"k8s.io/klog/v2"
 
-	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/metaserver"
+	"github.com/kubeedge/kubeedge/common/constants"
 )
 
 func TestGet(t *testing.T) {
 	kubeclient, err := kubernetes.NewForConfig(&rest.Config{
-		Host: "http://" + metaserver.Httpaddr,
+		Host: constants.DefaultMetaServerAddr,
```
zc2638, line 18: Need to add protocol?

TianTianBigWang: no

TianTianBigWang: may be yes

pkg/apis/componentconfig/edgecore/v1alpha1/types.go
```diff
@@ -383,8 +383,9 @@ type MetaManager struct {
 }
 
 type MetaServer struct {
-	Enable bool `json:"enable"`
-	Debug  bool `json:"debug"`
+	Enable bool   `json:"enable"`
+	Debug  bool   `json:"debug"`
+	Server string `json:"server"`
```
fisherxu, line 388: Do  we  need to spilit the server and port in config?

zc2638: No other place needs a port, it doesn’t have to be alone.

fisherxu (commented):
/approve

PR #3124, opened 2021-09-03
add iptables in dockerfile

subpathdev:
**What type of PR is this?**
/kind feature

adding the binary iptables in the cloudcore image


siredmar:
@subpathdev i created another PR with also iptables but also the priviliged flag for the cloudcore container. See #3129

PR #3126, opened 2021-09-03
single device only supports binding a single node

zc2638:
**What type of PR is this?**

/kind feature

```
In the actual scenario, no device needs to bind multiple gateways.
```

Fixes #

```release-note
crd resource need to update
```

### DIFF
--- a/build/crds/devices/devices_v1alpha2_device.yaml
+++ b/build/crds/devices/devices_v1alpha2_device.yaml
@@ -31,11 +31,10 @@ spec:
               description: 'Required: DeviceModelRef is reference to the device model
                 used as a template to create the device instance.'
               type: object
-            nodeSelector:
-              description: NodeSelector indicates the binding preferences between
-                devices and nodes. Refer to k8s.io/kubernetes/pkg/apis/core NodeSelector
-                for more details
-              type: object
+            nodeName:
+              description: NodeName indicates the binding preferences between
+                device and node.
+              type: string
             protocol:
               description: 'Required: The protocol configuration used to connect to
                 the device.'
@@ -392,7 +391,6 @@ spec:
               type: object
           required:
             - deviceModelRef
-            - nodeSelector
           type: object
         status:
           properties:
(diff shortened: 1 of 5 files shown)


cloud/pkg/apis/devices/v1alpha2/device_instance_types.go
```diff
@@ -38,10 +38,9 @@ type DeviceSpec struct {
 	// on edge node.
 	// +optional
 	Data DeviceData `json:"data,omitempty"`
-	// NodeSelector indicates the binding preferences between devices and nodes.
-	// Refer to k8s.io/kubernetes/pkg/apis/core NodeSelector for more details
+	// NodeName indicates the binding preferences between device and node.
 	// +optional
-	NodeSelector *v1.NodeSelector `json:"nodeSelector,omitempty"`
```
fisherxu, line 44: Now one device only bind to one node, but we need to leave the interface for binding to multi-node for future. We can have more discussion.

zc2638: Yes, we can discuss it later.

subpathdev:
Like @fisherxu  I think using a NodeSelector is better than setting a NodeName explicit. Can we discuss this in this PR?

build/crds/devices/devices_v1alpha2_device.yaml
```diff
@@ -392,7 +391,6 @@ spec:
               type: object
           required:
             - deviceModelRef
-            - nodeSelector
```
subpathdev, line 395: If we want to use NodeName instead, I would prefer if that is also required.

kevin-wangzefeng:
The nodeselector is a preference here, we don't have an explicit binding indicator so far, which is a problem.
Could you describe more about your use case and scenarios and we can discuss how to improve the API.

zc2638:
> The nodeselector is a preference here, we don't have an explicit binding indicator so far, which is a problem.
> Could you describe more about your use case and scenarios and we can discuss how to improve the API.

So far, the device associated node has only realized a single node, and its purpose is very fixed. It may not be changed for a long time in the future. It is better to use a more concise correlation method.

In terms of scenarios, most of what we see are scenarios from devices to single nodes. I'm not sure if there will be other multi-node scenarios, we can have more discussion here.

PR #3130, opened 2021-09-03
Added CustomInterfaceEnabled and CustomInterfaceName for edgecore

### LINKED ISSUE #3127: Readd deprecated `InterfaceName` option in edgecore.yaml
I'd like to have the ability to specify a specific network interface to have the control over the IP address that is used to register to tunnel server. 

**Why is this needed**:
The scenario is: Over different local networks multiple edge devices have the same IP address, e.g. 192.168.0.5. The iptables rules redirect from the node IP to the node cloudcore runs on. If multiple device share the same IP address within their private network, this concept breaks. 
Possible solution: if there is a CNI installation like e.g. flannel, each kubeedge node will get a uniqe IP address for the cluster. The IP address of the flannel network interface could be used to connect to the tunnel server. However, this is only possible if one can specify the network interface cloudcore uses for its connection.

**Proposal**:
I propose to readd the deprecated setting `InterfaceName` as `CustomInterfaceName` and make it enabled with a new bool value `CustomInterfaceEnabled` that defaults to false to use the current standard behavior. I propose to basically use the removed code from #2238 but with some control logic around it. I'd leave the keadm tool untouched as well, as i think this might not be the standard use case for most keadm users.

siredmar:

**What type of PR is this?**
/kind feature

This PR adds the option `EnableCustomInterface` and `CustomInterfaceName`. If enabled the IP address will be obtained directly from the specified interface. This is needed when edgecore needs to connect with a specific IP address, e.g. flannel interface's IP address instead any IP address the system may find.

Fixes #3127 

This partly reintegrates code deleted in #2238

```release-note
Added the ability to define a custom network interface for connecting to tunnel server.
```


edge/pkg/edged/edged_pods.go
```diff
@@ -737,7 +738,13 @@ func (e *edged) podFieldSelectorRuntimeValue(fs *v1.ObjectFieldSelector, pod *v1
...
 		return pod.Spec.ServiceAccountName, nil
 	case "status.hostIP":
-		hostIP, err := pkgutil.GetLocalIP(e.nodeName)
+		var hostIP string
+		var err error
+		if e.customInterfaceEnabled {
+			hostIP, err = e.getHostIPByInterface()
+		} else {
+			hostIP, err = pkgutil.GetLocalIP(e.nodeName)
+		}
```
zc2638, line 747: Can you try to use `netutil.ChooseBindAddressForInterface` instead?

edge/pkg/edged/edged.go
```diff
@@ -179,6 +179,8 @@ type edged struct {
 	namespace                 string
 	nodeName                  string
 	runtimeCache              kubecontainer.RuntimeCache
+	customInterfaceName       string
+	customInterfaceEnabled    bool //siredmar
```
zc2638, line 183: If `customInterfaceName` is not empty, it will take effect. There is no need for `customInterfaceEnabled` parameter

siredmar: you are right, i will remove that enable flag

edge/pkg/edged/edged_pods.go
siredmar: alright, will use this one instead.

siredmar:
@zc2638 would you rereview this? I made some more changes and found a bug on the previous version

common/constants/default.go
```diff
@@ -55,6 +55,7 @@ const (
 	DefaultMaximumDeadContainersPerPod = 1
 	DefaultHostnameOverride            = "default-edge-node"
 	DefaultRegisterNodeNamespace       = "default"
+	DefaultCustomInterfaceName         = ""
```
zc2638, line 58: This constant is unnecessary

edge/pkg/edged/edged_pods.go
```diff
@@ -788,8 +801,17 @@ func (e *edged) makeBlockVolumes(pod *v1.Pod, container *v1.Container, podVolume
...
+	var err error
+	var ip net.IP
+	if e.customInterfaceName != constants.DefaultCustomInterfaceName {
+		ip, err = netutil.ChooseBindAddressForInterface(e.customInterfaceName)
+		if ip != nil {
+			hostIP = ip.String()
+		}
+	} else {
+		hostIP, err = pkgutil.GetLocalIP(e.nodeName)
+	}
```
zc2638, line 814: Can you try to encapsulate a method into `pkgutil` package?
And if the `ChooseBindAddressForInterface` acquisition fails, will you consider using the `GetLocalIP ` method to attempt to obtain it?

edge/cmd/edgecore/app/server.go
```diff
@@ -82,12 +85,19 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
...
+			// Defaults to the local IP from the default interface by the default config
+			if config.Modules.Edged.CustomInterfaceName != constants.DefaultCustomInterfaceName {
+				var ip net.IP
+				var localIP string
+				klog.Infof("Getting IP address by custom interface: %s", config.Modules.Edged.CustomInterfaceName)
+				ip, err = netutil.ChooseBindAddressForInterface(config.Modules.Edged.CustomInterfaceName)
+				if ip != nil {
+					localIP = ip.String()
+				}
 				config.Modules.Edged.NodeIP = localIP
```
fisherxu, line 98: Is is possible to set nodeIP directly instead of get it through inerface name?

edge/cmd/edgecore/app/server.go
```diff
@@ -82,12 +85,19 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 				}
 			}
 
-			// get edge node local ip
-			if config.Modules.Edged.NodeIP == "" {
-				hostnameOverride := util.GetHostname()
-				localIP, _ := util.GetLocalIP(hostnameOverride)
+			// Get edge node local ip only when the custiomInterfaceName has been set.
+			// Defaults to the local IP from the default interface by the default config
+			if config.Modules.Edged.CustomInterfaceName != constants.DefaultCustomInterfaceName {
```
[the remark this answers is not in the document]
siredmar: No, the nodeIP is set by default to the IP address using `util.GetLocalIP` in `pkg/apis/componentconfig/edgecore/v1alpha1/default.go`. This gets called in `edge/cmd/edgecore/app/server.go` by `config, err := opts.Config()`.
Your mentioned code section above just checks if a CustomInterface was set and  patches the default configuration by the custom interface's IP.

edge/cmd/edgecore/app/server.go
siredmar: I explicitly want to use the interface name. However it is still possible to manually set the NodeIP, if the `CustomInterfaceName` is empty in the config.

edge/pkg/edged/edged_pods.go
siredmar: I can encapsulate this, sure.

And if the `ChooseBindAddressForInterface` fails, this means there is an error of some sort. In this case i would exit with the error because it is higly possible that there is a user misconfiguration e.g. wrong interface name or no IP address for that interface. 
It is better to exit in a defined way than run anyway with some sort of unknown behavior. The user expects the interface `wlan0` but gets `eth0`. As a user that would confuse me quite a bit.

common/constants/default.go
siredmar: will remove it

edge/cmd/edgecore/app/server.go
```diff
@@ -82,12 +85,19 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
...
+				var ip net.IP
+				var localIP string
+				klog.Infof("Getting IP address by custom interface: %s", config.Modules.Edged.CustomInterfaceName)
+				ip, err = netutil.ChooseBindAddressForInterface(config.Modules.Edged.CustomInterfaceName)
+				if ip != nil {
+					localIP = ip.String()
+				}
```
gy95, line 97: If `err != nil` or `ip == nil` , do we need stop the process?

edge/cmd/edgecore/app/server.go
```diff
@@ -82,12 +85,19 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
...
 				config.Modules.Edged.NodeIP = localIP
 			}
+			klog.Infof("Local IP: %s", config.Modules.Edged.NodeIP)
```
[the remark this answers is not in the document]
subpathdev: I think the ip adresses should printed without setting the verbosity explicit. Why do you think the address should only printed in verbosity level 4?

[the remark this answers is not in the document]
siredmar: I consent. Setting a custom interface is a setting i as a user do want to see in the logs by default.

edge/cmd/edgecore/app/server.go
subpathdev: I think that feature can be added in the feature with separate pr

edge/pkg/edged/edged_pods.go
```diff
@@ -1438,3 +1460,29 @@ func (pk *podKillerWithChannel) PerformPodKillingWork() {
 		}
 	}
 }
+
+// getHostIPByInterface gets the IP address of a the custom network interface
+func (e *edged) getHostIPByInterface() (string, error) {
```
fisherxu, line 1465: Seems the method is not used :)

edge/cmd/edgecore/app/server.go
fisherxu: +1, please handle the err case.

siredmar: Ok. will exit the process in the error case.

edge/pkg/edged/edged_pods.go
siredmar: Not anymore, already deleted locally. Will be removed with the next commit.

edge/pkg/edged/edged_pods.go
```diff
@@ -69,8 +70,10 @@ import (
 	"k8s.io/kubernetes/third_party/forked/golang/expansion"
 	utilfile "k8s.io/utils/path"
 
+	"github.com/kubeedge/kubeedge/common/constants"
 	edgedconfig "github.com/kubeedge/kubeedge/edge/pkg/edged/config"
 	pkgutil "github.com/kubeedge/kubeedge/pkg/util"
+	netutil "k8s.io/apimachinery/pkg/util/net"
```
gy95, line 76: this import package need to be moved to the second part, together with the below packages
```
"k8s.io/kubernetes/third_party/forked/golang/expansion"
	utilfile "k8s.io/utils/path"
```
or, will not pass the gci lint checks

siredmar: ok, will fix this

edge/cmd/edgecore/app/server.go
subpathdev: Oh I think I get it now

subpathdev: @fisherxu you can set the IP Address by hand in the configuration file already. Do you mean that?

pkg/util/util.go
```diff
@@ -61,6 +62,28 @@ func GetLocalIP(hostName string) (string, error) {
 	return ipAddr.String(), nil
 }
 
+// GetNodeIP returns node ip
+func GetNodeIP(hostname string, interfaceName string) (string, error) {
+	var hostIP string
```
gy95, line 67: i think it's better to use using `edged` as receiver, and the above block to get ip can also use this function.
```
if config.Modules.Edged.CustomInterfaceName != "" {
				klog.Infof("Getting IP address by custom interface: %s", config.Modules.Edged.CustomInterfaceName)
				ip, err := 
....
```

siredmar: If i understand you correct, you asking me to move this GetNodeIP from pkgutil to edged as receiver. I can do that, Ok.
However the code section in your comment refers to the `server.go` at the very beginning. At his time there is no edged component existend at all. So i would suggest leaving this as it is the best option here.

gy95: got it, you are right.

siredmar:
This can be reviewed again. I think i got covered everything this time.

edge/cmd/edgecore/app/server.go
fisherxu: Yes, because I noticed in k8s/kubelet there's no interface name flag, only has nodeIP. I have another question, if users specify both nodeIP and interfaceName, seems now the nodeIP will be overided, we can add some comments that interfaceName has a higher priority if both specified.

edge/cmd/edgecore/app/server.go
```diff
@@ -82,12 +83,18 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
...
-				localIP, _ := util.GetLocalIP(hostnameOverride)
-				config.Modules.Edged.NodeIP = localIP
+			// Get edge node local ip only when the custiomInterfaceName has been set.
+			// Defaults to the local IP from the default interface by the default config
+			if config.Modules.Edged.CustomInterfaceName != "" {
+				klog.Infof("Getting IP address by custom interface: %s", config.Modules.Edged.CustomInterfaceName)
+				ip, err := netutil.ChooseBindAddressForInterface(config.Modules.Edged.CustomInterfaceName)
+				if err != nil {
+					klog.Errorf("Failed to get IP address by custom interface: %s: %v", config.Modules.Edged.CustomInterfaceName, err)
+					os.Exit(1)
```
fisherxu, line 93:
```suggestion
					klog.Fatal("Failed to get IP address by custom interface: %s: %v", config.Modules.Edged.CustomInterfaceName, err)
```

edge/cmd/edgecore/app/server.go
fisherxu: Already has comments now :)

edge/cmd/edgecore/app/server.go
subpathdev: I think printing the stacktrace, doesn't help to debug this, so an error message and exiting with in failure state is better.

edge/pkg/edged/edged_pods.go
```diff
@@ -1438,3 +1435,25 @@ func (pk *podKillerWithChannel) PerformPodKillingWork() {
...
+		hostIP = ip.String()
+	} else {
+		hostIP, err = pkgutil.GetLocalIP(hostname)
+		if err != nil {
+			klog.Errorf("Cannot obtain the IP Address using the hostname %q: %+v", hostname, err)
+			return "", err
+		}
+	}
+	return hostIP, nil
+}
```
fisherxu, line 1459:
```suggestion
func (e *edged) getNodeIP(hostname string, interfaceName string) (string, error) {
	if interfaceName != "" {
		ip, err := utilnet.ChooseBindAddressForInterface(interfaceName)
		if err != nil {
			klog.Errorf("Cannot obtain the IP Address using the interface %q: %+v", interfaceName, err)
			return "", err
		}
		return ip.String(), nil
	} 
	
	hostIP, err := pkgutil.GetLocalIP(hostname)
	if err != nil {
		klog.Errorf("Cannot obtain the IP Address using the hostname %q: %+v", hostname, err)
		return "", err
	}
	
	return hostIP, nil
}
```

edge/cmd/edgecore/app/server.go
fisherxu: Sounds good. :)

fisherxu (commented):
Others lgtm, thanks for the feature :)

fisherxu (commented):
/approve
Leave to @subpathdev @zc2638 @gy95 for final review.

PR #3131, opened 2021-09-04
add experimental feature

zc2638:
**What type of PR is this?**

/kind feature

```
// Experimental indicates the experimental features that need to be enabled.
// e.g. ProcessStatusSync=true
```

Fixes #

```release-note
Users can use the experimental feature function by configuring the featureGates field in the configuration file.
e.g.

kind: CloudCore
featureGates:
  ProcessStatusSync: true
...
```


subpathdev:
@zc2638 can you add a little bit more description in this PR please. I think it is useful to have a experimental flag in the edgecore. Those a flog can also be useful in the cloudcore. Can you add these flag there too?

fisherxu (commented):
How about importing feature gate from k8s? Then we can control the feature start or stop.  Ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L782

zc2638:
> How about importing feature gate from k8s? Then we can control the feature start or stop. Ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L782

ok, I'll revise it.

subpathdev:
\lgtm
Can you add a description at the `user-facing change` field, please?

pkg/features/features.go
```diff
@@ -0,0 +1,33 @@
...
+
+func init() {
+	runtime.Must(DefaultMutableFeatureGate.Add(defaultFeatureGates))
+}
+
+const (
+	// ProcessStatusSync
+	// owner: @zc2638
+	// alpha: v1.8.0.beta
+	ProcessStatusSync featuregate.Feature = "ProcessStatusSync"
```
fisherxu, line 25: Could we add these lines alone with the feature code?

fisherxu (commented):
LGTM, thanks for add this to kubeedge!!

fisherxu (commented):
/lgtm
/approve

PR #3132, opened 2021-09-06
simply return an error

subpathdev:
**What type of PR is this?**

/kind cleanup

It isn't helpful to create an error first and then return that error


fisherxu (commented):
/approve

fisherxu (commented):
/lgtm

PR #3136, opened 2021-09-06
Idenfifying session not by IP of node

siredmar:
Using the IP address of the node the pod is running on for identifying the session.

**What type of PR is this?**

/kind bug

Currently the session key is idenfitied by the IP address of the node. However this breaks if there are two nodes with the same IP address, e.g. two different LANs with nodes connected who share the same IP address by accident. This PR changes the session key to the nodes name the pod is running on. This way cloudcore can address the correct node when trying to read the logs. The current behavior can fail if it addresses a node by IP which does not have the pod running.

### DIFF
--- a/cloud/pkg/cloudstream/streamserver.go
+++ b/cloud/pkg/cloudstream/streamserver.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cloudstream
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
@@ -26,10 +27,12 @@ import (
 	"strings"
 
 	"github.com/emicklei/go-restful"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/klog/v2"
 
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudstream/config"
+	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/common/constants"
 	"github.com/kubeedge/kubeedge/pkg/stream/flushwriter"
 )
@@ -106,7 +109,23 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
 		}
 	}()
 
-	sessionKey := strings.Split(r.Request.Host, ":")[0]
+	// extract pod namespace and pod name from request
+	meta := strings.Split(r.Request.URL.Path, "/")
+	namespaceName := meta[2]
+	podName := meta[3]
+
+	kubeClient := client.GetKubeClient()
+	if kubeClient == nil {
+		klog.Fatalf("cannot get kube client\n")
+		return
+	}
+
+	pod, err := kubeClient.CoreV1().Pods(namespaceName).Get(context.Background(), podName, v1.GetOptions{})
+	if err != nil {
+		klog.Errorf("get pod %s/%s failed: %v\n", namespaceName, podName, err)
+		return
+	}
+	sessionKey := pod.Spec.NodeName
 	session, ok := s.tunnel.getSession(sessionKey)
 	if !ok {
 		err = fmt.Errorf("can not find %v session ", sessionKey)


cloud/pkg/cloudstream/streamserver.go
```diff
@@ -106,7 +109,23 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
...
 
-	sessionKey := strings.Split(r.Request.Host, ":")[0]
+	// extract pod namespace and pod name from request
+	meta := strings.Split(r.Request.URL.Path, "/")
+	namespaceName := meta[2]
+	podName := meta[3]
+
+	kubeClient := client.GetKubeClient()
+	if kubeClient == nil {
+		klog.Fatalf("cannot get kube client\n")
```
gy95, line 119: need to use `fatal` and stop the process?

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -106,7 +109,23 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
...
+		return
+	}
+
+	pod, err := kubeClient.CoreV1().Pods(namespaceName).Get(context.Background(), podName, v1.GetOptions{})
+	if err != nil {
+		klog.Errorf("get pod %s/%s failed: %v\n", namespaceName, podName, err)
```
gy95, line 125: no need to use newline character when klog print logs

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -106,7 +109,23 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
...
+		return
+	}
+	sessionKey := pod.Spec.NodeName
 	session, ok := s.tunnel.getSession(sessionKey)
```
gy95, line 129: when use nodeName, IMO, we also need to use nodeName as key to store in the tunnel map cache.

gy95:
Another question, `s.tunnel.getSession(sessionKey)` this function is also  called by many other codes, do we also need to modify all those callers?

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -106,7 +109,23 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
 		}
 	}()
 
-	sessionKey := strings.Split(r.Request.Host, ":")[0]
+	// extract pod namespace and pod name from request
+	meta := strings.Split(r.Request.URL.Path, "/")
+	namespaceName := meta[2]
+	podName := meta[3]
+
+	kubeClient := client.GetKubeClient()
```
zc2638, line 117: Can use `informer` here?

cloud/pkg/cloudstream/streamserver.go
subpathdev: He is using `Fatalf()` this will exit the process and print a stacktrace. I don't get your point

cloud/pkg/cloudstream/streamserver.go
subpathdev: Yes, klog checks if a new line is added at the end of the string and if is missing, klog will add one. So this is something like a programming style and I think this is not a problem.

cloud/pkg/cloudstream/streamserver.go
subpathdev: Yes you are right, but the nodeName is similiar to the config parameter `hostnameOveride` from the edge and this will already be used in the tunnel map.

subpathdev:
> Another question, `s.tunnel.getSession(sessionKey)` this function is also called by many other codes, do we also need to modify all those callers?

In my opinion, we have to use the unique parameter to identify the edges. The IP-Address isn't unique because multiple edges can use the same CIDR and so they can have the same IP-Address. So yes, we have to use the nodeName, which is unique by design, instead.

cloud/pkg/cloudstream/streamserver.go
subpathdev: Can you describe your point a little bit more or can you add a reference to the informer, please?

zc2638: Please see `cloud/pkg/common/informers`.
It may be better for `get` or `list` to request data from the cache.

subpathdev: Yes, requesting data from the cache might be better. But actually there are not the information which we need. From the `kubectl log` request we know the name of the namespace and the name of the pod. So currently I think it is easier to query the apiserver. The usage of the cache can be added in a different request. To track this we can create an issue. Is this ok for you?

cloud/pkg/cloudstream/streamserver.go
siredmar: When you can't get the Kube Client you have a pretty much broken system you don't wont to work with.
So in my opinion he correct behavior is to exit.

cloud/pkg/cloudstream/streamserver.go
siredmar: Ok, will remove it

cloud/pkg/cloudstream/streamserver.go
siredmar: as @subpathdev mentioned the hostname and the IP are already used as keys for the sessions when registering the edge to the tunnel server

cloud/pkg/cloudstream/streamserver.go
siredmar: @zc2638 let's fix this bug first and create an issue for the caching later on

siredmar:
@fisherxu can we use this fix for now and try using the informer later on?

fisherxu (commented):
/approve
Sorry for the delay!

PR #3139, opened 2021-09-07
fix message format

monsteredc:
**What type of PR is this?**
/kind bug

The message string seems not to be formatted, just format the parameters.

Fixes #

```release-note

```

### DIFF
--- a/pkg/stream/message.go
+++ b/pkg/stream/message.go
@@ -76,7 +76,7 @@ func (m *Message) Bytes() []byte {
 }
 
 func (m *Message) String() string {
-	return fmt.Sprintf("MESSAGE: connectid %v MessageType %s", m.ConnectID, m.MessageType)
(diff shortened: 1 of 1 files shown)


pkg/stream/message.go
```diff
@@ -76,7 +76,7 @@ func (m *Message) Bytes() []byte {
 }
 
 func (m *Message) String() string {
-	return fmt.Sprintf("MESSAGE: connectid %v MessageType %s", m.ConnectID, m.MessageType)
+	return fmt.Sprintf("MESSAGE: ConnectID %v MessageType %s", m.ConnectID, m.MessageType)
```
subpathdev, line 79: can you use either `connectID` or `ConnectID` i would preffer `connectID`

monsteredc: @subpathdev Done. Thanks for the suggestion!

subpathdev (approved):

fisherxu (commented):
/approve

PR #3144, opened 2021-09-07
fix kubeedge license

zc2638:
**What type of PR is this?**

/kind bug

Fixes #

```release-note

```


cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2014 The Kubernetes Authors.
+Copyright 2020 The Kubeedge Authors.
```
subpathdev, line 2: can you use the current year with 2021?

zc2638: My fault, it should be 2019.This file was created at that time.

subpathdev: Why we have to use the date, when the file was created? Can we used the last modification date instead? And that is in the current year ....

zc2638: `Copyright` usually starts with the first date and ends with the second date.

subpathdev: Yes you are right

fisherxu (commented):
This is because we copied some codes like `updatePodStatus` from K8s/kubelet, so we used the k8s license here. So can we add a new line for kubeedge and keep the k8s here?

zc2638:
> This is because we copied some codes like `updatePodStatus` from K8s/kubelet, so we used the k8s license here. So can we add a new line for kubeedge and keep the k8s here?

I suggest no need to keep.
Like other projects that quote or directly copy part of the k8s code, most of them use their own license.
I did not find anything about kubernetes in the license.
What do you think?

fisherxu:
> > This is because we copied some codes like `updatePodStatus` from K8s/kubelet, so we used the k8s license here. So can we add a new line for kubeedge and keep the k8s here?
> 
> I suggest no need to keep.
> Like other projects that quote or directly copy part of the k8s code, most of them use their own license.
> I did not find anything about kubernetes in the license.
> What do you think?

In fact in edged we also keep to use the K8s license in most files, if you copy the code from a project, it's necessary to make declaration.

kevin-wangzefeng:
Respecting the copyright of upstream code is required according to Apache 2.0 license, we shall not remove tuat.

Adding a new line to claim KubeEdge copyright shall be fine.

zc2638:
> Respecting the copyright of upstream code is required according to Apache 2.0 license, we shall not remove that.
> 
> Adding a new line to claim KubeEdge copyright shall be fine.

Okey.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -1,4 +1,5 @@
 /*
+Copyright 2019 The Kubeedge Authors.
```
fisherxu, line 2:
```suggestion
Copyright 2019 The KubeEdge Authors.
```

zc2638: done.

fisherxu (commented):
Need rebase to master to fix the failed tests :)

fisherxu (commented):
/approve

fisherxu (commented):
/lgtm

PR #3146, opened 2021-09-08
update golangci-lint version to v1.42.0

gy95:

**What type of PR is this?**
/kind feature

golangci-lint v1.30.0  has more than one year, so update it to v1.42.0
and when exec `make lint`, we should ensure not only golangci-lint has been installed, but also version must be v1.42.0

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #3148, opened 2021-09-08
fix some code smells

WillardHu:
- upstream.go error strings should not be capitalized
- util_test.go avoid using reflect.DeepEqual with errors

**What type of PR is this?**

/kind bug

Fixes #

```release-note

```


subpathdev:
can you fix the unit test, please

gy95:
seems not a right  way to use `errors.Is` to judge error is equal or not.

WillardHu:
> can you fix the unit test, please

OK

WillardHu:
> seems not a right way to use `errors.Is` to judge error is equal or not.

Yes, you're right

fisherxu (commented):
/approve

PR #3150, opened 2021-09-08
fix parameter verification of func GoToTarget

WillardHu:
**What type of PR is this?**

/kind bug

Data parameters could not be correctly verified  before repair.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/router/provider/eventbus/eventbus.go
+++ b/cloud/pkg/router/provider/eventbus/eventbus.go
@@ -112,18 +112,24 @@ func (*EventBus) Forward(target provider.Target, data interface{}) (response int
 
 func (eb *EventBus) GoToTarget(data map[string]interface{}, stop chan struct{}) (interface{}, error) {
 	messageID, ok := data["messageID"].(string)
-	body, ok := data["data"].([]byte)
-	param, ok := data["param"].(string)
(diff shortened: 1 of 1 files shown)


cloud/pkg/router/provider/eventbus/eventbus.go
```diff
@@ -112,18 +112,30 @@ func (*EventBus) Forward(target provider.Target, data interface{}) (response int
 
 func (eb *EventBus) GoToTarget(data map[string]interface{}, stop chan struct{}) (interface{}, error) {
 	messageID, ok := data["messageID"].(string)
-	body, ok := data["data"].([]byte)
-	param, ok := data["param"].(string)
+	if !ok {
+		err := errors.New("data transform failed, messageID must be not null")
```
gy95, line 116: better to print type is not matched, !ok happens when  maybe value is nil or type does not match string

WillardHu: Ok, fixed

fisherxu (commented):
/approve

WillardHu:
> WillardHu force-pushed the WillardHu:fix/eventbus branch from b273023 to 19c4e28 40 seconds ago

Batter error message.

PR #3157, opened 2021-09-09
feat(edge): node ephemeral storage info

### LINKED ISSUE #3156: edgecore doesn't report node ephemeral storage

edgecore reports ephemeral storage which is the behavior of kubelet

image

**Why is this needed**:

To keep parity with kubelet. And the infomation is useful for admin and scheduler

stingshen:
Signed-off-by: shenyi <shenyi@cambricon.com>

**What type of PR is this?**

/kind feature

Edgecore reports ephemeral storage which is the behavior of kubelet. The behavior keeps parity with kubelet. And the info is useful for scheduler.

Fixes #3156

N/A

```release-note
Ephemeral storage capacity and allocatable is reported by edgecore
```

### DIFF
--- a/edge/pkg/edged/edged_status.go
+++ b/edge/pkg/edged/edged_status.go
@@ -37,6 +37,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/kubelet/cadvisor"
 
 	"github.com/kubeedge/beehive/pkg/core"
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
Thanks for the fixing, do you have usage scenario for this feature?

stingshen:
> Thanks for the fixing, do you have usage scenario for this feature?

There are two use case in out scenario:

1. kube-scheduler should know how much resource is left on the node. CPU, memory, ephemeral storage requirements should be set on every pods deployed on the node. If the allocatable ephemeral storage doesn't set on the node, kube-scheduler couldn't compute the ephemeral storage left on node and schedule the pod properly.
2. cluster admin needs to use kubectl or other k8s clients to check the capacity of CPU, memory, ephemeral storage on each node. It helps if ephemeral storage is set on the k8s node capacity and allocatable field.

stingshen:
> Thanks for the fixing, do you have usage scenario for this feature?

There are two use case in out scenario:

1. kube-scheduler should know how much resource is left on the node. CPU, memory, ephemeral storage requirements should be set on every pods deployed on the node. If the allocatable ephemeral storage doesn't set on the node, kube-scheduler couldn't compute the ephemeral storage left on node and schedule the pod properly.
2. cluster admin needs to use kubectl or other k8s clients to check the capacity of CPU, memory, ephemeral storage on each node. It helps if ephemeral storage is set on the k8s node capacity and allocatable field.

fisherxu (commented):
/lgtm

PR #3068, opened 2021-08-30
Support sendToGroup printing module name

zc2638:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
> Support sendToGroup printing module name

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3161, opened 2021-09-10
cherry pick of #3068: support sendToGroup printing module name 

zc2638:
cherry pick of #3068: support sendToGroup printing module name.
We need this added record to verify the problem in version 1.8.


gy95:
we could use script `hack/cherry_pick_pull.sh` to auto cherry-pick commit to another branch, it seems that this PR is not opened using provided script

zc2638:
> we could use script `hack/cherry_pick_pull.sh` to auto cherry-pick commit to another branch, it seems that this PR is not opened using provided script

Thx. I'll do it.

PR #3169, opened 2021-09-13
support configure edged version

gy95:

**What type of PR is this?**
/kind feature

Now, edged version is format by `fmt.Sprintf("%s-kubeedge-%s", constants.CurrentSupportK8sVersion, version.Get())`, for example, just like `v1.19.3-kubeedge-v1.7.1-5+9307dc7d3557eb`

we could support user customize version format like `1.11.2`

So if this version configuration is empty by default, version is generated using the old way. Or will be customized by the Edged configuration

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -443,6 +443,13 @@ func newEdged(enable bool) (*edged, error) {
 
 	metaClient := client.New()
 
+	edgedVersion := ""
+	if edgedconfig.Config.Version != "" {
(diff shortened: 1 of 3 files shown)


fisherxu (commented):
edged version should be auto generated, which reflect the version of the binary,  
and it's a fixed value after building.

gy95:
> edged version should be auto generated, which reflect the version of the binary,
> and it's a fixed value after building.

Got it.
/close

PR #3172, opened 2021-09-15
fix: use path.Join(..) to set resource string

WillardHu:

**What type of PR is this?**

/kind bug
/kind test

Make the code better for performance and readability, and add unit test case when `HubInfo` is nil.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/common/model/types.go
+++ b/cloud/pkg/cloudhub/common/model/types.go
@@ -4,7 +4,7 @@ import (
 
 	// Mapping value of json to struct member
 	_ "encoding/json"
-	"fmt"
+	"path"
 	"strings"
 
 	"github.com/kubeedge/beehive/pkg/core/model"
@@ -72,14 +72,15 @@ type HubInfo struct {
 
 // NewResource constructs a resource field using resource type and ID
 func NewResource(resType, resID string, info *HubInfo) string {
-	var prefix string
+	paths := make([]string, 0, 4)
 	if info != nil {
-		prefix = fmt.Sprintf("%s/%s/", model.ResourceTypeNode, info.NodeID)
+		paths = append(paths, model.ResourceTypeNode, info.NodeID)
 	}
-	if resID == "" {
-		return fmt.Sprintf("%s%s", prefix, resType)
+	paths = append(paths, resType)
+	if resID != "" {
+		paths = append(paths, resID)
 	}
-	return fmt.Sprintf("%s%s/%s", prefix, resType, resID)
+	return path.Join(paths...)
 }
 
 // IsNodeStopped indicates if the node is stopped or running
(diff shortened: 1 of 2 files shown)


cloud/pkg/cloudhub/common/model/types.go
```diff
@@ -72,14 +72,15 @@ type HubInfo struct {
 
 // NewResource constructs a resource field using resource type and ID
 func NewResource(resType, resID string, info *HubInfo) string {
-	var prefix string
+	paths := make([]string, 4)
 	if info != nil {
-		prefix = fmt.Sprintf("%s/%s/", model.ResourceTypeNode, info.NodeID)
+		paths = append(paths, model.ResourceTypeNode, info.NodeID)
```
Iceber, line 77: `append` will still cause the slice to expand.
fix: 
```go
paths = make([]string, 0 , 4)
```

Iceber (changes requested):

cloud/pkg/cloudhub/common/model/types.go
```diff
@@ -72,14 +72,15 @@ type HubInfo struct {
...
 	}
-	if resID == "" {
-		return fmt.Sprintf("%s%s", prefix, resType)
+	paths = append(paths, resType)
+	if resID != "" {
+		paths = append(paths, resID)
 	}
-	return fmt.Sprintf("%s%s/%s", prefix, resType, resID)
+	return path.Join(paths...)
```
Iceber, line 83: Although this may not cause any problems，it is worth noting that this PR will cause the result of the function to change.

When `NewResource("", "id",nil)` is called, the return value will change from `/id` to `id`.

Regardless of whether `resType` can be empty, the format should be `resType/resID` by semantics, even if the type is empty, the `/` should be preserved.

@fisherxu What do you think?

cloud/pkg/cloudhub/common/model/types.go
WillardHu: Fixed, thx.

cloud/pkg/cloudhub/common/model/types.go
WillardHu:
```golang
var prefix string
if info != nil {
    prefix = fmt.Sprintf("%s/%s/", model.ResourceTypeNode, info.NodeID)
}
if resID == "" {
    return fmt.Sprintf("%s%s", prefix, resType)
}
return fmt.Sprintf("%s%s/%s", prefix, resType, resID)
```
By the way, in the historical version of the code , when ‘resID == “”’, the constructor return value will be:
  1.ResourceTypeNode/info.NodeID/resType
  2. resType
  3. ""
Waiting for the outcome of the discussion, thx.

fisherxu: Make sense, I'd prefer not modify this, have not seen significant improvement, and also not quite sure about the impact.

WillardHu: Ok~

PR #3175, opened 2021-09-17
Grant access to cloudcore for PV and PVC

cuirunxing-hub:
**What type of PR is this?**
/kind bug

when use local-volume plugin,if edged cannot get pv&pvc,the volume mount action cannot go on

Fixes #

```release-note

```


fisherxu (commented):
/approve

PR #3176, opened 2021-09-17
support envFrom configmap in edge pods

### LINKED ISSUE #3020: Can't get configmap when the pod in the edgenode
I create a deployment , the deployment use a  configmapRef  like below:
 envFrom:
        - configMapRef:
            name:  config-env
when the deployment is run in the cloud node , it  can get the configmap, but when it run in the edge node, It can't get the config map. 
Deploy the kubeedge  as the doc  https://kubeedge.io/zh/docs/setup/keadm/.
And then create a configmap, and create a deployment to use the configmap, and  add a nodeSelector to let  it  run in the edge node.
I deploy the kubeedge  use the keadm,  and deploy the kubernets use the kubeadm. and use the calico  as the CNI
**Environment**:
- Kubernetes version (use `kubectl version`): 1.19.3  and   1.16.0   hase the same  issue
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): 1.7.1

haoheipi:
**What type of PR is this?**
/kind feature

Currently the pods running at edge can not set the environment variables from envFrom tag,
this pr is to support set environment variables with envFrom configmap.
- Reading from configmap
```
envFrom:
- configMapRef:
    name: special-config
```

Fixes #3020

### DIFF
--- a/edge/pkg/edged/edged_pods.go
+++ b/edge/pkg/edged/edged_pods.go
@@ -699,6 +699,9 @@ func (e *edged) makeEnvironmentVariables(pod *v1.Pod, container *v1.Container, p
 
 			invalidKeys := []string{}
 			for k, v := range configMap.Data {
+				if len(envFrom.Prefix) > 0 {
+					k = envFrom.Prefix + k
+				}
 				if errMsgs := utilvalidation.IsEnvVarName(k); len(errMsgs) != 0 {
 					invalidKeys = append(invalidKeys, k)
 					continue
--- a/edge/pkg/edged/edged_pods.go
+++ b/edge/pkg/edged/edged_pods.go
@@ -49,6 +49,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/sets"
+	utilvalidation "k8s.io/apimachinery/pkg/util/validation"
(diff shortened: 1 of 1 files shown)


edge/pkg/edged/edged_pods.go
```diff
@@ -672,6 +673,45 @@ func (e *edged) makeEnvironmentVariables(pod *v1.Pod, container *v1.Container, p
...
+						// ignore error when marked optional
+						continue
+					}
+					return result, err
+				}
+				configMaps[name] = configMap
+			}
+
+			invalidKeys := []string{}
+			for k, v := range configMap.Data {
```
fisherxu, line 701: Do we also need to add these lines here?
```go
if len(envFrom.Prefix) > 0 {
    k = envFrom.Prefix + k
}
```

haoheipi: "Prefix" is used to prevent name conflicts when multiple ConfiMaps are referenced.
Because I don't see an example of this field in the configmap documentation of K8S, I understand that this is rarely used by others so I remove this codes. But It was like the previous behavior of removing envfrom, so I decided to add it, what do you think？

fisherxu: Yes, let's add it back and keep consistent with K8s :)

fisherxu (commented):
/approve

fisherxu (commented):
/lgtm

PR #2867, opened 2021-05-31
Cloudcore support ha

fisherxu:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Support cloudcore run in multi-instances mode, cloud-edge tunnel can also work for multi-instances.

**Which issue(s) this PR fixes**:

Fixes #
Ref: #1560

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Support cloudcore run as multi instaces.
```

PR #3184, opened 2021-09-19
add stream/tunnel port for cloudcore

allenleu:
add stream/tunnel port for cloudcore


build/cloud/07-deployment.yaml
```diff
@@ -26,6 +26,12 @@ spec:
             - containerPort: 10000
               name: cloudhub
               protocol: TCP
+            - containerPort: 10003
+              name: stream
+              protocol: TCP
+            - containerPort: 10004
+              name: tunnel
+              protocol: TCP
```
fisherxu, line 34: I think we can delete the ports here, that's useless here, we have used `hostnetwork` here.

build/cloud/ha/03-ha-deployment.yaml.example
```diff
@@ -52,6 +52,12 @@ spec:
             - containerPort: 10002
               name: certandreadyz
               protocol: TCP
+            - containerPort: 10003
+              name: stream
+              protocol: TCP
+            - containerPort: 10004
+              name: tunnel
+              protocol: TCP
```
fisherxu, line 60: ditto

allenleu (commented):
It currently affects the normal startup of cloudcore

build/cloud/07-deployment.yaml
fisherxu: Can the ports be removed here? cc @zhu733756 Please help take a look :)

zhu733756: > Can the ports be removed here? cc @zhu733756 Please help take a look :)

I think it cannot be removed, and the related service type needs to use nodeport or loadblancer.  Therefore, these exposed services can be accessed by the edge nodes.

zhu733756:
@allenleu I have pushed a pr in #3304 to add helm support , we can also discuss there.

build/cloud/07-deployment.yaml
allenleu: @zhu733756 I just withdrew my submission. Do THE Stream and tunnel ports need to be reserved? Also should readinessGates be removed?

zhu733756: > @zhu733756 I just withdrew my submission. Do THE Stream and tunnel ports need to be reserved? Also should readinessGates be removed?

Yeah. And I think you also need to expose nodeport for the cloudcore service.

allenleu: @zhu733756 Ok, I need to think about it. There is another problem. I have seen that Cloudcore has functions related to leader-election in the 1.6 version of the code, but the 1.7 version has deleted all the relevant codes. Do you know the reason?

zhu733756: > @zhu733756 Ok, I need to think about it. There is another problem. I have seen that Cloudcore has functions related to leader-election in the 1.6 version of the code, but the 1.7 version has deleted all the relevant codes. Do you know the reason?

KubeEdge support ha-cloudcore now . See #2867

allenleu: Okey，thanks.

fisherxu: @allenleu All the leader selection related code can be removed. :)

fisherxu: Also you can take a look for #3304, we can use helm to deploy the cloudcore. @allenleu

PR #3185, opened 2021-09-19
Upgrading the API version of the CRDs to apiextensions.k8s.io/v1

prodanlabs:
Signed-off-by: prodan <pengshihaoren@gmail.com>

**What type of PR is this?**
/kind feature

kubernetes v1.22 Removed apiextensions.k8s.io/v1batev1
https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/

Upgrade `devices` and `reliablesyncs` CRD API version to `apiextensions.k8s.io/v1`

I think it would be cooler to use tools like kubebuilder to generate CRDs.


fisherxu:
@prodanlabs Seems the e2e test failed, could we generated the yaml with kubebuilder?

PR #3187, opened 2021-09-22
fix some problems of socket context

gy95:

**What type of PR is this?**
/kind bug

1. socket config structure should use `json` to define, and need to assign values to the global variable `buildinModuleConfig`
2. when `StartModules`, we need to passin `MsgCtxTypeUS`, or modules that use socket way will not work well.
3. when module start, we need to use `moduleKeeper` function, to try to re-connect when socket connection offline.
4. function `IsModuleEnabled` should read from configure file to predict whether enabled, this function is only used by socket context.
5. `NewWrapper` readerType and writerType assign values bugs
6. `context_factory.go` `Send` and `SendSync` should use module name to choose contextType, but not use message type. Because when use socket way to communicate, some message type is also empty or `""`
7. `NewRespByMessage` should `SetType` also

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/core.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/core.go
@@ -13,6 +13,7 @@ import (
 
 // StartModules starts modules that are registered
 func StartModules() {
+	// only register channel mode, if want to use socket mode, we should also pass in common.MsgCtxTypeUS parameter
 	beehiveContext.InitContext([]string{common.MsgCtxTypeChannel})
 
 	modules := GetModules()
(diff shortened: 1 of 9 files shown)


gy95:
@wackxu

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // StartModules starts modules that are registered
 func StartModules() {
-	beehiveContext.InitContext([]string{common.MsgCtxTypeChannel})
+	beehiveContext.InitContext([]string{common.MsgCtxTypeChannel, common.MsgCtxTypeUS})
```
fisherxu, line 16:
```suggestion
	beehiveContext.InitContext([]string{common.MsgCtxTypeChannel})
```

We haven't used the `MsgCtxTypeUS` context, no need to init here. If users use it, just add by themselves. Can just add some comments above.

fisherxu (commented):
/lgtm
/approve

PR #3190, opened 2021-09-23
update golang to 1.16

gy95:

**What type of PR is this?**
/kind feature

upgrade golang from 1.14 to 1.16

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: "^1.14.x"
+          go-version: "^1.16.x"
       - run: go version
       # Runs a set of commands to initialize and analyze with FOSSA
       - name: run FOSSA analysis
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -18,7 +18,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14.x
+          go-version: 1.16.x
 
       - name: Checkout code
         uses: actions/checkout@v2
@@ -50,7 +50,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14.x
+          go-version: 1.16.x
 
       - uses: actions/cache@v2
         with:
@@ -85,7 +85,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14.x
+          go-version: 1.16.x
 
       - uses: actions/cache@v2
         with:
@@ -115,7 +115,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14.x
+          go-version: 1.16.x
 
       - uses: actions/cache@v2
         with:
@@ -145,7 +145,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14.x
+          go-version: 1.16.x
 
       - uses: actions/cache@v2
         with:
@@ -177,7 +177,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14.x
+          go-version: 1.16.x
 
       - uses: actions/cache@v2
         with:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ services:
   - docker
 
 go:
-  - 1.14
+  - 1.16
 
 env:
   - HOME=/home/travis
--- a/Makefile
+++ b/Makefile
@@ -281,7 +281,7 @@ edgeimage:
 	tar -xzf ./build/edge/tmp/qemu-${QEMU_ARCH}-static.tar.gz -C ./build/edge/tmp
 	docker build -t kubeedge/edgecore:${IMAGE_TAG} \
 	--build-arg GO_LDFLAGS=${GO_LDFLAGS} \
-	--build-arg BUILD_FROM=${ARCH}/golang:1.14-alpine3.11 \
+	--build-arg BUILD_FROM=${ARCH}/golang:1.16-alpine3.13 \
 	--build-arg RUN_FROM=${ARCH}/docker:dind \
 	-f build/edge/Dockerfile .
 
--- a/build/admission/Dockerfile
+++ b/build/admission/Dockerfile
@@ -4,7 +4,7 @@ ARG GO_LDFLAGS
 
 COPY . /go/src/github.com/kubeedge/kubeedge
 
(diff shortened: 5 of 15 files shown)


fisherxu (commented):
/approve

siredmar:
@gy95 Is there any reason why not bump directly to version 1.17?

fisherxu:
> @gy95 Is there any reason why not bump directly to version 1.17?

Now kubeedge depends on k8s v1.21, which use go 1.16, I think we can follow k8s and bump to 1.16? 1.17 maybe to new for users now. :)  @siredmar @gy95

gy95:
> > @gy95 Is there any reason why not bump directly to version 1.17?
> 
> Now kubeedge depends on k8s v1.21, which use go 1.16, I think we can follow k8s and bump to 1.16? 1.17 maybe to new for users now. :) @siredmar @gy95

yes, bump to 1.16

gy95:
when I  add some below codes for debug details
```
if !reflect.DeepEqual(got.Transport.(*http.Transport).TLSClientConfig.RootCAs, tt.want.Transport.(*http.Transport).TLSClientConfig.RootCAs) {
	t.Errorf("NewHTTPClientWithCA????() = %#v, want %#v", got.Transport.(*http.Transport).TLSClientConfig.RootCAs, tt.want.Transport.(*http.Transport).TLSClientConfig.RootCAs)
}
```
the print info is as follows
```
NewHTTPClientWithCA????() = &x509.CertPool{byName:map[string][]int{"0 1\r0\v\x06\x03U\x04\x06\x13\x04test1\x0f0\r\x06\x03U\x04\n\x13\x06testor":[]int{0}}, lazyCerts:[]x509.lazyCert{x509.lazyCert{rawSubject:[]uint8{0x30, 0x20, 0x31, 0xd, 0x30, 0xb, 0x6, 0x3, 0x55, 0x4, 0x6, 0x13, 0x4, 0x74, 0x65, 0x73, 0x74, 0x31, 0xf, 0x30, 0xd, 0x6, 0x3, 0x55, 0x4, 0xa, 0x13, 0x6, 0x74, 0x65, 0x73, 0x74, 0x6f, 0x72}, getCert:(func() (*x509.Certificate, error))(0x5ef640)}}, haveSum:map[x509.sum224]bool{x509.sum224{0xe3, 0x80, 0xf2, 0x4a, 0x20, 0x55, 0x6b, 0xb6, 0x4c, 0xb, 0xfc, 0xc0, 0x39, 0xc8, 0xbc, 0x9d, 0x15, 0x6f, 0x9e, 0xd1, 0x1c, 0x45, 0x1a, 0x36, 0xb5, 0x7f, 0x24, 0xaa}:true}}, want &x509.CertPool{byName:map[string][]int{"0 1\r0\v\x06\x03U\x04\x06\x13\x04test1\x0f0\r\x06\x03U\x04\n\x13\x06testor":[]int{0}}, lazyCerts:[]x509.lazyCert{x509.lazyCert{rawSubject:[]uint8{0x30, 0x20, 0x31, 0xd, 0x30, 0xb, 0x6, 0x3, 0x55, 0x4, 0x6, 0x13, 0x4, 0x74, 0x65, 0x73, 0x74, 0x31, 0xf, 0x30, 0xd, 0x6, 0x3, 0x55, 0x4, 0xa, 0x13, 0x6, 0x74, 0x65, 0x73, 0x74, 0x6f, 0x72}, getCert:(func() (*x509.Certificate, error))(0x5ef640)}}, haveSum:map[x509.sum224]bool{x509.sum224{0xe3, 0x80, 0xf2, 0x4a, 0x20, 0x55, 0x6b, 0xb6, 0x4c, 0xb, 0xfc, 0xc0, 0x39, 0xc8, 0xbc, 0x9d, 0x15, 0x6f, 0x9e, 0xd1, 0x1c, 0x45, 0x1a, 0x36, 0xb5, 0x7f, 0x24, 0xaa}:true}}
```
not sure why the two RootCAs is not deep equal. I compare the two value, they are equal

gy95:
And I use the below codes to test `reflect.DeepEqual`, And the output is 
```
not equal
&x509.CertPool{byName:map[string][]int{"":[]int{0}}, lazyCerts:[]x509.lazyCert{x509.lazyCert{rawSubject:[]uint8{}, getCert:(func() (*x509.Certificate, error))(0x77cca0)}}, haveSum:map[x509.sum224]bool{x509.sum224{0xd1, 0x4a, 0x2, 0x8c, 0x2a, 0x3a, 0x2b, 0xc9, 0x47, 0x61, 0x2, 0xbb, 0x28, 0x82, 0x34, 0xc4, 0x15, 0xa2, 0xb0, 0x1f, 0x82, 0x8e, 0xa6, 0x2a, 0xc5, 0xb3, 0xe4, 0x2f}:true}}
&x509.CertPool{byName:map[string][]int{"":[]int{0}}, lazyCerts:[]x509.lazyCert{x509.lazyCert{rawSubject:[]uint8{}, getCert:(func() (*x509.Certificate, error))(0x77cca0)}}, haveSum:map[x509.sum224]bool{x509.sum224{0xd1, 0x4a, 0x2, 0x8c, 0x2a, 0x3a, 0x2b, 0xc9, 0x47, 0x61, 0x2, 0xbb, 0x28, 0x82, 0x34, 0xc4, 0x15, 0xa2, 0xb0, 0x1f, 0x82, 0x8e, 0xa6, 0x2a, 0xc5, 0xb3, 0xe4, 0x2f}:true}}
```
```
package main

import (
	"crypto/x509"
	"fmt"
	"reflect"
)

func main() {
	a := x509.NewCertPool()
	a.AddCert(&x509.Certificate{})

	b := x509.NewCertPool()
	b.AddCert(&x509.Certificate{})

	if !reflect.DeepEqual(a, b) {
		fmt.Println("not equal")
		fmt.Printf("%#v\n", a)
		fmt.Printf("%#v\n", b)
	}
}
```

How about temporarily comment out the result comparing codes, to pass the checks 😢

gy95:
golang 1.16 modify the structure and introduce a func field, 
structure changed from 
```
type CertPool struct {
	byName map[string][]int
	certs  []*Certificate
}
```
to 
```
// CertPool is a set of certificates.
type CertPool struct {
	byName map[string][]int // cert.RawSubject => index into lazyCerts

	// lazyCerts contains funcs that return a certificate,
	// lazily parsing/decompressing it as needed.
	lazyCerts []lazyCert

	// haveSum maps from sum224(cert.Raw) to true. It's used only
	// for AddCert duplicate detection, to avoid CertPool.contains
	// calls in the AddCert path (because the contains method can
	// call getCert and otherwise negate savings from lazy getCert
	// funcs).
	haveSum map[sum224]bool
}
```
and according to the official doc: 
https://pkg.go.dev/reflect#DeepEqual
`Func values are deeply equal if both are nil; otherwise they are not deeply equal.` So the two variables are not deep equal anymore.


refer this commit:
https://github.com/golang/go/commit/e8379ab5964a920e59dbcc5bc4eaa1bbf5a88e90

fisherxu (commented):
/lgtm

PR #3192, opened 2021-09-23
Automated cherry pick of #3136 origin release 1.8

siredmar:
Using the IP address of the node the pod is running on for identifying the session.

**What type of PR is this?**

/kind bug

Currently the session key is idenfitied by the IP address of the node. However this breaks if there are two nodes with the same IP address, e.g. two different LANs with nodes connected who share the same IP address by accident. This PR changes the session key to the nodes name the pod is running on. This way cloudcore can address the correct node when trying to read the logs. The current behavior can fail if it addresses a node by IP which does not have the pod running.

### DIFF
--- a/cloud/pkg/cloudstream/streamserver.go
+++ b/cloud/pkg/cloudstream/streamserver.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cloudstream
 
 import (
+	"context"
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
/lgtm
/approve

PR #3193, opened 2021-09-23
fix beehive context send to group

gy95:

**What type of PR is this?**
/kind bug

Now in beehive context, both function `SendToGroup` and function `SendToGroupSync` use message type to choose context to send messages, 
but in actual senarios, message type sometimes is empty, but in socket mode, so will cause choose the incorrect context, 
So we need to store `group <-> contextType` mapping in a map `groupContextType`, and decide which context we need to use by reading the map cache

Fixes #

@wackxu PTAL

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #3196, opened 2021-09-23
reschedule messages to node if node is unreachable

### LINKED ISSUE #3045: Pods remain in state "Pending" when added while node is offline in v1.7.x
**What happened**: All pods that are deployed while the respective edge node is offline remain in state "Pending", in v1.7.1 (Cloudcore and Edgecore). Re-deploying the pods in question while the edge node is online succeeds. 

**What you expected to happen**: Pods should be running as soon as the edge node gets online.

* Stop edgecore
* Deploy a pod on the respective node
* Start edgecore
* -> The pod remains in state "Pending"

**Anything else we need to know?**: Comparing the cloudcore logs of v1.6.1 and v1.7.1 (edgecore version is always v1.7.1), there is one deviating line that appears only with v1.7.1:

```text
W0819 08:43:43.234425       1 upstream.go:789] message: 7aead20a-bb32-426c-99a2-94592d268fa2 process failure, update node failed with error: nodes "<nodename>" is forbidden: User "system:serviceaccount:kubeedge:cloudcore" cannot patch resource "nodes" in API group "" at the cluster scope, namespace: , name: <nodename>
```

That message is kind of expected, since "patch" is not specified in `clusterrole.yaml`. However, adding "patch" to the clusterrole only results in:

```text
W0819 09:12:07.219506       1 upstream.go:789] message: 27fb7366-37f1-473c-949f-6f8bd7f42b1f process failure, update node failed with error: Operation cannot be fulfilled on nodes " <nodename>": the object has been modified; please apply your changes to the latest version and try again, namespace: , name:  <nodename>
``` 

The pod does not start running whether "patch" is added to the clusterrole or not. I tried also Cloudcore version v1.7.2, but the problem remains.

If Cloudcore is restarted after the deployment and the respective edge node is online, the pod will start running on the edge node.

I am not sure if and how this is related to the line that appears only in the logs of v1.7.x. Is this is a general problem with v1.7.x or needs the setup to be adjusted for v1.7.x?

**Environment**:
- Kubernetes version (use `kubectl version`): v1.18.12
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): KubeEdge v1.7.1 (Cloudcore and Edgecore)

fnordic:
**What type of PR is this?**

/kind bug

Fixes #3045

Without this line, the "break" in the line below drops messages.
For example a pod creation event gets lost otherwise.
Only on next cloudcore restart (or major resync) a new message is created for example pod sync.
In a setup with many unreachable nodes and i.e. scheduled pods the queue will be again more busy, i guess.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -456,6 +456,7 @@ func (mh *MessageHandle) MessageWriteLoop(info *model.HubInfo, stopServe chan Ex
 			conn, ok := mh.nodeConns.Load(info.NodeID)
 			if !ok {
 				if retry == 1 {
+					nodeQueue.AddRateLimited(key.(string))
 					break
 				}
 				retry++


gitolicious:
@fisherxu this was introduced by your change in https://github.com/kubeedge/kubeedge/commit/eade99a1c3039a5830f20607541976ee5f2619fa so you would probably be the best reviewer here.

fisherxu:
> @fisherxu this was introduced by your change in [eade99a](https://github.com/kubeedge/kubeedge/commit/eade99a1c3039a5830f20607541976ee5f2619fa) so you would probably be the best reviewer here.

Sure, will have a look :)

cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -456,6 +456,7 @@ func (mh *MessageHandle) MessageWriteLoop(info *model.HubInfo, stopServe chan Ex
 			conn, ok := mh.nodeConns.Load(info.NodeID)
 			if !ok {
 				if retry == 1 {
+					nodeQueue.AddRateLimited(key.(string))
```
fisherxu, line 459: Thanks for the fixing! And in active-active cloudcore mode, cloudcore will fetch all edge nodes related objects, and cloudcore will omit the object if edgenode haven't connect to the related cloudcore. So here we need to break.

And in synccontroller, we will check whether oject created successfully, and if not, send the created message. But the feature still haven't implemented.

ref: https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/synccontroller/createfailedobject.go#L5

gitolicious: Does this mean the proposed change breaks active-active scenarios or it just wouldn't fix the error for those but helps for single cloudcore mode? In latter case, could we merge it until a proper solution in the synccontroller is implemented?
Is there a planned timeline for the synccontroller implementation?

elrosar: This PR seems to fix at least the issue described in [#3045](https://github.com/kubeedge/kubeedge/issues/3045).

@fisherxu, I do not fully understand your comment, does this mean that [#3045](https://github.com/kubeedge/kubeedge/issues/3045) and related issues will be solved once the feature is implemented in synccontroller? Is there any chance using this PR as transistion solution until the feature in synccontroller is implemented or would it break other things?

gitolicious: @fisherxu, any updates on this? Will the full solution in synccontroller be implemented in KubeEdge 1.19, oder can we use @elrosar's solution in the meantime?

fnordic: We use now cloudcore 1.9 the bug still persist with this new version.
To clarify and reproduce: 
stop edgecore wait until it has k8s status "NotReady" than create a deployment. Wait for approximately 20 Seconds. Start edgecore.
expected result: pod from deployment starts after device has k8s status "Ready".
actual result: pod hangs in state pending forever.

fisherxu:
@fnordic Sorry for the delay, I think the issue have been fixed in #3244 , please take a look :)

fnordic:
Not sure if  https://github.com/kubeedge/kubeedge/pull/3244 fixes this issue.
I'm closing this issue since it is no longer relevant for me.

PR #3217, opened 2021-10-11
fix: replace magic strings with constants

WillardHu:

**What type of PR is this?**
/kind bug

The map key of the source resource can be easily maintained.

Fixes #

```release-note

```


WillardHu:
> WillardHu force-pushed the fix/use-constants branch from a28bc4e to c625794 26 seconds ago
Modified string formatting.

PR #3222, opened 2021-10-13
replace deprecated package io/ioutil with packages io and os

gy95:

**What type of PR is this?**
/kind cleanup

package `io/ioutil` is deprecated in Go1.16, and according to the official doc(https://golang.org/doc/go1.16#ioutil), we could use definitions in `io` and `os`.
So replace all codes using `io/ioutil` with package `os` and `io`

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/admissioncontroller/admission.go
+++ b/cloud/pkg/admissioncontroller/admission.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"io/ioutil"
 	"net/http"
+	"os"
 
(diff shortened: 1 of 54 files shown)


gy95:
go.mod has already specified the golang version to go1.16
https://github.com/kubeedge/kubeedge/blob/4ff2d99da342e3623141665696f23ca85e12d675/go.mod#L1-L6
this patch can be pushed in progress.

gy95:
ping @fisherxu @zc2638 PTAL

fisherxu (commented):
/approve

PR #3244, opened 2021-10-26
fix/reliable-message-delivery: ObjectSync must be created when the resource first sends to CloudHub

### LINKED ISSUE #3243: There is a problem with the reliable messgae delivery
**What happened**:

Because when a resource is delivered for the first time, an ACK from edgecore is required to create the corresponding ObjectSync. So the following could happen:
When you apply a Deployment or a single Pod when the node status is ready but the edgecore has been disconnected, no ObjectSync has been created, so the Pod will not be delivered any more even after the edgecore is connected.

I expect that the Pod might be synchronized to the  side. 

1. Make edgenode break.
2. Apply a Deployment or a single Pod, and the Pod schedules to the disconnected node. 
3. Make edgenode reconnection.

**Environment**:
- It's not important.

My-pleasure:
…source first sends to CloudHub

**What type of PR is this?**

/kind bug

When the resource message arrives at CloudHub for the first time, check whether there is a corresponding ObjectSync. If not, create an ObjectSync with zero `ObjectResourceVersion` to trigger the SyncController.

Fixes #3243

NONE

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -132,7 +132,7 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
 						Name: objectSyncName,
 					},
 					Spec: v1alpha1.ObjectSyncSpec{
-						ObjectAPIVersion: util.GetMessageAPIVerison(msg),
+						ObjectAPIVersion: util.GetMessageAPIVersion(msg),
 						ObjectKind:       util.GetMessageResourceType(msg),
 						ObjectName:       resourceName,
 					},
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -19,7 +19,6 @@ import (
 	crdClientset "github.com/kubeedge/kubeedge/cloud/pkg/client/clientset/versioned"
 	reliablesyncslisters "github.com/kubeedge/kubeedge/cloud/pkg/client/listers/reliablesyncs/v1alpha1"
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/model"
-	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/modules"
 	"github.com/kubeedge/kubeedge/cloud/pkg/dynamiccontroller/application"
 	edgeconst "github.com/kubeedge/kubeedge/cloud/pkg/edgecontroller/constants"
@@ -44,11 +43,11 @@ type ChannelMessageQueue struct {
 }
 
 // NewChannelMessageQueue initializes a new ChannelMessageQueue
-func NewChannelMessageQueue(objectSyncLister reliablesyncslisters.ObjectSyncLister, clusterObjectSyncLister reliablesyncslisters.ClusterObjectSyncLister) *ChannelMessageQueue {
+func NewChannelMessageQueue(objectSyncLister reliablesyncslisters.ObjectSyncLister, clusterObjectSyncLister reliablesyncslisters.ClusterObjectSyncLister, crdClient crdClientset.Interface) *ChannelMessageQueue {
 	return &ChannelMessageQueue{
 		objectSyncLister:        objectSyncLister,
 		clusterObjectSyncLister: clusterObjectSyncLister,
-		crdClient:               client.GetCRDClient(),
+		crdClient:               crdClient,
 	}
 }
 
@@ -148,10 +147,8 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
 					klog.Errorf("Failed to update objectSync: %s, err: %v", objectSyncName, err)
 				}
 			}
-		}
-
-		// Check if message is older than already in store, if it is, discard it directly
-		if exist {
+		} else {
+			// Check if message is older than already in store, if it is, discard it directly
 			msgInStore := item.(*beehiveModel.Message)
 			if isDeleteMessage(msgInStore) || synccontroller.CompareResourceVersion(msg.GetResourceVersion(), msgInStore.GetResourceVersion()) <= 0 {
 				return
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -1,10 +1,12 @@
 package channelq
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"sync"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/cache"
@@ -13,14 +15,18 @@ import (
 
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
 	beehiveModel "github.com/kubeedge/beehive/pkg/core/model"
+	"github.com/kubeedge/kubeedge/cloud/pkg/apis/reliablesyncs/v1alpha1"
+	crdClientset "github.com/kubeedge/kubeedge/cloud/pkg/client/clientset/versioned"
 	reliablesyncslisters "github.com/kubeedge/kubeedge/cloud/pkg/client/listers/reliablesyncs/v1alpha1"
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/model"
+	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/modules"
 	"github.com/kubeedge/kubeedge/cloud/pkg/dynamiccontroller/application"
 	edgeconst "github.com/kubeedge/kubeedge/cloud/pkg/edgecontroller/constants"
 	edgemessagelayer "github.com/kubeedge/kubeedge/cloud/pkg/edgecontroller/messagelayer"
 	"github.com/kubeedge/kubeedge/cloud/pkg/synccontroller"
 	commonconst "github.com/kubeedge/kubeedge/common/constants"
+	"github.com/kubeedge/kubeedge/pkg/metaserver/util"
 )
 
 // ChannelMessageQueue is the channel implementation of MessageQueue
@@ -33,6 +39,8 @@ type ChannelMessageQueue struct {
 
 	objectSyncLister        reliablesyncslisters.ObjectSyncLister
 	clusterObjectSyncLister reliablesyncslisters.ClusterObjectSyncLister
+
+	crdClient crdClientset.Interface
 }
 
 // NewChannelMessageQueue initializes a new ChannelMessageQueue
@@ -40,6 +48,7 @@ func NewChannelMessageQueue(objectSyncLister reliablesyncslisters.ObjectSyncList
 	return &ChannelMessageQueue{
 		objectSyncLister:        objectSyncLister,
 		clusterObjectSyncLister: clusterObjectSyncLister,
+		crdClient:               client.GetCRDClient(),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -33,13 +39,16 @@ type ChannelMessageQueue struct {
...
+
+	crdClient crdClientset.Interface
 }
 
 // NewChannelMessageQueue initializes a new ChannelMessageQueue
 func NewChannelMessageQueue(objectSyncLister reliablesyncslisters.ObjectSyncLister, clusterObjectSyncLister reliablesyncslisters.ClusterObjectSyncLister) *ChannelMessageQueue {
 	return &ChannelMessageQueue{
 		objectSyncLister:        objectSyncLister,
 		clusterObjectSyncLister: clusterObjectSyncLister,
+		crdClient:               client.GetCRDClient(),
```
zc2638, line 51: I think it would be better to assign this field as a parameter.

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -108,14 +117,36 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
...
+				if err != nil {
+					klog.Errorf("Failed to create objectSync: %s, err: %v", objectSyncName, err)
+					return
+				}
+				objectSyncStatus.Status.ObjectResourceVersion = "0"
+				if _, err := q.crdClient.ReliablesyncsV1alpha1().ObjectSyncs(resourceNamespace).UpdateStatus(context.Background(), objectSyncStatus, metav1.UpdateOptions{}); err != nil {
+					klog.Errorf("Failed to update objectSync: %s, err: %v", objectSyncName, err)
+				}
 			}
 		}
```
zc2638, line 151: We shouldn't need to judge `exists` again, we can use `else` directly

cloud/pkg/cloudhub/channelq/channelq.go
My-pleasure: Sorry, what do you mean about assigning this field as a parameter?

cloud/pkg/cloudhub/channelq/channelq.go
My-pleasure: Here in line 146-149, I just update the ObjectSync's Status without judging `exists`. Can you say it more clearly?

zc2638: Sorry, I mean, there are `if !exist` in line 118 and `if exist` in line 154 in the code, can we merge them with `else`?

cloud/pkg/cloudhub/channelq/channelq.go
zc2638:
```golang
func NewChannelMessageQueue(objectSyncLister reliablesyncslisters.ObjectSyncLister, clusterObjectSyncLister reliablesyncslisters.ClusterObjectSyncLister, crdClient crdClientset.Interface) *ChannelMessageQueue
```
Like this `crdClient`, and initialize the incoming parameters in the upper layer.

cloud/pkg/cloudhub/channelq/channelq.go
My-pleasure: Ok, I know.

fisherxu:
> I found another problem, In the func `MessageWriteLoop`(https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/handler/messagehandler.go#L420), if there is no websocket connection or func `sendMsg` returns nil(etc. the message fails to be sent after five attempts), the key in the nodeQueue will be removed directly, but the message will not be removed from the nodeStore. Althogh the SyncController keeps sending messages to the CloudHub, due to a judgment in the func `addMessageToQueue`(https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/channelq/channelq.go#L125), it returns directly instead of adding it to the nodeQueue. Which results in, the first resource creation message still cannot be delivered to the edgenode. So, I think we should delete the message from nodeStore as well as the nodeQueue.

So this bug still haven't fixed in this PR now? :)

My-pleasure:
> > I found another problem, In the func `MessageWriteLoop`(https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/handler/messagehandler.go#L420), if there is no websocket connection or func `sendMsg` returns nil(etc. the message fails to be sent after five attempts), the key in the nodeQueue will be removed directly, but the message will not be removed from the nodeStore. Althogh the SyncController keeps sending messages to the CloudHub, due to a judgment in the func `addMessageToQueue`(https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/channelq/channelq.go#L125), it returns directly instead of adding it to the nodeQueue. Which results in, the first resource creation message still cannot be delivered to the edgenode. So, I think we should delete the message from nodeStore as well as the nodeQueue.
> 
> So this bug still haven't fixed in this PR now? :)

To fix this bug, I think we should hold the key in nodeQueue. However, this will cause MessageWriteLoop Goroutine to continue working even if the WebSocket connection no longer exists, affecting CloudHub performance.
So I would like to add a sync.cond for every node. And it'll wait when the WebSocket connection is not exist, util the edgenode connects with the CloudHub.
```
func (mh *MessageHandle) MessageWriteLoop(info *model.HubInfo, stopServe chan ExitCode) {
    ...
    for {
        for conn, ok := mh.nodeConns.Load(info.NodeID); !ok; {
            c, _ := mh.nodeCond.Load(info.NodeID)
            c.(*sync.Cond).Wait()
        }
        key, quit := nodeQueue.Get()
        ...
    }
}

func (mh *MessageHandle) OnRegister(connection conn.Connection) {
        ...
	if _, ok := mh.nodeCond.Load(nodeID); !ok {
		mh.nodeCond.Store(nodeID, sync.NewCond(&sync.Mutex{}))
	}

	if _, ok := mh.nodeRegistered.Load(nodeID); ok {
		if conn, exist := mh.nodeConns.Load(nodeID); exist {
			if err := conn.(hubio.CloudHubIO).Close(); err != nil {
				klog.Errorf("failed to close connection %v, err is %v", conn, err)
			}
		}
		mh.nodeConns.Store(nodeID, io)
		c, _ := mh.nodeCond.Load(nodeID)
		c.(*sync.Cond).Signal()
		return
	}
	mh.nodeConns.Store(nodeID, io)
	go mh.ServeConn(&model.HubInfo{ProjectID: projectID, NodeID: nodeID})
}
```
How about this idea? And should we talk about it in another issue?

My-pleasure:
How about my idea that adding a sync.Cond for every websocket connection?   /PTAL @fisherxu @zc2638 
First, sync.Cond can stop `MessageWriteLoop` when the websocket connection is broken and let `MessageWriteLoop` continue to work only when the websocket connection is reconnected.
Second, it can solve another problem which i said before.

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -108,19 +116,39 @@ func (q *ChannelMessageQueue) addMessageToQueue(nodeID string, msg *beehiveModel
...
+			objectSync, err := q.objectSyncLister.ObjectSyncs(resourceNamespace).Get(objectSyncName)
 			if err == nil && objectSync.Status.ObjectResourceVersion != "" && synccontroller.CompareResourceVersion(msg.GetResourceVersion(), objectSync.Status.ObjectResourceVersion) <= 0 {
 				return
+			} else if err != nil && apierrors.IsNotFound(err) {
+				objectSync := &v1alpha1.ObjectSync{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: objectSyncName,
+					},
+					Spec: v1alpha1.ObjectSyncSpec{
+						ObjectAPIVersion: util.GetMessageAPIVerison(msg),
```
fisherxu, line 135: Seems build failed because of this line :)

My-pleasure: Yes, the error says "GetMessageAPIVerison not declared by package util (typecheck)". But this line  is same as https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/handler/messagehandler.go#L559. And I can build and run it on my computer. I don't know why？

fisherxu: Can you fetch  the latest master branch then rebase to it and then force push, let's see the result.

fisherxu: Just have a check, `GetMessageAPIVerison` in which `Verison` is a typo, and fixed last month, you branch is based a old branch, and the ci will rebase your commit to latest master, so build will failed.

gy95: yes, https://github.com/My-pleasure/kubeedge/blob/fix/reliable-message-delivery/pkg/metaserver/util/util.go#L121 need rebase to the upstream/master

My-pleasure: Ok, I know. But I may have made a mistack when I merge and rebase.... I'll try to fix it tomorrow.

gy95: thanks a lot, we'd better use `git rebase` cmd to rebase to the head, but not using `git merge`, this will make commits too much.

cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -445,30 +466,12 @@ func (mh *MessageHandle) MessageWriteLoop(info *model.HubInfo, stopServe chan Ex
...
-					info.NodeID, dumpMessageMetadata(copyMsg), err.Error())
-				nodeQueue.AddRateLimited(key.(string))
-				time.Sleep(time.Second * 2)
-			}
-			break
+		err := mh.sendMsg(conn.(hubio.CloudHubIO), info, copyMsg, msg, nodeStore)
+		if err != nil {
+			klog.Errorf("Failed to send event to node: %s, affected event: %s, err: %s",
+				info.NodeID, dumpMessageMetadata(copyMsg), err.Error())
+			nodeQueue.Add(key.(string))
```
fisherxu, line 473: I think can we add another judgement here, like:
```
if strings.Contains(err.Error(), "use of closed network connection") {
		mh.nodeConns.Delete(info.NodeID)
                continue
}
```
To avoid print too many err logs when edgenode offline, and the keepalive still haven't detect it. 
And the `use of closed network connection` seems should be a standard out put. Ref: https://github.com/kubeedge/kubeedge/blob/master/vendor/golang.org/x/net/http2/server.go#L674
@My-pleasure

fisherxu: the err logs:
```
E0210 21:14:43.231232 2402450 messagehandler.go:471] Failed to send event to node: edge-node, affected event: id: 2c9c451f-7c21-40f6-bceb-552d8e288d56, parent_id: , group: resource, source: edgecontroller, resource: default/pod/nginx-deployment-7d5bf4bc7-rlgqn, operation: update, err: use of closed network connection
E0210 21:14:45.231488 2402450 messagehandler.go:471] Failed to send event to node: edge-node, affected event: id: 2c9c451f-7c21-40f6-bceb-552d8e288d56, parent_id: , group: resource, source: edgecontroller, resource: default/pod/nginx-deployment-7d5bf4bc7-rlgqn, operation: update, err: use of closed network connection
E0210 21:14:47.231661 2402450 messagehandler.go:471] Failed to send event to node: edge-node, affected event: id: 2c9c451f-7c21-40f6-bceb-552d8e288d56, parent_id: , group: resource, source: edgecontroller, resource: default/pod/nginx-deployment-7d5bf4bc7-rlgqn, operation: update, err: use of closed network connection
E0210 21:14:49.231859 2402450 messagehandler.go:471] Failed to send event to node: edge-node, affected event: id: 2c9c451f-7c21-40f6-bceb-552d8e288d56, parent_id: , group: resource, source: edgecontroller, resource: default/pod/nginx-deployment-7d5bf4bc7-rlgqn, operation: update, err: use of closed network connection
E0210 21:14:51.232059 2402450 messagehandler.go:471] Failed to send event to node: edge-node, affected event: id: 2c9c451f-7c21-40f6-bceb-552d8e288d56, parent_id: , group: resource, source: edgecontroller, resource: default/pod/nginx-deployment-7d5bf4bc7-rlgqn, operation: update, err: use of closed network connection
E0210 21:14:53.232728 2402450 messagehandler.go:471] Failed to send event to node: edge-node, affected event: id: 2c9c451f-7c21-40f6-bceb-552d8e288d56, parent_id: , group: resource, source: edgecontroller, resource: default/pod/nginx-deployment-7d5bf4bc7-rlgqn, operation: update, err: use of closed network connection
E0210 21:14:55.233382 2402450 messagehandler.go:471] Failed to send event to node: edge-node, affected event: id: 2c9c451f-7c21-40f6-bceb-552d8e288d56, parent_id: , group: resource, source: edgecontroller, resource: default/pod/nginx-deployment-7d5bf4bc7-rlgqn, operation: update, err: use of closed network connection
E0210 21:14:57.234131 2402450 messagehandler.go:471] Failed to send event to node: edge-node, affected event: id: 2c9c451f-7c21-40f6-bceb-552d8e288d56, parent_id: , group: resource, source: edgecontroller, resource: default/pod/nginx-deployment-7d5bf4bc7-rlgqn, operation: update, err: use of closed network connection

```

My-pleasure: Good idea

fisherxu (commented):
/lgtm
/approve

Let's merged it now, blocked so long time :)

fisherxu:
In cloudcore HA mode, I'm wondering whether there're bugs in the current codes. 

In cloudcore HA mode, when the node connect to another cloudcore B, the current cloudcore A will waits here, and the key will still exists in the queue. Now the node is connecting in cloudcore B, will also receive the message from cloud.

When the node disconnects, and re-connect to cloudcore A, I think the obj in queue will send to edge? This is the dirty data. @My-pleasure

My-pleasure:
Yes, there is indeed dirty data generation. I'll find a way to solve it later.

fisherxu:
> Yes, there is indeed dirty data generation. I'll find a way to solve it later.

Okey, let's find a solution to resolve it.

PR #3255, opened 2021-10-29
fix not use return values when error occurs

gy95:

**What type of PR is this?**

when calling list interface failed, we should not use the return values. so just return or continue to the next loop

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/synccontroller/synccontroller.go
+++ b/cloud/pkg/synccontroller/synccontroller.go
@@ -107,12 +107,14 @@ func (sctl *SyncController) reconcile() {
 	allClusterObjectSyncs, err := sctl.clusterObjectSyncLister.List(labels.Everything())
 	if err != nil {
 		klog.Errorf("Filed to list all the ClusterObjectSyncs: %v", err)
+		return
 	}
 	sctl.manageClusterObjectSync(allClusterObjectSyncs)
 
 	allObjectSyncs, err := sctl.objectSyncLister.List(labels.Everything())
 	if err != nil {
 		klog.Errorf("Failed to list all the ObjectSyncs: %v", err)
+		return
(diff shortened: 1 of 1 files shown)


cloud/pkg/synccontroller/synccontroller.go
```diff
@@ -107,12 +107,14 @@ func (sctl *SyncController) reconcile() {
 	allClusterObjectSyncs, err := sctl.clusterObjectSyncLister.List(labels.Everything())
 	if err != nil {
 		klog.Errorf("Filed to list all the ClusterObjectSyncs: %v", err)
+		return
```
zc2638, line 110: Do we need to consider, when `ClusterObjectSync` resource acquisition fails, we should also try to acquire `ObjectSync` for the next step?

fisherxu: Actually from the source code of list, it will never return err if selector is labels.Everything(). And if return err, the ret still can contains some objs, so do we need to return here? Also from the point of code style, this is reasonable to return.

ref: https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/client/listers/reliablesyncs/v1alpha1/clusterobjectsync.go#L51

zc2638: I prefer not to add return, even if the returned object is `nil`, the following code can work well.
What's your opinion?
@fisherxu @gy95

gy95: my concern is that if return error, the return slice result maybe be incomprehensive, and we couldn't use these incorrect values. 
And accoding to the above comment, the err is never meet. and the following code works well,
So we dont need to add return.

PR #3256, opened 2021-10-29
avoid print goroutine stack trace when component start failed

gy95:

**What type of PR is this?**
/kind cleanup

`klog.Fatal` will print error info, including a stack trace of all running goroutines, then calls os.Exit(255).
trace contains too much info, about hundreds of lines, and these trace are helpless for debug why component progress exit abnormally.
But `klog.Exit` will only print logs, and calls os.Exit(1). And dont print stack trace.
And so it's helpful and more easily for us to find the definite reason why component start failed.
So we replace log functions as below
```
klog.Fatal => klog.Exit
klog.Fatalf => klog.Exitf
```

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #3257, opened 2021-10-29
Fix device does not exist in the downstream controller after it is added

TianTianBigWang:

**What type of PR is this?**
/kind bug

After a new device is added, the device Controller on the cloud cannot automatically obtain the device ID. As a result, device events uploaded by edge nodes cannot be processed.
The phenomenon is as follows:
image

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -616,13 +616,13 @@ func (dc *DownstreamController) deviceUpdated(device *v1alpha2.Device) {
 		if isDeviceUpdated(cachedDevice, device) {
 			// if node selector updated delete from old node and create in new node
 			if isNodeSelectorUpdated(cachedDevice.Spec.NodeSelector, device.Spec.NodeSelector) {
-				dc.deviceAdded(device)
 				deletedDevice := &v1alpha2.Device{ObjectMeta: cachedDevice.ObjectMeta,
 					Spec:     cachedDevice.Spec,
 					Status:   cachedDevice.Status,
 					TypeMeta: device.TypeMeta,
 				}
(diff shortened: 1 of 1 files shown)


cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -616,13 +616,13 @@ func (dc *DownstreamController) deviceUpdated(device *v1alpha2.Device) {
...
 			// if node selector updated delete from old node and create in new node
 			if isNodeSelectorUpdated(cachedDevice.Spec.NodeSelector, device.Spec.NodeSelector) {
-				dc.deviceAdded(device)
 				deletedDevice := &v1alpha2.Device{ObjectMeta: cachedDevice.ObjectMeta,
 					Spec:     cachedDevice.Spec,
 					Status:   cachedDevice.Status,
 					TypeMeta: device.TypeMeta,
 				}
 				dc.deviceDeleted(deletedDevice)
+				dc.deviceAdded(device)
```
fisherxu, line 625: Looks like the code here will only be called when the device is updated, and from the pr description, it's related to the device creation?

TianTianBigWang: In our usage scenario, we will create the device first, and then bind the node. Therefore, the device will be added first, but the device is updated due to the binding node, so the code for updating the device will be entered. 

The `deviceAdded` and `deviceDeleted` here both use the **deviceName** to operate `sync.Map`, which leads to the problem that the device cannot be queried after the node is bound.

fisherxu: Sounds good.

fisherxu (commented):
/lgtm
/approve

PR #3262, opened 2021-11-02
support labelselector for metaserver

chenchunxiu:
**What type of PR is this?**
/kind bug
/kind feature

If the network of edge-node is ok, we send /list request with labelSelector, it works ok.
If edge-node is offline, metaserver does not support labelSelector filter, and the full amount of data will be returned.

Fixes #

### DIFF
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/conversion"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
@@ -71,9 +72,8 @@ func (s *store) Get(ctx context.Context, key string, opts storage.GetOptions, ob
 	return nil
 }
 
-//TODO: implement it
 func (s *store) GetToList(ctx context.Context, key string, opts storage.ListOptions, listObj runtime.Object) error {
-	panic("GetToList implement me")
+	return s.List(ctx, key, opts, listObj)
 }
 
 func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, listObj runtime.Object) error {
@@ -103,6 +103,15 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, 
 
 		labelSet := labels.Set(unstrObj.GetLabels())
 		if !opts.Predicate.Label.Matches(labelSet) {
+			continue
+		}
+
+		// only support metadata.name & metadata.namespace
+		fieldSet := fields.Set{
+			"metadata.name":      unstrObj.GetName(),
+			"metadata.namespace": unstrObj.GetNamespace(),
+		}
+		if !opts.Predicate.Field.Matches(fieldSet) {
 			continue
 		}
 
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/conversion"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/apiserver/pkg/storage"
@@ -99,6 +100,18 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, 
 		if err != nil {
 			return err
 		}
+
+		if !opts.Predicate.Label.Empty() {
+			tmpObjLabels := unstrObj.GetLabels()
+			if tmpObjLabels == nil {
+				continue
+			}
+			var objLabels = (labels.Set)(tmpObjLabels)
+			if !opts.Predicate.Label.Matches(objLabels) {
+				continue
+			}
+		}
+
 		unstrList.Items = append(unstrList.Items, unstrObj)
 	}
 	rv := strconv.FormatUint(resp.Revision, 10)
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
@@ -101,15 +101,9 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions, 
 			return err
 		}
 
-		if !opts.Predicate.Label.Empty() {
-			tmpObjLabels := unstrObj.GetLabels()
-			if tmpObjLabels == nil {
-				continue
-			}
-			var objLabels = (labels.Set)(tmpObjLabels)
-			if !opts.Predicate.Label.Matches(objLabels) {
-				continue
-			}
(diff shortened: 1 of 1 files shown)


edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
```diff
@@ -99,6 +100,18 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
...
+		if !opts.Predicate.Label.Empty() {
+			tmpObjLabels := unstrObj.GetLabels()
+			if tmpObjLabels == nil {
+				continue
+			}
+			var objLabels = (labels.Set)(tmpObjLabels)
+			if !opts.Predicate.Label.Matches(objLabels) {
+				continue
+			}
+		}
```
zc2638, line 113: Thanks for the fixing.
Can we use the `Matches` method directly?
```suggestion
		labelSet := labels.Set(unstrObj.GetLabels())
		if !opts.Predicate.Label.Matches(labelSet) {
			continue
		}
```

chenchunxiu: Thanks for the suggestion. Done

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
```diff
@@ -99,6 +100,12 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
 		if err != nil {
 			return err
 		}
+
+		labelSet := labels.Set(unstrObj.GetLabels())
+		if !opts.Predicate.Label.Matches(labelSet) {
+			continue
+		}
```
fisherxu, line 107: Thanks for the fixing, make sense for the `laberselector` bug. And I think we also need to add the field selector here.
We can only add `name` and `namespace` field selector here. Would you like to add them? @chenchunxiu  Or we can add in other PR :) 

Ref: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/selection_predicate.go#L104

chenchunxiu: Done

fisherxu (commented):
/approve
Thanks for the fixing. Leave to @zc2638 for final reviewing.

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
```diff
@@ -99,6 +100,22 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
...
+
+		// TODO: only support metadata.name & metadata.namespace
+		fieldMap := make(map[string]string)
+		fieldMap["metadata.name"] = unstrObj.GetName()
+		if unstrObj.GetNamespace() != "" {
```
zc2638, line 112: Why should we judge that it is not empty, I understand that it should not match if it is empty.

chenchunxiu: Thanks for  your review.  the judgement has been removed.

fisherxu: In K8s, it also has the code to determine whether has the namespace. ref: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/registry/generic/matcher.go#L25

I'm also not sure why need the `hasNamespaceField` here..

chenchunxiu: the judgement is really useless. If  unsupported field-selector is specified, it will return empty whether the judgment is added or not.
In k8s, field selector not supported error occurs. so here i return  the same error.

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
```diff
@@ -99,6 +101,24 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
 		if err != nil {
 			return err
 		}
+
+		labelSet := labels.Set(unstrObj.GetLabels())
+		if !opts.Predicate.Label.Matches(labelSet) {
+			continue
+		}
+
+		// TODO: only support metadata.name & metadata.namespace
```
zc2638, line 110: As a remark, should `TODO` be removed ?

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
```diff
@@ -99,6 +101,24 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
...
+		if _, found := opts.Predicate.Field.RequiresExactMatch("metadata.namespace"); found && unstrObj.GetNamespace() == "" {
+			return errors.New("field label not supported: metadata.namespace")
+		}
+		fieldMap := make(map[string]string)
```
zc2638, line 114:
```suggestion
		fieldMap := make(field.Set)
```

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
```diff
@@ -99,6 +101,24 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
...
 			return err
 		}
+
+		labelSet := labels.Set(unstrObj.GetLabels())
+		if !opts.Predicate.Label.Matches(labelSet) {
+			continue
+		}
+
+		// TODO: only support metadata.name & metadata.namespace
+		if _, found := opts.Predicate.Field.RequiresExactMatch("metadata.namespace"); found && unstrObj.GetNamespace() == "" {
```
zc2638, line 111: `opts.Predicate.Field.RequiresExactMatch("metadata.namespace")`
I think it is more appropriate to put this logic before the loop processing.

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
zc2638: @fisherxu What do you think about it?

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
chenchunxiu: Done

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
chenchunxiu: Done

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
chenchunxiu: Done

zc2638:
/lgtm
/hold
wait for @fisherxu

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
```diff
@@ -93,12 +95,31 @@ func (s *store) List(ctx context.Context, key string, opts storage.ListOptions,
...
 		}
+
+		labelSet := labels.Set(unstrObj.GetLabels())
+		if !opts.Predicate.Label.Matches(labelSet) {
+			continue
+		}
+
+		// only support metadata.name & metadata.namespace
+		if nsFound && unstrObj.GetNamespace() == "" {
+			return errors.New("field label not supported: metadata.namespace")
```
fisherxu, line 113: The err message above seems indicates that we don't support `metadata.namespace` field selector...

I think there's no need to return error here? If obj has no namespace, do the unstrObj.GetNamespace() will return err? If don't return err, I think we can remove the error above and just do the `!opts.Predicate.Field.Matches(fieldSet)` directly. 

If no matched obj, return the empty list, how do you think? @zc2638 @chenchunxiu

chenchunxiu: If obj has no namespace, do the unstrObj.GetNamespace() will not return err. And I remove the error here.

fisherxu:
/hold cancel
We can add some tests in future :)

PR #3263, opened 2021-11-02
Fix: msg from edgecontroller overwrite  by msg which from dynamiccontroller 

chenchunxiu:
If the network of edge-node is OK, we send /list request with labelSelector, it works ok,
If edge-node is offline, metaserver does not support labelSelector filter, and the full amount of data will be returned.

**What type of PR is this?**
/kind bug

somtimes message from edgecontroller will be overwrite by message from dynamiccontroller because they have the same messagekey

Fixes #

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -142,6 +142,9 @@ func getMsgKey(obj interface{}) (string, error) {
 		resourceType, _ := edgemessagelayer.GetResourceType(*msg)
 		resourceNamespace, _ := edgemessagelayer.GetNamespace(*msg)
 		resourceName, _ := edgemessagelayer.GetResourceName(*msg)
+		if msg.Router.Source == modules.DynamicControllerModuleName {
+			return strings.Join([]string{resourceType, resourceNamespace, resourceName, msg.Router.Source}, "/"), nil
+		}
 		return strings.Join([]string{resourceType, resourceNamespace, resourceName}, "/"), nil
 	}
 


cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -142,6 +142,9 @@ func getMsgKey(obj interface{}) (string, error) {
 		resourceType, _ := edgemessagelayer.GetResourceType(*msg)
 		resourceNamespace, _ := edgemessagelayer.GetNamespace(*msg)
 		resourceName, _ := edgemessagelayer.GetResourceName(*msg)
+		if msg.Router.Source == modules.DynamicControllerModuleName {
+			return strings.Join([]string{resourceType, resourceNamespace, resourceName, msg.Router.Source}, "/"), nil
+		}
 		return strings.Join([]string{resourceType, resourceNamespace, resourceName}, "/"), nil
```
fisherxu, line 148: Thanks for the fixing, and could you please provide more details for this bug? Because we have the deduplication mechanism in cloudhub, so it will has no problem.

fisherxu:
cc @My-Pleasure

cloud/pkg/cloudhub/channelq/channelq.go
chenchunxiu: Thanks for the review.
[the deduplication mechanism in cloudhub] use the message key to save the message in queue,  the newer message(revision is newer) will overwrite the older message， and only keep the newest message which has the newest revision.

The message key cannot distinguish whether the message comes from edgecontroller or dynamiccontroller. 
If downstream(edgecontroller) watched a new resource be deleted, cloudhub store the message (msg1: for delete) which to be sent to the edge.
If the cloud has not yet processed msg1, at this time the edge query the same resource from metaserver, the dynamic controller will generate a new msg2 (message key is the same to msg1). 
At this time, msg2 will overwrite msg1. The edge also cannot receive msg1(msg1: for delete)

My-pleasure: Your worry will not happen, because the operation of that occurs in dynamiccontroller is `ApplicationResp`(see https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/dynamiccontroller/application/application.go#L547). And the msg will be judged as a list message when it enters the channel of CloudHub(see https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/channelq/channelq.go#L166).
You say `getMsgKey` is used in Message Channel, and the msg from dynamiccontroller will send to List Message Channel. They will not affect each other.

chenchunxiu: Thanks for your reply. 
msg from dynamiccontroller/ edgecontroller has different key,because they has different type.(pod or pods)
i will close this pr later.

PR #3267, opened 2021-11-04
cleanup and simplify keadm code

gy95:

**What type of PR is this?**
/kind cleanup

keadm is only for terminal users, so we don't need to use `klog` package to record logs, just need to print infos to the std out terminal.
New Command functions input parameters always std.out and nil pointers, so don't need input parameters, and just remove these.

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #3268, opened 2021-11-04
add bind-address config option and remove address in websocket/quic/https

### LINKED ISSUE #1642: Add advertise address for cloudcore
We have added the feature of certificate issuance in kubeedge, and now we need to add a filed `advertise-address` to specify the service address of the cloudcore, which will be used in the SANs of the cloudcore certificate.

And now in cloudhub, websocket/quic all have [Address](https://github.com/kubeedge/kubeedge/blob/master/pkg/apis/componentconfig/cloudcore/v1alpha1/types.go#L117) field, when we add the `advertise -address`, is the `Address` in websocket/quic still needed?

The `advertise -address` will be a specific address like 1.2.3.4, but  `Address` in websocket/quic now default as `0.0.0.0`.

/cc @kevin-wangzefeng @kadisi @chendave @subpathdev  @GsssC @ls889 @XJangel 

Task List:
- [x] Add advertise address in cloudcore/cloudhub #1643
- [x] Update the docs for the usage of advertise address #1656 
- [ ] Add bind-address config option and remove address in websocket/quic/httpserver
- [x] Support config domain URL for cloudcore #2126

gy95:

**What type of PR is this?**
/kind feature

websocket/quic/https configures of cloudhub all have address config, so combine these to one config bind-address
add bind-address config option and remove address in websocket/quic/https

part of #1642

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/servers/httpserver/server.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/server.go
@@ -44,7 +44,12 @@ func StartHTTPServer() {
 	router.HandleFunc(constants.DefaultCertURL, edgeCoreClientCert).Methods(http.MethodGet)
 	router.HandleFunc(constants.DefaultCAURL, getCA).Methods(http.MethodGet)
 
-	addr := fmt.Sprintf("%s:%d", hubconfig.Config.BindAddress, hubconfig.Config.HTTPS.Port)
+	var addr string
+	if hubconfig.Config.BindAddress != "" {
+		addr = fmt.Sprintf("%s:%d", hubconfig.Config.BindAddress, hubconfig.Config.HTTPS.Port)
+	} else {
+		addr = fmt.Sprintf("%s:%d", hubconfig.Config.HTTPS.Address, hubconfig.Config.HTTPS.Port)
+	}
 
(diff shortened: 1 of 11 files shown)


TianTianBigWang:
If bindAddress has two addresses, such as the intranet IP of my node: 10.222.111.111, and the local loopback address 127.0.0.1, will it cause a port conflict and the service cannot be started?

gy95:
> If bindAddress has two addresses, such as the intranet IP of my node: 10.222.111.111, and the local loopback address 127.0.0.1, will it cause a port conflict and the service cannot be started?

oh, yes.. 😞

fisherxu (commented):
This change will break the compatible, maybe need to add the bind addr and keep the old addr for several releases.

gy95:
> This change will break the compatible, maybe need to add the bind addr and keep the old addr for several releases.

feels that it has no influence on the compatibility 😂 could you please explain the reason more in deep :)

gy95:
> > > This change will break the compatible, maybe need to add the bind addr and keep the old addr for several releases.
> > 
> > 
> > feels that it has no influence on the compatibility 😂 could you please explain the reason more in deep :)
> 
> Users use the config file in old version to the new releasae, will not work.

got it, I see.

PR #3277, opened 2021-11-08
change kubeletEndpoint port when cloudstream conneted

chenchunxiu:
Signed-off-by: chenchunxiu <chenchunxiu_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind bug

If cloudcore is multi-active (cloudcore1, cloudcore2, cloudcore3) and LB is used.
In bellowing case, kubectl exec/logs failed.

LB forward connection of cloudhub (from edge-node-01) to cloudcore1.    --- cloudhub port 10000
LB forward connection of cloudstream (from edge-node-01) to cloudcore2. --- cloudstream port 10004 

So, we should change kubeletEndpoint when cloudstrem conneted, not cloudhub connected.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudstream/tunnelserver.go
+++ b/cloud/pkg/cloudstream/tunnelserver.go
@@ -23,14 +23,15 @@ import (
 	"encoding/pem"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/emicklei/go-restful"
 	"github.com/gorilla/websocket"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog/v2"
 
@@ -38,6 +39,12 @@ import (
 	streamconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudstream/config"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/pkg/stream"
+)
+
+const (
+	// The amount of time the tunnelserver should sleep between retrying node status updates
+	retrySleepTime        = 20 * time.Second
+	nodeStatusUpdateRetry = 3
 )
 
 type TunnelServer struct {
@@ -178,23 +185,23 @@ func (s *TunnelServer) Start() {
 }
 
 func (s *TunnelServer) updateNodeKubeletEndpoint(nodeName string) {
-	getNode, err := client.GetKubeClient().CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
-	if errors.IsNotFound(err) {
-		klog.Warningf("node %s not found", nodeName)
-		return
-	}
-	if err != nil {
-		klog.Warningf("get node %s failed with error: %s", nodeName, err)
-		return
-	}
+	if err := wait.PollImmediate(retrySleepTime, nodeStatusUpdateRetry, func() (bool, error) {
+		getNode, err := client.GetKubeClient().CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
+		if err != nil {
+			klog.Errorf("Failed while getting a Node to retry updating node KubeletEndpoint Port, node: %s, error: %v", nodeName, err)
+			return false, nil
+		}
 
-	getNode.Status.DaemonEndpoints.KubeletEndpoint.Port = int32(s.tunnelPort)
-	node, err := client.GetKubeClient().CoreV1().Nodes().UpdateStatus(context.Background(), getNode, metav1.UpdateOptions{})
-	if err != nil {
-		klog.Errorf("update node KubeletEndpoint Port failed with error: %s, node: %v, tunnelPort:%v",
-			err, getNode.Name, s.tunnelPort)
-		return
+		getNode.Status.DaemonEndpoints.KubeletEndpoint.Port = int32(s.tunnelPort)
+		_, err = client.GetKubeClient().CoreV1().Nodes().UpdateStatus(context.Background(), getNode, metav1.UpdateOptions{})
+		if err != nil {
+			klog.Errorf("Failed to update node KubeletEndpoint Port, node: %s, tunnelPort: %s, err: %v", nodeName, s.tunnelPort, err)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		klog.Errorf("Update KubeletEndpoint Port of Node '%v' error: %v. ", nodeName, err)
+		os.Exit(1)
 	}
-	klog.Infof("update node KubeletEndpoint Port success. node: %s, tunnelPort:%v->%v", node.Name,
-		getNode.Status.DaemonEndpoints.KubeletEndpoint.Port, node.Status.DaemonEndpoints.KubeletEndpoint.Port)
+	klog.V(4).Infof("Update node KubeletEndpoint Port successfully, node: %s, tunnelPort: %s", nodeName, s.tunnelPort)
 }
--- a/cloud/pkg/cloudstream/tunnelserver.go
+++ b/cloud/pkg/cloudstream/tunnelserver.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cloudstream
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
@@ -28,11 +29,14 @@ import (
 
 	"github.com/emicklei/go-restful"
 	"github.com/gorilla/websocket"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog/v2"
 
 	hubconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/config"
 	streamconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudstream/config"
+	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/pkg/stream"
 )
 
@@ -42,12 +46,14 @@ type TunnelServer struct {
 	sync.Mutex
 	sessions   map[string]*Session
 	nodeNameIP sync.Map
+	tunnelPort int
 }
 
-func newTunnelServer() *TunnelServer {
+func newTunnelServer(tunnelPort int) *TunnelServer {
 	return &TunnelServer{
-		container: restful.NewContainer(),
-		sessions:  make(map[string]*Session),
+		container:  restful.NewContainer(),
+		sessions:   make(map[string]*Session),
+		tunnelPort: tunnelPort,
 		upgrader: websocket.Upgrader{
 			HandshakeTimeout: time.Second * 2,
 			ReadBufferSize:   1024,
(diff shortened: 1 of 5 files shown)


chenchunxiu:
anybody has time to review? @fisherxu

zc2638:
Thanks for fixing.
Can you add more descriptions of this patch?

Is it because the node has been created, and then `LB` needs to be added to the cloud, and the `kubeletEndpoint` needs to be changed at this time?

chenchunxiu:
> Thanks for fixing. Can you add more descriptions of this patch?
> 
> Is it because the node has been created, and then `LB` needs to be added to the cloud, and the `kubeletEndpoint` needs to be changed at this time?

When cloudcore is multi-active and LB is used, there is no guarantee that LB will forward the edgehub connection and edgestream connection of the same node to the same cloudcore.
If LB forwards both edgehub and edgestream connections to the same cloudcore, kubectl logs/exec works normally
image

If LB forwards both edgehub and edgestream connections to different cloudcore, kubectl logs/exec will not work because the kubectl logs/exec request is forwarded to the cloudcore without stream connection.
image

Therefore, we should keep the setting of kubeletEndpointPort consistent with the cloudcore where cloudstream is connected.
image

zc2638:
I have some doubts. 

The original logic is to update the `kubeletEndpoint` when the node is created or the node status is updated.
Now, when the edge node connects to the `cloud stream`, first update the `kubeletEndpoint`.

What is the difference between them? 
What is the difference between processing this logic in different places?

zc2638:
For edge nodes, only the address of LB.
When `edgestream` connects to the corresponding `cloudstream`, confirm the proxy port of the cloud.

e.g. maybe `edgestream` connects to `cloudstream A`, so we update `kubeletEndpoint` to 10351. 
e.g. maybe `edgestream` connects to `cloudstream B`, so we update `kubeletEndpoint` to 10352. 

This is my understanding of your explanation, is that the case?
@chenchunxiu

kevin-wangzefeng:
This is caused by inconsistent load-balancing on multiple requests from a some edge node.
To quick fix, you may enable features like "sticky session", or "route by source IP" provided by the LB.

fisherxu:
> For edge nodes, only the address of LB. When `edgestream` connects to the corresponding `cloudstream`, confirm the proxy port of the cloud.
> 
> e.g. maybe `edgestream` connects to `cloudstream A`, so we update `kubeletEndpoint` to 10351. e.g. maybe `edgestream` connects to `cloudstream B`, so we update `kubeletEndpoint` to 10352.
> 
> This is my understanding of your explanation, is that the case? @chenchunxiu

CloudStream/EdgeStream use one connection, and the cloudhub/edgehub use another connection. The two connections may on two servers behide LB, and we use the cloud/edge hub to update the nodestatus, so CloudStream/EdgeStream will not work. @zc2638

fisherxu (commented):
Looks good, please resolve the confict :) @chenchunxiu

chenchunxiu:
> Looks good, please resolve the confict :) @chenchunxiu

Done

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -169,3 +176,25 @@ func (s *TunnelServer) Start() {
...
+	}
+	if err != nil {
+		klog.Warningf("get node %s failed with error: %s", nodeName, err)
+		return
+	}
+
+	getNode.Status.DaemonEndpoints.KubeletEndpoint.Port = int32(s.tunnelPort)
+	node, err := client.GetKubeClient().CoreV1().Nodes().UpdateStatus(context.Background(), getNode, metav1.UpdateOptions{})
+	if err != nil {
+		klog.Errorf("update node KubeletEndpoint Port failed with error: %s, node: %v, tunnelPort:%v",
```
fisherxu, line 194: We should have some try then exit directly if err occurs in this func. @chenchunxiu 
Ref:  https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L792

chenchunxiu: Done

chenchunxiu:
ping  @fisherxu @zc2638

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -169,3 +184,30 @@ func (s *TunnelServer) Start() {
...
+	if err := wait.PollImmediate(retrySleepTime, nodeStatusUpdateRetry, func() (bool, error) {
+		getNode, err := client.GetKubeClient().CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
+		if errors.IsNotFound(err) {
+			klog.Warningf("node %s not found", nodeName)
+			return false, nil
+		}
+		if err != nil {
+			klog.Errorf("Failed while getting a Node to retry updating node KubeletEndpoint Port. error: %v", nodeName, err)
+			return false, err
+		}
```
fisherxu, line 198:
```suggestion
		if err != nil {
			klog.Errorf("Failed while getting a Node to retry updating node KubeletEndpoint Port, node: %s, error: %v", nodeName, err)
			return false, nil
		}
```

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -169,3 +184,30 @@ func (s *TunnelServer) Start() {
...
+
+		getNode.Status.DaemonEndpoints.KubeletEndpoint.Port = int32(s.tunnelPort)
+		_, err = client.GetKubeClient().CoreV1().Nodes().UpdateStatus(context.Background(), getNode, metav1.UpdateOptions{})
+		if err != nil {
+			klog.Errorf("update node KubeletEndpoint Port failed with error: %s, node: %v, tunnelPort: %v",
```
fisherxu, line 203:
```suggestion
			klog.Errorf("Failed to update node KubeletEndpoint Port, node: %s, tunnelPort: %s, err: %v", getNode.Name, s.tunnelPort, err)
```

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -169,3 +184,30 @@ func (s *TunnelServer) Start() {
...
+				err, getNode.Name, s.tunnelPort)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		klog.Errorf("Update KubeletEndpoint Port of Node '%v' error: %v. ", nodeName, err)
+		os.Exit(1)
+	}
+	klog.Infof("update node KubeletEndpoint Port success. node: %s, tunnelPort: %v", nodeName, s.tunnelPort)
```
fisherxu, line 212:
```suggestion
	klog.V(2).Infof("Update node KubeletEndpoint Port successfully, node: %s, tunnelPort: %s", nodeName, s.tunnelPort)
```

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -169,3 +184,30 @@ func (s *TunnelServer) Start() {
...
+		getNode.Status.DaemonEndpoints.KubeletEndpoint.Port = int32(s.tunnelPort)
+		_, err = client.GetKubeClient().CoreV1().Nodes().UpdateStatus(context.Background(), getNode, metav1.UpdateOptions{})
+		if err != nil {
+			klog.Errorf("update node KubeletEndpoint Port failed with error: %s, node: %v, tunnelPort: %v",
+				err, getNode.Name, s.tunnelPort)
+			return false, nil
+		}
+		return true, nil
+	}); err != nil {
+		klog.Errorf("Update KubeletEndpoint Port of Node '%v' error: %v. ", nodeName, err)
```
fisherxu, line 209:
```suggestion
		klog.Errorf("Failed to update node KubeletEndpoint Port, node: %s, tunnelPort: %s, err: %v", nodeName, s.tunnelPort, err)
```

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -17,37 +17,51 @@ limitations under the License.
...
 
 	hubconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/config"
 	streamconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudstream/config"
+	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/pkg/stream"
 )
 
+const (
+	// The amount of time the tunnelserver should sleep between retrying node status updates
+	retrySleepTime        = 20 * time.Millisecond
```
fisherxu, line 47:
```suggestion
	retrySleepTime        = 20 * time.Second
```
The retry time can be 20s here, if there is a delay in edge node creation, here need to wait more.

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -169,3 +184,30 @@ func (s *TunnelServer) Start() {
 		return
 	}
 }
+
+func (s *TunnelServer) updateNodeKubeletEndpoint(nodeName string) {
+	if err := wait.PollImmediate(retrySleepTime, nodeStatusUpdateRetry, func() (bool, error) {
```
fisherxu, line 189:
```suggestion
	if err := wait.Poll(retrySleepTime, nodeStatusUpdateRetry, func() (bool, error) {
```
Let's use `wait.Poll` here, PollImmediate will exit when err occur, here we need to retry.

chenchunxiu: Thanks for the reply. 
I test in my machine, if err happens retry will execute.
`wait.PollImmediate` will run the func immediately, and  `wait.Poll` will wait for the interval firstly.

fisherxu: Okey, sounds good, and so we need to return `false,nil` as I comment above. Or the retry will exit directly.

cloud/pkg/cloudstream/tunnelserver.go
fisherxu: V(4) maybe more fine, or large scale edge nodes will print lots of success logs.

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -17,37 +17,51 @@ limitations under the License.
...
+	nodeStatusUpdateRetry = 3
+)
+
 type TunnelServer struct {
 	container *restful.Container
 	upgrader  websocket.Upgrader
 	sync.Mutex
 	sessions   map[string]*Session
 	nodeNameIP sync.Map
+	tunnelPort int
```
gy95, line 57: my question is that this `tunnelPort` is from the configuration? and need to be configured by users themselves?

chenchunxiu: No,  users do not need to configure manually.
tunnelPort will be automatically generated when cloudcore is started at first time and saved in configmap. It will be read directly from configmap the next time cloudcore started.

cloud/pkg/cloudstream/tunnelserver.go
chenchunxiu: Done

cloud/pkg/cloudstream/tunnelserver.go
chenchunxiu: Done

cloud/pkg/cloudstream/tunnelserver.go
chenchunxiu: Done

cloud/pkg/cloudstream/tunnelserver.go
chenchunxiu: Done

cloud/pkg/cloudstream/tunnelserver.go
chenchunxiu: Thanks for the reply, I have changed the `return false, err` to `return false, nil`.

cloud/pkg/cloudstream/tunnelserver.go
chenchunxiu: Done

fisherxu (commented):
/lgtm
/approve
Sorry for the delay.

PR #3278, opened 2021-11-08
devicetwin code cleanup

TianTianBigWang:

**What type of PR is this?**
/kind cleanup

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/communicate.go
+++ b/edge/pkg/devicetwin/dtmanager/communicate.go
@@ -153,17 +153,16 @@ func (cw CommWorker) checkConfirm(context *dtcontext.DTContext, msg interface{})
 		dtmsg, ok := value.(*dttype.DTMessage)
 		klog.V(2).Info("has msg")
 		if !ok {
-
-		} else {
-			klog.V(2).Info("redo task due to no recv")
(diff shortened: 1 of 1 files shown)


edge/pkg/devicetwin/dtmanager/communicate.go
```diff
@@ -152,13 +152,10 @@ func (cw CommWorker) checkConfirm(context *dtcontext.DTContext, msg interface{})
 	context.ConfirmMap.Range(func(key interface{}, value interface{}) bool {
 		dtmsg, ok := value.(*dttype.DTMessage)
 		klog.V(2).Info("has msg")
-		if !ok {
-
-		} else {
```
fisherxu, line 157: If not ok, do we need to print err log and return false? :)

zc2638: Maybe `warining` log?

TianTianBigWang: The log can be printed, but there should be no need to return false. Because the confirmMap may still have some messages that are normally waiting to be processed.

fisherxu: /approve

PR #3279, opened 2021-11-08
Still allowed delete rule when source not exist

lvchenggang:
What type of PR is this?
/kind feature

When router module monitor the rule delete event, if through getSourceOfRule method not get the source of the rule, we should skip UnregisterListener stage.
But the operation which delete rule from cache named "rules" shoud be allowed.

NONE

### DIFF
--- a/cloud/pkg/router/rule/rule.go
+++ b/cloud/pkg/router/rule/rule.go
@@ -147,14 +147,16 @@ func delRule(namespace, name string) {
 	rule := v.(*routerv1.Rule)
 
 	source, err := getSourceOfRule(rule)
-	if err != nil {
-		klog.Error(err)
-		return
+	// if source not exist, skip UnregisterListener
+	if err == nil {
+		klog.V(4).Infof("delRule: source of rule:%s exist, do UnregisterListener", rule.Spec.Source)
+		source.UnregisterListener()
+	} else {
+		klog.Warningf("delRule: source of rule:%s not exist, unnecessary do UnregisterListener:%v", rule.Spec.Source, err)
 	}
-	source.UnregisterListener()
 
 	rules.Delete(ruleKey)
-	klog.Infof("delete rule success: %s", ruleKey)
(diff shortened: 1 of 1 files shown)


cloud/pkg/router/rule/rule.go
```diff
@@ -147,11 +147,13 @@ func delRule(namespace, name string) {
 	rule := v.(*routerv1.Rule)
 
 	source, err := getSourceOfRule(rule)
-	if err != nil {
-		klog.Error(err)
-		return
+	// if source not exist, skip UnregisterListener
+	if err == nil {
+		klog.Infof("delRule: source of rule:%s exist, do UnregisterListener", rule.Spec.Source)
```
fisherxu, line 152:
```suggestion
		klog.V(4).Infof("delRule: source of rule:%s exist, do UnregisterListener", rule.Spec.Source)
```

cloud/pkg/router/rule/rule.go
```diff
@@ -147,11 +147,13 @@ func delRule(namespace, name string) {
...
+		source.UnregisterListener()
+	} else {
+		klog.Warningf("delRule: source of rule:%s not exist, unnecessary do UnregisterListener:%v", rule.Spec.Source, err)
 	}
-	source.UnregisterListener()
 
 	rules.Delete(ruleKey)
 	klog.Infof("delete rule success: %s", ruleKey)
```
fisherxu, line 159:
```suggestion
	klog.V(4).Infof("delete rule success: %s", ruleKey)
```

lvchenggang: Accept and complete modification.

cloud/pkg/router/rule/rule.go
lvchenggang: Accept and complete modification.

fisherxu (commented):
/lgtm
/approve

PR #3287, opened 2021-11-10
Add require rule and ruleendpoint resource from api

lvchenggang:

What type of PR is this?
/kind feature

There exist a situation when rule and ruleendpoint crd resource create after cloudcore start, the cache in router module cannot save the rule and ruleendpoint resource where k8s subsequently create.
So this patch add getResourceFromApi method to solve this problem.

NONE

### DIFF
--- a/cloud/pkg/router/constants/constants.go
+++ b/cloud/pkg/router/constants/constants.go
@@ -5,4 +5,7 @@ const (
 	EventbusProvider   string = "eventbus"
 	GroupResource      string = "resource"
 	ServicebusProvider string = "servicebus"
+
+	ResourceTypeRules         string = "rules"
+	ResourceTypeRuleEndpoints string = "ruleendpoints"
 )
--- a/cloud/pkg/router/rule/rule.go
+++ b/cloud/pkg/router/rule/rule.go
@@ -1,15 +1,19 @@
 package rule
 
 import (
+	"context"
 	"fmt"
 	"sync"
 	"time"
 
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 
 	"github.com/kubeedge/beehive/pkg/core/model"
 	routerv1 "github.com/kubeedge/kubeedge/cloud/pkg/apis/rules/v1"
+	keclient "github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/modules"
+	routerConst "github.com/kubeedge/kubeedge/cloud/pkg/router/constants"
 	"github.com/kubeedge/kubeedge/cloud/pkg/router/listener"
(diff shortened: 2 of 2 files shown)


TianTianBigWang:
This looks like the content of the informer synchronization mechanism, why not use the crd informer of k8s to synchronize the rules and endpoints created subsequently?

lvchenggang:
> This looks like the content of the informer synchronization mechanism, why not use the crd informer of k8s to synchronize the rules and endpoints created subsequently?

Well, use informer synchronization mechanism is a method, but the amount of code will expand a lot.

fisherxu (commented):
> There exist a situation when rule and ruleendpoint crd resource create after cloudcore start, the cache in router module cannot save the rule and ruleendpoint resource where k8s subsequently create.

If cloudcore restart or start after the crd created, edgecontroller will list all the obj and send to router, so in theory router can get all the crds, do we have any tests for this situation?

lvchenggang:
> > There exist a situation when rule and ruleendpoint crd resource create after cloudcore start, the cache in router module cannot save the rule and ruleendpoint resource where k8s subsequently create.
> 
> If cloudcore restart or start after the crd created, edgecontroller will list all the obj and send to router, so in theory router can get all the crds, do we have any tests for this situation?

The situation has been verified that if cloudcore restart or start after the crd created, edgecontroller will list all the obj and send to router, and could get all the crds.
But the restart of cloudcore can not be a frequent operation in production environment, so this patch solution the potential problem until next upgrade in prod env.

fisherxu:
> > > There exist a situation when rule and ruleendpoint crd resource create after cloudcore start, the cache in router module cannot save the rule and ruleendpoint resource where k8s subsequently create.
> > 
> > 
> > If cloudcore restart or start after the crd created, edgecontroller will list all the obj and send to router, so in theory router can get all the crds, do we have any tests for this situation?
> 
> The situation has been verified that if cloudcore restart or start after the crd created, edgecontroller will list all the obj and send to router, and could get all the crds. But the restart of cloudcore can not be a frequent operation in production environment, so this patch solution the potential problem until next upgrade in prod env.

We can test it in the test env  :)

lvchenggang:
> > > > There exist a situation when rule and ruleendpoint crd resource create after cloudcore start, the cache in router module cannot save the rule and ruleendpoint resource where k8s subsequently create.
> > > 
> > > 
> > > If cloudcore restart or start after the crd created, edgecontroller will list all the obj and send to router, so in theory router can get all the crds, do we have any tests for this situation?
> > 
> > 
> > The situation has been verified that if cloudcore restart or start after the crd created, edgecontroller will list all the obj and send to router, and could get all the crds. But the restart of cloudcore can not be a frequent operation in production environment, so this patch solution the potential problem until next upgrade in prod env.
> 
> We can test it in the test env :)

Yes, this sence can be prevent in advance.

PR #3292, opened 2021-11-11
Code improve: ticker should be stopped

hainesc:
**What type of PR is this?**
/kind cleanup

Code improvement

Fixes #

```release-note
No change
```

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -138,12 +138,10 @@ func environmentCheck() error {
 	}
 
 	for _, process := range processes {
-		// if kubelet is running, return error
-		if process.Executable() == "kubelet" {
+		switch process.Executable() {
+		case "kubelet": // if kubelet is running, return error
 			return errors.New("kubelet should not running on edge node when running edgecore")
-		}
-		// if kube-proxy is running, return error
-		if process.Executable() == "kube-proxy" {
+		case "kube-proxy": // if kube-proxy is running, return error
 			return errors.New("kube-proxy should not running on edge node when running edgecore")
 		}
 	}
--- a/edge/pkg/edgestream/edgestream.go
+++ b/edge/pkg/edgestream/edgestream.go
@@ -73,23 +73,27 @@ func (e *edgestream) Start() {
 		Path:   "/v1/kubeedge/connect",
 	}
 	// TODO: Will improve in the future
-	ok := <-edgehub.HasTLSTunnelCerts
-	if ok {
-		cert, err := tls.LoadX509KeyPair(config.Config.TLSTunnelCertFile, config.Config.TLSTunnelPrivateKeyFile)
-		if err != nil {
-			klog.Exitf("Failed to load x509 key pair: %v", err)
-		}
-		tlsConfig := &tls.Config{
-			InsecureSkipVerify: true,
-			Certificates:       []tls.Certificate{cert},
-		}
+	if ok := <-edgehub.HasTLSTunnelCerts; !ok {
+		klog.Exitf("Failed to find cert key pair")
+	}
 
-		for range time.NewTicker(time.Second * 2).C {
-			select {
-			case <-beehiveContext.Done():
-				return
-			default:
-			}
+	cert, err := tls.LoadX509KeyPair(config.Config.TLSTunnelCertFile, config.Config.TLSTunnelPrivateKeyFile)
+	if err != nil {
+		klog.Exitf("Failed to load x509 key pair: %v", err)
+	}
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: true,
+		Certificates:       []tls.Certificate{cert},
+	}
+
+	ticker := time.NewTicker(time.Second * 2)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-beehiveContext.Done():
+			return
+		case <-ticker.C:
 			err := e.TLSClientConnect(serverURL, tlsConfig)
 			if err != nil {
 				klog.Errorf("TLSClientConnect error %v", err)


hainesc:
ping @daixiang0 @zc2638 PTAL

edge/cmd/edgecore/app/server.go
```diff
@@ -138,12 +138,10 @@ func environmentCheck() error {
...
 	for _, process := range processes {
-		// if kubelet is running, return error
-		if process.Executable() == "kubelet" {
+		switch process.Executable() {
+		case "kubelet": // if kubelet is running, return error
 			return errors.New("kubelet should not running on edge node when running edgecore")
-		}
-		// if kube-proxy is running, return error
-		if process.Executable() == "kube-proxy" {
+		case "kube-proxy": // if kube-proxy is running, return error
```
zc2638, line 144: There are only these two situations, I think there is no need to use `switch` to extend the judgment.

edge/pkg/edgestream/edgestream.go
```diff
@@ -73,27 +73,32 @@ func (e *edgestream) Start() {
...
-		if err != nil {
-			klog.Exitf("Failed to load x509 key pair: %v", err)
-		}
-		tlsConfig := &tls.Config{
-			InsecureSkipVerify: true,
-			Certificates:       []tls.Certificate{cert},
-		}
+	if ok := <-edgehub.HasTLSTunnelCerts; !ok {
+		klog.Exitf("Failed to write cert key pair into disk")
+	}
```
zc2638, line 78: What needs to be achieved here is the processing without the certificate, can you try it?

edge/pkg/edgestream/edgestream.go
```diff
@@ -73,27 +73,32 @@ func (e *edgestream) Start() {
...
+
+	ticker := time.NewTicker(time.Second * 2)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-beehiveContext.Done():
+			return
+		case <-ticker.C:
```
zc2638, line 96: `time.After` is it more comfortable?

edge/cmd/edgecore/app/server.go
hainesc: Yes, but for each process in processes, the former calls process.Executable() twice while the right only calls once.

edge/pkg/edgestream/edgestream.go
hainesc: No, I don't think so. time.After does not work that way, it not a ticker.

hainesc: Every time you call time.After, it new a Timer object, ticker is the standard solution here.

edge/pkg/edgestream/edgestream.go
hainesc: The happy path is left-aligned, so log and exit when ok is not true.

hainesc: What should we do if no certs? http instead of https?

fisherxu: Should exits the process, I think you are right.

edge/cmd/edgecore/app/server.go
zc2638:
```golang
func (p *UnixProcess) Executable() string {
	return p.binary
}
```
This method directly returns the structure field without additional processing logic.

edge/pkg/edgestream/edgestream.go
zc2638:
```golang
func (eh *EdgeHub) Start() {
    ...
    HasTLSTunnelCerts <- true
    close(HasTLSTunnelCerts)
```
In the edgehub module, it is hard-coded, and this channel will only receive `true` forever.

zc2638: Like `TLSClientConnect` method.
You can try to implement a `ws` connection instead of a secure `wss` connection.
However, we need to consider whether non secure connections are necessary?
@hainesc @fisherxu

fisherxu: It's the cloud-edge connection, I think secure wss is needed. :)

edge/pkg/edgestream/edgestream.go
```diff
@@ -73,27 +73,32 @@ func (e *edgestream) Start() {
...
 			err := e.TLSClientConnect(serverURL, tlsConfig)
 			if err != nil {
 				klog.Errorf("TLSClientConnect error %v", err)
 			}
+
```
fisherxu, line 101: Please remove this line to fix the failed CI test.

edge/cmd/edgecore/app/server.go
hainesc: If you think the code is so beautiful, I will quit. We should never talk about the detail of `Executable()` here, it should be transparent to the code what we talk about.

edge/pkg/edgestream/edgestream.go
hainesc: @zc2638 The code here should not dependent the hard code of the channel `HasTLSTunnelCerts`, as @fisherxu says, if we should exits the process if we get `false` from the channel, we should exit here. The hard code is a temporary work and somebody may improve it.

edge/pkg/edgestream/edgestream.go
hainesc: Done

edge/pkg/edgestream/edgestream.go
zc2638: Should we remove the `HasTLSTunnelCerts ` channel? 
Since it must be `true`, I think there is no need to exist here.

fisherxu: Through the exit will not be triggered, but I think it's more reasonable to add  the exit codes. If there are some bugs due to code modification, it can exit here.

But the exit message seems wrong. Should be `Failed to obtain certs`?

zc2638: Since it has been determined that only TLS is supported, we should delete the channel logic for judging TLS.
The following logic is sufficient to determine whether the certificate exists and is valid.
```golang
cert, err := tls.LoadX509KeyPair(config.Config.TLSTunnelCertFile, config.Config.TLSTunnelPrivateKeyFile)
if err != nil {
    klog.Exitf("Failed to load x509 key pair: %v", err)
}
```

zc2638: Since it has been determined that only TLS is supported, we should delete the channel logic for judging TLS.
The following logic is sufficient to determine whether the certificate exists and is valid.
```golang
cert, err := tls.LoadX509KeyPair(config.Config.TLSTunnelCertFile, config.Config.TLSTunnelPrivateKeyFile)
if err != nil {
    klog.Exitf("Failed to load x509 key pair: %v", err)
}
```

hainesc: I agree with @fisherxu

hainesc: @fisherxu exit message modified.

edge/cmd/edgecore/app/server.go
fisherxu: Switch and if all fine to me :)

edge/pkg/edgestream/edgestream.go
fisherxu: > Since it has been determined that only TLS is supported, we should delete the channel logic for judging TLS.
The following logic is sufficient to determine whether the certificate exists and is valid.

No, the certs is generated in edgehub module, so we must wait edgehub completed, then trigger here to load.

fisherxu (commented):
/lgtm
/approve

PR #3294, opened 2021-11-13
[WIP] add nodelease feature gate

### LINKED ISSUE #3086: Use NodeLease when updating heartbeat timestamp of node
When edged starts, a goroutine called [`syncNodeStatus`](https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edged/edged.go#L322) will continuously send msgs to the cloudcore to update nodeStatus every `NodeStatusUpdateFrequency` which defined in `edgecore.yaml` and its default value is 10s.

I found it that `MetaManager` will not send msgs to the cloudcore when the update msgs satisfy [`resourceUnchanged`](https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/metamanager/process.go#L153-L158). Unfortunately, for NodeStatus updating msgs, it's useless, because the `LastHeartbeatTime` field always changes.

I used tcpdump to get msgs sent to cloud. The results is something like this:
Screenshot_20210831_175238_p

At line 1, 4, 10, 16, 21, 27, 33, 36, 42,  we can see a msg with size of 1074 (about 1KB) was sent to the cloudcore (`10.1.11.140` in my cluster), and the interval between them is 10s.

Maybe we can use NodeLease msg to maintain the `LastHeartbeatTime`,  the size of which is just about 0.1KB.  

**Why is this needed**:

For Kubernetes, where a good network environment is assumed, updating NodeStatus every 10s is a heavy job. Kubeedge assumes that the network between edge and cloud is weak and can easily break. In this assumption, sending NodeStatus msgs to the cloud frequently will make the situation even worse.

Congrool:
**What type of PR is this?**
Add a new feature gate `NodeLease` to kubeedge, users can enable the feature gate to use nodelease to update heartbeat rather than using nodeStatus.

As we can see in the #3086, kubeedge used to update heartbeat with nodeStatus which is about 1KB. The frequent transmission of the nodeStatus(deafult every 10s) can result in overload on the weak cloud-edge network. With the `NodeLease` feature gate, we can reduce the heartbeat update packet size down to 0.2KB. I demonstrated it with tcpdump catching the packets sent from edgecore to the cloudcore during 5 min. The tcpdump output is as follows. The red box represents the heartbeat message and the yellow box represents the nodestatus message.

Screenshot_20211113_130541

Also I give the changes of `lastHeartbeatTime` and lease `renewTime` during serval minutes to display their relationship.  
**lastHearbeatTime of node0**  
```bash
[root@master ~]# while true; do
> kubectl get node node0 -o=go-template --template="lastHeartTime: {{ (index .status.conditions 0).lastHeartbeatTime}}" && echo ""
> sleep 10s
> done
lastHeartTime: 2021-11-13T03:48:38Z
lastHeartTime: 2021-11-13T03:48:38Z
lastHeartTime: 2021-11-13T03:48:38Z
lastHeartTime: 2021-11-13T03:48:38Z
lastHeartTime: 2021-11-13T03:49:38Z
lastHeartTime: 2021-11-13T03:49:38Z
lastHeartTime: 2021-11-13T03:49:38Z
lastHeartTime: 2021-11-13T03:49:38Z
lastHeartTime: 2021-11-13T03:49:38Z
lastHeartTime: 2021-11-13T03:49:38Z
lastHeartTime: 2021-11-13T03:50:38Z
lastHeartTime: 2021-11-13T03:50:38Z
lastHeartTime: 2021-11-13T03:50:38Z
lastHeartTime: 2021-11-13T03:50:38Z
lastHeartTime: 2021-11-13T03:50:38Z
lastHeartTime: 2021-11-13T03:50:38Z
```   
**lease renewTime**  
```bash
[root@master ~]# while true; do
> kubectl get lease node0 -n kube-node-lease -o=go-template --template="renewTime: {{.spec.renewTime}}" && echo ""
> sleep 10s
> done
renewTime: 2021-11-13T03:53:56.000000Z
renewTime: 2021-11-13T03:54:11.000000Z
renewTime: 2021-11-13T03:54:26.000000Z
renewTime: 2021-11-13T03:54:26.000000Z
renewTime: 2021-11-13T03:54:41.000000Z
renewTime: 2021-11-13T03:54:56.000000Z
renewTime: 2021-11-13T03:54:56.000000Z
renewTime: 2021-11-13T03:55:11.000000Z
renewTime: 2021-11-13T03:55:26.000000Z
renewTime: 2021-11-13T03:55:26.000000Z
renewTime: 2021-11-13T03:55:41.000000Z
renewTime: 2021-11-13T03:55:56.000000Z
renewTime: 2021-11-13T03:55:56.000000Z
```
We can find that the renewTime of nodelease is updated every 15s (the same as default `heartbeat` of edgehub), and the `lastHeartbeatTime` changes every 1 min (can be set in the config, which I'll describe bellow).  
This because a new nodelease controller has been introduced in kubeedge, to be specified, in cloudcore. It will continuously accept heartbeat msgs from all edge nodes connecting to it and then update corresponding nodeleases. So, the interval between every two lease renewTime is the same as `heartbeat` of edgehub.  
The interval of every two lastHeartbeatTime can be set through `nodeStatusReportFrequency` of edged. The default value is 60s currently.  When the `NodeLease` feature is enabled, the previous `nodeStatusUpdateFrequency` will be like a check interval. Edged will check the node status every `nodeStatusUpdateFrequency`. If there's no change, edged will not send a nodestatus update message until reaching the `nodeStatusReportFrequency`. Otherwise, it will send the most recent nodestatus to cloudcore.

Fixes #3086 

Now the `NodeLease` is in its alpha version and defaultly disabled. Users can enable the `NodeLease` feature through setting `NodeLease: true` in the `featureGates` filed of cloudcore and edgecore. Give an example here:
**cloudcore.yaml**
```yaml
commonConfig:
  tunnelPort: 10350
kind: CloudCore
kubeAPIConfig:
  burst: 200
  contentType: application/vnd.kubernetes.protobuf
  kubeConfig:
  master: ""
  qps: 100
featureGates:
  NodeLease: true
modules:
  nodeLeaseController:
    enable: true
...
```
**edgecore.yaml**
```yaml
apiVersion: edgecore.config.kubeedge.io/v1alpha1
database:
  aliasName: default
  dataSource: /tmp/var/lib/kubeedge/edgecore.db
  driverName: sqlite3
kind: EdgeCore
featureGates:
  NodeLease: true
modules:
  edgeHub:
    heartbeat: 15
    ...
  edged:
    nodeStatusReportFrequency: 60
    nodeStatusUpdateFrequency: 10
    ...
```
And run cloudcore and edgecore with `cloudcore.yaml` and `edgecore.yaml` respectively, then you can use the `NodeLease` feature.  
Note: if the `NodeLease` feature gate is not enabled, the fields of `nodeLeaseController.enable` and `edged.nodeStatusReportFrequency` will be ignored. 

/kind feature

### DIFF
--- a/cloud/cmd/cloudcore/app/server.go
+++ b/cloud/cmd/cloudcore/app/server.go
@@ -30,6 +30,7 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/pkg/devicecontroller"
 	"github.com/kubeedge/kubeedge/cloud/pkg/dynamiccontroller"
 	"github.com/kubeedge/kubeedge/cloud/pkg/edgecontroller"
+	"github.com/kubeedge/kubeedge/cloud/pkg/nodeleasecontroller"
 	"github.com/kubeedge/kubeedge/cloud/pkg/router"
 	"github.com/kubeedge/kubeedge/cloud/pkg/synccontroller"
 	"github.com/kubeedge/kubeedge/common/constants"
@@ -89,6 +90,11 @@ kubernetes controller which manages devices so that the device metadata/status d
 
 			gis := informers.GetInformersManager()
 
+			if !features.DefaultMutableFeatureGate.Enabled(features.NodeLease) {
+				// If not enable NodeLease feature gate, nodelease controller should not run.
+				config.Modules.NodeLeaseController.Enable = false
+			}
+
 			registerModules(config)
 
 			// IptablesManager manages tunnel port related iptables rules
@@ -133,6 +139,7 @@ func registerModules(c *v1alpha1.CloudCoreConfig) {
 	cloudstream.Register(c.Modules.CloudStream)
 	router.Register(c.Modules.Router)
 	dynamiccontroller.Register(c.Modules.DynamicController)
+	nodeleasecontroller.Register(c.Modules.NodeLeaseController)
 }
 
 func NegotiateTunnelPort() (*int, error) {
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -28,8 +28,10 @@ import (
 	deviceconst "github.com/kubeedge/kubeedge/cloud/pkg/devicecontroller/constants"
 	edgeconst "github.com/kubeedge/kubeedge/cloud/pkg/edgecontroller/constants"
 	edgemessagelayer "github.com/kubeedge/kubeedge/cloud/pkg/edgecontroller/messagelayer"
+	nodelease "github.com/kubeedge/kubeedge/cloud/pkg/nodeleasecontroller"
 	"github.com/kubeedge/kubeedge/cloud/pkg/synccontroller"
 	"github.com/kubeedge/kubeedge/common/constants"
+	"github.com/kubeedge/kubeedge/pkg/features"
 	"github.com/kubeedge/kubeedge/pkg/metaserver/util"
 	"github.com/kubeedge/viaduct/pkg/conn"
 	"github.com/kubeedge/viaduct/pkg/mux"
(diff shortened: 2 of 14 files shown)


Congrool:
I'll move nodelease update logic from cloudcore to edgecore and submit a new pr. So now close this.

PR #3297, opened 2021-11-15
Add type check for each implementation of Module interface

hainesc:
Signed-off-by: Haines Chan <hainescx@gmail.com>

**What type of PR is this?**

Improve code readable, see https://golang.org/doc/effective_go#blank_implements

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/cloudhub.go
+++ b/cloud/pkg/cloudhub/cloudhub.go
@@ -26,6 +26,8 @@ type cloudHub struct {
 	messageq             *channelq.ChannelMessageQueue
 }
 
+var _ core.Module = (*cloudHub)(nil)
+
 func newCloudHub(enable bool) *cloudHub {
 	crdFactory := informers.GetInformersManager().GetCRDInformerFactory()
 	// declare used informer
--- a/cloud/pkg/cloudstream/cloudstream.go
+++ b/cloud/pkg/cloudstream/cloudstream.go
@@ -28,6 +28,8 @@ type cloudStream struct {
 	enable bool
 }
 
+var _ core.Module = (*cloudStream)(nil)
+
 func newCloudStream(enable bool) *cloudStream {
 	return &cloudStream{
 		enable: enable,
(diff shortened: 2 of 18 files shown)


gy95:
this is redundant, if a specifics module doesn't implement all the methods of interface `Module`, will not pass the compiling, and the compilers will report the compiler errors as below
`*cloudHub does not implement "github.com/kubeedge/beehive/pkg/core".Module (missing xxxx method)`

gy95:
ref official docs: `By convention, such declarations are only used when there are **no static conversions** already present in the code, which is a rare event.`

hainesc:
@gy95 see https://github.com/uber-go/guide/pull/86

gy95:
we've tried to use struct `cloudhub` as interface `Module` here, so this is not requisite.
https://github.com/kubeedge/kubeedge/blob/a9d0a01c0c7940997e8eefe9e75b86e2b801251d/cloud/pkg/cloudhub/cloudhub.go#L46

And the interface check is used when there's no static conversions present in the code.

fisherxu (commented):
/approve

hainesc:
> we've tried to use struct `cloudhub` as interface `Module` here, so this is not requisite.
> 
> https://github.com/kubeedge/kubeedge/blob/a9d0a01c0c7940997e8eefe9e75b86e2b801251d/cloud/pkg/cloudhub/cloudhub.go#L46
> 
> And the interface check is used when there's no static conversions present in the code.

See here: https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance
Of course, it not the guide of kubeedge, it's the guide of uber, but I think it makes sense. Refer the discussion following the upper link.

hainesc:
> /approve

Thanks

fisherxu:
Any comments @zc2638? :)

zc2638:
Under normal circumstances, we need to add interface implementation verification.
In the project, `core.Register` has already done this part of the work, we can also omit it.
Of course, there is no problem adding additional interface implementation verification.

LGTM

fisherxu:
> Under normal circumstances, we need to add interface implementation verification. In the project, `core.Register` has already done this part of the work, we can also omit it. Of course, there is no problem adding additional interface implementation verification.
> 
> LGTM

OK, I think it's more readable that indicates modules implement the Module interface.
/lgtm

Issue #3254, opened 2021-10-29
Enhance Router Module and ServiceBus


ttlv:
**What would you like to be added/modified**:
I want to enhance Router module which in CloudCore.I want to send message between ServiceBus and Rest just like we can send message between Rest to EventBus.So.it means the ServiceBus will upgrade to a  HTTP Server in the EdgeCore。What's more the ServiceBus provider in Router which belongs to CloucCore should implement Source interface 
**Why is this needed**:
I have an application A which run in edge site and at the same time I have an application B which is a http or https server run in cloud site。application A want to communicate to application B and because of the complicated network which causes A can't send http request to B directly so servibus should be given an ability to become a http request proxy between  edge and cloud.
This enhancement can bring some benefits to us: 
1.  Applications or devices in edge site communicate to cloud will be very easily.
2. Skip complicated authentication.Applications and Devices are credible in edge site,they can send request any time.
3. If the network is not avaliable between cloud and edge and time out of the request will be easy to handle
4. MQTT is not the only way for communication between edge and cloud and rich communication protocol in edge site.
5. ServiceBus  upgrade to a http server,we can do more things in the edge site.

Here is the design
![image](https://user-images.githubusercontent.com/20856105/139382132-d4fb55cd-5431-4026-af94-39ac699df296.png) @fisherxu @kevin-wangzefeng

shenguangfu0906:
how  can  i  use  the  servicebus  edge to  cloud  ，how to create yamls  。there have any document？  the  kubeedge.io  do not  have document  about servicebus  for  edge to cloud  ,there only cloud to edge

PR #3301, opened 2021-11-16
feat/servicebus: issue3254 enhance router module and servicebus

ttlv:
https://github.com/kubeedge/kubeedge/issues/3254
https://github.com/kubeedge/kubeedge/issues/3254
```

### DIFF
--- a/cloud/pkg/router/provider/servicebus/servicebus.go
+++ b/cloud/pkg/router/provider/servicebus/servicebus.go
@@ -12,6 +12,7 @@ import (
 
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
 	"github.com/kubeedge/beehive/pkg/core/model"
+
 	v1 "github.com/kubeedge/kubeedge/cloud/pkg/apis/rules/v1"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/modules"
 	"github.com/kubeedge/kubeedge/cloud/pkg/router/constants"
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: cloud/pkg/router/provider/servicebus/servicebus.go [hunk 1]
```diff
@@ -21,38 +24,97 @@ type servicebusFactory struct{}
...
 }
 
-func (factory *servicebusFactory) GetTarget(ep *v1.RuleEndpoint, targetResource map[string]string) provider.Target {
+func (sf *servicebusFactory) GetSource(ep *v1.RuleEndpoint, sourceResource map[string]string) provider.Source {
+	targetURL, exist := sourceResource["target_url"]
+	if !exist {
+		klog.Errorf("source resource attributes \"target_url\" does not exist")
+		return nil
+	}
+	nodeName, exist := sourceResource["node_name"]
```
fisherxu, line 48: Is there any const for nodename and target_url?

edge/pkg/servicebus/servicebus.go
```diff
@@ -79,7 +111,7 @@ func (sb *servicebus) Start() {
 		}
 
 		// build new message with required field & send message to servicebus
-		klog.Infof("ServiceBus receive msg")
+		klog.Infof("servicebus receive msg")
```
fisherxu, line 114:
```suggestion
		klog.V(4).Infof("servicebus receive msg")
```

PR #3302, opened 2021-11-17
Fix the log's syntax problem in edged

lvchenggang:
What type of PR is this?
/kind feature

Some logs in edged.go have syntax error, which will cause misunderstanding. So need to fix them.

NONE


fisherxu (commented):
/approve

fisherxu (commented):
/lgtm

PR #3313, opened 2021-11-18
remove unused member variable of device DownstreamController

jidalong:
**What type of PR is this?**

/kind feature

In devicecontroller/controller/downstream.go, all we use is kubeClient instead of crdClient of DownstreamController struct. Thus, crdClient is useless.

NONE

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #3316, opened 2021-11-19
syncPod method add message content type choice in edged.go

lvchenggang:

What type of PR is this?
/kind feature

syncPod method add message  content type choice in edged.go.
the message content type probably more than one []byte type, if the message come from cloudhub udsserver, then the type is string.
We need to consider this situation.

NONE

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -1121,10 +1121,18 @@ func (e *edged) syncPod() {
 		}
 		op := result.GetOperation()
 
-		content, err := result.GetContentData()
-		if err != nil {
-			klog.Errorf("get message content data failed: %v", err)
-			continue
+		var content []byte
+		switch result.Content.(type) {
+		case []byte:
+			content = result.GetContent().([]byte)
+		case string:
+			content = []byte(result.GetContent().(string))
+		default:
+			content, err = json.Marshal(result.Content)
+			if err != nil {
+				klog.Errorf("marshal message content failed: %v", err)
+				continue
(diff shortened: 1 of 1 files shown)


lvchenggang:
> What type of PR is this? /kind feature
> 
> What this PR does / why we need it: syncPod method add message content type choice in edged.go. the message content type probably more than one []byte type, if the message come from cloudhub udsserver, then the type is string. We need to consider this situation.
> 
> Does this PR introduce a user-facing change?: NONE

gy95:
what's difference? it seems that they are the same. using json.Marshal or []byte directively transformed to []byte

lvchenggang:
> what's difference? it seems that they are the same. using json.Marshal or []byte directively transformed to []byte

The difference lies in message's content type, it maybe string type, such as the message from cloudhub udsserver.
The previous method result.GetContentData() only transfer the message's content to []byte but not consider the string type, which will make parse failed.
The code of GetContentData() is:
image

gy95:
can't we use `json.Marshal` to convert `string` to `[]byte` format?
IMHO, I think the two methods are equal. 😃

lvchenggang:
This two methods essentially have same function.

fisherxu:
> This two methods essentially have same function.

As @gy95 said, string can be json.Marshal to the expect byte, so I think there's no need to add the code for string. :) WDYT?

fisherxu:
@lvchenggang

lvchenggang:
Yes, these add codes are no need, this patch will be closed

Issue #3322, opened 2021-11-19
Builds failing on macOS


mjlshen:
**How to reproduce it (as minimally and precisely as possible)**:
`make test`

Results in these errors and warnings on macOS:
```
# github.com/shirou/gopsutil/cpu
vendor/github.com/shirou/gopsutil/cpu/cpu_darwin_cgo.go:13:5: warning: 'TARGET_OS_MAC' is not defined, evaluates to 0 [-Wundef-prefix=TARGET_OS_]
# github.com/shirou/gopsutil/disk
iostat_darwin.c:28:2: warning: 'IOMasterPort' is deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:132:1: note: 'IOMasterPort' has been explicitly marked deprecated here
# github.com/shirou/gopsutil/disk
vendor/github.com/shirou/gopsutil/disk/disk_darwin.go:67:51: cannot use stat.Mntfromname[:] (type []byte) as type []int8 in argument to common.IntToString
vendor/github.com/shirou/gopsutil/disk/disk_darwin.go:68:49: cannot use stat.Mntonname[:] (type []byte) as type []int8 in argument to common.IntToString
vendor/github.com/shirou/gopsutil/disk/disk_darwin.go:69:50: cannot use stat.Fstypename[:] (type []byte) as type []int8 in argument to common.IntToString
vendor/github.com/shirou/gopsutil/disk/disk_darwin.go:85:43: cannot use stat.Fstypename[:] (type []byte) as type []int8 in argument to common.IntToString
```

and this build failure:
```
FAIL	github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/debug [build failed]
```

The errors have been fixed upstream by https://github.com/shirou/gopsutil/issues/976 and https://github.com/shirou/gopsutil/issues/958. So it can likely be fixed by just bumping the version of gopsutil

PR #3325, opened 2021-11-19
Builds are only supported for GOOS=linux, so set it explicitly

mjlshen:
Signed-off-by: Michael Shen <mishen@umich.edu>

**What type of PR is this?**
/kind bug

Since only target operating systems of linux are supported, so explicitly setting this will allow builds from other operating systems, like macOS to succeed. Furthermore, I noticed there was an error in the integrationtest make target.

Related to #3322

```release-note
NONE
```

### DIFF
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -214,7 +214,7 @@ kubeedge::golang::build_binaries() {
     echo "building $bin"
     local name="${bin##*/}"
     set -x
-    GOOS="linux" CGO_ENABLED=1 go build -o ${KUBEEDGE_OUTPUT_BINPATH}/${name} -gcflags="${gogcflags:-}" -ldflags "${goldflags:-}" $bin
+    GOOS="linux" go build -o ${KUBEEDGE_OUTPUT_BINPATH}/${name} -gcflags="${gogcflags:-}" -ldflags "${goldflags:-}" $bin
     set +x
   done
 }
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -214,10 +214,9 @@ kubeedge::golang::build_binaries() {
     echo "building $bin"
     local name="${bin##*/}"
     set -x
-    go build -o ${KUBEEDGE_OUTPUT_BINPATH}/${name} -gcflags="${gogcflags:-}" -ldflags "${goldflags:-}" $bin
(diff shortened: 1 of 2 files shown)


gy95:
please sign off the commit, ref: https://github.com/kubeedge/kubeedge/pull/3325/checks?check_run_id=4268491656

mjlshen:
> please sign off the commit, ref: https://github.com/kubeedge/kubeedge/pull/3325/checks?check_run_id=4268491656 😄

Oops, thanks, done! @gy95

hack/lib/golang.sh
```diff
@@ -214,10 +214,9 @@ kubeedge::golang::build_binaries() {
     echo "building $bin"
     local name="${bin##*/}"
     set -x
-    go build -o ${KUBEEDGE_OUTPUT_BINPATH}/${name} -gcflags="${gogcflags:-}" -ldflags "${goldflags:-}" $bin
+    GOOS="linux" CGO_ENABLED=1 go build -o ${KUBEEDGE_OUTPUT_BINPATH}/${name} -gcflags="${gogcflags:-}" -ldflags "${goldflags:-}" $bin
```
gy95, line 217: for mac os, must this `CGO_ENABLED=1` be required?

Makefile
```diff
@@ -143,7 +143,7 @@ integrationtest:
 	@echo "$$INTEGRATION_TEST_HELP_INFO"
 else
 integrationtest:
-	hack/make-rules/build.sh edgecore
+	hack/make-rules/build.sh edgecore; \
```
gy95, line 146: could you please provide some materials about this `;`, is this required? :)

hack/lib/golang.sh
mjlshen: Not needed for macOS, I saw that some of the binaries needed it. I can remove `CGO_ENABLED=1` if they don't all need CGO

Makefile
mjlshen: I think I was mistaken, I thought that `execute.sh` wasn't being run

fisherxu: So what's meaning of the `; \` here?

hack/lib/golang.sh
fisherxu: I think now we have no conclusion that kubeedge only supports Linux, so if we add `GOOS="linux"` here, all the other os   need to be tested.

Makefile
mjlshen: I reverted it, thought it was needed to run both commands

PR #3326, opened 2021-11-22
Make csi driver name configurable

lvchenggang:

What type of PR is this?
/kind feature

Make csi driver name configurable, which will be more flexible.

NONE

### DIFF
--- a/edge/pkg/edged/volume/csi/csi_controller.go
+++ b/edge/pkg/edged/volume/csi/csi_controller.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"k8s.io/klog/v2"
+
+	edgedconfig "github.com/kubeedge/kubeedge/edge/pkg/edged/config"
 )
 
 type Controller struct {
@@ -30,7 +32,7 @@ type Controller struct {
 func (c *Controller) CreateVolume(req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	if c.csiClient == nil {
 		var err error
-		c.csiClient, err = newCsiDriverClient(csiDriverName("csi-hostpath"))
+		c.csiClient, err = newCsiDriverClient(csiDriverName(edgedconfig.Config.CsiDriverName))
 		if err != nil {
 			klog.Errorf("failed to create newCsiDriverClient: %v", err)
 			return nil, err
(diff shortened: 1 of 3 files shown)


pkg/apis/componentconfig/edgecore/v1alpha1/types.go
```diff
@@ -235,6 +235,9 @@ type Edged struct {
 	// EnableMetrics indicates whether enable the metrics
 	// default true
 	EnableMetrics bool `json:"enableMetrics,omitempty"`
+	// CsiDriverName indicates the csi driver name used by edge node
+	// default "csi-hostpath"
+	CsiDriverName string `json:"csiDriverName,omitempty"`
```
gy95, line 240: cannot be `omitempty`, we need a not-empty value here.

gy95: And may I ask for more materials about application scenarios?

lvchenggang: The usable csi driver of edge node not just hostpath type, it may be lvm, nfs .etc.
We can not fixed as csi-hostpath in code.

lvchenggang:
Could it be merged? @gy95 @GsssC

fisherxu:
cc @moolen Please also take a look  :)

moolen:
LGTM from my point of view, we need that ;)

edge/pkg/edged/volume/csi/csi_controller.go
```diff
@@ -30,7 +32,7 @@ type Controller struct {
 func (c *Controller) CreateVolume(req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
 	if c.csiClient == nil {
 		var err error
-		c.csiClient, err = newCsiDriverClient(csiDriverName("csi-hostpath"))
+		c.csiClient, err = newCsiDriverClient(csiDriverName(edgedconfig.Config.CsiDriverName))
```
fisherxu, line 35: Here we only can support one csi driver? @lvchenggang @moolen

fisherxu: Is it possible to read the name from some parameters?

moolen: Yes, right now `edged` can only talk to one specific csi driver. Because the driver name is hard-coded `kubeedge` does 
 support only the `csi-hostpath` driver. But we can potentially support `nfs` or any other driver. 
The csi-driver name can be configured through `/etc/kubeedge/config/edgecore.yaml`.

gy95: /hold
keadm need to fix also, although this parameter is configurable, keadm should pass in parameter to fix this default value.

fisherxu: > We could potentially support multiple csi drivers on the edge side but there is a lot of work to be done in order to support that.

The storage driver name maybe should from volume spec, and then edgecore calls the related driver? Configured from `/etc/kubeedge/config/edgecore.yaml` is not that extensible.

moolen: > Configured from /etc/kubeedge/config/edgecore.yaml is not that extensible

i agree.

> The storage driver name maybe should from volume spec

Unfortunately it's not embedded in the CreateVolume request. we probably need to embed the `Volume*` with additional metadta (the to-be-called csi driver on the edge)

I opened #3471 to discuss the high-level approach there.

lvchenggang: Agree with that, the modification of this PR is not extensible.

PR #3330, opened 2021-11-22
delete useless device-v1alpha1 code while device is update to v1alpha2

jidalong:
**What type of PR is this?**
/kind cleanup

Since device is update to alpha2 from alpha1, while the annotation code is only used for alpha1, thus remove them.

NONE

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #3335, opened 2021-11-22
add stream/tunnel port and nodeport for cloudcore

allenleu:
add stream/tunnel port for deployent.yaml and service.yaml


build/cloud/07-deployment.yaml
```diff
@@ -26,6 +26,12 @@ spec:
             - containerPort: 10000
               name: cloudhub
               protocol: TCP
+            - containerPort: 10003
+              name: stream
+              protocol: TCP
+            - containerPort: 10004
+              name: tunnel
+              protocol: TCP
```
fisherxu, line 34: The all containerPorts are no need in the deployment yaml I think, any ideas? @zc2638 @zhu733756

zhu733756: I think on the mode of cloudcore host network, service can be disabled, and no need to add containerPorts, and in this case,  the containerports are equal to hostports. However, If we use outside Iptables manager or the cloudcore pod IP is not the host ip, we also need a service to expose it outside the cluster. Since the edges may not be able to communicate with cloudside.

zhu733756: This is a test for disabled service and inner container ports:

9ffcf0f2f81cf3f79782d0d12b86077

allenleu: yes,I have added nodeport to service.yaml.example.

zc2638:
Hi @allenleu ,

    The installation tool `keadm` will be refactored in the near future, and this PR may not receive feedback.
    Are you interested in participating in the refactoring of `keadm`?
    Related issue #3475

allenleu:
> 



> Hi @allenleu, The installation tool `keadm` will be refactored in the near future, and this PR may not receive feedback. Are you interested in participating in the refactoring of `keadm`? Related issue #3475

ok,Is it because this pr conflicts with refactoring Keadm? I can revoke this pr.
I'm sorry I don't have any other time to participate in your work now.

fisherxu (commented):
port `10002` still need to be exposed @allenleu  :)

fisherxu:
Ref: https://github.com/kubeedge/kubeedge/blob/master/build/helm/cloudcore/templates/deployment_cloudcore.yaml#L52

PR #3346, opened 2021-11-23
fix "keadm get nodes -A"  printing namespaces while nodes without namespaces actually

### LINKED ISSUE #3345: keadm get nodes -A print namespace while node resource whithout namespace actually
**What happened**:
When I excute “keadm get nodes/node -A/--all-namespaces ”，keadm may print the namespace with null while in original k8s， the resource of node does not have any namespaces.
image
The following is **kubectl** get nodes -A in original k8s cluster:
image

“keadm get nodes/node -A/--all-namespaces ” may print without namespaces

jidalong:
**What type of PR is this?**
/kind bug

When  excute “keadm get nodes/node -A/--all-namespaces ”，keadm may not print the namespace with null while in original k8s，the resource of node does not have any namespaces.
before:
image
after this PR, not print NAMESPACE column anymore:
image

Fixes #3345

NONE

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/debug/get.go
+++ b/keadm/cmd/keadm/app/cmd/debug/get.go
@@ -180,7 +180,7 @@ func (g *GetOptions) Run(args []string) error {
 		}
 	}
 
-	if g.AllNamespace {
+	if g.AllNamespace && resType != "nodes" && resType != "node" {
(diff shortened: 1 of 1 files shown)


gy95:
LGTM

fisherxu (commented):
/approve

PR #3348, opened 2021-11-23
log err if convert membership failed

TianTianBigWang:

**What type of PR is this?**
/kind bug

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

edge/pkg/devicetwin/dtmanager/membership.go
```diff
@@ -240,7 +240,8 @@ func addDevice(context *dtcontext.DTContext, toAdd []dttype.Device, baseMessage
 		addDeviceResult := dttype.MembershipUpdate{BaseMessage: baseMessage, AddDevices: addDeviceDevices}
 		result, err := dttype.MarshalMembershipUpdate(addDeviceResult)
 		if err != nil {
-
+			klog.Errorf("add device %s failed, marshal membership err: %s", device.ID, err)
+			continue
```
gy95, line 244: not correct, line 250 has a unlock operation.
```
if delta {
			context.Unlock(device.ID)
		}
```
maybe this will result in not release the  lock and  dead lock. :)

TianTianBigWang: seem that fix in #3353

PR #3350, opened 2021-11-23
fix keadm info print format

jidalong:
**What type of PR is this?**
/kind bug

fix the print  infomation by command “keadm -h ” is not formatting：
image

NONE

```release-note

```


zhu733756:
Seems great, more clear ! :)

PR #3358, opened 2021-11-24
fix grammar in pkg module

xujingjing-cmss:
Signed-off-by: xujingjing xujingjing_yewu@cmss.chinamobile.com

What type of PR is this?
/kind cleanup

fix spelling error and remove redundant code.

Which issue(s) this PR fixes:
Fixes #

### DIFF
--- a/pkg/metaserver/key.go
+++ b/pkg/metaserver/key.go
@@ -116,7 +116,7 @@ func ParseKey(key string) (gvr schema.GroupVersionResource, namespace string, na
 	slices := strings.Split(sl, "/")
 	length := len(slices)
 	if len(slices) == 0 || slices[0] != "" {
-		//klog.Errorf("[metaserver]falied to parse key: format error, %v",key)
+		//klog.Errorf("[metaserver]failed to parse key: format error, %v",key)
 		return
 	}
 	var (


pkg/util/flag/flags.go
```diff
@@ -52,7 +52,7 @@ func (m *ConfigValue) Set(s string) error {
 }
 
 func (m *ConfigValue) String() string {
-	return fmt.Sprintf("%v", bool(*m == ConfigTrue))
```
fisherxu, line 55: This file is from K8s repo, let's keep it as above :) Ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/util/flag/flags.go

pkg/version/verflag/verflag.go
```diff
@@ -64,7 +64,7 @@ func (v *VersionValue) String() string {
 	if *v == VersionRaw {
 		return strRawVersion
 	}
-	return fmt.Sprintf("%v", bool(*v == VersionTrue))
```
fisherxu, line 67: Ditto.

pkg/util/flag/flags.go
xujingjing-cmss: done

pkg/version/verflag/verflag.go
xujingjing-cmss: done

PR #3359, opened 2021-11-24
fix keadm check mqtt install bug

jidalong:
**What type of PR is this?**
/kind bug

keadm join may first check if host has already install mqtt, by exec the following code:
>~~~go
>cmd := NewCommand("ps aux |awk '/mosquitto/ {print $1}' | awk '/mosquit/ {print}'")
>if err := cmd.Exec(); err != nil {
>	return err
>}
>
>if stdout := cmd.GetStdOut(); stdout != "" {
>	fmt.Println("Host has", stdout, "already installed and running. Hence skipping the installation steps !!!")
>	return nil
>}
>~~~
But "ps aux |awk '/mosquitto/ {print $1}' | awk '/mosquit/ {print}'" cannot checkout whether if  the host has installed mqtt where mqtt service was under 'root' by editing the "/etc/mosquitto/mosquitto.conf" with "user root":
>~~~go
>user root
>~~~
image
because we can see the screen output command "/usr/sbin/mosquitto" is in 11st column not 1st column : 
image
So we should change  **"{print $1}"** to **"{print $11}"** instead, we can compare after this PR:
image

NONE

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/debinstaller.go
+++ b/keadm/cmd/keadm/app/cmd/util/debinstaller.go
@@ -41,7 +41,7 @@ func (d *DebOS) SetKubeEdgeVersion(version semver.Version) {
 
 // InstallMQTT checks if MQTT is already installed and running, if not then install it from OS repo
 func (d *DebOS) InstallMQTT() error {
-	cmd := NewCommand("ps aux |awk '/mosquitto/ {print $1}' | awk '/mosquit/ {print}'")
+	cmd := NewCommand("ps aux |awk '/mosquitto/ {print $11}' | awk '/mosquit/ {print}'")
 	if err := cmd.Exec(); err != nil {
 		return err
 	}
--- a/keadm/cmd/keadm/app/cmd/util/pacmaninstaller.go
+++ b/keadm/cmd/keadm/app/cmd/util/pacmaninstaller.go
@@ -23,7 +23,7 @@ func (o *PacmanOS) SetKubeEdgeVersion(version semver.Version) {
 
 // InstallMQTT checks if MQTT is already installed and running, if not then install it from OS repo
 func (o *PacmanOS) InstallMQTT() error {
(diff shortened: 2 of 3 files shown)


TianTianBigWang:
image
seem that i have not this problem?

jidalong:
> image seem that i have not this problem?

jidalong:
> image seem that i have not this problem?

Thanks, I tried on different OS, and find that OS I used BC-Linux based on Centos，where the mosquito process was owned by 'root' not 'mosquit' like the following. And I also think if the mosquitto service was installed by user manually it may probalbly under 'root' :
image

TianTianBigWang:
> > image seem that i have not this problem?
> 
> @TianTianBigWang Thanks, I tried on different OS[CentOS7.6, ubuntu18.04,BC-Linux7.6(CentOS)]，where the mosquito process was owned by 'mosquit+' not 'root' like yours. But After I edit the user in file "/etc/mosquitto/mosquitto.conf" to set the owner of mqtt service:
> 
> > ```go
> > user root
> > ```
> 
> which may cause that the mosquitto service was under **'root'** not 'mosquit+' like the following: image In such environment, keadm could not check whether if the host has installed mqtt, although mqtt was installed under root, which prints: image After this PR change "{print $1}" to "{print $11}" instead , no matter mosquitto is under 'mosquit+' or 'root', it can checkout accurately. image

sounds good.

keadm/cmd/keadm/app/cmd/util/debinstaller.go
```diff
@@ -41,7 +41,7 @@ func (d *DebOS) SetKubeEdgeVersion(version semver.Version) {
 
 // InstallMQTT checks if MQTT is already installed and running, if not then install it from OS repo
 func (d *DebOS) InstallMQTT() error {
-	cmd := NewCommand("ps aux |awk '/mosquitto/ {print $1}' | awk '/mosquit/ {print}'")
```
fisherxu, line 44: I wonder if we can implement this with golang :)

Issue #3138, opened 2021-09-07
improve the test coverage


gy95:
**What would you like to be added/modified**:
add more test to improve the test coverage.
**Why is this needed**:
improve the test coverage
kubeedge has already provided a script to display the test coverage. 
just use `make test PROFILE=coverprofile` to print coverage
```
ok  	github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dtmodule	0.015s	coverage: 77.8% of statements
ok  	github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dtcommon	0.019s	coverage: 100.0% of statements
ok  	github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dtclient	0.016s	coverage: 96.1% of statements
```
If we want to view coverage in detail, we could use `go tool cover -func=coverprofile` to find more details, and the output is like this. this will print every function test coverage
```
github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/model/types.go:74:			NewResource			100.0%
github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/model/types.go:86:			IsNodeStopped			100.0%
github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/model/types.go:99:			IsFromEdge			100.0%
github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common/model/types.go:110:			IsToEdge			100.0%
```

PR #3366, opened 2021-11-24
Refactor computing the SHA512 checksum to go instead of bash

mjlshen:
**What type of PR is this?**
/kind cleanup
/kind test

Refactors untestable bash --> testable Go

Work on #3138

```release-note
NONE
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -19,6 +19,7 @@ package util
 import (
 	"archive/tar"
 	"compress/gzip"
+	"crypto/sha512"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -438,12 +439,30 @@ func hasSystemd() bool {
...
+
+	return fmt.Sprintf("%x", h.Sum(nil)), nil
+}
+
 func checkSum(filename, checksumFilename string, version semver.Version, tarballPath string) (bool, error) {
 	//Verify the tar with checksum
 	fmt.Printf("%s checksum: \n", filename)
-	getActualCheckSum := NewCommand(fmt.Sprintf("cd %s && sha512sum %s | awk '{split($0,a,\"[ ]\"); print a[1]}'", tarballPath, filename))
-	if err := getActualCheckSum.Exec(); err != nil {
-		return false, err
```
gy95, line 446: seems work well by this function `computeSHA512Checksum`

keadm/cmd/keadm/app/cmd/util/common_test.go
```diff
@@ -192,3 +192,37 @@ func TestFileExists(t *testing.T) {
...
+			name:     "a",
+			expected: "1f40fc92da241694750979ee6cf582f2d5d7d28e18335de05abc54d0560e0f5302860c652bf08d560252aa5e74210546f369fbbbce8c12cfc7957b2652fe9a75",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			tmpfile, err := ioutil.TempFile("", test.name)
+			if err != nil {
+				t.Errorf("Unable to create temp file: %v", err)
```
gy95, line 211: need return
and `tmpfile.Close()` should be added to every return branch.

mjlshen: Thanks for the review! I don't think I need a return since `t.Errorf()` fails the test if it gets there. Kind of like how other tests in that file fail with `t.Fatalf()`. I did change tmpfile.Close() to `defer tmpfile.Close()` to ensure it runs though.

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -165,15 +166,22 @@ func RunningModule() (types.ModuleRunning, error) {
 
 // GetLatestVersion return the latest non-prerelease, non-draft version of kubeedge in releases
 func GetLatestVersion() (string, error) {
-	//Download the tar from repo
-	versionURL := "curl -k " + latestReleaseVersionURL
-	cmd := exec.Command("sh", "-c", versionURL)
-	latestReleaseData, err := cmd.Output()
+	// curl https://kubeedge.io/latestversion
+	resp, err := http.Get(latestReleaseVersionURL)
```
gy95, line 170: By my test, throuth method `http.Get(latestReleaseVersionURL)` result is equal to directly exec shell command `"curl -k " + latestReleaseVersionURL`
Is this refactor work OK? @fisherxu 
If OK, but I  also think we need to open another PR to refactor this function, but not mix it with other functions.

mjlshen: Sure, I can make an issue for this and put it in another PR

mjlshen: I have removed this code from this PR and made #3385 to address this - I can work on it if @fisherxu agrees.

fisherxu (commented):
/lgtm
/approve
Thanks for the great fixing! :) @mjlshen  And Congratulations on becoming KubeEdge contributor 👍

PR #3368, opened 2021-11-25
fix daemonset pods not synced to edge

chenchunxiu:
**What type of PR is this?**
/kind bug

After the edge-node broken and the db file is lost, if edge-node reconnect to the cloudcore again, the daemonset pod will not be synced to the edge-node, but the pod status is still running

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -31,12 +31,14 @@ import (
 	stderrors "errors"
 	"fmt"
 	"sort"
+	"strconv"
 	"time"
 
 	authenticationv1 "k8s.io/api/authentication/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


gy95:
this is not correct, upstream controller is for syncing edge messages from edge to cloud.

chenchunxiu:
> this is not correct, upstream controller is for syncing edge messages from edge to cloud.

yes，upstream controller is uaually for syncing edge messages from edge to cloud.
but，downstream controller cannot watch the event of daemonset pods，so here i sync the ds pods to edge only when edge-node register agagin

gy95:
If the database file is lost, edge node does not work well, what else do we need to consider? :)

chenchunxiu:
> If the database file is lost, edge node does not work well, what else do we need to consider? :)
I may not get what you mean.
The scenario I am considering here is:  After the node broken and the db file lost, the daemonset pod can be restored after the edge-node is reconnected to cloudcore.

chenchunxiu:
> If the database file is lost, edge node does not work well, what else do we need to consider? :)

When db file lost, we usually cannot restore it, but some users will access cloudcore with the same nodeName after restoring the node. 
After reconnect to cloudcore, the pod of the existing daemonset will not be synched to the edge-node, here is mainly to solve this problem.

PR #3370, opened 2021-11-25
Remove useless parentheses

lvchenggang:

What type of PR is this?
/kind feature

Remove useless parentheses in edged module.

NONE

### DIFF
--- a/edge/pkg/edged/volume/csi/csi_plugin.go
+++ b/edge/pkg/edged/volume/csi/csi_plugin.go
@@ -226,7 +226,7 @@ func (p *csiPlugin) Init(host volume.VolumeHost) error {
 		}
 	}
 
-	var migratedPlugins = map[string](func() bool){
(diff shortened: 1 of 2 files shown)


zc2638:
There should be many such problems.
Can you try to add configuration to CI, as much as possible to automate to avoid the occurrence of such problems in the future.

lvchenggang:
> There should be many such problems. Can you try to add configuration to CI, as much as possible to automate to avoid the occurrence of such problems in the future.

I will try it, if the implementation is successful, I will submit to next PR.

lvchenggang:
Could this PR be merged?

gy95:
some codes are derived from kubernetes, so better keep compatible with upstream code :)

lvchenggang:
> some codes are derived from kubernetes, so better keep compatible with upstream code :)

......

PR #3372, opened 2021-11-25
DealDeviceTwin cleanup

TianTianBigWang:
**What type of PR is this?**
/kind cleanup

needn't check dealType twice, if err != nil return err is ok

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/twin.go
+++ b/edge/pkg/devicetwin/dtmanager/twin.go
@@ -207,14 +207,12 @@ func DealDeviceTwin(context *dtcontext.DTContext, deviceID string, eventID strin
 		}
 	}
 
-	if err != nil && dealType == RestDealType {
-		updateResult, _ := dttype.BuildDeviceTwinResult(dttype.BaseMessage{EventID: eventID, Timestamp: now}, dealTwinResult.Result, dealType)
-		dealUpdateResult(context, deviceID, eventID, dtcommon.InternalErrorCode, err, updateResult)
-		return err
-	}
 	if dealType == RestDealType {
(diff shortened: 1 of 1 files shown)


gy95:
lgtm

edge/pkg/devicetwin/dtmanager/twin.go
```diff
@@ -207,14 +207,12 @@ func DealDeviceTwin(context *dtcontext.DTContext, deviceID string, eventID strin
...
+	if dealType == RestDealType {
 		updateResult, _ := dttype.BuildDeviceTwinResult(dttype.BaseMessage{EventID: eventID, Timestamp: now}, dealTwinResult.Result, dealType)
 		dealUpdateResult(context, deviceID, eventID, dtcommon.InternalErrorCode, err, updateResult)
-		return err
 	}
-	if dealType == RestDealType {
-		updateResult, _ := dttype.BuildDeviceTwinResult(dttype.BaseMessage{EventID: eventID, Timestamp: now}, dealTwinResult.Result, dealType)
-		dealUpdateResult(context, deviceID, eventID, dtcommon.InternalErrorCode, nil, updateResult)
+	if err != nil {
+		return err
```
gy95, line 215: should be like this: `if err != nil && dealType == RestDealType`
Or logic will change and need more considerations

considering one circumstance: if  dealType==SyncDealType and err != nil, this modification will return directly but will continue to process without this patch.

TianTianBigWang: Yes, I thought about it. It does cause a return when `err!=nil && dealType == 1`, without executing subsequent code, so I put `if err != nil` in the `dealType` judgment

fisherxu (commented):
/approve

PR #3376, opened 2021-11-25
support device message attr error return

orangegzx:
**What type of PR is this?**

/kind feature

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/device.go
+++ b/edge/pkg/devicetwin/dtmanager/device.go
@@ -3,6 +3,7 @@ package dtmanager
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


edge/pkg/devicetwin/dtmanager/device.go
```diff
@@ -159,6 +160,9 @@ func UpdateDeviceAttr(context *dtcontext.DTContext, deviceID string, attributes
 		return nil, nil
 	}
 	dealAttrResult := DealMsgAttr(context, Device.ID, attributes, dealType)
+	if dealAttrResult.Err != nil {
+		return nil, err
+	}
```
gy95, line 165: seems reasonable, if err occur, Add, Delete, Update field will be nil.

gy95: better `return nil, nil`
the original logic is always returned `nil, nil`. 😄

orangegzx: already edited

fisherxu (commented):
/approve

PR #3382, opened 2021-11-25
Enhanced disk check function

xujingjing-cmss:
Signed-off-by: xujingjing xujingjing_yewu@cmss.chinamobile.com

What type of PR is this?
/kind feature

Only disk which mount on / use to be checked. Now all mounted part will be checked.

Which issue(s) this PR fixes:
Fixes #


keadm/cmd/keadm/app/cmd/debug/check.go
```diff
@@ -230,21 +230,24 @@ func CheckDisk() error {
...
+	for _, part := range parts {
+		usageInfo, err := disk.Usage(part.Mountpoint)
+		if err != nil {
+			return err
+		}
+		fmt.Printf("Device: %s, MountPoint: %s, Total: %.2f MB, Free Total: %.2f MB, Usage Rate: %.2f\n", part.Device, part.Mountpoint, float32(usageInfo.Total)/common.MB, float32(usageInfo.Free)/common.MB, usageInfo.UsedPercent/100)
+		if usageInfo.Total >= common.AllowedValueDisk &&
+			usageInfo.Free >= common.AllowedCurrentValueDisk &&
+			usageInfo.UsedPercent/100 <= common.AllowedCurrentValueDiskRate {
+			checkFlag = true
```
gy95, line 244: just return nil, don not need this bool flag

xujingjing-cmss: I think that at least one part should meet the requirement

gy95: yes, as long as one part meets the requirement, we don't need to continue to loop to check another part whether meets the requirement. :) So just return or break the loop

xavierxjj: The loop here aims to list all parts. If we return or break the loop, not all part's usage will be printed

xujingjing-cmss: The loop here aims to print all part's usage. If we do return or break, some of parts may not be printed.

PR #3390, opened 2021-11-26
Explicitly implement the pod.Manager interface and add test coverage

mjlshen:
**What type of PR is this?**
/kind test
/kind cleanup

Edged was depending on k8s.io/kubernetes/pkg/kubelet/pod/testing to implement the "Manager" interface. I explicitly implemented it and added comments + test coverage.

Works on #3138 

```release-note
NONE
```


edge/pkg/edged/podmanager/pod_manager_test.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"reflect"
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestGetSetPods(t *testing.T) {
```
fisherxu, line 12: Does this has already been tested in k8s？

PR #3393, opened 2021-11-26
Add testing to keadm version

mjlshen:
**What type of PR is this?**
/kind cleanup
/kind test

Improves test coverage by testing `keadm version`, using the exact same code from testing `kubeadm version` https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/cmd/version_test.go.

Works on #3138 

```release-note
NONE
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cmd.go
+++ b/keadm/cmd/keadm/app/cmd/cmd.go
@@ -17,6 +17,8 @@ limitations under the License.
 package cmd
 
 import (
+	"io"
+
 	"github.com/spf13/cobra"
 
 	cloud "github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/cloud"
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -33,6 +33,7 @@ require (
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.8.1
 	github.com/paypal/gatt v0.0.0-20151011220935-4ae819d591cf
+	github.com/pkg/errors v0.9.1
```
gy95, line 36: we would not use this package, just use Go native error packages instead.

mjlshen: Good catch, I've removed that. Maybe I can go make that same fix for `kubeadm` haha

mjlshen:
Need help re-running this CI workflow @fisherxu

zc2638:
Hi @mjlshen ,
The installation tool keadm will be refactored in the near future, and this PR may not receive feedback.
Are you interested in participating in the refactoring of keadm?
Related issue #3475

PR #3401, opened 2021-11-29
add lock for status_manager

muxuelan:
**What type of PR is this?**
/kind feature

It's better add lock

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/edged/status/status_manager.go
+++ b/edge/pkg/edged/status/status_manager.go
@@ -1,6 +1,7 @@
 package status
 
 import (
+	"sync"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


edge/pkg/edged/status/status_manager.go
```diff
@@ -73,6 +74,8 @@ func (m *manager) updatePodStatus() {
 			}
 			continue
 		}
+		m.mu.Lock()
+		defer m.mu.Unlock()
```
gy95, line 78: not correct, in range loop using the one lock, it will cause deadlock.

muxuelan: oh, sorry~

edge/pkg/edged/status/status_manager.go
```diff
@@ -21,7 +22,7 @@ import (
 // inherit it's method but refactored Start() function to periodicity update status to IEF
 type manager struct {
 	status.Manager
-	// TODO: consider need lock?
```
gy95, line 24: seems no need to use a lock, we only use this variable in only one goroutine.

PR #3405, opened 2021-11-29
remove "huawei.com/gpu-status" and "huawei.com/gpu-decision" related code

jidalong:
**What type of PR is this?**
/kind feature

"huawei.com/gpu-status" and "huawei.com/gpu-decision" aren't the origin  gpu resources, I guess there are from IEF  and would be better removed in kubeedge.  Only "nvidia.com/gpu" is enough.

NONE

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged_status.go
+++ b/edge/pkg/edged/edged_status.go
@@ -137,22 +137,6 @@ func (e *edged) setInitNode(node *v1.Node) {
 	initNode.Status = *node.Status.DeepCopy()
 }
 
-// Retrieve node status
-func retrieveDevicePluginStatus(s string) (string, error) {
-	tagLen := len(apis.StatusTag)
-	if len(s) <= tagLen {
-		return "", fmt.Errorf("no node status wrapped in")
-	}
-
-	tag := s[:tagLen]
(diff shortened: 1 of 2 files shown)


edge/pkg/edged/edged_status.go
```diff
@@ -171,19 +155,10 @@ func (e *edged) getNodeStatusRequest(node *v1.Node) (*edgeapi.NodeStatusRequest,
...
-				if node.Annotations == nil {
-					node.Annotations = make(map[string]string)
-				}
-				node.Annotations[apis.NvidiaGPUStatusAnnotationKey] = status
-				klog.Infof("Setting node annotation to add node status list to Scheduler")
-				continue
-			}
+		if nameSet.Has(removedResource) {
+			klog.Infof("Remove capacity for %s", removedResource)
+			delete(node.Status.Capacity, v1.ResourceName(removedResource))
```
fisherxu, line 160: We should set the resource to 0 instead remove them? ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/nodestatus/setters.go#L349

fisherxu: Ping  :)

PR #3409, opened 2021-11-29
Remove dependency on os/exec and curl in favor of net/http

### LINKED ISSUE #3385: Refactor sh -c curl --> Go net.http
https://github.com/kubeedge/kubeedge/blob/8629fe4de73aa985d5383121b6020b01299cd27c/keadm/cmd/keadm/app/cmd/util/common.go#L167-L177

In this function, there's no need to depend on the shell and curl being installed when we can just use Go's net.http instead.

**Why is this needed**:
This removes a dependency on curl and furthermore we should probably not be doing `curl -k` anyway. I would propose that we do verify the TLS certificate in the net.http implementation to lower the risk of MITM attacks.

mjlshen:
**What type of PR is this?**
/kind cleanup

Removes a dependency on curl being installed and the insecure check (-k/--insecure) flag to mitigate the risk of man-in-the-middle attacks.

Fixes #3385

```release-note
NONE
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -23,8 +23,8 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -166,15 +166,22 @@ func RunningModule() (types.ModuleRunning, error) {
...
+	if resp.StatusCode == http.StatusOK {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return "", err
+		}
+		return string(body), nil
 	}
 
-	return string(latestReleaseData), nil
+	return "", fmt.Errorf("failed to get latest version from %s, expected %d, got status code: %d", latestReleaseVersionURL, http.StatusOK, resp.StatusCode)
```
gy95, line 184: /lgtm
nit: whether better adjust the sequence of two blocks, just like below:
```
if resp.StatusCode != http.StatusOK {
    return "", fmt.Errorf("failed to get latest version from %s, expected %d, got status code: %d", latestReleaseVersionURL, http.StatusOK, resp.StatusCode)
}

body, err := ioutil.ReadAll(resp.Body)
if err != nil {
    return "", err
}
return string(body), nil
```

/assign @fisherxu final decision

zc2638: Agree

mjlshen: Sure! Easy to change 😄

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -166,15 +166,22 @@ func RunningModule() (types.ModuleRunning, error) {
...
+		return "", fmt.Errorf("failed to get latest version from %s: %v", latestReleaseVersionURL, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("failed to get latest version from %s, expected %d, got status code: %d", latestReleaseVersionURL, http.StatusOK, resp.StatusCode)
 	}
 
-	return string(latestReleaseData), nil
+	body, err := ioutil.ReadAll(resp.Body)
```
zc2638, line 180: Could use `io.ReadAll` instead of `ioutil.ReadAll`?
After golang 1.16, it is more recommended to use `io` instead of `ioutil`

mjlshen: Done! I found another ioutil call in the same file as well to fix as well

zc2638:
Thanks for fixing.
/lgtm 
/approve

PR #3412, opened 2021-11-30
Optimize script when create stream cert

gujun4990:
Signed-off-by: jun.gu <gujun4990@outlook.com>

### DIFF
--- a/build/tools/certgen.sh
+++ b/build/tools/certgen.sh
@@ -67,8 +67,9 @@ stream() {
     readonly STREAM_KEY_FILE=${certPath}/stream.key
     readonly STREAM_CSR_FILE=${certPath}/stream.csr
     readonly STREAM_CRT_FILE=${certPath}/stream.crt
-    readonly K8SCA_FILE=/etc/kubernetes/pki/ca.crt
-    readonly K8SCA_KEY_FILE=/etc/kubernetes/pki/ca.key
+
+    readonly K8SCA_FILE=${K8SCA_FILE:-/etc/kubernetes/pki/ca.crt}
+    readonly K8SCA_KEY_FILE=${K8SCA_KEY_FILE:-/etc/kubernetes/pki/ca.key}
 
     if [ -z ${CLOUDCOREIPS} ]; then
         echo "You must set CLOUDCOREIPS Env,The environment variable is set to specify the IP addresses of all cloudcore"
@@ -84,7 +85,7 @@ stream() {
         SUBJECTALTNAME="${SUBJECTALTNAME}IP.${index}:${ip}"
     done
 
-    cp /etc/kubernetes/pki/ca.crt ${caPath}/streamCA.crt
+    cp ${K8SCA_FILE} ${caPath}/streamCA.crt
     echo $SUBJECTALTNAME > /tmp/server-extfile.cnf
 
     openssl genrsa -out ${STREAM_KEY_FILE}  2048


gy95:
welcome @gujun4990 
need sign off the commit to pass the DCO 😄 
ref: https://github.com/kubeedge/kubeedge/pull/3412/checks?check_run_id=4364943896

build/tools/certgen.sh
```diff
@@ -67,8 +67,9 @@ stream() {
     readonly STREAM_KEY_FILE=${certPath}/stream.key
     readonly STREAM_CSR_FILE=${certPath}/stream.csr
     readonly STREAM_CRT_FILE=${certPath}/stream.crt
-    readonly K8SCA_FILE=/etc/kubernetes/pki/ca.crt
-    readonly K8SCA_KEY_FILE=/etc/kubernetes/pki/ca.key
+
+    readonly k8sCaFile=${K8SCA_FILE:-/etc/kubernetes/pki/ca.crt}
+    readonly k8sCaKeyFile=${K8SCA_KEY_FILE:-/etc/kubernetes/pki/ca.key}
```
gy95, line 72: why need to fix this? IMHO, K8SCA_FILE is never set. :)

gujun4990: Consider the path of k8sCA maybe different, so optimize the code. If the file is never set, maybe the patch is not necessary.

zc2638: Maintain uniform variable naming rules?

gujun4990: Modify k8sCaFile to K8SCA_FILE, k8sCaKeyFile to K8SCA_KEY_FILE? Like this?

zc2638: yeah, You can also change it all into humps.

gujun4990: OK, I will fix this.

build/tools/certgen.sh
```diff
@@ -67,8 +67,9 @@ stream() {
     readonly STREAM_KEY_FILE=${certPath}/stream.key
     readonly STREAM_CSR_FILE=${certPath}/stream.csr
     readonly STREAM_CRT_FILE=${certPath}/stream.crt
-    readonly K8SCA_FILE=/etc/kubernetes/pki/ca.crt
-    readonly K8SCA_KEY_FILE=/etc/kubernetes/pki/ca.key
+
+    readonly K8SCA_FILE=${K8SCA_FILE:-/etc/kubernetes/pki/ca.crt}
+    readonly K8SCA_KEY_FILE=${K8SCA_KEY_FILE:-/etc/kubernetes/pki/ca.key}
```
gy95, line 72: this is redundant, just readonly K8SCA_FILE=/etc/kubernetes/pki/ca.crt, K8SCA_FILE is first defined.

build/tools/certgen.sh
```diff
@@ -84,7 +85,7 @@ stream() {
         SUBJECTALTNAME="${SUBJECTALTNAME}IP.${index}:${ip}"
     done
 
-    cp /etc/kubernetes/pki/ca.crt ${caPath}/streamCA.crt
+    cp ${K8SCA_FILE} ${caPath}/streamCA.crt
```
gy95, line 88: this is reasonable :)

build/tools/certgen.sh
gujun4990: If having the code, It's necessary to set K8SCA_FILE and K8SCA_KEY_FILE environment.

gy95: WDYT @zc2638

fisherxu (commented):
/approve
Leave @zc2638  for final check.

build/tools/certgen.sh
zc2638: It makes sense, and I am using it too. 
Many private environments, their certificates are not in the default k8s path.
The script is executed by changing the required path by defining variables.
@gy95

fisherxu (commented):
/lgtm

PR #3413, opened 2021-11-30
beehive: fixes SendSync channel allocation behaviour

### LINKED ISSUE #3016: Deadlock occurs using beehiveContext.SendSync
**What happened**:
Deadlock occurs using beehiveContext.SendSync，Have a small probability
Resolve deadlock
using beehiveContext.SendSync，But the cloudcore response time is just the critical point of the timeout time

**Environment**:
- Kubernetes version (use `kubectl version`):v1.20.4
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`):v1.7.1
- 
**solution**:
Problem code block：github.com/kubeedge/beehive/pkg/core/context/context_channel.go

```go
 func (ctx *ChannelContext) SendResp(message model.Message) {
 	anonName := getAnonChannelName(message.GetParentID())
 
 	ctx.anonChsLock.RLock()
 	defer ctx.anonChsLock.RUnlock()
 	if channel, exist := ctx.anonChannels[anonName]; exist {
 		channel <- message
 		return
 	}
 
 	klog.V(4).Infof("Get bad anonName:%s when sendresp message, do nothing", anonName)
 }
```
Personal solution：
```go
 func (ctx *ChannelContext) SendResp(message model.Message) {
 	anonName := getAnonChannelName(message.GetParentID())
 	respTimer := time.NewTimer(MessageTimeoutDefault)
 	ctx.anonChsLock.RLock()
 	defer ctx.anonChsLock.RUnlock()
 	if channel, exist := ctx.anonChannels[anonName]; exist {
 		select {
		case channel <- message:
 		case <- respTimer.C:
 			klog.Warningf("the messages ParentID:%s return resp timeout", anonName)
 		}
 		return
 	}
 
 	klog.V(4).Infof("Get bad anonName:%s when sendresp message, do nothing", anonName)
}
```

ankitrgadiya:
**What type of PR is this?**
/kind bug

The Beehive Channel implementation for SendSync sends the message first
and then allocates the anonymous channel to receive the response. If it
so happens that the responding module is able to send response before
the channel is allocated then it won't find the channel in the map.

This results in the SendResp function just printing a log and move
forward. But the SendSync function will get stuck until the Timeout is
reached. This will lead to the caller Module not receiving the message
at all.

This commit fixes this by pre-allocating the channel and saving in the
map before it sends the request message.

Fixes #3016 

**Does this PR introduce a user-facing change?**: None

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go
@@ -121,14 +121,6 @@ func (ctx *ChannelContext) SendSync(module string, message model.Message, timeou
 		return model.Message{}, fmt.Errorf("bad request module name(%s)", module)
 	}
 
-	sendTimer := time.NewTimer(timeout)
-	select {
-	case reqChannel <- message:
-	case <-sendTimer.C:
-		return model.Message{}, errors.New("timeout to send message")
-	}
-	sendTimer.Stop()
-
 	// new anonymous channel for response
 	anonChan := make(chan model.Message)
 	anonName := getAnonChannelName(message.GetID())
@@ -142,14 +134,18 @@ func (ctx *ChannelContext) SendSync(module string, message model.Message, timeou
 		ctx.anonChsLock.Unlock()
 	}()
 
+	select {
+	case reqChannel <- message:
+	case <-time.After(timeout):
+		return model.Message{}, errors.New("timeout to send message")
+	}
+
 	var resp model.Message
-	respTimer := time.NewTimer(time.Until(deadline))
 	select {
 	case resp = <-anonChan:
-	case <-respTimer.C:
+	case <-time.After(time.Until(deadline)):
 		return model.Message{}, errors.New("timeout to get response")
 	}
-	respTimer.Stop()
 
 	return resp, nil
 }


gy95:
welcome @ankitrgadiya Thanks for your fixing. 😄 
LGTM

IMHO the previous codes will also not result in deadlock? This patch is to solve the resp channel not found  bug?

ankitrgadiya:
@gy95 I think Deadlock is just slightly incorrect description in the original issue. But it defines the same problem of channel not found.

gy95:
thanks a lot
/assign @zc2638 @fisherxu

staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go
```diff
@@ -142,6 +134,14 @@ func (ctx *ChannelContext) SendSync(module string, message model.Message, timeou
 		ctx.anonChsLock.Unlock()
 	}()
 
+	sendTimer := time.NewTimer(timeout)
+	select {
+	case reqChannel <- message:
+	case <-sendTimer.C:
```
zc2638, line 140: Although this situation basically does not happen, but it is a great patch.

By the way, can you use `time.After` instead of `time.NewTimer` to streamline the code?

ankitrgadiya: I've made that change.

gy95: now there's one `do-not-merge/invalid-commit-message` label, need to fix this

gy95: kindly ping @ankitrgadiya , please squash the commits to one, and simplify the commit message to `beehive: fixes SendSync channel allocation behaviour`

ankitrgadiya: Got it, I'll make the change

ankitrgadiya: @zc2638 @gy95 Can you please re-review the changes?

zc2638:
Please wait for @fisherxu

fisherxu (commented):
/approve
Sorry for the delay.

PR #3415, opened 2021-12-01
fix deviceTwin device update fields

TianTianBigWang:

**What type of PR is this?**
/kind bug

The first err is not processed

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/device_test.go
+++ b/edge/pkg/devicetwin/dtmanager/device_test.go
@@ -237,7 +237,7 @@ func TestDealDeviceStateUpdate(t *testing.T) {
 			updateReturnInt:  int64(1),
 			updateReturnErr:  nil,
 			queryTableReturn: querySeterMock,
-			times:            2,
+			times:            1,
 		},
 	}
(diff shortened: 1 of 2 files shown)


gy95:
/lgtm
but need to fix test error

gy95:
maybe too much modification, IMHO, we just fix a little, and maybe ut just need a little fix, now ut fix too much.

edge/pkg/devicetwin/dtmanager/device_test.go
```diff
@@ -237,14 +244,13 @@ func TestDealDeviceStateUpdate(t *testing.T) {
 			updateReturnInt:  int64(1),
 			updateReturnErr:  nil,
 			queryTableReturn: querySeterMock,
-			times:            2,
```
gy95, line 240: I try to fix this from 2 to 1, and ut passed.

TianTianBigWang: 😲seem that you're right

fisherxu (commented):
/approve

PR #3417, opened 2021-12-01
fix pod exec is not close conn when edgecore is closed

lvfei103650:
**What type of PR is this?**
/kind bug

When we use the exec pod function, when edgecore fails (for example, systemctl stop edgecore), exec will hang, the command line cannot be entered, and the tcp connection between apiserver and cloudcore's 10003 port is still connected.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudstream/containerexec_connection.go
+++ b/cloud/pkg/cloudstream/containerexec_connection.go
@@ -123,6 +123,7 @@ func (c *ContainerExecConnection) Serve() error {
 			msg := stream.NewMessage(connector.GetMessageID(), stream.MessageTypeData, data[:n])
 			if err := c.WriteToTunnel(msg); err != nil {
 				klog.Errorf("%s failed to write to tunnel server, err: %v", c.String(), err)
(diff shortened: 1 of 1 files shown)


lvfei103650:
exec_hang
企业微信截图_16383497762296
the connection between apiserver and cloudcore:10003 canot closed


image

if edgecore is stoped, the second for{} cannot break and exit, so this case <-c.EdgePeerDone()  can't enter

so we add break for{} when c.WriteToTunnel(msg) is failed.

test reult:
exec_ok

企业微信截图_1638349804475

fisherxu (commented):
Thanks for the fixing, looks good, and have you tried that if edgecore stopped and then restarted, will the terminal exit when type something?

fisherxu (commented):
/lgtm
/approve

lvfei103650:
> Thanks for the fixing, looks good, and have you tried that if edgecore stopped and then restarted, will the terminal exit when type something?

yes,the terminal exit when type something

PR #3419, opened 2021-12-02
add label noproxy=edgemesh in the kubeedge-admission-service

pengbinbin1:
**What type of PR is this?**
/kind feature

when i deploy edgemesh on my service,I found that I can not visit the kubeedge-admission-service.I checked the log and found the request is captured by the edgemesh, and edgemesh can not support the https protocol. We can add the label noproxy=edgemesh so the edgemesh will ingnore the kubeedge-admission-service.

### DIFF
--- a/build/admission/service.yaml
+++ b/build/admission/service.yaml
@@ -3,6 +3,7 @@ kind: Service
 metadata:
   labels:
     app: kubeedge-admission
+    noproxy: edgemesh
   name: kubeedge-admission-service
   namespace: kubeedge
 spec:


zc2638:
I think not everyone uses `edgemesh`, you can build specific style components in your own way.
This can ensure the versatility of community components.

build/admission/service.yaml
```diff
@@ -3,6 +3,7 @@ kind: Service
 metadata:
   labels:
     app: kubeedge-admission
+    noproxy: edgemesh
```
fisherxu, line 6: @Poorunga Ptal of this.

Poorunga: EdgeMesh can support https transmission, please refer to: https://edgemesh.netlify.app/zh/guide/test-case.html#https. And i agree with @zc2638 suggestion

pengbinbin1: Yes,i think so

PR #2812, opened 2021-04-25
synccontroller: trigger the delete operation when the relate resource does not exist

zc2638:
Signed-off-by: zc <ce.zheng@daocloud.io>

**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
 When the edge side communication is disconnected and the pod is restarted, edge cannot receive the request and confirm. The historical pod is deleted and a new pod is generated. When the edge is ready, the pod will be distributed and a new object resource will be generated. The historical resource redundancy is only judged to skip and should be deleted.
 Edgecontroller manages object resources. It should handle invalid object resources.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3420, opened 2021-12-02
delete objectsync when obj not found

chenchunxiu:
Signed-off-by: chenchunxiu <chenchunxiu_yewu@cmss.chinamobile.com>

**What type of PR is this?**
/kind bug

When some object has already been deleted, but objectsync still remains, we should delete the objectsync

Fixes #

```release-note

```


zc2638:
This is not appropriate if the resources at the edge are still running.
I think the PR(#2812) operation process is more reasonable. Should we reopen?

chenchunxiu:
> This is not appropriate if the resources at the edge are still running. I think the PR(#2812) operation process is more reasonable. Should we reopen?

I think PR(#2812) is better.

PR #3435, opened 2021-12-07
integration test: for pod status, both Running and Succeeded are OK

gy95:
**What type of PR is this?**
/kind test

encountered one integration test failure, error logs are as below.
```
• Failure [253.012 seconds]
Application deployment in edgecore Testing
/home/runner/work/kubeedge/kubeedge/edge/test/integration/appdeployment/application_test.go:39
  Test application deployment and delete deployment
  /home/runner/work/kubeedge/kubeedge/edge/test/integration/appdeployment/application_test.go:41
    TC_TEST_APP_DEPLOYMENT_5: Test application deployment delete from edgecore [It]
    /home/runner/work/kubeedge/kubeedge/edge/test/integration/appdeployment/application_test.go:100

    Timed out after 240.002s.
    Application Deployment is Unsuccessful, Pod has not come to Running State
    Expected
        <string>: Succeeded
    to equal
        <string>: Running

    /home/runner/work/kubeedge/kubeedge/edge/test/integration/utils/helpers/helpers.go:363
```

and according to the official comment, for pod status, both `Succeeded` and `Running` are OK.
```
// The phase of a Pod is a simple, high-level summary of where the Pod is in its lifecycle.
// The conditions array, the reason and message fields, and the individual container status
// arrays contain more detail about the pod's status.
// There are five possible phase values:
//
// Pending: The pod has been accepted by the Kubernetes system, but one or more of the
// container images has not been created. This includes time before being scheduled as
// well as time spent downloading images over the network, which could take a while.
// Running: The pod has been bound to a node, and all of the containers have been created.
// At least one container is still running, or is in the process of starting or restarting.
// Succeeded: All containers in the pod have terminated in success, and will not be restarted.
// Failed: All containers in the pod have terminated, and at least one container has
// terminated in failure. The container either exited with non-zero status or was terminated
// by the system.
// Unknown: For some reason the state of the pod could not be obtained, typically due to an
// error in communicating with the host of the pod.
//
// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#pod-phase
// +optional
Phase PodPhase `json:"phase,omitempty" protobuf:"bytes,1,opt,name=phase,casttype=PodPhase"`
```

So in test, both `Running` and `Succeeded` are OK

Fixes #

```release-note

```

### DIFF
--- a/edge/test/integration/utils/helpers/helpers.go
+++ b/edge/test/integration/utils/helpers/helpers.go
@@ -21,6 +21,7 @@ import (
 	"crypto/tls"
 	"database/sql"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
The pod images are `nginx` and `redis`, in theory it should not be `Succeed`, not sure why the `Succeed` occurs here.

edge/test/integration/utils/helpers/helpers.go
```diff
@@ -349,18 +350,20 @@ func GetPods(EdgedEndpoint string) (v1.PodList, error) {
...
-	gomega.Eventually(func() string {
-		var status string
+	gomega.Eventually(func() error {
 		pods, _ := GetPods(EdgedEndPoint)
 		for index := range pods.Items {
 			pod := &pods.Items[index]
 			if podname == pod.Name {
-				status = string(pod.Status.Phase)
 				common.Infof("PodName: %s PodStatus: %s", pod.Name, pod.Status.Phase)
+				if pod.Status.Phase == v1.PodRunning || pod.Status.Phase == v1.PodSucceeded {
```
fisherxu, line 359: For redis and nginx, `PodSucceeded` is the wrong Status, so close this first.
/close

PR #3437, opened 2021-12-07
add rules to 01-ha-prepare.yaml for cloudcore ha deployment

Rachel-Shao:
**What type of PR is this?**
/kind bug

When I tried to deploy cloudcore high availability, I needed to make some `list/watch/get` requests for crd resources, but I got an error:
When I execute: `curl http://127.0.0.1:10550/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions` in edgenode
I get:
~~~
W1207 14:25:50.920859   16813 storage.go:135] [metaserver/reststorage] failed to list obj from cloud, successfully to add listener but failed to get current list, customresourcedefinitions.apiextensions.k8s.io is forbidden: User "system:serviceaccount:kubeedge:cloudcore" cannot list resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope; try local
I1207 14:25:50.920995   16813 store.go:80] get a list req, key=/apiextensions.k8s.io/v1beta1/customresourcedefinitions/null/null
I1207 14:25:50.921054   16813 imitator.go:142] apiextensions.k8s.io/v1beta1, Resource=customresourcedefinitions,,
E1207 14:25:50.921318   16813 store.go:93] <nil>
I1207 14:25:50.921336   16813 storage.go:140] [metaserver/reststorage] successfully process list req (/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions) at local
I1207 14:26:19.918482   16813 edged.go:944] worker [4] get pod addition item [kube-flannel-ds-7pcgc]
~~~

This seems to be because there is no operation authority assigned to the `customresourcedefinitions` resource of the `apiextensions.k8s.io` group, so I added relevant fields in the deployment template yaml

none

none

none

### DIFF
--- a/build/cloud/03-clusterrole.yaml
+++ b/build/cloud/03-clusterrole.yaml
@@ -30,3 +30,6 @@ rules:
 - apiGroups: ["rules.kubeedge.io"]
   resources: ["rules", "ruleendpoints", "rules/status", "ruleendpoints/status"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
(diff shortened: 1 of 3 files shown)


gy95:
we need to fix here also:
https://github.com/kubeedge/kubeedge/blob/4ff2d99da342e3623141665696f23ca85e12d675/build/helm/cloudcore/templates/rbac_cloudcore.yaml#L31-L35

and maybe here 😂 
https://github.com/kubeedge/kubeedge/blob/master/build/cloud/03-clusterrole.yaml

afterwards, we should unify these deployments from different places to the only one 😂

fisherxu (commented):
/lgtm
/approve
We can use the helm, and remove the files under `build/cloud` in future.

PR #3441, opened 2021-12-07
enable golangci-lint for codes under staging

gy95:
**What type of PR is this?**
/kind cleanup

now we only check format issue under staging dir, this patch will enable golangci-lint checks under staging dir. and fix related lint errors.

Fixes #

```release-note

```

### DIFF
--- a/hack/lib/lint.sh
+++ b/hack/lib/lint.sh
@@ -21,6 +21,7 @@ set -o nounset
 set -o pipefail
 
 SED_CMD=""
+KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 
 if [[ "$OSTYPE" == "darwin"* ]]
 then
@@ -69,7 +70,7 @@ kubeedge::lint::check() {
     echo "check any issue by golangci-lint ..."
     GOOS="linux" golangci-lint run -v
 
-    # only check format issue under staging dir
-    echo "check any issue under staging dir by gofmt ..."
-    gofmt -l -w staging
+    # check codes under staging dir, this will also use .golangci.yaml in the {KUBEEDGE_ROOT} dir
+    cd "${KUBEEDGE_ROOT}/staging/src/github.com/kubeedge/beehive" && GOOS="linux" golangci-lint run -v
+    cd "${KUBEEDGE_ROOT}/staging/src/github.com/kubeedge/viaduct" && GOOS="linux" golangci-lint run -v
 }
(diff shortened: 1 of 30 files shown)


hack/lib/lint.sh
```diff
@@ -69,7 +70,7 @@ kubeedge::lint::check() {
     echo "check any issue by golangci-lint ..."
     GOOS="linux" golangci-lint run -v
 
-    # only check format issue under staging dir
-    echo "check any issue under staging dir by gofmt ..."
-    gofmt -l -w staging
```
fisherxu, line 74: The golangci-lint will auto skip the staging dir?

fisherxu: Can open an issue in upstream to check the issue why golangci-lint skip the files under staging.

gy95: https://github.com/golangci/golangci-lint/issues/2409

gy95: https://github.com/golangci/golangci-lint/issues/2409#issuecomment-991550432

gy95: according to the golangci-lint upstream community, golangci-lint works as Go: it's module aware.
So it will not check beehive module and viaduct module, we need to check those manually :)

staging/src/github.com/kubeedge/beehive/pkg/core/socket/context_socket.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/kubeedge/beehive/pkg/core/socket/wrapper"
 )
 
+//nolint:revive
```
fisherxu, line 20: Why skip the check?

gy95: `revive` check recommend us to fix structure name from `SocketContext` to `Context`
Because `SocketContext` contains package name `socket`, it's recommended to remove prefix `socket`

fisherxu: Can we add a space like "// nolint:revive"? So keep same with other comments.

gy95: fixed.

fisherxu (commented):
Looks like there're too many `//nolint:xx`..

gy95:
all nolint fixed. 😊

fisherxu (commented):
/lgtm
/approve

PR #3455, opened 2021-12-10
keadm: support the installation of cloudcore in systemd

zc2638:
**What type of PR is this?**

/kind feature

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
+++ b/keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
@@ -3,6 +3,7 @@ package util
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/kubeedge/kubeedge/common/constants"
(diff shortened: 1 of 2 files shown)


gy95:
To stop the cloudcore, we maybe also need to modify the function synchronously.
maybe here 😂 
https://github.com/kubeedge/kubeedge/blob/c3dd45071f5963ece6046528ad03076aa2fe9a42/keadm/cmd/keadm/app/cmd/util/common.go#L365-L373

zc2638:
> To stop the cloudcore, we maybe also need to modify the function synchronously. maybe here 😂
> 
> https://github.com/kubeedge/kubeedge/blob/c3dd45071f5963ece6046528ad03076aa2fe9a42/keadm/cmd/keadm/app/cmd/util/common.go#L365-L373

I will do it.

fisherxu (commented):
I think we can modify to install cloudcore as container?

fisherxu:
Now we already support helm installation, is it possible to use helm in keadm？

zc2638:
Perhaps helm has become a standard for k8s application deployment, 
but I think there are still a large number of people who don’t use helm or even understand helm. 
And some people don’t like to use helm.

More discussion may be needed here, and keadm may be changed widely?

fisherxu:
> More discussion may be needed here, and keadm may be changed widely?

Yes, I think whether using helm or not, cloudcore should be deployed as ccontainer, how do you think? :) @zc2638 @zhu733756

zhu733756:
> > More discussion may be needed here, and keadm may be changed widely?
> 
> Yes, I think whether using helm or not, cloudcore should be deployed as ccontainer, how do you think? :) @zc2638 @zhu733756

I think container deployment is preferred for security reasons, However, as an open-source project, we can offer a variety of alternatives.

zc2638:
My idea is that `kubeedge` is based on `k8s`, so we can support native resource deployment and helm deployment at the same time.
Process deployment may not be preferred here, but we can provide responsive documents to support process deployment.

zc2638:
will refactor keadm in #3475.
/close

PR #3343, opened 2021-11-23
Csi driver bug fix

lvchenggang:
Signed-off-by: lvchenggang lvchenggang_yewu@cmss.chinamobile.com

What type of PR is this?
/kind bug

What this PR does / why we need it:
The socket communication between csidriver and cloudhub's udsserver not encrypted by base64 algorithm.
Currently processing way will affect the response chan about csidriver->provisioner->k8s apiserver.
More details can see issue: #3336 

Which issue(s) this PR fixes:
Fixes #3336 

Does this PR introduce a user-facing change?:
NONE

Issue #3336, opened 2021-11-23
Csi driver report error: "illegal base64 data at input byte 0"


lvchenggang:
Step1: Use Dockerfile exist in build/csidriver to build csidriver image used in cloud at project root directory.
![image](https://user-images.githubusercontent.com/68538306/142959573-a706ba05-e55f-4668-92c8-4290e6157d1e.png)
......
![image](https://user-images.githubusercontent.com/68538306/142959587-3c84cce0-db3f-4679-a6aa-05d0552ee4c0.png)

Step2: Use YAML build/csisamples/hostpath/deploy/csicontroller.yaml to create sts in cloud. Meanwhile replace the csi-driver container's image to above new build image: kubeedge/csidriver:v1.1.1.
![image](https://user-images.githubusercontent.com/68538306/142959852-2eeb60aa-a0ec-49f0-8312-2077dead7c7f.png)

Step3: Create Storageclass and PVC:
![image](https://user-images.githubusercontent.com/68538306/142959937-e80a27b8-2609-437a-a394-b6e99d2f9950.png)
![image](https://user-images.githubusercontent.com/68538306/142959956-1ac11da0-9907-4cb6-9bed-5b3956a64e1a.png)
![image](https://user-images.githubusercontent.com/68538306/142959977-392d4d1b-e20f-4f45-b49f-32aa2c086133.png)

Step4: Watch the log of csicontroller's csi-driver container, found error log:
![image](https://user-images.githubusercontent.com/68538306/142960065-bb2309d2-db6b-4d52-8d8f-13edbf45191d.png)

Phenomenon analysis:
It seems that the uds communication between csidriver and cloudhub's udsserver not encrypted by base64 algorithm.

Impact Analysis:
The error will affect the response chan about csidriver->provisioner->k8s apiserver.

lvchenggang:
It seems that the uds communication between csidriver and cloudhub's udsserver not encrypted by base64 algorithm

moolen:
Hey @lvchenggang, i found a couple of more issues with the CSI integration:
1. cloudhub/channelq drops messages towards edgecore that do not contain a resourceVersion [see here](https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/channelq/channelq.go#L91)
[...]

PR #3458, opened 2021-12-10
edged: fix volume message de-/serialization

moolen:
**What type of PR is this?**
/kind bug

(1) The bug: We can not use `GetContentData()` from above because the embedded value is a string: It is marshaled and now contains a **escaped** json string which can not be unmarshaled later on ([e.g. here](https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edged/edged.go#L1196-L1197))!
(2) We want to be a good citizen and return an error to the caller in case a Volume request fails. Otherwise csidriver sees a timeout. A large csidriver latency causes `external-provisioner` to retry that causes volumes to be provisioned multiple times (will be addressed in a separate PR). external-provisioner timeouts and retry interval [is configurable](https://github.com/kubernetes-csi/external-provisioner).
(3) We take care of serializing the VolumeResponse in a central place using jsonpb. csidriver is changed in #3343 accordingly.

Relates to #3336

This PR breaks the csidriver -> edged api by changing the message serialization. This is addressed in #3343. 

```release-note
NONE
```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -38,6 +38,7 @@ import (
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/protobuf/jsonpb"
+	"github.com/golang/protobuf/proto"
 	cadvisorapi "github.com/google/cadvisor/info/v1"
 	cadvisorapi2 "github.com/google/cadvisor/info/v2"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


fisherxu:
cc @lvchenggang

moolen:
@fisherxu / @lvchenggang can i get feedback on this, please?

moolen:
This is part of #3559, closing this one out.

PR #3459, opened 2021-12-11
drop import of github.com/stretchr/testify

gy95:
**What type of PR is this?**
/kind cleanup

drop import of github.com/stretchr/testify, instead use go native.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/membership_test.go
+++ b/edge/pkg/devicetwin/dtmanager/membership_test.go
@@ -19,11 +19,11 @@ package dtmanager
 import (
 	"encoding/json"
 	"errors"
+	"reflect"
 	"sync"
 	"testing"
 
 	"github.com/golang/mock/gomock"
-	"github.com/stretchr/testify/assert"
 
 	"github.com/kubeedge/beehive/pkg/core/model"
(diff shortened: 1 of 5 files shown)


edge/pkg/devicetwin/dtmanager/membership_test.go
```diff
@@ -87,8 +93,9 @@ func TestDealMembershipDetailInvalidMsg(t *testing.T) {
 	}
 
 	err := dealMembershipDetail(dtc, "t", m)
-	assert.Error(t, err)
-	assert.Equal(t, errors.New("assertion failed"), err)
+	if !reflect.DeepEqual(err, errors.New("assertion failed")) {
```
fisherxu, line 96:
```suggestion
	if err.Error()!="assertion failed" {
```
How about this?

gy95: both are ok, but I think it's better to judge `error` itself, that's more accurate.

fisherxu: /approve

fisherxu (commented):
/lgtm

PR #3472, opened 2021-12-15
fix commonName using const

dsy3502:
using const for string “KubeEdge”

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/join.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join.go
@@ -24,6 +24,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	"github.com/kubeedge/kubeedge/common/constants"
 	types "github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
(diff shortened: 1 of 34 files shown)


gy95:
welcome, :) please signoff the commit to pass the DCO
ref: https://github.com/kubeedge/kubeedge/pull/3472/checks?check_run_id=4533842935

keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -18,6 +18,7 @@ package cmd
 
 import (
 	"fmt"
+	"github.com/kubeedge/kubeedge/common/constants"
```
gy95, line 21: kubeedge own package should be placed in the third part. For example: line 28 29

dsy3502: I've modified him

dsy3502: I've modified it

gy95: also need fix the lint error, and please squash the commits to only one.

zc2638:
/kind cleanup
And please fix the CI error

PR #3418, opened 2021-12-01
support helm ci test

zhu733756:
Signed-off-by: zhu733756 <zhu733756@kubesphere.io>

**What type of PR is this?**
Add helm ci test.
/kind feature

**What this PR does / why we need it**:
Enhance #3304 
**Which issue(s) this PR fixes**:

Part of #3300 

**Special notes for your reviewer**:
```
None
```

**Does this PR introduce a user-facing change?**:

```release-note
None
```

Issue #3300, opened 2021-11-16
Plan for containerized deployment for cloudcore on k8s platform.


zhu733756:
Plans:
- [x] Support helm repo for KubeEdge cloudcore. https://github.com/kubeedge/kubeedge/pull/3304
- [ ] Helm CI test.  #3418 
- [ ] Custom keadm build for the edgecore. https://github.com/kubeedge/kubeedge/pull/3320
- [x] External iptables-manager component integration, considers a complete externality later and needs to optimize the code.  https://github.com/kubeedge/kubeedge/pull/3265

zhu733756:
Can we open this issue? @gy95 @fisherxu, and also take a look at the two uncompleted prs.

Thanks！

zhu733756:
Will  be refactored in https://github.com/kubeedge/kubeedge/issues/3475

PR #3469, opened 2021-12-15
using lateset codes to do keadm_e2e

gy95:
**What type of PR is this?**
/kind cleanup

**What this PR does / why we need it**:
before we use the official release version to do keadm_e2e, that's not right, that's using last release version kubeedge to do e2e test.
we should use the latest codes to build a new release, and then with it do keadm_e2e test
/hold waiting for https://github.com/kubeedge/kubeedge/pull/3467
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3467, opened 2021-12-14
add script for build release

gy95:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
add script for build release
And for simplicity, not taking cross platform in consideration, make release only support compile the release of the local operating system.
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3473, opened 2021-12-16
Helm ci test 

zhu733756:
**What type of PR is this?**
/kind test

Add another patch for #3418,  reusing the keadm e2e test codes to implement helm keadm e2e test. 

Related to #3300 

See #3418

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -169,6 +169,45 @@ jobs:
 
       - run: make keadm_e2e
 
+  helm_keadm_e2e_test:
+    runs-on: ubuntu-18.04
+    timeout-minutes: 30
+    name: Helm Keadm e2e test
+    env:
+      GO111MODULE: on
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.x
+
+      - uses: actions/cache@v2
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+
+      - name: Install dependences
+        run: |
+          command -v ginkgo || go install github.com/onsi/ginkgo/ginkgo@latest
+          go install sigs.k8s.io/kind@v0.11.1
+          curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.19.3/bin/linux/amd64/kubectl && sudo install kubectl /usr/local/bin/kubectl
+          export RELEASE_VERSION=$(wget -qO - https://kubeedge.io/latestversion | cat)
+          sudo wget -qP /etc/kubeedge https://github.com/kubeedge/kubeedge/releases/download/${RELEASE_VERSION}/checksum_kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz.txt
+          sudo wget -qP /etc/kubeedge https://github.com/kubeedge/kubeedge/releases/download/${RELEASE_VERSION}/kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz
+          echo "$(cat /etc/kubeedge/checksum_kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz.txt) /etc/kubeedge/kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz" | sha512sum -c
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.0
+
+      - run: make helm_keadm_e2e
+
   docker_build:
     runs-on: ubuntu-18.04
     timeout-minutes: 30
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -191,10 +191,6 @@ jobs:
           command -v ginkgo || go install github.com/onsi/ginkgo/ginkgo@latest
           go install sigs.k8s.io/kind@v0.11.1
           curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.19.3/bin/linux/amd64/kubectl && sudo install kubectl /usr/local/bin/kubectl
-          export RELEASE_VERSION=$(wget -qO - https://kubeedge.io/latestversion | cat)
-          sudo wget -qP /etc/kubeedge https://github.com/kubeedge/kubeedge/releases/download/${RELEASE_VERSION}/checksum_kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz.txt
-          sudo wget -qP /etc/kubeedge https://github.com/kubeedge/kubeedge/releases/download/${RELEASE_VERSION}/kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz
-          echo "$(cat /etc/kubeedge/checksum_kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz.txt) /etc/kubeedge/kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz" | sha512sum -c
 
       - name: Checkout code
         uses: actions/checkout@v2
--- a/build/helm/cloudcore/templates/rbac_cloudcore.yaml
+++ b/build/helm/cloudcore/templates/rbac_cloudcore.yaml
@@ -8,7 +8,7 @@ metadata:
   {{- end }}
 rules:
 - apiGroups: [""]
-  resources: ["nodes", "nodes/status", "configmaps", "pods", "pods/status", "secrets", "endpoints", "services","persistentvolumes","persistentvolumeclaims"]
+  resources: ["nodes",  "nodes/status", "serviceaccounts/token", "configmaps", "pods", "pods/status", "secrets", "endpoints", "services","persistentvolumes","persistentvolumeclaims"]
   verbs: ["get", "list", "watch", "create", "update"]
 - apiGroups: [""]
   resources: ["namespaces"]
--- a/build/helm/cloudcore/templates/rbac_cloudcore.yaml
+++ b/build/helm/cloudcore/templates/rbac_cloudcore.yaml
@@ -14,7 +14,7 @@ rules:
   resources: ["namespaces"]
   verbs: ["get", "create"]
 - apiGroups: [""]
-  resources: ["pods/status"]
+  resources: ["nodes", "pods/status"]
   verbs: ["patch"]
 - apiGroups: [""]
   resources: ["pods"]
@@ -34,6 +34,9 @@ rules:
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: ["networking.istio.io"]
+  resources: ["*"]
+  verbs: ["get", "list", "watch"]
 
 ---
 apiVersion: v1
--- a/tests/e2e/scripts/helm_keadm_e2e.sh
+++ b/tests/e2e/scripts/helm_keadm_e2e.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2020 The KubeEdge Authors.
+# Copyright 2021 The KubeEdge Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -16,11 +16,13 @@
 
 KUBEEDGE_ROOT=$PWD
 WORKDIR=$(dirname $0)
+HELM_DIR=$KUBEEDGE_ROOT/build/helm
 E2E_DIR=$(realpath $(dirname $0)/..)
+IMAGE_TAG=$(date +%Y%m%d)
 
 function cleanup() {
   sudo pkill edgecore || true
-  sudo pkill cloudcore || true
+  helm uninstall cloudcore -n kubeedge && kubectl delete ns kubeedge  || true
   kind delete cluster --name test
   sudo rm -rf /var/log/kubeedge /etc/kubeedge /etc/systemd/system/edgecore.service $E2E_DIR/keadm/keadm.test $E2E_DIR/config.json
 }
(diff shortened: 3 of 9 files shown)


tests/e2e/scripts/keadm_e2e.sh
```diff
@@ -35,7 +35,7 @@ function prepare_cluster() {
   kind create cluster --name test
 
   echo "wait the control-plane ready..."
-  kubectl wait --for=condition=Ready node/test-control-plane --timeout=60s
```
fisherxu, line 38: Why delete the timeout here? :)

tests/e2e/scripts/keadm_e2e.sh
```diff
@@ -110,11 +110,16 @@ set -Ee
 trap cleanup EXIT
 trap cleanup ERR
 
+echo -e "\nBuild keadm..."
```
fisherxu, line 113:
```suggestion
echo -e "\nBuilding keadm..."
```

tests/e2e/scripts/keadm_e2e.sh
```diff
@@ -110,11 +110,16 @@ set -Ee
...
 build_keadm
 
 export KUBECONFIG=$HOME/.kube/config
+
+echo -e "\nPrepare cluster..."
```
fisherxu, line 118:
```suggestion
echo -e "\nPreparing cluster..."
```

tests/e2e/scripts/keadm_e2e.sh
```diff
@@ -110,11 +110,16 @@ set -Ee
...
 prepare_cluster
 
+echo -e "\nStart kubeedge..."
```
fisherxu, line 121:
```suggestion
echo -e "\nStarting KubeEdge..."
```

tests/e2e/scripts/keadm_e2e.sh
```diff
@@ -110,11 +110,16 @@ set -Ee
...
 start_kubeedge
 
-run_test
+echo -e "\nRun test..."
```
fisherxu, line 124:
```suggestion
echo -e "\nRunning test..."
```

tests/e2e/scripts/keadm_e2e.sh
```diff
@@ -110,11 +110,16 @@ set -Ee
...
+run_test
```
fisherxu, line 125:
```suggestion
run_test

```

tests/e2e/scripts/helm_keadm_e2e.sh
```diff
@@ -0,0 +1,139 @@
...
+
+echo -e "\nBuild keadm..."
+build_keadm
+
+export KUBECONFIG=$HOME/.kube/config
+
+echo -e "\nPrepare cluster..."
+prepare_cluster
+
+echo -e "\nBuild cloud image..."
```
fisherxu, line 132:
```suggestion
echo -e "\nBuilding cloud image..."
```

tests/e2e/scripts/helm_keadm_e2e.sh
```diff
@@ -0,0 +1,139 @@
...
+build_cloud_image
+
+echo -e "\nStart kubeedge..."
+start_kubeedge
+
+echo -e "\nRun test..."
```
fisherxu, line 138:
```suggestion
echo -e "\nRunning test..."
```

tests/e2e/scripts/helm_keadm_e2e.sh
```diff
@@ -0,0 +1,139 @@
...
+
+export KUBECONFIG=$HOME/.kube/config
+
+echo -e "\nPrepare cluster..."
+prepare_cluster
+
+echo -e "\nBuild cloud image..."
+build_cloud_image
+
+echo -e "\nStart kubeedge..."
```
fisherxu, line 135:
```suggestion
echo -e "\nStarting KubeEdge..."
```

tests/e2e/scripts/helm_keadm_e2e.sh
```diff
@@ -0,0 +1,139 @@
...
+set -Ee
+trap cleanup EXIT
+trap cleanup ERR
+
+echo -e "\nBuild keadm..."
+build_keadm
+
+export KUBECONFIG=$HOME/.kube/config
+
+echo -e "\nPrepare cluster..."
```
fisherxu, line 129:
```suggestion
echo -e "\nPreparing cluster..."
```

tests/e2e/scripts/helm_keadm_e2e.sh
```diff
@@ -0,0 +1,139 @@
...
+      echo "Integration suite successfully passed all the tests !!"
+      exit 0
+  fi
+}
+
+set -Ee
+trap cleanup EXIT
+trap cleanup ERR
+
+echo -e "\nBuild keadm..."
```
fisherxu, line 124:
```suggestion
echo -e "\nBuilding keadm..."
```

fisherxu (commented):
Overall looks good, thanks for the great test :)

.github/workflows/main.yaml
```diff
@@ -169,6 +169,45 @@ jobs:
...
+
+      - name: Install dependences
+        run: |
+          command -v ginkgo || go install github.com/onsi/ginkgo/ginkgo@latest
+          go install sigs.k8s.io/kind@v0.11.1
+          curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.19.3/bin/linux/amd64/kubectl && sudo install kubectl /usr/local/bin/kubectl
+          export RELEASE_VERSION=$(wget -qO - https://kubeedge.io/latestversion | cat)
+          sudo wget -qP /etc/kubeedge https://github.com/kubeedge/kubeedge/releases/download/${RELEASE_VERSION}/checksum_kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz.txt
+          sudo wget -qP /etc/kubeedge https://github.com/kubeedge/kubeedge/releases/download/${RELEASE_VERSION}/kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz
+          echo "$(cat /etc/kubeedge/checksum_kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz.txt) /etc/kubeedge/kubeedge-${RELEASE_VERSION}-linux-amd64.tar.gz" | sha512sum -c
```
gy95, line 197: we should use the latest codes to do test, but not the latest release version to do test.

tests/e2e/scripts/helm_keadm_e2e.sh
```diff
@@ -0,0 +1,139 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+KUBEEDGE_ROOT=$PWD
+WORKDIR=$(dirname $0)
+HELM_DIR=$KUBEEDGE_ROOT/build/helm
+E2E_DIR=$(realpath $(dirname $0)/..)
+IMAGE_TAG=$(date +%Y%m%d)
```
gy95, line 21: this is not precise, if we trigger the CI twice in one day, Isn't this conflict with each other, IMHO, we'd better use some others like git commit info.

tests/e2e/scripts/helm_keadm_e2e.sh
```diff
@@ -0,0 +1,139 @@
...
+  # ensure tokensecret is generated
+  while true; do
+      sleep 3
+      kubectl get secret -nkubeedge 2>/dev/null | grep -q tokensecret && break
+  done
+  
+  cd $KUBEEDGE_ROOT
+  export TOKEN=$(sudo _output/local/bin/keadm gettoken --kube-config=$KUBECONFIG)
+  sudo systemctl set-environment CHECK_EDGECORE_ENVIRONMENT="false"
+  sudo -E CHECK_EDGECORE_ENVIRONMENT="false" _output/local/bin/keadm join --token=$TOKEN --cloudcore-ipport=$MASTER_IP:10000 --edgenode-name=edge-node
```
gy95, line 73: this join command will use the latest version release as edgecore, but not the latest codes. we'd better build a new release with the latest codes and use --kubeedge-version flag to specify which new release we will use.

.github/workflows/main.yaml
gy95: and I think this block is redundant, we donot need to download the release manually. Will not helm use images but not binary release?

zhu733756:
Will be fixed.

zhu733756:
@gy95 Thanks for your reply.
Since I can't find the reply places for each review question. I filled all as follows:
1. The cloud side deployed by helm would use images only, the edge side not changed.
2. I think the image tag doesn't matter, it is mainly used in the action test, and we also can use it by overwriting the variable,

The other questions will have a check to reply you.

gy95:
> @gy95 Thanks for your reply. Since I can't find the reply places for each review question. I filled all as follows:
> 
> 1. The cloud side deployed by helm would use images only, the edge side not changed.
> 2. I think the image tag doesn't matter, it is mainly used in the action test, and we also can use it by overwriting the variable,
> 
> The other questions will have a check to reply you.

2. Yes, but what I worried about is that if we trigger two CI test concurrently, it will cause race condition :) maybe the two CI will use the same overwritten image. and it maybe cause confusion.
3. about using latest codes to do test. I've done some tries in https://github.com/kubeedge/kubeedge/pull/3469 and https://github.com/kubeedge/kubeedge/pull/3467. we can do this based on the two PRs. If OK, any ideas about the two PRs?

fisherxu:
Keadm uses latest binary for edgecore is reasonable.  And in futrue, this test will replace the keadm-e2e test once keadm support to use helm. Can we merge this first, and do the `using latest binary for edgecore` in #3469 ?

zhu733756:
> > @gy95 Thanks for your reply. Since I can't find the reply places for each review question. I filled all as follows:
> > 
> > 1. The cloud side deployed by helm would use images only, the edge side not changed.
> > 2. I think the image tag doesn't matter, it is mainly used in the action test, and we also can use it by overwriting the variable,
> > 
> > The other questions will have a check to reply you.
> 
> 2. Yes, but what I worried about is that if we trigger two CI test concurrently, it will cause race condition :) maybe the two CI will use the same overwritten image. and it maybe cause confusion.
> 3. about using latest codes to do test. I've done some tries in [[WIP]using lateset codes to do keadm_e2e #3469](https://github.com/kubeedge/kubeedge/pull/3469) and [add script for build release #3467](https://github.com/kubeedge/kubeedge/pull/3467). we can do this based on the two PRs. If OK, any ideas about the two PRs?

It is just used in kind, have not pushed outside, doesn't matter. The two CI tests I think it was in two isolated environments.

zhu733756:
> Keadm uses latest binary for edgecore is reasonable. And in futrue, this test will replace the keadm-e2e test once keadm support to use helm. Can we merge this first, and do the `using latest binary for edgecore` in #3469 ?

Agree with you, let's move there to discuss it. @gy95 @fisherxu 
I think the mentioned points need to fix in @gy95 's prs.

.github/workflows/main.yaml
fisherxu, line 197: Ref: #3487

tests/e2e/scripts/helm_keadm_e2e.sh
fisherxu: We can just build a kubeedge release package and please it to /etc/kubeedge now..  but `/etc/kubeedge` is not a good place for the package.

fisherxu (commented):
/approve

fisherxu:
Need rebase..

zhu733756:
> Need rebase..

Done.

fisherxu:
Any comments? @gy95

tests/e2e/scripts/helm_keadm_e2e.sh
```diff
@@ -0,0 +1,139 @@
...
+function prepare_cluster() {
+  kind create cluster --name test
+
+  echo "wait the control-plane ready..."
+  kubectl wait --for=condition=Ready node/test-control-plane --timeout=60s
+
+  kubectl create clusterrolebinding system:anonymous --clusterrole=cluster-admin --user=system:anonymous
+
+  # edge side don't support kind cni now, delete kind cni plugin for workaround
+  # kubectl delete daemonset kindnet -nkube-system
```
gy95, line 45: this line is commented out? why.

tests/e2e/scripts/helm_keadm_e2e.sh
zhu733756: Currently, we only test the hostnetwork case, maybe need to test the external iptablesmgr, any ideas?

tests/e2e/scripts/helm_keadm_e2e.sh
zhu733756: In our history tests, it blocks the connection as the comment said "edge side don't support kind cni now". With cloudcore deployed with container type, I think the kind cni would not block it. In additional, I consider using the iptablemgr component outside, in that case, we would use service and the cloudcore will not share hostnetwork.

gy95:
LGTM

build/helm/cloudcore/templates/rbac_cloudcore.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
   {{- end }}
 rules:
 - apiGroups: [""]
-  resources: ["nodes", "nodes/status", "configmaps", "pods", "pods/status", "secrets", "endpoints", "services","persistentvolumes","persistentvolumeclaims"]
+  resources: ["nodes",  "nodes/status", "serviceaccounts/token", "configmaps", "pods", "pods/status", "secrets", "endpoints", "services","persistentvolumes","persistentvolumeclaims"]
   verbs: ["get", "list", "watch", "create", "update"]
```
fisherxu, line 12: We still need `node` patch verb, can just help add ut here. Add a new resource as in https://github.com/kubeedge/kubeedge/pull/3466 or add verbs in L12 looks all fine, Add a new resource maybe better?

zhu733756: > 

OK, I will do the helm CRD patch. The CRDs in build/cloud dir will need @zivAnyvision‘s help

build/helm/cloudcore/templates/rbac_cloudcore.yaml
```diff
@@ -34,6 +34,9 @@ rules:
 - apiGroups: ["apiextensions.k8s.io"]
   resources: ["customresourcedefinitions"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: ["networking.istio.io"]
```
zhu733756, line 37: Todo: maybe we need to do some combined tests with edgemesh.

zhu733756:
Why e2e fail? Is it occurred by accident？ It seems to just add empty whitespace to the former comment with forced-push.

fisherxu:
> Why e2e fail? Is it occurred by accident？ It seems to just add a whitespace to the former comment with forced-push.
> 
> See https://github.com/kubeedge/kubeedge/runs/4590380659?check_suite_focus=true

It should be occurred by accident I think.

zhu733756:
> > Why e2e fail? Is it occurred by accident？ It seems to just add a whitespace to the former comment with forced-push.
> > See https://github.com/kubeedge/kubeedge/runs/4590380659?check_suite_focus=true
> 
> It should be occurred by accident I think.

I agree with you.

fisherxu (commented):
/lgtm

PR #3483, opened 2021-12-20
supplement the configuration of haproxy

allenleu:
add Haproxy configuration items to the document

### DIFF
--- a/build/cloud/ha/README.md
+++ b/build/cloud/ha/README.md
@@ -79,33 +79,37 @@ The `keepalived` configuration we recommend is as following. You can adjust it a
 ! Configuration File for keepalived
 
 global_defs {
-  router_id lb01
-  vrrp_mcast_group4 224.0.0.19
+  vrrp_version 2
+  vrrp_garp_master_delay 1
+  vrrp_garp_master_refresh 1
 }
-# CloudCore
-vrrp_script CloudCore_check {
-  script "/etc/keepalived/check_cloudcore.sh" # the script for health check
+
+vrrp_script check_haproxy_alive {
+  script "/usr/bin/curl -sLk -o /dev/null -w %{response_code} 10.10.102.101:1984" # this is haproxy service port
+  timeout 5
   interval 2
-  weight 2
   fall 2
-  rise 2
+  rise 10
 }
-vrrp_instance CloudCore {
+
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
Thanks for the enhancement, and do we also need to make haproxy highly available using like  keepalived?

allenleu:
> Thanks for the enhancement, and do we also need to make haproxy highly available using like keepalived?

Yes, we still need to use Keepalived.

zhu733756:
It seems a static configuration example.   Can it be generated with a script? Will it also support a profile for the newer keadm? /cc @fisherxu

build/cloud/ha/README.md
```diff
@@ -158,3 +158,33 @@ fi
 
 ```
 
+## haproxy
```
fisherxu, line 161: I think the haproxy need to config with keeppalived together?

fisherxu:
> It seems a static configuration example. Can it be generated with a script? Will it also support a profile for the newly keadm? /cc @fisherxu

For ha deploy, we can provide the static docs. In future we can also support ha deploy in keadm? @zhu733756

zhu733756:
> > It seems a static configuration example. Can it be generated with a script? Will it also support a profile for the newly keadm? /cc @fisherxu
> 
> For ha deploy, we can provide the static docs. In future we can also support ha deploy in keadm? @zhu733756

Maybe it can work by giving a list of ha IPs to do dynamically configure.  What do you think? @fisherxu @allenleu

allenleu:
> > > It seems a static configuration example. Can it be generated with a script? Will it also support a profile for the newly keadm? /cc @fisherxu
> > 
> > 
> > For ha deploy, we can provide the static docs. In future we can also support ha deploy in keadm? @zhu733756
> 
> Maybe it can work by giving a list of ha IPs to do dynamically configure. Cloudcore IPs can get by kubeclient. What do you think? @fisherxu @allenleu

Yes, maybe we can dynamically configure it. But now we add an example to the document, is it easier for users to understand?

build/cloud/ha/README.md
allenleu: > I think the haproxy need to config with keeppalived together?

Yes, I can sort out the configuration information of keepalived.

Issue #3498, opened 2021-12-23
Identify session by node name instead of ip


stingshen:
**What would you like to be added/modified**:

Identify session by node name instead of ip. Currently, session is identified by node ip when exec and get metrics. It breaks if two edge nodes with the same internal ip are added to the cluster. https://github.com/kubeedge/kubeedge/pull/3136 fixes the behavior on logs. Maybe exec and get metrics could be fixed similarly.

**Why is this needed**:

It is possible that two edge nodes have the same ip in kubeedge cluster. When it is the case, `kubectl logs` and `kubectl top` break because the duplicated internal ip is used to get the session.

Kubesphere uses a sophisticated IPAM to solve the issue: <https://segmentfault.com/a/1190000040467814>. However, I think this issue should be fixed on kubeedge side.

sdghchj:
The blog article you mentioned is a sulotion given by my team in China mobile Co.
We used kubesphere 1.3 which integrated kubeedge 1.6  half a year ago and ran into the edge ip conflict issue, so we posted that alternative sulotion to kubeSphere. I am not sure whether kubesphere adopted the sugested solution in their new version.
But, that solution removes the real ip of edge node, which will break the functions of edgemesh, although in our product, we didn't need edgemesh.
Finally, kubeedge fixed it in 1.8.1 in another way, identifying a node session with the unique node name. This solution keeps the real ip of edge node.

PR #3499, opened 2021-12-23
Identify session not by IP of node for exec

stingshen:
Signed-off-by: shenyi <shenyi@cambricon.com>

**What type of PR is this?**

/kind bug

Addresses #3498

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudstream/streamserver.go
+++ b/cloud/pkg/cloudstream/streamserver.go
@@ -105,27 +105,16 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
 	defer func() {
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
-			klog.Errorf(err.Error())
+			klog.Errorf("Failed to get container logs, err: %v", err)
 		}
 	}()
 
-	// extract pod namespace and pod name from request
-	meta := strings.Split(r.Request.URL.Path, "/")
-	namespaceName := meta[2]
-	podName := meta[3]
-
-	kubeClient := client.GetKubeClient()
-	if kubeClient == nil {
-		klog.Exitf("cannot get kube client\n")
-		return
-	}
-
-	pod, err := kubeClient.CoreV1().Pods(namespaceName).Get(context.Background(), podName, v1.GetOptions{})
+	sessionKey, err := s.getSessionKey(r.Request.URL.Path)
 	if err != nil {
-		klog.Errorf("get pod %s/%s failed: %v\n", namespaceName, podName, err)
+		err = fmt.Errorf("can not get session key: %v", err)
 		return
 	}
-	sessionKey := pod.Spec.NodeName
+
 	session, ok := s.tunnel.getSession(sessionKey)
 	if !ok {
 		err = fmt.Errorf("can not find %v session ", sessionKey)
@@ -172,7 +161,7 @@ func (s *StreamServer) getMetrics(r *restful.Request, w *restful.Response) {
 	defer func() {
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
-			klog.Errorf(err.Error())
+			klog.Errorf("Failed to get metrics, err: %v", err)
 		}
 	}()
 
@@ -224,11 +213,15 @@ func (s *StreamServer) getExec(request *restful.Request, response *restful.Respo
 	defer func() {
 		if err != nil {
 			response.WriteHeader(http.StatusInternalServerError)
-			klog.Errorf(err.Error())
+			klog.Errorf("Failed to get exec, err: %v", err)
 		}
 	}()
 
-	sessionKey := strings.Split(request.Request.Host, ":")[0]
+	sessionKey, err := s.getSessionKey(request.Request.URL.Path)
+	if err != nil {
+		err = fmt.Errorf("can not get session key: %v", err)
+		return
+	}
 	session, ok := s.tunnel.getSession(sessionKey)
 	if !ok {
 		err = fmt.Errorf("exec: can not find %v session ", sessionKey)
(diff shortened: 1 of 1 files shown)


gy95 (commented):
LGTM

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -228,7 +228,23 @@ func (s *StreamServer) getExec(request *restful.Request, response *restful.Respo
...
+		klog.Exitf("cannot get kube client\n")
+		return
+	}
+
+	pod, err := kubeClient.CoreV1().Pods(namespaceName).Get(context.Background(), podName, v1.GetOptions{})
+	if err != nil {
+		klog.Errorf("get pod %s/%s failed: %v\n", namespaceName, podName, err)
+		return
+	}
+	sessionKey := pod.Spec.NodeName
```
fisherxu, line 247: Looks good, thanks for the fixing, can we wrap a function to get the sessionKey? So we can reuse it in `logs/exec/metrics`.

stingshen: done. But only logs/exec is modified. I do not have a good idea for metrics. Most of metrics url doesn't contain node name:

image

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -109,23 +109,12 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
...
-	if kubeClient == nil {
-		klog.Exitf("cannot get kube client\n")
-		return
-	}
-
-	pod, err := kubeClient.CoreV1().Pods(namespaceName).Get(context.Background(), podName, v1.GetOptions{})
+	sessionKey, err := s.getSessionKey(r.Request.URL.Path)
 	if err != nil {
-		klog.Errorf("get pod %s/%s failed: %v\n", namespaceName, podName, err)
+		err = fmt.Errorf("can not get session key: %v", err)
```
fisherxu, line 114:
```suggestion
		klog.Errorf("Failed to get session key, err: %v", err)
```
We need to print the err log?

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -228,7 +217,11 @@ func (s *StreamServer) getExec(request *restful.Request, response *restful.Respo
 		}
 	}()
 
-	sessionKey := strings.Split(request.Request.Host, ":")[0]
+	sessionKey, err := s.getSessionKey(request.Request.URL.Path)
+	if err != nil {
+		err = fmt.Errorf("can not get session key: %v", err)
```
fisherxu, line 222: ditto

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -282,6 +275,29 @@ func (s *StreamServer) getExec(request *restful.Request, response *restful.Respo
...
+	if len(meta) < 4 {
+		return "", fmt.Errorf("can not get pod name from url path: %s", urlPath)
+	}
+	namespaceName := meta[2]
+	podName := meta[3]
+
+	kubeClient := client.GetKubeClient()
+	if kubeClient == nil {
+		klog.Exitf("can not get kube client\n")
+		return "", nil
```
fisherxu, line 290:
```suggestion
		return "", fmt.Errorf("can not get kube client")
```

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -282,6 +275,29 @@ func (s *StreamServer) getExec(request *restful.Request, response *restful.Respo
...
+	}
+
+	pod, err := kubeClient.CoreV1().Pods(namespaceName).Get(context.Background(), podName, v1.GetOptions{})
+	if err != nil {
+		klog.Errorf("get pod %s/%s failed: %v\n", namespaceName, podName, err)
+		return "", err
```
fisherxu, line 296:
```suggestion
		return "", fmt.Errorf("get pod %s/%s failed: %v", namespaceName, podName, err)
```

fisherxu (commented):
Overall looks good!

cloud/pkg/cloudstream/streamserver.go
stingshen: The err log seems to be printed at defer statement?

https://github.com/kubeedge/kubeedge/blob/0ec91cf419ddfac72da8d840ee1c03018ffc76b5/cloud/pkg/cloudstream/streamserver.go#L105-L110

cloud/pkg/cloudstream/streamserver.go
stingshen: done

cloud/pkg/cloudstream/streamserver.go
stingshen: done

cloud/pkg/cloudstream/streamserver.go
fisherxu: The error will not be print in func `getSessionKey`, it should be printed here.
And if we don't print here, what's the usage of this line `err = fmt.Errorf("can not get session key: %v", err)`? and the next line is return :)

stingshen: The error will be printed in the `defer` clause:

https://github.com/kubeedge/kubeedge/blob/0ec91cf419ddfac72da8d840ee1c03018ffc76b5/cloud/pkg/cloudstream/streamserver.go#L103-L116

I did an experiment locally and confirm the behavior that the err would be printed at https://github.com/kubeedge/kubeedge/blob/0ec91cf419ddfac72da8d840ee1c03018ffc76b5/cloud/pkg/cloudstream/streamserver.go#L108

Similarly, other existing error reporting logic has the same pattern. For example: 

https://github.com/kubeedge/kubeedge/blob/0ec91cf419ddfac72da8d840ee1c03018ffc76b5/cloud/pkg/cloudstream/streamserver.go#L128

I don't know why the error printing logic is designed like this previously. But it works.

fisherxu: Ohh, sorry, I missed the log in defer which you also mentioned before...  And can you fix the error meessage, it need starts with a uppercase.

gy95: error starts with a lowercase, log info starts with a uppercase
So no need to fix case :)

stingshen: Yes, per [Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#error-strings), error starts with a lowercase:

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

fisherxu: This err will be printed directly in L108

fisherxu: Or we can modify L108 as: klog.Errorf("Failed to get container logs, err: %v", err)

stingshen: done. `modify L108 as: klog.Errorf("Failed to get container logs, err: %v", err)` should be the right way :)

cloud/pkg/cloudstream/streamserver.go
stingshen: done. I have modifed error log message in defer clause

fisherxu (commented):
/approve
/lgtm

PR #3504, opened 2021-12-24
fix node status ephemeral-storage will be zero  when edgedconfig.Config.EnableMetrics is false

lvfei103650:
**What type of PR is this?**
/kind bug  /kind feature

node status ephemeral-storage will be zero  when edgedconfig.Config.EnableMetrics is false.
if edgedconfig.Config.EnableMetrics is false, reconstruction cadvisor.GetDirFsInfo function.

Fixes #

```release-note

```


fisherxu (commented):
The cadvisorClient here is just a fake/stub client, this patch add the codes for ephemeral-storage, I wonder whether there's any issues if we depend on this cadvisor client? @lvfei103650

PR #3509, opened 2021-12-26
add matchPodNodeName for edged

### LINKED ISSUE #3489: cloud container be scheduled to the edge node
**What happened**:
this is my cluster info, it work on kubeedge in v1.8.2 version.
1 normal-first

#### Testing process #3293 
1. prepare
`systemctl stop edgecore` on all of the edgenode.
and clean all of the config on the edgenode
```shell
rm -f ca/*
rm -f certs/*
rm -fr edgemesh/
rm -f /var/lib/kubeedge/edgecore.db
```
and remove the `reliablesyncs.kubeedge.io` and it's children node from etcd
1 1 remove-reliablesyncs-kubeedge-io

and clean cloudimage on cloud node
`docker rmi kubeedge/cloudcore:v1.8.2`

2. modify these two files according to 3293
modify the `cloud/pkg/dynamiccontroller/application/listener.go` and `pkg/metaserver/util/util.go` 
In order to find the problem easier, I modified the `edged.go` file and added the print content:
```go
// klog.Infof("result content is %s", result.Content)
klog.Infof("result content is %s, resource: %s, group: %s, source: %s, operation: %s", result.Content, result.GetResource(), result.GetGroup(), result.GetSource(), result.GetOperation())
```
then rebuild the cloudimage and edgecore binary
```shell
make cloudimage
make all WHAT=edgecore
```
3. restart the cloudcore and add edgecore
2 add-edgenode
exec the `kubectl get pod -owide -nkubeedge -w` on the cloud

4. start the edgemesh-server and edgemesh-agent
At this moment, there is a problem that various cloud containers are scheduled to edge nodes.
It is found that containers such as `cloudcore` and `edgemesh` are also failing and restarting continuously;
3 watch-pod-status
The container being pulled up can be seen on the edge node: `docker ps -a`
4 docker-pa-a-on-edgenode

5. I keep the journal log files on the edge node
[edge-container-list.log](https://github.com/kubeedge/kubeedge/files/7749877/edge-container-list.log)
[edge-journal-3293.log](https://github.com/kubeedge/kubeedge/files/7749881/edge-journal-3293.log)

I have tested a lot of this, and it has nothing to do with the number of edge nodes. The current positioning is the problem caused by fix #3293; because I don't know much about `synccontroller` and `dynamiccontroller`, I sent the log for answers;

**Environment**:
v1.17.17

v1.8.2 + #3293 

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

yz271544:
**What type of PR is this?**
/kind bug

In order to ensure that the function of edge's `watch` is not affected, add a function `matchPodNodeName` for edged to fix the issue of #3489 that the cloud application is scheduled to the edge;

Fixes #3489 

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -1118,7 +1118,8 @@ func (e *edged) syncPod() {
 			klog.Errorf("get message content data failed: %v", err)
 			continue
 		}
-		klog.Infof("result content is %s", result.Content)
+		klog.Infof("result resource is %s, group is %s, operation is %s, source is %s, content is %s",
+			result.GetResource(), result.GetGroup(), result.GetOperation(), result.GetSource(), result.Content)
 		switch resType {
 		case model.ResourceTypePod:
 			if op == model.ResponseOperation && resID == "" && result.GetSource() == metamanager.MetaManagerModuleName {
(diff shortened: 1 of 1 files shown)


gy95 (commented):
do we also need to filter node pods in the dynamicController?

gy95:
If  I understand correctly, The pod message from cloudcore is first transported to edgehub, and then transferred to metamanager(store in database), and then notify the edged, Is this right @Rachel-Shao 
So If we want to filter msg in the edgecore, I think it's better to filter by nodes in the metamanager but not in the edged. But it seems that not proper.
So better filter in the cloud, so cloud will not send edge node not related message to the edge.

yz271544:
ok, i only want to know about case of integration_test Job, it seem has not `NodeName` field, how filter it?

gy95:
not understanding,by common, msg from cloud to edge will carry the node name.

Rachel-Shao:
> If I understand correctly, The pod message from cloudcore is first transported to edgehub, and then transferred to metamanager(store in database), and then notify the edged, Is this right @Rachel-Shao So If we want to filter msg in the edgecore, I think it's better to filter by nodes in the metamanager but not in the edged. But it seems that not proper. So better filter in the cloud, so cloud will not send edge node not related message to the edge.

yes it is, but setting the filtering in the cloud means that I cannot watch resources that do not belong to the node on the edge side. I am not sure whether there is such a demand?

PR #3513, opened 2021-12-28
add a basic image for building various components of KubeEdge

zc2638:

**What type of PR is this?**

/kind feature

As a basic image for building various components of KubeEdge.
used for edgecore and keadm multi arch build.

Fixes #

```release-note

```

### DIFF
--- a/build/docker/build-tools.dockerfile
+++ b/build/docker/build-tools.dockerfile
@@ -1,0 +1,14 @@
+# As a basic image for building various components of KubeEdge
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install -y wget \
+    vim git make gcc \
+    upx-ucl gcc-aarch64-linux-gnu libc6-dev-arm64-cross gcc-arm-linux-gnueabi libc6-dev-armel-cross &&\
+    apt-get autoremove -y &&\
+    apt-get clean &&\
+    rm -rf /var/lib/apt/lists/*
+RUN wget -c https://dl.google.com/go/go1.17.linux-amd64.tar.gz -O - | tar -xz -C /usr/local
+ENV GO111MODULE=on
+ENV GOPROXY=https://goproxy.io,direct
+ENV GOROOT=/usr/local/go
+ENV GOPATH=/go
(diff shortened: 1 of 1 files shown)


fisherxu:
> used for edgecore and keadm multi arch build.

We also need to multi arch build for all the components, ref: #3518 when you do the #3518, please take a look of this PR @gy95 :)

build/docker/crossbuild.dockerfile
```diff
@@ -0,0 +1,14 @@
+# As a basic image for building various components of KubeEdge
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install -y wget \
+    vim git make gcc \
+    upx-ucl gcc-aarch64-linux-gnu libc6-dev-arm64-cross gcc-arm-linux-gnueabi libc6-dev-armel-cross &&\
+    apt-get autoremove -y &&\
+    apt-get clean &&\
+    rm -rf /var/lib/apt/lists/*
+RUN wget -c https://dl.google.com/go/go1.17.linux-amd64.tar.gz -O - | tar -xz -C /usr/local
```
gy95, line 9:
```
As a basic image for building various components of KubeEdge.
used for edgecore and keadm multi arch build.
```
I'm not quite sure about how this Dockerfile worked. 
But here we emphasize the golang arch: linux-amd64, Is it can be used for image build and cross platform image build?

zc2638: This is the basic image used to build the `edgecore` multi-architecture binary, not the basic image used to build the multi-architecture image.
Deploying edgecore adopts a process deployment method, which requires binary rather than running the image directly.

fisherxu (commented):
Not only crossbuild can use this image, I think all the build can use it? Can we named the file like `build-tools` in istio?

zc2638:
> Not only crossbuild can use this image, I think all the build can use it? Can we named the file like `build-tools` in istio?

Although I think build-tools is not a good name and can be misleading. 
However, this PR has seriously affected the follow-up progress.
I will modify it.

fisherxu (commented):
/lgtm
/approve

PR #3514, opened 2021-12-28
Doc: Update the  CloudCore high-availability deployment document

### LINKED ISSUE #1560: HA support of CloudCore

Rachel-Shao:
**What type of PR is this?**
/kind documentation

Updated the documentation for cloudcore high-availability deployment

Fixes #1560

none

none


Rachel-Shao:
It was possible to observe the CloudCore currently connected to the edge node by viewing the `status.daemonEndpoints.kubeletEndpoint.Port` of the edge node, but because the `stream` is not enable, there is currently no test for the cloudcore high-availability operation result.   @fisherxu  @gy95  @yz271544

fisherxu:
cc @allenleu  Would you like to have a look of this PR? :)

PR #3515, opened 2021-12-28
supporting crossbuild all components

fisherxu:

**What type of PR is this?**

Supporting crossbuild all components, now we can crossbuild  all components as below:
```
#   make crossbuild
#   make crossbuild HELP=y
#   make crossbuild WHAT=edgecore
#   make crossbuild WHAT=edgecore GOARM=GOARM7
#   make crossbuild WHAT=keadm GOARM=GOARM8
```

Fixes #

cc @chendave @zhu733756 @gy95 

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,6 @@ integrationtest:
 	edge/test/integration/scripts/execute.sh
 endif
 
-CROSSBUILD_COMPONENTS=edgecore
 GOARM_VALUES=GOARM7 \
 	GOARM8
 
@@ -155,7 +154,7 @@ define CROSSBUILD_HELP_INFO
 # cross build components.
 #
 # Args:
-#   WHAT: Component names to be lint check. support: $(CROSSBUILD_COMPONENTS)
+#   WHAT: Component names to be lint check. support: $(BINARIES)
 #         If not specified, "everything" will be cross build.
 #
 # GOARM: go arm value, now support:$(GOARM_VALUES)
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -220,22 +220,6 @@ kubeedge::golang::build_binaries() {
 
 }
 
-
-KUBEEDGE_ALL_CROSS_BINARIES=(
-edgecore
-)
-
-kubeedge::golang::is_cross_build_binary() {
-  local key=$1
-  for bin in "${KUBEEDGE_ALL_CROSS_BINARIES[@]}" ; do
-    if [ "${bin}" == "${key}" ]; then
-      echo ${YES}
-      return
-    fi
-  done
-  echo ${NO}
-}
-
 KUBEEDGE_ALL_CROSS_GOARMS=(
 8
 7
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -155,18 +155,18 @@ define CROSSBUILD_HELP_INFO
...
+# GOARM: arm version for go crossbuild, now support:$(GOARM_VALUES)
+#        If not specified, default use ARM=GOARM8
 #
 #
 # Example:
 #   make crossbuild
 #   make crossbuild HELP=y
 #   make crossbuild WHAT=edgecore
-#   make crossbuild WHAT=edgecore GOARM=GOARM7
+#   make crossbuild WHAT=edgecore ARM=GOARM7
```
fisherxu, line 169: `GOARM` is a golang env value, so use ARM here to avoid conflicting with golang. Besides, if I set `GOARM=GOARM8`, will exit with err like: `Invalid GOARM value. Must be 5, 6, or 7.`

gy95:
make crossbuild WHAT=edgecore error
```
hack/make-rules/clean.sh
hack/make-rules/crossbuild.sh edgecore 
edgecore does not support cross build
make: *** [Makefile:178: crossbuild] Error 1
```

hack/lib/golang.sh
```diff
@@ -220,11 +220,6 @@ kubeedge::golang::build_binaries() {
...
 }
 
-
-KUBEEDGE_ALL_CROSS_BINARIES=(
-edgecore
-)
-
 kubeedge::golang::is_cross_build_binary() {
   local key=$1
   for bin in "${KUBEEDGE_ALL_CROSS_BINARIES[@]}" ; do
```
gy95, line 225: here also use this deleted KUBEEDGE_ALL_CROSS_BINARIES variable, also need fix

gy95: to fix `make crossbuild WHAT=edgecore` error.

fisherxu: Done.

gy95: LGTM test OK

Makefile
```diff
@@ -155,18 +155,18 @@ define CROSSBUILD_HELP_INFO
 # cross build components.
 #
 # Args:
-#   WHAT: Component names to be lint check. support: $(CROSSBUILD_COMPONENTS)
+#   WHAT: Component names to be lint check. support: $(BINARIES)
```
zhu733756, line 158: Use ${BINARIES} here, CROSSBUILD_COMPONENTS seems can be removed? Line L150

fisherxu: Done. Have removed :)

Makefile
```diff
@@ -147,26 +147,25 @@ integrationtest:
...
 #         If not specified, "everything" will be cross build.
 #
-# GOARM: go arm value, now support:$(GOARM_VALUES)
-#        If not specified ,default use GOARM=GOARM8
+# GOARM: arm version for go crossbuild, now support:$(GOARM_VALUES)
+#        If not specified, default use ARM=GOARM8
```
chendave, line 161: Hmm... GOARM is the original args, but you have updated this to "ARM",  so it would be something like,

```
ARM: arm version....
```

But I still think using `ARM` here is a little misleading.

Makefile
chendave: But why it works before?  Let me have a try.

fisherxu: If I set `GOARM=GOARM8`, will exit with err `Invalid GOARM value. Must be 5, 6, or 7.`, before for armv8, just haven't set the  GOARM value, default is armv8

Makefile
fisherxu: Yes, I also think `ARM` maybe misleading.. ;-)

Makefile
chendave: agree, but the code should not crush if we pass the value. I create a simple quick fix for this, pls  let me know whether that make sense to you: https://github.com/kubeedge/kubeedge/pull/3522

chendave: And how about define the variable as "ARM_VERSION" instead of "ARM"?  so it is more descriptive.

gy95 (approved):

Makefile
fisherxu: With your pr, I think we can still use `GOARM` :) PTAL @chendave

chendave:
/lgtm
/approve

thanks!

Issue #3518, opened 2021-12-28
Support make image and make crossbuild image


fisherxu:
**What would you like to be added/modified**:
Now we support make image in makefile like:
```
make cloudimage
make admissionimage
```
how about unnify the image cmd like `make binary` as blow?
```
make image WHAT=cloudcore
make image WHAT=admission
```

We can also add the make crossbuild image for arm platform:
```
make crossbuildimage  WHAT=cloudcore
```

How do you think? cc @zhu733756 @gy95 @zc2638

PR #3519, opened 2021-12-28
simplify edgecore image build

gy95:
**What type of PR is this?**
/kind cleanup

simplify edgecore image build,  just like other components. So we can use one command `docker build -f Dockerfile` to build images for all components
As pre of https://github.com/kubeedge/kubeedge/issues/3518

Fixes #

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -326,15 +326,7 @@ iptablesmgrimage:
 
 .PHONY: edgeimage
 edgeimage:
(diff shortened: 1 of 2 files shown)


fisherxu (commented):
/lgtm
/approve

We also need to check whether edegcore can run as excepted in container, including run the application like `kind`.

PR #3522, opened 2021-12-29
Fix the issue when cross-building for ARMv8

chendave:

**What type of PR is this?**

/kind bug

```make crossbuild WHAT=edgecore GOARM=GOARM8```
will hit below exception like this, 
```Invalid GOARM value. Must be 5, 6, or 7.```

should clear the env when building the binary for armv8

Fixes #

```release-note
NONE
```

@fisherxu

### DIFF
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -294,6 +294,7 @@ kubeedge::golang::cross_build_place_binaries() {
     local name="${bin##*/}"
     if [ "${goarm}" == "8" ]; then
       set -x
+      GOARM="" # need to clear the value since golang compiler doesn't allow this env when building the binary for ARMv8.
       GOARCH=arm64 GOOS="linux" CGO_ENABLED=1 CC=aarch64-linux-gnu-gcc go build -o ${KUBEEDGE_OUTPUT_BINPATH}/${name} -ldflags "$ldflags" $bin
       set +x
     elif [ "${goarm}" == "7" ]; then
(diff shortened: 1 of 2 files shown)


hack/lib/golang.sh
```diff
@@ -294,6 +294,7 @@ kubeedge::golang::cross_build_place_binaries() {
     local name="${bin##*/}"
     if [ "${goarm}" == "8" ]; then
       set -x
+      GOARM="" # need to clear the value since golang compiler doesn't allow this env when building the binary for ARMv8.
       GOARCH=arm64 GOOS="linux" CGO_ENABLED=1 CC=aarch64-linux-gnu-gcc go build -o ${KUBEEDGE_OUTPUT_BINPATH}/${name} -ldflags "$ldflags" $bin
```
gy95, line 298: we could specify `GOARM=""` just like set `GOARM=${goarm}` specifically for GOARM7 .
ref: https://github.com/golang/go/wiki/GoArm we should not set GOARM for GOARM8

gy95 (commented):
LGTM

hack/lib/golang.sh
chendave: yes, that is why the command like this will fail, 
```make crossbuild WHAT=edgecore GOARM=GOARM8```

fisherxu: make sense, in golang `GOARM=GOARM8` will failed.. I'm wondering whether it's reasonable ;-)

fisherxu (commented):
/lgtm
/approve

PR #3524, opened 2021-12-29
unify image build

gy95:
**What type of PR is this?**
/kind cleanup

part of https://github.com/kubeedge/kubeedge/issues/3518
unify the image cmd like make binary as below
`make image WHAT=cloudcore`

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -307,33 +307,26 @@ ARCH ?= amd64
 IMAGE_TAG ?= $(shell git describe --tags)
 GO_LDFLAGS='$(shell hack/make-rules/version.sh)'
 
-.PHONY: cloudimage
-cloudimage:
-	docker build --build-arg GO_LDFLAGS=${GO_LDFLAGS} -t kubeedge/cloudcore:${IMAGE_TAG} -f build/cloud/Dockerfile .
-
-.PHONY: admissionimage
-admissionimage:
-	docker build --build-arg GO_LDFLAGS=${GO_LDFLAGS} -t kubeedge/admission:${IMAGE_TAG} -f build/admission/Dockerfile .
-
-.PHONY: csidriverimage
-csidriverimage:
-	docker build --build-arg GO_LDFLAGS=${GO_LDFLAGS} -t kubeedge/csidriver:${IMAGE_TAG} -f build/csidriver/Dockerfile .
-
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -307,33 +307,26 @@ ARCH ?= amd64
 IMAGE_TAG ?= $(shell git describe --tags)
 GO_LDFLAGS='$(shell hack/make-rules/version.sh)'
```
fisherxu, line 308: Seems all these value can be removed?

fisherxu:
cc @zhu733756

hack/make-rules/image.sh
```diff
@@ -0,0 +1,91 @@
...
+  admission:admission:build/admission/Dockerfile
+  edgecore:edgecore:build/edge/Dockerfile
+  edgesite-agent:edgesite-agent:build/edgesite/agent-build.Dockerfile
+  edgesite-server:edgesite-server:build/edgesite/server-build.Dockerfile
+  csidriver:csidriver:build/csidriver/Dockerfile
+  iptablesmanager:iptables-manager:build/iptablesmanager/Dockerfile
+)
+
+GO_LDFLAGS="$(${KUBEEDGE_ROOT}/hack/make-rules/version.sh)"
+IMAGE_TAG=$(git describe --tags)
```
fisherxu, line 37: Now the `IMAGE_TAG` is from `git describe --tags`, for daily build, what IMAGE_TAG should used? cc @zhu733756 @zc2638 Is the `git describe --tags` still ok?

Makefile
gy95: done, and now both helm_e2e and `make image` cmd use default image tag defined by cmd `shell git describe --tags`

hack/make-rules/image.sh
zhu733756: > Now the `IMAGE_TAG` is from `git describe --tags`, for daily build, what IMAGE_TAG should used? cc @zhu733756 @zc2638 Is the `git describe --tags` still ok?

I think it is ok for me.

fisherxu (commented):
/lgtm
/approve

zc2638:
If it is only used to generate a release version, then it is reasonable.
If it is compatible with cicd, it needs the current commit hash.
@fisherxu @gy95

PR #3530, opened 2021-12-30
support multi architecture image build

gy95:
**What type of PR is this?**
/kind feature

support multi architecture image build
use Buildx to build multi arch images: this holds all arch images into one.
But we need to open docker buildx function first: this is not easy and need many steps, I follow this blog https://medium.com/@artur.klauser/building-multi-architecture-docker-images-with-buildx-27d80f7e2408 to enable buildx
And buildx will directly push built images to the registry, but not store them locally, so we need an image repo to hold the images.

Fixes #

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -323,6 +323,27 @@ image:
 	hack/make-rules/image.sh $(WHAT)
 endif
 
+define CROSS_IMAGE_HELP_INFO
+# Use Buildx to build multi-architecture docker images.
+#
+# Args:
+#   WHAT: component names to build. support: $(BINARIES)
+#         If not specified, "everything" will be built.
+#
+# Example:
+#   make crossbuildimage
+#   make crossbuildimage HELP=y
+#   make crossbuildimage WHAT=cloudcore
+endef
+.PHONY: crossbuildimage
+ifeq ($(HELP),y)
+crossbuildimage:
+	@echo "CROSS_IMAGE_HELP_INFO"
+else
+crossbuildimage:
+	hack/make-rules/crossbuildimage.sh $(WHAT)
+endif
+
 define INSTALL_HELP_INFO
 # install
 #
--- a/hack/make-rules/crossbuildimage.sh
+++ b/hack/make-rules/crossbuildimage.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+
+###
+#Copyright 2021 The KubeEdge Authors.
+#
+#Licensed under the Apache License, Version 2.0 (the "License");
+#you may not use this file except in compliance with the License.
+#You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#Unless required by applicable law or agreed to in writing, software
+#distributed under the License is distributed on an "AS IS" BASIS,
+#WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#See the License for the specific language governing permissions and
+#limitations under the License.
+###
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
+IMAGE_TAG=$(git describe --tags)
+GO_LDFLAGS="$(${KUBEEDGE_ROOT}/hack/make-rules/version.sh)"
+IMAGE_REPO_NAME="kubeedge"
+
+ALL_IMAGES_AND_TARGETS=(
+  #{target}:{IMAGE_NAME}:{DOCKERFILE_PATH}
+  cloudcore:cloudcore:build/cloud/Dockerfile
+  admission:admission:build/admission/Dockerfile
+  edgecore:edgecore:build/edge/Dockerfile
+  edgesite-agent:edgesite-agent:build/edgesite/agent-build.Dockerfile
+  edgesite-server:edgesite-server:build/edgesite/server-build.Dockerfile
+  csidriver:csidriver:build/csidriver/Dockerfile
+  iptablesmanager:iptables-manager:build/iptablesmanager/Dockerfile
+)
+
+function get_imagename_by_target() {
+  local key=$1
+  for bt in "${ALL_IMAGES_AND_TARGETS[@]}" ; do
+    local binary="${bt%%:*}"
+    if [ "${binary}" == "${key}" ]; then
+      local name_path="${bt#*:}"
+      echo "${name_path%%:*}"
+      return
+    fi
+  done
+  echo "can not find image name: $key"
+  exit 1
+}
+
+function get_dockerfile_by_target() {
+  local key=$1
+  for bt in "${ALL_IMAGES_AND_TARGETS[@]}" ; do
+    local binary="${bt%%:*}"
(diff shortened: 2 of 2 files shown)


zc2638:
Will this operation be used in cicd?

zc2638:
If used in cicd, I'm afraid you can't use `--push` directly

gy95:
> Will this operation be used in cicd?

no, not used in CI/CD, this is too slow

> Will this operation be used in cicd?

No, buildx costs too much and is too slow. And if we need to store multi arch images, we must specify the `--push`
blog says like this:
```
The docker buildx build subcommand has a number of flags which determine where the final image will be stored. By default, i.e. if none of the flags are specified, the resulting image will remain captive in docker’s internal build cache. This is unlike the regular docker build command which stores the resulting image in the local docker images list. The important flags are:
--load: This flag instructs docker to load the resulting image into the local docker images list. However, this currently only works for single-architecture images. If you try this with multi-architecture images you’ll get an export error:
$ docker buildx build … --load …
…
 => ERROR exporting to oci image format 0.0s
 — — — 
 > exporting to oci image format:
 — — — 
failed to solve: rpc error: code = Unknown desc = docker
 exporter does not currently support exporting manifest lists
--push: This flag tells docker to push the resulting image to a docker registry. Your image tag has to contain the proper reference to the registry and repository name. This currently is the best way to store multi-architecture images.
```

zc2638:
We should use the `manifest` directly so that the image can be generated locally.
This step is necessary in cicd, otherwise it is impossible to verify whether the image is available or whether it can be successfully built

gy95:
> We should use the `manifest` directly so that the image can be generated locally. This step is necessary in cicd, otherwise it is impossible to verify whether the image is available or whether it can be successfully built

we can use `make image`, but this is only used for build local arch images.
`make crossbuildimage` can be used for build multi-arch or cross-platform images.

gy95:
fix https://github.com/kubeedge/kubeedge/issues/3518
support multi architecture images build

hack/make-rules/crossbuildimage.sh
```diff
@@ -0,0 +1,94 @@
...
+       targets+=("${bt%%:*}")
+     done
+  fi
+
+  for arg in "${targets[@]}"; do
+    IMAGE_NAME="$(get_imagename_by_target ${arg})"
+    DOCKERFILE_PATH="$(get_dockerfile_by_target ${arg})"
+
+    set -x
+    docker buildx build --build-arg GO_LDFLAGS="${GO_LDFLAGS}" -t ${IMAGE_REPO_NAME}/${IMAGE_NAME}:${IMAGE_TAG} -f ${DOCKERFILE_PATH} --platform linux/amd64,linux/arm64,linux/arm/v7 --push .
```
fisherxu, line 85: The binaries compiled for arm and arm64 are different, list all the archs here, so the binary will all in image?

gy95: yes

zhu733756:
Marked to https://github.com/docker/setup-buildx-action.  Seems that we can use this action to test buildx.

fisherxu:
> Marked to https://github.com/docker/setup-buildx-action. Seems that we can use this action to test buildx.

With this we can copy the image to local host? @zhu733756

zhu733756:
> > Marked to https://github.com/docker/setup-buildx-action. Seems that we can use this action to test buildx.
> 
> With this we can copy the image to local host? @zhu733756

Cannot,  but we can test the process and push the image to the public registry like dockerhub by using an account with permission. And I think this solves the problem of periodic mirror push.

fisherxu:
> Cannot, but we can test the process and push the image to the public registry like dockerhub by using an account with permission. And I think this solves the problem of periodic mirror push.

`make image` will runs in ci only for `amd64`, which can check the image built process.
How about add this `make crossbuild` to a daily built action? So no need to run for every pr and will push the daily build image.  cc @zc2638 @gy95 @zhu733756

fisherxu:
I'm wondering how long does it take to run `make crossbuildimage` for one components a cloudcore?

zhu733756:
> > Cannot, but we can test the process and push the image to the public registry like dockerhub by using an account with permission. And I think this solves the problem of periodic mirror push.
> 
> `make image` will runs in ci only for `amd64`, which can check the image built process. How about add this `make crossbuild` to a daily built action? So no need to run for every pr and will push the daily build image. cc @zc2638 @gy95 @zhu733756

Agree with you. Multiple architectures can be built on a regular basis, regardless of running time.

gy95:
Using docker/setup-buildx-action@v1 to install buildx 
Now it seems to work, only need to process the  authentication problems of `docker push images`
And we can trigger this `make crossbuildimage` action only when a new tag is created.
@fisherxu @zhu733756

zhu733756:
> Using docker/setup-buildx-action@v1 to install buildx Now it seems to work, only need to process the authentication problems of `docker push images` And we can trigger this `make crossbuildimage` action only when a new tag is created. @fisherxu @zhu733756

Do you mean using a label to trigger, not daily built?

gy95:
now we release a version using a new tag,
so we can trigger `buildx ci` to build multi-arch images only when a new tag is created, based on this new tag.
we can do this like below. and based on the new tag, run `make crossbuildimage` cmd

```
name: Image Release

#only triggered when a new tag is created.
on:
  create:
    tags:
      - '*'

jobs:
  buildx:
    runs-on: ubuntu-18.04
    timeout-minutes: 30
    name: Buildx to build multi-arch images
    steps:
      - uses: actions/cache@v2
        with:
          path: ~/go/pkg/mod
          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}

      - name: Checkout
        uses: actions/checkout@v2
        with:
          fetch-depth: 0

      - uses: docker/setup-buildx-action@v1
        id: buildx
        with:
          install: true

      - name: Build
        run: |
          make crossbuildimage 
```

gy95:
> > Using docker/setup-buildx-action@v1 to install buildx Now it seems to work, only need to process the authentication problems of `docker push images` And we can trigger this `make crossbuildimage` action only when a new tag is created. @fisherxu @zhu733756
> 
> Do you mean using a label to trigger, not a daily build?

yes, this cost too much time, and IMO we don't need daily build at all, we only need to do this when release a new version, and host images to the official docker repo. so users can pull images.

zhu733756:
Thx.  Seems good.  Why not move it out to a single temple?

gy95:
this patch only add `make crossbuildimage` cmd,  integrate with buildx in https://github.com/kubeedge/kubeedge/pull/3538

Makefile
```diff
@@ -323,6 +323,27 @@ image:
 	hack/make-rules/image.sh $(WHAT)
 endif
 
+define CROSS_IMAGE_HELP_INFO
+# Use Buildx multi-architecture docker images.
+#	Usage of Buildx, please ref: https://medium.com/@artur.klauser/building-multi-architecture-docker-images-with-buildx-27d80f7e2408
```
fisherxu, line 328: We can remove this comment, which has been added in script.

gy95: fixed.

fisherxu (commented):
/lgtm
/approve

Merge this first, let's discuss about the CICD later.

PR #3533, opened 2022-01-06
Patch cloudcore rbac

zhu733756:
Signed-off-by: zhu733756 <zhu733756@kubesphere.io>

/kind bug

See https://github.com/kubeedge/edgemesh/issues/252#issuecomment-1000006246

/cc @fisherxu @gy95 @Poorunga

### DIFF
--- a/build/cloud/03-clusterrole.yaml
+++ b/build/cloud/03-clusterrole.yaml
@@ -11,7 +11,7 @@ rules:
   verbs: ["get", "list", "watch", "create", "update"]
 - apiGroups: [""]
   resources: ["namespaces"]
-  verbs: ["get", "create"]
+  verbs: ["get", "create", "list", "watch"]
 - apiGroups: [""]
   resources: ["nodes", "pods/status"]
   verbs: ["patch"]
--- a/build/cloud/ha/01-ha-prepare.yaml
+++ b/build/cloud/ha/01-ha-prepare.yaml
@@ -30,7 +30,7 @@ rules:
     verbs: ["get", "list", "watch", "create", "update"]
   - apiGroups: [""]
     resources: ["namespaces"]
-    verbs: ["get", "create"]
+    verbs: ["get", "create", "list", "watch"]
   - apiGroups: [""]
     resources: ["nodes", "pods/status"]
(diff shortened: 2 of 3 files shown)


gy95 (commented):
this file need fixing too
https://github.com/kubeedge/kubeedge/blob/eecaacaf9e0e1067189098b68a348db6a5155a8b/build/cloud/ha/01-ha-prepare.yaml#L32-L33

zhu733756:
> this file need fixing too
> 
> https://github.com/kubeedge/kubeedge/blob/eecaacaf9e0e1067189098b68a348db6a5155a8b/build/cloud/ha/01-ha-prepare.yaml#L32-L33

Fixed

gy95:
LGTM

fisherxu (commented):
Looks good for this pr.
And we need a mechanism to manage the role, if add a client in edge, then always update the config file here, maybe it's not the right way..
/approve

zhu733756:
> Looks good for this pr. And we need a mechanism to manage the role, if add a client in edge, then always update the config file here, maybe it's not the right way.. /approve

I think we can use a verify script integrated into CI to check these CRDs or K8s resources, developers will be reminded to find the changes.

zhu733756:
> > Looks good for this pr. And we need a mechanism to manage the role, if add a client in edge, then always update the config file here, maybe it's not the right way.. /approve
> 
> I think we can use a verify script integrated into CI to check these CRDs or K8s resources, developers will be reminded to find the changes.

Or later, we can integrate a make cmd to do these changes at the same time.

fisherxu:
> Or later, we can integrate a make cmd to do these changes at the same time.

There will be other resources that users use we don't know, so we need a extensionable mechanism to do this, it will be a big feature.

zhu733756:
> > Or later, we can integrate a make cmd to do these changes at the same time.
> 
> There will be other resources that users use we don't know, so we need a extensionable mechanism to do this, it will be a big feature.

K8s use the kubebuilder annotations/markers to create roles.  The problem is how we catch the permissions, then we can remind users or patch it automatically.

fisherxu (commented):
/lgtm

Issue #3475, opened 2021-12-16
Plans for Installation and Deployment Optimization


zhu733756:
Hey guys, we have plans for installation and deployment optimization for KubeEdge, the installation tool keadm will be refactored.

If you have features to achieve or to get involved in the development, please see the project form below, and you can add other todos here as you want:

- [ ] 1. **Cloud side**: use tools like `istioctl`  to integrate cloud components and provide stronger functions including the existing helm installation, and the newer keadm will only install containerized Cloudcore;
- [ ] 2. **Cloud side**: As a supplement, we will provide documents for installing CloudCore in process mode.
- [ ] 3. **Edge side**: Provide rpm domestic/international source integration, to download managed static files and binary files;
- [ ] 4. **Edge side**: Provides the public mirror repository service to host static files and binary files.
- [ ] 5.**CI & CD**: Mirror warehouse synchronization, regular mirror building, automatic release, and other auxiliary functions development and construction with Github actions.

/cc @fisherxu @zc2638 @zhu733756

gy95:
some work about cloudcore container deployment: maybe we can define required k8s resources in a string, or just use go `embed`
https://github.com/kubeedge/kubeedge/pull/2853

zc2638:
> some work about cloudcore container deployment: maybe we can define required k8s resources in a string, or just use go `embed` #2853

Because we need to support `helm` deployment, there should not be two yaml definitions, so a better way is to generate yaml resources through `helm`

gy95:
> > some work about cloudcore container deployment: maybe we can define required k8s resources in a string, or just use go `embed` #2853
> 
> Because we need to support `helm` deployment, there should not be two yaml definitions, so a better way is to generate yaml resources through `helm`

you mean the new keadm will only support install cloudcore using helm? : https://github.com/kubeedge/kubeedge/blob/master/build/helm/cloudcore/README.md#install-examples
or maybe also need to support common pod deployment?

zhu733756:
> > > some work about cloudcore container deployment: maybe we can define required k8s resources in a string, or just use go `embed` #2853
> > 
> > 
> > Because we need to support `helm` deployment, there should not be two yaml definitions, so a better way is to generate yaml resources through `helm`
> 
> you mean the new keadm will only support install cloudcore using helm? : https://github.com/kubeedge/kubeedge/blob/master/build/helm/cloudcore/README.md#install-examples or maybe also need to support common pod deployment?

[...]

PR #3537, opened 2022-01-07
add keadm beta reset command

gy95:
**What type of PR is this?**

add reset-beta command for keadm
a first simple version  for `keadm reset-beta` command implementation
waiting for init-beta command implementation finished, and maybe need an update

part of  #3475 

cc @zhu733756 

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/beta/reset.go
+++ b/keadm/cmd/keadm/app/cmd/beta/reset.go
@@ -1,0 +1,223 @@
+/*
+Copyright 2022 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package beta
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	dockertypes "github.com/docker/docker/api/types"
+	dockerfilters "github.com/docker/docker/api/types/filters"
+	dockerclient "github.com/docker/docker/client"
+	"github.com/spf13/cobra"
+	phases "k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/reset"
+	utilruntime "k8s.io/kubernetes/cmd/kubeadm/app/util/runtime"
+	utilsexec "k8s.io/utils/exec"
+
+	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
+	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/util"
+)
+
+var (
+	resetLongDescription = `
+keadm reset command can be executed in both cloud and edge node
+In cloud node it shuts down the cloud processes of KubeEdge
+In edge node it shuts down the edge processes of KubeEdge
+`
+	resetExample = `
+For cloud node:
+keadm beta reset
+
+For edge node:
+keadm beta reset
+`
+)
+
+func newResetOptions() *common.ResetOptions {
+	opts := &common.ResetOptions{}
+	opts.Kubeconfig = common.DefaultKubeConfig
+	return opts
+}
+
+func NewKubeEdgeResetBeta() *cobra.Command {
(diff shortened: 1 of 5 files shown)


keadm/cmd/keadm/app/cmd/reset.go
```diff
@@ -52,6 +53,68 @@ func newResetOptions() *common.ResetOptions {
...
+				if err := s.Err(); err != nil {
+					return err
+				}
+				if strings.ToLower(s.Text()) != "y" {
+					return fmt.Errorf("aborted reset operation")
+				}
+			}
+			// 1. kill cloudcore/edgecore process.
+			// For edgecore, don't delete node from K8S
+			if err := TearDownKubeEdgeV2(IsEdgeNode, reset.Kubeconfig); err != nil {
```
zhu733756, line 94: 1 helm delete releaseName -n namespace
2 kubectl delete ns or etcd delete ns
3 In other cases, if there have workloads on the edges, edges must reset firstly. We can add docs for this case.

gy95: 1. https://github.com/kubeedge/kubeedge/tree/master/build/helm/cloudcore#uninstall
we should run `helm uninstall` cmd ?
2. the previous function `keadm reset` doesn't process evict the pods, only remove all the pods and database on the edge

zhu733756: > we should run `helm uninstall` cmd ?

Because  init-beta will install as a chart, we had better do that.

> the previous function keadm reset doesn't process evict the pods, only remove all the pods and database on the edge

Does the previous function work on the cloud? If so, we need to implement it and add a parameter like compoent=cloud/edge to split the cloud reset and edge reset.

keadm/cmd/keadm/app/cmd/util/cloudinstaller.go
```diff
@@ -112,13 +113,23 @@ func (cu *KubeCloudInstTool) TearDown() error {
 	cu.SetKubeEdgeVersion(cu.ToolVersion)
 
 	// Kill cloudcore process
-	if err := cu.KillKubeEdgeBinary(KubeCloudBinaryName); err != nil {
-		return err
+	if cu.HelmInstallMode {
+		// For cloudcore container install, we need to run `helm uninstall cloudcore -n kubeedge` command and clean kubeedge namespace
+		uninstall := NewCommand("helm uninstall cloudcore -n kubeedge")
```
zhu733756, line 118: I think we had better use the helm API to uninstall it. Users will not install helm.  I will pr keadm init-beta later.

zhu733756:
See #3540, but need to fix the ci problem, will be fixed tomorrow.

gy95:
cc @fisherxu @zhu733756

keadm/cmd/keadm/app/cmd/beta/reset.go
```diff
@@ -0,0 +1,223 @@
...
+		Common: util.Common{
+			KubeConfig: kubeConfig,
+		},
+	}
+	if isEdgeNode {
+		ke = &util.KubeEdgeInstTool{Common: util.Common{}}
+	}
+
+	err := ke.TearDown()
+	if err != nil {
```
zhu733756, line 167: Does the ke.TearDown() remove the CRDs?

gy95: no, just clean up `kubeedge` namespace, binary install mode also doesn't process CRDs

zhu733756:
LGTM

keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -142,6 +143,28 @@ func GetOSInterface() types.OSTypeInstaller {
...
+	}
+
+	edgeCoreRunning, err := osType.IsKubeEdgeProcessRunning(KubeEdgeBinaryName)
+	if edgeCoreRunning {
+		return types.KubeEdgeEdgeRunning, nil
+	} else if err != nil {
+		return types.NoneRunning, err
+	}
+
+	return types.NoneRunning, nil
```
zc2638, line 165: If it is not in the cloud, it defaults to the edge.
HDYT?

gy95: if edgecore also is not running, we need do nothing.So it's better to distinguish these. :)

fisherxu (commented):
/approve

gy95:
@zc2638 PTAL again, need LGTM 😄

gy95:
@fisherxu , need to cherry pick to release-1.10? I see 1.10 branch has been created. or do nothing?

fisherxu:
> @fisherxu , need to cherry pick to release-1.10? I see 1.10 branch has been created. or do nothing?

Can cherry-pick to release-1.10.

PR #3538, opened 2022-01-08
add buildx to github action

gy95:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/buildx.yaml
+++ b/.github/workflows/buildx.yaml
@@ -1,0 +1,33 @@
+name: Image Release
+
+#only triggered when a new tag is created.
+on:
+  create:
+    tags:
+      - '*'
+
+jobs:
+  buildx:
+    runs-on: ubuntu-18.04
+    timeout-minutes: 30
+    name: Buildx to build multi-arch images
+    steps:
+      - uses: actions/cache@v2
+        with:
+          path: ~/go/pkg/mod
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -221,3 +221,20 @@ jobs:
...
+        uses: actions/checkout@v2
+      - uses: docker/setup-buildx-action@v1
+        id: buildx
+        with:
+          install: true
+      - name: Build
+        run: |
+          docker build -f /home/runner/work/${{ github.repository }}/build/cloud/Dockerfile # will run buildx
+        with:
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
```
fisherxu, line 240: Should do the make command from makefile.

gy95: got it, i'll integrate with buildx in https://github.com/kubeedge/kubeedge/pull/3530

.github/workflows/main.yaml
```diff
@@ -221,3 +221,20 @@ jobs:
           fetch-depth: 0
 
       - run: make image
+
+  buildx:
```
zhu733756, line 225: Seems good. Have 3 questions to ask.
1 Does this work if the users don't have push permission for the image?  Do we need a script for docker logining?
2 If this CI is too slow to build and may affect other CI actions, why not move it out? 
3 Does this CI need to be daily built?  If so, we need to add a schedule for it.

PR #3548, opened 2022-01-12
make the namespace optional

### LINKED ISSUE #3547: Fix tokensecret created in cloudcore hardcode in kubeedge namespace
**What happened**:
If users deploy the cloudcore in other namespace, tokensecret and other certs will still created in `kubeedge` ns.
 
Create them in same ns with cloudcore.

cc @gy95 @zhu733756

zhu733756:
Signed-off-by: zhu733756 <zhu733756@kubesphere.io>

Fix #3547 

## How to test
```
helm upgrade --install cloudcore ./cloudcore --namespace kubeedge001 --create-namespace -f ./cloudcore/values.yaml --set cloudCore.modules.cloudHub.advertiseAddress[0]=192.168.88.6 --set cloudCore.image.repository=zhu733756/cloudcore --set cloudCore.image.tag=v1.9.1-beta-2
```

## Screen Captures

image

### DIFF
--- a/common/constants/default.go
+++ b/common/constants/default.go
@@ -1,6 +1,7 @@
 package constants
 
 import (
+	"sync"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -15,9 +16,9 @@ const (
 	NodeName = "NodeName"
 
 	ProjectName = "KubeEdge"
+	SystemName  = "kubeedge-system"
 
-	SystemName      = "kubeedge"
-	SystemNamespace = SystemName
+	NamespaceEnvKey = "NAMESPACE"
 )
 
 // Resources
@@ -146,3 +147,20 @@ const (
 	// MessageSuccessfulContent is the successful content value of Message struct
 	MessageSuccessfulContent string = "OK"
 )
+
+var (
+	SystemNamespace = SystemName
+
+	once sync.Once
+)
+
+func SystemNamespaceInit(ns string) {
+	once.Do(
+		func() {
+			if ns == "" {
+				return
+			}
+			SystemNamespace = ns
+		},
+	)
+}
--- a/pkg/apis/componentconfig/cloudcore/v1alpha1/helper.go
+++ b/pkg/apis/componentconfig/cloudcore/v1alpha1/helper.go
@@ -23,6 +23,8 @@ import (
 
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
+
+	"github.com/kubeedge/kubeedge/common/constants"
 )
 
 type IptablesMgrMode string
@@ -44,6 +46,14 @@ func (c *CloudCoreConfig) Parse(filename string) error {
 		return err
 	}
 	return nil
+}
+
+func (c *CloudCoreConfig) GetNamespace() string {
+	if ns := os.Getenv(constants.NamespaceEnvKey); ns != "" {
+		return ns
+	}
+
+	return constants.SystemName
 }
 
 func (in *IptablesManager) UnmarshalJSON(data []byte) error {
--- a/build/cloud/01-namespace.yaml
+++ b/build/cloud/01-namespace.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: kubeedge
+  name: kubeedge-system
   labels:
     k8s-app: kubeedge
--- a/build/cloud/02-serviceaccount.yaml
+++ b/build/cloud/02-serviceaccount.yaml
@@ -5,4 +5,4 @@ metadata:
     k8s-app: kubeedge
     kubeedge: cloudcore
   name: cloudcore
-  namespace: kubeedge
+  namespace: kubeedge-system
--- a/build/cloud/04-clusterrolebinding.yaml
+++ b/build/cloud/04-clusterrolebinding.yaml
@@ -12,4 +12,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: cloudcore
-  namespace: kubeedge
+  namespace: kubeedge-system
--- a/build/cloud/05-configmap.yaml
+++ b/build/cloud/05-configmap.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: cloudcore
-  namespace: kubeedge
+  namespace: kubeedge-system
   labels:
     k8s-app: kubeedge
     kubeedge: cloudcore
--- a/build/cloud/07-deployment.yaml
+++ b/build/cloud/07-deployment.yaml
@@ -5,7 +5,7 @@ metadata:
     k8s-app: kubeedge
     kubeedge: cloudcore
   name: cloudcore
-  namespace: kubeedge
+  namespace: kubeedge-system
 spec:
   selector:
     matchLabels:
--- a/build/cloud/08-service.yaml.example
+++ b/build/cloud/08-service.yaml.example
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: cloudcore
-  namespace: kubeedge
+  namespace: kubeedge-system
   labels:
     k8s-app: kubeedge
     kubeedge: cloudcore
--- a/build/cloud/ha/01-ha-prepare.yaml
+++ b/build/cloud/ha/01-ha-prepare.yaml
@@ -14,7 +14,7 @@ metadata:
     k8s-app: kubeedge
     kubeedge: cloudcore
   name: cloudcore
-  namespace: kubeedge
+  namespace: kubeedge-system
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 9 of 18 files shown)


common/constants/default.go
```diff
@@ -147,3 +147,21 @@ const (
...
+	once sync.Once
+)
+
+func init() {
+	once.Do(
+		func() {
+			ns := os.Getenv(NamespaceEnvKey)
+			if ns != "" {
+				SystemNamespace = ns
+			}
```
zc2638, line 164: How about parse from config file before running?

zhu733756: Just have a see:
```
/ # cat /etc/kubeedge/config/cloudcore.yaml 
apiVersion: cloudcore.config.kubeedge.io/v1alpha2
kind: CloudCore
kubeAPIConfig:
  kubeConfig: ""
  master: ""
modules:
  cloudHub:
    advertiseAddress:
    - 192.168.88.6
    nodeLimit: 1000
    tlsCAFile: /etc/kubeedge/ca/rootCA.crt
    tlsCertFile: /etc/kubeedge/certs/edge.crt
    tlsPrivateKeyFile: /etc/kubeedge/certs/edge.key
    unixsocket:
      address: unix:///var/lib/kubeedge/kubeedge.sock
      enable: true
    websocket:
      address: 0.0.0.0
      enable: true
      port: 10000
    quic:
      address: 0.0.0.0
      enable: false
      maxIncomingStreams: 10000
      port: 10001
    https:
      address: 0.0.0.0
      enable: true
      port: 10002
  cloudStream:
    enable: true
    streamPort: 10003
    tunnelPort: 10004
  dynamicController:
    enable: false
  router:
    enable: false
```

But not have a namespace parameter.

zc2638: add one?

zhu733756: Like this?
```
metadata:
   namespace: kubeedge
```
What do you think ? @fisherxu @gy95

fisherxu: Add a config field in cloudcore for namespace maybe a little weird, as we already have the ns value from the pod. Can we add a new value in helm Value?

zhu733756: Helm no need to add a namespace.  It already can work now. :)

fisherxu: Okey, just seen helm have the --namespace option :) If users don't specify the --namespace option, helm will deploy the chart to `default` ns？

zhu733756: > Okey, just seen helm have the --namespace option :) If users don't specify the --namespace option, helm will deploy the chart to `default` ns？

Yeah.

gy95 (commented):
lgtm, so skillful method.

zhu733756:
> lgtm, so skillful method.

Haha, fix everything by monkey patching.

common/constants/default.go
zc2638: We should provide a way for users of process deployment to customize the namespace, configuration files are obviously preferred.
@fisherxu @zhu733756

fisherxu (commented):
/approve
Just have a check, k8s will not set the ns env variable. Below is the default env variable in pod:
```
KUBERNETES_SERVICE_PORT_HTTPS=443
KUBERNETES_SERVICE_PORT=443
HOSTNAME=nginx-deployment-6758b685d8-9p9k6
REDIS_DOWNLOAD_SHA=5b2b8b7a50111ef395bf1c1d5be11e6e167ac018125055daa8b5c2317ae131ab
PWD=/data
HOME=/root
KUBERNETES_PORT_443_TCP=tcp://10.96.0.1:443
REDIS_VERSION=6.2.6
GOSU_VERSION=1.12
TERM=xterm
REDIS_DOWNLOAD_URL=http://download.redis.io/releases/redis-6.2.6.tar.gz
SHLVL=1
KUBERNETES_PORT_443_TCP_PROTO=tcp
KUBERNETES_PORT_443_TCP_ADDR=10.96.0.1
KUBERNETES_SERVICE_HOST=10.96.0.1
KUBERNETES_PORT=tcp://10.96.0.1:443
KUBERNETES_PORT_443_TCP_PORT=443
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env
```

common/constants/default.go
fisherxu: Process deployment should be deprecated in future :) now most users all use deployment to deploy the cloudcore, how do you think? @zc2638  @gy95 @zhu733756

zc2638:
There should be no single configuration, we should unify all configuration sources. If support for environment variables is required, we should add it as an add-on.
I don't agree that only namespace is independent and only environment variables are supported.
@fisherxu @zhu733756 @gy95

common/constants/default.go
zhu733756: Process deployment may use export NAMESPACE=kubeedge001. 
Read from cloudcore.yaml,  Cloudcore configmap need to be changed to
```
metadata:
   namespace: {{.Release.Namespace}}
```

zhu733756: I think both are ok.

zhu733756:
> There should be no single configuration, we should unify all configuration sources. If support for environment variables is required, we should add it as an add-on. I don't agree that only namespace is independent and only environment variables are supported. @fisherxu @zhu733756 @gy95

Do we need to support both? If given a NAMESPACE env, will read it, or will try to parse the yaml to get ns.

zhu733756:
Screen Captures updated.

pkg/apis/componentconfig/cloudcore/v1alpha1/helper.go
```diff
@@ -46,6 +48,18 @@ func (c *CloudCoreConfig) Parse(filename string) error {
...
+func (c *CloudCoreConfig) GetNamespace() string {
+	if c.Namespace != "" {
+		return c.Namespace
+	}
+
+	if ns := os.Getenv(constants.NamespaceEnvKey); ns != "" {
+		return ns
+	}
+
+	return constants.SystemName
```
zc2638, line 60: We should override the configuration file fields with environment variables first

zhu733756: Yeah

zhu733756: > We should override the configuration file fields with environment variables first

I think we no need do that, 
```
if ns := os.Getenv(constants.NamespaceEnvKey); ns != "" {
	return ns
}
```
can be removed.

Because the external iptables can directly get env key and call SystemNamespaceInit:
```
constants.SystemNamespaceInit(config.GetNamespace())
```

What do you think?

zhu733756: For cloudcore, we will read from cloudcore.yaml. For external iptablesmgr, we will read from env key NAMESPACE to get the given namespace.

fisherxu: cloudcore is also no need to add new config field for namespace I think.. @zhu733756 @zc2638

common/constants/default.go
```diff
@@ -17,7 +16,9 @@ const (
 	NodeName = "NodeName"
 
 	ProjectName = "KubeEdge"
-	SystemName  = "kubeedge"
+	SystemName  = "kubeedge-system"
```
zhu733756, line 19: modified to kubeedge-system

gy95: will this cause version compatibility issue?

zhu733756: On the cloud, we init this namespace field by executing NamespaceInit function.  On the edge, if the namespace kubeedge cannot get, will try the namespace kubeedge-system.

gy95: this is OK, but there're some deployment YAMLs in dir https://github.com/kubeedge/kubeedge/tree/master/build/cloud still using ns `kubeedge`, do we also need to fix those synchronously?

gy95: or maybe we can just focus on helm install.

zhu733756: > this is OK, but there're some deployment YAMLs in dir https://github.com/kubeedge/kubeedge/tree/master/build/cloud still using ns kubeedge, do we also need to fix those synchronously?

I will fix it.

zhu733756: Do the labels and the namespace all needed to modify?
```
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    k8s-app: kubeedge  => k8s-app: kubeedge-system
    kubeedge: cloudcore =>  kubeedge-system:  cloudcore
  name: cloudcore
  namespace: kubeedge  => kubeedge-system
```

gy95: IMHO, we just need to modify the namespace. 👀

zhu733756:
Is this pr still needed? /cc

fisherxu:
> Is this pr still needed? /cc

Sure.

kevin-wangzefeng:
> Is this still relevant? If so, what is blocking it? Is there anything you can do to help move it forward?
> 
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

We still need this.

PR #3559, opened 2022-01-17
csidriver: handle topology constraints

### LINKED ISSUE #3471: csidriver: support multiple edge nodes /multiple csi drivers per edge node
related: original [csi design doc](https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go#L802)

tl;dr
* csidriver needs to be able to send `Volume*` requests to specific edge nodes
* we can use the CSI topology feature to address specific nodes
* csidriver needs a way to support multiple csi drivers (like hostpath, nfs etc.) on a given edge node.

**Why is this needed**:
`csidriver` in its current form supports only a single node (specified via `--node-id`). csidriver should be smarter and provision volumes on many edge nodes. There is one central problem to solve: **how do we know where a request should go to**?

We can do this by using `volumeBindingMode=WaitForFirstConsumer` while attaching correct labels to `Kind=Node` and configuring `Kind=CSINode` correctly.

Example storage class: We must enable `volumeBindingMode=WaitForFirstConsumer` to delay volume binding until a pod was  scheduled on a edge node. Now we know where the pod lives and create a volume there.

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-kubeedge
provisioner: kubeedge-csidriver
volumeBindingMode: WaitForFirstConsumer
parameters:
  edgeDriverName: csi-hostpath # <- used to select the correct csi driver on the edge, see below
```

external-provisioner uses `Kind=CSINode` to populate the `csi.CreateVolumeRequest` ([spec](https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go#L863-L874)) with `AccessibilityRequirements` [source](https://github.com/kubernetes-csi/external-provisioner/blob/master/pkg/controller/topology.go#L131-L134). That means for us we need to set up the CSINodes correctly in order to integrate correctly with external-provisioner.

```yaml
apiVersion: storage.k8s.io/v1
kind: CSINode
metadata:
  name: edge-1
spec:
  drivers:
  - name: csi-hostpath
    nodeID: edge-1
    topologyKeys:
    - csi.kubeedge.io/nodeid # <- this is something we can set from edged
    - topology.hostpath.csi/node
```
Further, external-provisioner relies on correct `Kind=Node` labels (the one specified above in the `topologyKeys`).

```yaml
apiVersion: v1
kind: Node
metadata:
  labels:
    csi.kubeedge.io/nodeid: edge-1 # <-- this is required and must be set by the operator.
    kubernetes.io/arch: amd64
    kubernetes.io/hostname: edge-1
    kubernetes.io/os: linux
    node-role.kubernetes.io/agent: ""
    node-role.kubernetes.io/edge: ""
```

How do we support multiple csi drivers **on the edge**?

* define multiple storage classes: e.g. one for hostpath, one for nfs.
* a user has to specify `parameter.edgeDriverName` to select the correct driver on the edge
* TODO: we need to find a way to pass that metadata down to the edge, we can not simply serialize the `Volume*` requests anymore.

## Alternatives

* run all csi-related controllers on edge nodes. That we we circumvent the routing problem in `csidriver` on the cloud side
* implement our own external-provisioner that is aware of the edge nodes so we don't need to rely on topology

CC @lvchenggang @fisherxu @gy95

moolen:
**What type of PR is this?**
/kind feature

(1) `csidriver` in its current form supports only a single node (specified via `--node-id`). csidriver should be smarter and provision volumes for many edge nodes. There is one central problem to solve: how do we know where a request should go to?
We can do this by using `volumeBindingMode=WaitForFirstConsumer` while attaching correct labels to `Kind=Node` and configuring `Kind=CSINode` correctly.

(2) We have to store state between `CreateVolume()` and `DeleteVolume()` requests because we can neither access `VolumeContext` from `DeleteVolume()` nor is the NodeID sent with it. Hence we store it as state on the filesystem.

(3) `inflight`: as per spec we must ensure that CreateVolume() requests are rejected if a particular volume is in the process of being created.

Fixes #3471

This PR makes #3343 and #3458 obsolete.
This PR is only about handling Volume requests from external-provisioner. How `Kind=CSINode` is created is part of a separate PR #3558.

```release-note
NONE
```

### DIFF
--- a/cloud/cmd/csidriver/app/options/options.go
+++ b/cloud/cmd/csidriver/app/options/options.go
@@ -24,9 +24,10 @@ import (
 type CSIDriverOptions struct {
 	Endpoint         string
 	DriverName       string
-	NodeID           string
 	KubeEdgeEndpoint string
 	Version          string
+	StatePath        string
+	TopologyKey      string
 }
 
 // NewCSIDriverOptions returns options object
@@ -39,7 +40,8 @@ func (o *CSIDriverOptions) Flags() (fss cliflag.NamedFlagSets) {
 	fs := fss.FlagSet("csidriver")
 	fs.StringVar(&o.Endpoint, "endpoint", "unix:///csi/csi.sock", "CSI endpoint")
 	fs.StringVar(&o.DriverName, "drivername", "csidriver", "name of the driver")
-	fs.StringVar(&o.NodeID, "nodeid", "", "node id determines which node will be used to create/delete volumes")
+	fs.StringVar(&o.StatePath, "state-path", "/kubeedge/csidriver", "path to the state storage")
+	fs.StringVar(&o.TopologyKey, "topology-key", "csi.kubeedge.io/nodeid", "topology key to use when considering accessibility requirements. The topology value must be the kubeedge node name which is used for routing CSI requests. The topology key must be defined")
 	fs.StringVar(&o.KubeEdgeEndpoint, "kubeedge-endpoint", "unix:///kubeedge/kubeedge.sock", "kubeedge endpoint")
 	fs.StringVar(&o.Version, "version", "dev", "version")
 	return
--- a/cloud/pkg/csidriver/controllerserver.go
+++ b/cloud/pkg/csidriver/controllerserver.go
@@ -17,374 +17,201 @@ limitations under the License.
 package csidriver
 
 import (
-	"encoding/base64"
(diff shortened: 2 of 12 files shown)


moolen:
closed in favor of using distributed provisioning

PR #3563, opened 2022-01-17
Fix no prompt while executing `make all` without sed

### LINKED ISSUE #3562: No prompt when executing `make all` without sed
**What happened**:

Currently, if we do not have sed installed and execute `make all`, we will get the result below. This is not what we want because `Please install gnu-sed (brew install gnu-sed)` does not show up.

截屏2022-01-18 上午12 24 20

echo "Please install gnu-sed (brew install gnu-sed)"

Execute `make all` on machines without gnu-sed/sed installed

See https://github.com/kubeedge/kubeedge/blob/f0ede113313f718505bd12e0803ce8577975d804/hack/lib/lint.sh#L28

The script exits directly after executing `which gsed`

JooKS-me:
Signed-off-by: zhukunshuai <jookunshuai@gmail.com>

**What type of PR is this?**

/kind bug

If sed/gnu-sed is not installed, it will not exit immediately, but will prompt that it is not installed.

image

Fixes #3562

### DIFF
--- a/hack/lib/lint.sh
+++ b/hack/lib/lint.sh
@@ -25,19 +25,21 @@ KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 
 if [[ "$OSTYPE" == "darwin"* ]]
 then
-    SED_CMD=`which gsed`
-    if [ -z $SED_CMD ]
+    if ! which gsed >/dev/null 2>&1
     then
         echo "Please install gnu-sed (brew install gnu-sed)"
         exit 1
+    else
+        SED_CMD=`which gsed`
     fi
 elif [[ "$OSTYPE" == "linux"* ]]
 then
-    SED_CMD=`which sed`
-    if [ -z $SED_CMD ]
+    if ! which sed >/dev/null 2>&1
     then
         echo "Please install sed"
         exit 1
+    else
+        SED_CMD=`which sed`
     fi
 else
     echo "Unsupported OS $OSTYPE"
--- a/hack/lib/lint.sh
+++ b/hack/lib/lint.sh
@@ -20,33 +20,36 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SED_CMD=""
-KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
+function kubeedge::lint::init() {
+    SED_CMD=""
+    KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 
-if [[ "$OSTYPE" == "darwin"* ]]
-then
-    if ! which gsed >/dev/null 2>&1
+    if [[ "$OSTYPE" == "darwin"* ]]
     then
-        echo "Please install gnu-sed (brew install gnu-sed)"
-        exit 1
-    else
-        SED_CMD=`which gsed`
-    fi
-elif [[ "$OSTYPE" == "linux"* ]]
-then
-    if ! which sed >/dev/null 2>&1
+        if ! which gsed >/dev/null 2>&1
+        then
+            echo "Please install gnu-sed (brew install gnu-sed)"
+            exit 1
+        else
+            SED_CMD=`which gsed`
+        fi
+    elif [[ "$OSTYPE" == "linux"* ]]
     then
-        echo "Please install sed"
-        exit 1
+        if ! which sed >/dev/null 2>&1
+        then
+            echo "Please install sed"
+            exit 1
+        else
+            SED_CMD=`which sed`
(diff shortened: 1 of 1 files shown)


gy95:
on my local machine, there's one error
```
# make lint
hack/make-rules/lint.sh
checking golangci-lint
found golangci-lint, version: golangci-lint has version 1.42.0 built from c6142e38 on 2021-08-17T11:47:22Z
start lint ...
check any whitenoise ...
xargs: s/[ \t]*$//: No such file or directory
Makefile:124: recipe for target 'lint' failed
make: *** [lint] Error 127
```

gy95 (commented):
in line 56 or line 53, variable SED_CMD is not defined if fixing in this way.

gy95:
what's `which gsed` cmd output on your mac os machine. It should  output `Please install gnu-sed (brew install gnu-sed)` if sed is not installed.

JooKS-me:
> in line 56 or line 53, variable SED_CMD is not defined if fixing in this way.

Yep, this is my oversight, I will fix it later.



> what's `which gsed` cmd output on your mac os machine. It should output `Please install gnu-sed (brew install gnu-sed)` if sed is not installed.

It will not output this, but exit the script directly, right?

fisherxu (commented):
L23-47 should be wrapped in a function like `kubeedge::lint::init`? `make all` should not check the sed cmd, just because source this script in `init.sh`.

JooKS-me:
> L23-47 should be wrapped in a function like `kubeedge::lint::init`? `make all` should not check the sed cmd, just because source this script in `init.sh`.

Yep, I think so too, and I've done it~

fisherxu (commented):
/lgtm
/approve
/hold
@gy95 for final review.

hack/lib/lint.sh
```diff
@@ -20,31 +20,36 @@ set -o errexit
...
+    if [[ "$OSTYPE" == "darwin"* ]]
     then
-        echo "Please install gnu-sed (brew install gnu-sed)"
-        exit 1
-    fi
-elif [[ "$OSTYPE" == "linux"* ]]
-then
-    SED_CMD=`which sed`
-    if [ -z $SED_CMD ]
+        if ! which gsed >/dev/null 2>&1
```
gy95, line 29: I don't think this is necessary. IMHO, these two are equal :)
```
SED_CMD=`which gsed`
    if [ -z $SED_CMD ]
```
`if ! which gsed >/dev/null 2>&1`

What is the output of  `which gsed` or `make lint` cmd if gsed is not installed?

fisherxu: you can try it in your linux host, modify the sed to as `sedddd`, seems needed.

hack/lib/lint.sh
```diff
@@ -20,31 +20,36 @@ set -o errexit
...
 
-SED_CMD=""
-KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
+function kubeedge::lint::init() {
+    SED_CMD=""
+    KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 
-if [[ "$OSTYPE" == "darwin"* ]]
-then
-    SED_CMD=`which gsed`
```
fisherxu, line 28: Can just modify this line? @JooKS-me  SED_CMD=`which gsed >/dev/null 2>&1`

hack/lib/lint.sh
JooKS-me: > I don't think this is necessary. IMHO, these two are equal :)
> 
> ```
> SED_CMD=`which gsed`
>     if [ -z $SED_CMD ]
> ```
> 
> `if ! which gsed >/dev/null 2>&1`
> 
> What is the output of `which gsed` or `make lint` cmd if gsed is not installed?

Hi, it will output this in master:
image

gy95: oh, i see, thanks for fixing

hack/lib/lint.sh
fisherxu: Omit the comment, seems no needed.

hack/lib/lint.sh
gy95: I guess it's due to `` mark
```
SED_CMD=`which sed`
    if [ -z $SED_CMD ]
```
```
SED_CMD="which sed"
    if [ -z "$SED_CMD" ]
```

JooKS-me: > you can try it in your linux host, modify the sed to as `sedddd`, seems needed.

I try it in my ubuntu:

image

image

gy95: if we want to get the result of cmd running using ``, we must ensure the cmd can be running successfully, so in the before way, the script exist unnormally when running `which sed` cmd.

gy95:
LGTM

PR #3566, opened 2022-01-18
fix lots of duplicate logs "Failed to get obj" in objectsync.go

### LINKED ISSUE #3555: Lots of duplicate logs "Failed to get obj" in objectsync.go L36
**What happened**:
Everytime we send a obj like Pod successfully to edge, cloudcore will create a `objectsync` cr in cloud. 
In normal condition, if we delete the obj in cloud, cloudcore will send the delete event to edge, if obj in edge deleted successfully, then delete the  `objectsync`.

But if edgenode is offline, at this point if users force delete the obj(Pod) in cloud, then can't deleted it in the edge. Now the  `objectsync` will be left.

So the synccontroller will always print the error log  "Failed to get obj".

We need to handle the case for edgenode offline then force delete the obj in cloud.

 edgenode offline ->  force delete the obj -> check the cloudcore log.

vincentgoat:
/kind bug

Fixes #3555

/cc @fisherxu 
/assign @fisherxu

### DIFF
--- a/cloud/pkg/synccontroller/objectsync.go
+++ b/cloud/pkg/synccontroller/objectsync.go
@@ -31,15 +31,20 @@ func (sctl *SyncController) manageObject(sync *v1alpha1.ObjectSync) {
 	}
 	resource := util.UnsafeKindToResource(sync.Spec.ObjectKind)
 	gvr := gv.WithResource(resource)
-
+	nodeName := getNodeName(sync.Name)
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
We should fix this bug in synccontroller, in synccontroller, if the k8s obj can't be found through objectsync, we should trigger a delete event to the edge, first delete the obj in edge, then delete objectsync in cloud.

PR #3571, opened 2022-01-20
fix remove os.Exit in updateNodeKubeletEndpoint

lvfei103650:
**What type of PR is this?**
/kind bug

Fixes #
If a edgenode registration fails, the node resource does not exist in the k8s cluster. When the edgestream of this node is connected to cloudstream (10004), it will updateNodeKubeletEndpoint, which will trigger os.Exit in this function, causing cloudcore to hang up, which will affect the connection status of other nodes

```release-note

```

### DIFF
--- a/cloud/pkg/cloudstream/tunnelserver.go
+++ b/cloud/pkg/cloudstream/tunnelserver.go
@@ -23,7 +23,6 @@ import (
 	"encoding/pem"
 	"fmt"
 	"net/http"
-	"os"
 	"strings"
 	"sync"
 	"time"
@@ -43,8 +42,8 @@ import (
 
 const (
 	// The amount of time the tunnelserver should sleep between retrying node status updates
-	retrySleepTime        = 20 * time.Second
-	nodeStatusUpdateRetry = 3
+	retrySleepTime          = 20 * time.Second
+	nodeStatusUpdateTimeout = 2 * time.Minute
 )
 
 type TunnelServer struct {
@@ -126,7 +125,16 @@ func (s *TunnelServer) connect(r *restful.Request, w *restful.Response) {
 		sessionID:     hostNameOverride,
 	}
 
-	s.updateNodeKubeletEndpoint(hostNameOverride)
+	err = s.updateNodeKubeletEndpoint(hostNameOverride)
+	if err != nil {
+		msg := stream.NewMessage(0, stream.MessageTypeCloseConnect, []byte(err.Error()))
+		if err := session.tunnel.WriteMessage(msg); err == nil {
+			klog.V(4).Infof("CloudStream send close connection message to edge successfully")
+		} else {
+			klog.Errorf("CloudStream failed to send close connection message to edge, error: %v", err)
+		}
+		return
+	}
 	s.addSession(hostNameOverride, session)
 	s.addSession(internalIP, session)
 	s.addNodeIP(hostNameOverride, internalIP)
@@ -185,8 +193,8 @@ func (s *TunnelServer) Start() {
 	}
 }
 
-func (s *TunnelServer) updateNodeKubeletEndpoint(nodeName string) {
-	if err := wait.PollImmediate(retrySleepTime, nodeStatusUpdateRetry, func() (bool, error) {
+func (s *TunnelServer) updateNodeKubeletEndpoint(nodeName string) error {
+	if err := wait.PollImmediate(retrySleepTime, nodeStatusUpdateTimeout, func() (bool, error) {
 		getNode, err := client.GetKubeClient().CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
 		if err != nil {
 			klog.Errorf("Failed while getting a Node to retry updating node KubeletEndpoint Port, node: %s, error: %v", nodeName, err)
@@ -202,7 +210,8 @@ func (s *TunnelServer) updateNodeKubeletEndpoint(nodeName string) {
 		return true, nil
 	}); err != nil {
 		klog.Errorf("Update KubeletEndpoint Port of Node '%v' error: %v. ", nodeName, err)
-		os.Exit(1)
+		return fmt.Errorf("failed to Update KubeletEndpoint Port")
 	}
 	klog.V(4).Infof("Update node KubeletEndpoint Port successfully, node: %s, tunnelPort: %s", nodeName, s.tunnelPort)
+	return nil
 }
--- a/edge/pkg/edgestream/tunnel.go
+++ b/edge/pkg/edgestream/tunnel.go
@@ -165,6 +165,10 @@ func (s *TunnelSession) Serve() error {
 			return err
 		}
 
+		if mess.MessageType == stream.MessageTypeCloseConnect {
+			return fmt.Errorf("close tunnel stream connection, error:%s", string(mess.Data))
+		}
+
 		if mess.MessageType < stream.MessageTypeData {
 			go s.ServeConnection(mess)
 		}
--- a/pkg/stream/message.go
+++ b/pkg/stream/message.go
@@ -33,6 +33,7 @@ const (
 	MessageTypeMetricConnect
 	MessageTypeData
 	MessageTypeRemoveConnect
+	MessageTypeCloseConnect
 )
 
 func (m MessageType) String() string {


lvfei103650:
Step 1:  The node name does not match the DNS subdomain name, resulting in unsuccessful registration
Step 2:  When updateNodeKubeletEndpoint in cloudstream, the node cannot be found, then os.Exit
image
image

Step 3:  Keep restarting cloudcore, affecting other nodes.

Therefore, "os.Exit" is unreasonable in this place, and "return" will be more appropriate

fisherxu (commented):
Make sense. And we need to ensure the KubeletEndpoint updated successfully. If return here, the edegstream will reconnect to cloudstream, then will try to update the KubeletEndpoint  again. Right? @lvfei103650 @zc2638

lvfei103650:
yes.
If return here, Edgestream establishes a connection with cloudstream（10004）.
But the node registration failed, so you need to modify the node name and restart edgecore.

fisherxu:
I've double check the code, if node registration successed, but here KubeletEndpoint update failed, edgecore/edgestream will not reconnect.

I think we need to send a error response to edge and then edgecore/edgestream reconnect to cloudstream with interval `1s 1s 2s 2s 3s 3s..` to avoid reconnect too frequently.

How do you think? @lvfei103650 @zc2638

zc2638:
From my point of view, the fix here is necessary anyway.
If it is not repaired, the entire cluster needs to be checked. After the repair, you only need to troubleshoot the node with the log problem.
The question raised by @fisherxu  is very important, this further fix can be done in another PR, can @lvfei103650 continue to fix it?

lvfei103650:
@zc2638 yes，i think the fix here is necessary anyway,too.And i can continue to fix it.
@fisherxu In the scenario you mentioned,  just expect the KubeletEndpoint to be updated successfully. I don't think reconnection is necessary, because the updateNodeKubeletEndpoint function has already tried to reconnect 3 times.And if updating KubeletEndpoint fails, only the functions of kubectl log/exec are affected.
How do you think? @lvfei103650 @zc2638

fisherxu:
> And if updating KubeletEndpoint fails, only the functions of kubectl log/exec are affected.

Yes, but we need to ensure the `kubectl log/exec` worked as excepted. Though it already tried to update kubeletEndPoint 3 times, but we can't ensure update successfully. 

Maybe two solutions,
One solution is tried to update kubeletEndPoint more times, but how many times should be or tried until successed?
Another one is print the err log and trigger edge reconnect. Users can get the errlog when debug logs/exec issue.

zc2638:
The retry operation experience is bad, my idea is that we can return an error and disconnect the `stream`.
`edgestream` can both catch errors and try to reconnect due to disconnection.
@fisherxu @lvfei103650

fisherxu:
> The retry operation experience is bad, my idea is that we can return an error and disconnect the `stream`. `edgestream` can both catch errors and try to reconnect due to disconnection. @fisherxu @lvfei103650

Make sense. Btw, the timeout in L188 seems incorrect, please also take a look @lvfei103650

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -126,7 +125,19 @@ func (s *TunnelServer) connect(r *restful.Request, w *restful.Response) {
...
+		klog.Errorf("Failed to Update KubeletEndpoint Port, send reconnect msg to edgeStream")
+		for retry := 0; retry < 3; retry++ {
+			msg := stream.NewMessage(0, stream.MessageTypeReconnect, []byte(err.Error()))
+			if err := session.tunnel.WriteMessage(msg); err != nil {
+				klog.Errorf("cloudStream send %s message error %v", msg.MessageType, err)
+			} else {
+				break
+			}
+		}
+		return
```
fisherxu, line 139:
```suggestion
		for retry := 0; retry < 3; retry++ {
			msg := stream.NewMessage(0, stream.MessageTypeReconnect, []byte(err.Error()))
			if err := session.tunnel.WriteMessage(msg); err == nil {
			        klog.V(4).Infof("CloudStream send reconnect message to edge successfully")
				return
			} 
			klog.Errorf("CloudStream failed to send reconnect message to edge, error: %v", err)
		}
		return
```

edge/pkg/edgestream/tunnel.go
```diff
@@ -165,6 +165,11 @@ func (s *TunnelSession) Serve() error {
 			return err
 		}
 
+		if mess.MessageType == stream.MessageTypeReconnect {
+			klog.Warningf("Recive message about tunnel reconnect because of update NodeKubeletEndpoint failed: %v", string(mess.Data))
+			return fmt.Errorf("update NodeKubeletEndpoint failed")
```
fisherxu, line 170:
```suggestion
			return fmt.Errorf("update NodeKubeletEndpoint failed, error: %s", string(mess.Data))
```
We can delete L169.

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -44,7 +43,7 @@ import (
 const (
 	// The amount of time the tunnelserver should sleep between retrying node status updates
 	retrySleepTime        = 20 * time.Second
-	nodeStatusUpdateRetry = 3
+	nodeStatusUpdateTimeout = 2 * time.Minute
```
fisherxu, line 46: File is not gofmted. Overall LGTM

fisherxu (commented):
/approve

/assign @zc2638 for final review.

lvfei103650:
@zc2638 please review.

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -126,7 +125,19 @@ func (s *TunnelServer) connect(r *restful.Request, w *restful.Response) {
...
+		klog.Errorf("Failed to Update KubeletEndpoint Port, send reconnect msg to edgeStream")
+		for retry := 0; retry < 3; retry++ {
+			msg := stream.NewMessage(0, stream.MessageTypeReconnect, []byte(err.Error()))
+			if err := session.tunnel.WriteMessage(msg); err == nil {
+				klog.V(4).Infof("CloudStream send reconnect message to edge successfully")
+				return
+			}
+			klog.Errorf("CloudStream failed to send reconnect message to edge, error: %v", err)
+		}
+		return
```
zc2638, line 139: First of all, does the retry here make sense?
Then, the `reconnect` expression is not appropriate, I think the disconnection type message should be sent directly, similar to `CloseMessage`.

lvfei103650: The first is to consider, retry if the send close message is unsuccessful。

zc2638: Yes, this practice is common.
But to be sure, is it necessary?
Under what circumstances will send fail?
Even if the sending fails, what will be the impact of not doing other processing?

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -126,7 +125,18 @@ func (s *TunnelServer) connect(r *restful.Request, w *restful.Response) {
...
 	}
 
-	s.updateNodeKubeletEndpoint(hostNameOverride)
+	err = s.updateNodeKubeletEndpoint(hostNameOverride)
+	if err != nil {
+		klog.Errorf("Failed to Update KubeletEndpoint Port, send reconnect msg to edgeStream")
+
+		msg := stream.NewMessage(0, stream.MessageTypeCloseConnect, []byte(err.Error()))
+		if err := session.tunnel.WriteMessage(msg); err == nil {
+			klog.V(4).Infof("CloudStream send reconnect message to edge successfully")
```
zc2638, line 134: message need fix.

edge/pkg/edgestream/tunnel.go
```diff
@@ -165,6 +165,10 @@ func (s *TunnelSession) Serve() error {
 			return err
 		}
 
+		if mess.MessageType == stream.MessageTypeCloseConnect {
+			return fmt.Errorf("update NodeKubeletEndpoint failed, error: %s", string(mess.Data))
```
zc2638, line 169: For the edge, it can only find the problem of closing the connection from the returned information, and cannot clearly indicate that this error is caused by `update NodeKubeletEndpoint`

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -126,7 +125,18 @@ func (s *TunnelServer) connect(r *restful.Request, w *restful.Response) {
 		sessionID:     hostNameOverride,
 	}
 
-	s.updateNodeKubeletEndpoint(hostNameOverride)
+	err = s.updateNodeKubeletEndpoint(hostNameOverride)
+	if err != nil {
+		klog.Errorf("Failed to Update KubeletEndpoint Port, send reconnect msg to edgeStream")
```
zc2638, line 130: No more questions, just one last point.
Unnecessary log. The log has been printed in the `updateNodeKubeletEndpoint ` method.

zc2638:
Please squash the commits.

PR #3581, opened 2022-01-21
fix: add missing metrics func

### LINKED ISSUE #3471: csidriver: support multiple edge nodes /multiple csi drivers per edge node

moolen:
**What type of PR is this?**
/kind bug

When we use distributed provisioning edged dies with a panic: 

```
Observed a panic: "invalid memory address or nil poi>
goroutine 1663 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x23847a0, 0x3fc23f0})
        /home/ubuntu/kubeedge/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:74 +0x85
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x30})
        /home/ubuntu/kubeedge/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:48 +0x75
panic({0x23847a0, 0x3fc23f0})
        /usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/kubeedge/kubeedge/edge/pkg/edged/volume/csi.(*csiMountMgr).GetMetrics(0xc0012b0f00)
        <autogenerated>:1 +0x29
[...]

```

This is due to the fact that the `Volume` interface embeds a `MetricsProvider` with `GetMetrics() (*Metrics, error)` which wasn't configured or stripped out. I took the upstream code from Kubernetes.

Fixes #3471

```release-note
NONE
```

### DIFF
--- a/edge/pkg/edged/volume/csi/csi_plugin.go
+++ b/edge/pkg/edged/volume/csi/csi_plugin.go
@@ -415,6 +415,8 @@ func (p *csiPlugin) NewMounter(
 	}
 	klog.V(4).Info(log("created path successfully [%s]", dataDir))
 
+	mounter.MetricsProvider = NewMetricsCsi(volumeHandle, dir, csiDriverName(driverName))
+
 	// persist volume info data for teardown
 	node := string(p.host.GetNodeName())
 	volData := map[string]string{
--- a/edge/pkg/edged/volume/csi/csi_metrics.go
+++ b/edge/pkg/edged/volume/csi/csi_metrics.go
@@ -1,0 +1,132 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package csi
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"google.golang.org/grpc"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/pkg/volume"
+	volumeutil "k8s.io/kubernetes/pkg/volume/util"
+)
+
+var _ volume.MetricsProvider = &metricsCsi{}
+
+// metricsCsi represents a MetricsProvider that calculates the used,free and
+// capacity information for volume using volume path.
+
+type metricsCsi struct {
+	// the directory path the volume is mounted to.
+	targetPath string
+
+	// Volume handle or id
+	volumeID string
+
+	//csiClient with cache
+	csiClientGetter
+}
+
+// NewMetricsCsi creates a new metricsCsi with the Volume ID and path.
+func NewMetricsCsi(volumeID string, targetPath string, driverName csiDriverName) volume.MetricsProvider {
+	mc := &metricsCsi{volumeID: volumeID, targetPath: targetPath}
+	mc.csiClientGetter.driverName = driverName
+	return mc
+}
+
+func (mc *metricsCsi) GetMetrics() (*volume.Metrics, error) {
+	currentTime := metav1.Now()
+	ctx, cancel := context.WithTimeout(context.Background(), csiTimeout)
+	defer cancel()
+	// Get CSI client
+	csiClient, err := mc.csiClientGetter.Get()
+	if err != nil {
+		return nil, err
+	}
+	// Check whether "GET_VOLUME_STATS" is set
+	volumeStatsSet, err := csiClient.NodeSupportsVolumeStats(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// if plugin doesnot support volume status, return.
+	if !volumeStatsSet {
+		return nil, volume.NewNotSupportedErrorWithDriverName(
+			string(mc.csiClientGetter.driverName))
+	}
+	// Get Volumestatus
+	metrics, err := csiClient.NodeGetVolumeStats(ctx, mc.volumeID, mc.targetPath)
+	if err != nil {
+		return nil, err
+	}
+	if metrics == nil {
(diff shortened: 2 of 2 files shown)


moolen:
keadm e2e tests failure seems unrelated:

```
execute keadm command failed:  EdgeCore is already running on this node, please run reset to clean up first
Error: EdgeCore is already running on this node, please run reset to clean up first
```

edge/pkg/edged/volume/csi/csi_plugin.go
```diff
@@ -415,6 +415,8 @@ func (p *csiPlugin) NewMounter(
 	}
 	klog.V(4).Info(log("created path successfully [%s]", dataDir))
 
+	mounter.MetricsProvider = NewMetricsCsi(volumeHandle, dir, csiDriverName(driverName))
```
fisherxu, line 418: Is it possible to import from k8s instead of copy in? :)

moolen: @fisherxu woud that be OK for you to copy/paste that part?

moolen:
@fisherxu ping

edge/pkg/edged/volume/csi/csi_plugin.go
fisherxu: The `csiDriverName` is not included in `[volume/csi/csi_metrics.go]`, so I just tested in my host:

in [csi_plugin.go]:

```
+       "k8s.io/kubernetes/pkg/volume/csi"
 
        "github.com/kubeedge/kubeedge/edge/pkg/edged/volume/csi/nodeinfomanager"
 )
@@ -415,6 +416,9 @@ func (p *csiPlugin) NewMounter(
        }
        klog.V(4).Info(log("created path successfully [%s]", dataDir))
 
+       mounter.MetricsProvider = csi.NewMetricsCsi(volumeHandle, dir, csiDriverName(driverName))
+       
+
        // persist volume info data for teardown
        node := string(p.host.GetNodeName())

```

Then run `hack/update-vendor.sh` to update the vendor. :) @moolen Would you like to have a try?

fisherxu: Just compile it, the error:
```
+ go build -o /root/codes/src/github.com/kubeedge/kubeedge/_output/local/bin/edgecore -gcflags= -ldflags '-s -w -buildid= -X github.com/kubeedge/kubeedge/pkg/version.buildDate=2022-02-18T09:35:08Z -X github.com/kubeedge/kubeedge/pkg/version.gitCommit=dd12ab16671269436c5af705316ef074535d7453 -X github.com/kubeedge/kubeedge/pkg/version.gitTreeState=dirty -X github.com/kubeedge/kubeedge/pkg/version.gitVersion=v1.9.0-168+dd12ab16671269-dirty -X github.com/kubeedge/kubeedge/pkg/version.gitMajor=1 -X github.com/kubeedge/kubeedge/pkg/version.gitMinor=9+' github.com/kubeedge/kubeedge/edge/cmd/edgecore
# github.com/kubeedge/kubeedge/edge/pkg/edged/volume/csi
edge/pkg/edged/volume/csi/csi_plugin.go:419:78: cannot use csiDriverName(driverName) (type csiDriverName) as type "k8s.io/kubernetes/pkg/volume/csi".csiDriverName in argument to "k8s.io/kubernetes/pkg/volume/csi".NewMetricsCsi

```

moolen: yes, that's exactly the error i'm seeing.

fisherxu (commented):
/approve

PR #3591, opened 2022-01-29
use testing.TempDir to create temp directory

gy95:
**What type of PR is this?**
/kind cleanup

we can call function `testing.TempDir` to create a temporary directory. this will simplify codes.
TempDir returns a temporary directory for the test to use. The directory is automatically removed by Cleanup when the test and all its subtests complete. 
ref: https://pkg.go.dev/testing#T.TempDir

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #3599, opened 2022-02-08
Fix pv's objectsync spec.ObjectKind is empty

neiba:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -561,6 +561,10 @@ func (mh *MessageHandle) saveSuccessPoint(msg *beehiveModel.Message, info *model
 					ObjectName:       resourceName,
 				},
 			}
+			if objectSync.Spec.ObjectKind == "" {
+				klog.Errorf("Failed to init objectSync: %s, ObjectKind is empty, msg content: %v", objectSyncName, msg.GetContent())
+				return
+			}
 			_, err := mh.crdClient.ReliablesyncsV1alpha1().ObjectSyncs(resourceNamespace).Create(context.Background(), objectSync, metav1.CreateOptions{})
 			if err != nil {
 				klog.Errorf("Failed to create objectSync: %s, err: %v", objectSyncName, err)
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -37,6 +37,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	apimachineryType "k8s.io/apimachinery/pkg/types"
 	k8sinformer "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
@@ -57,6 +58,7 @@ import (
 	common "github.com/kubeedge/kubeedge/common/constants"
 	edgeapi "github.com/kubeedge/kubeedge/common/types"
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/cloudcore/v1alpha1"
+	"github.com/kubeedge/kubeedge/pkg/metaserver/util"
 )
 
 // SortedContainerStatuses define A type to help sort container statuses based on container names.
@@ -571,6 +573,9 @@ func kubeClientGet(uc *UpstreamController, namespace string, name string, queryT
 		err := stderrors.New("wrong query type")
 		klog.Error(err)
 		return nil, err
+	}
+	if err == nil && obj != nil {
+		util.SetMetaType(obj.(runtime.Object))
 	}
 	return obj, err
 }
--- a/cloud/pkg/synccontroller/objectsync.go
+++ b/cloud/pkg/synccontroller/objectsync.go
@@ -44,8 +44,13 @@ func (sctl *SyncController) manageObject(sync *v1alpha1.ObjectSync) {
 		newObject.SetNamespace(sync.Namespace)
 		newObject.SetName(sync.Spec.ObjectName)
 		newObject.SetUID(types.UID(getObjectUID(sync.Name)))
-		msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, newObject)
-		beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		if msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, newObject); msg != nil {
+			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		} else {
+			if err := sctl.crdclient.ReliablesyncsV1alpha1().ObjectSyncs(sync.Namespace).Delete(context.Background(), sync.Name, *metav1.NewDeleteOptions(0)); err != nil {
+				klog.Errorf("Failed to delete objectsync %s: %v", sync.Name, err)
+			}
+		}
 		return
 	} else if err != nil || ret == nil {
 		klog.Errorf("failed to get obj(gvr:%v,namespace:%v,name:%v), %v", gvr, sync.Namespace, sync.Spec.ObjectName, err)
@@ -77,8 +82,9 @@ func sendEvents(err error, nodeName string, sync *v1alpha1.ObjectSync, resourceT
 	if err != nil && apierrors.IsNotFound(err) {
 		//trigger the delete event
 		klog.Infof("%s: %s has been deleted in K8s, send the delete event to edge", resourceType, sync.Spec.ObjectName)
-		msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, obj)
-		beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		if msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, obj); msg != nil {
+			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		}
 		return
 	}
 
--- a/build/helm/cloudcore/values.yaml
+++ b/build/helm/cloudcore/values.yaml
@@ -11,7 +11,7 @@ cloudCore:
     tag: "v1.9.1"
     pullPolicy: "IfNotPresent"
     pullSecrets: []
-  securityContext: 
+  securityContext:
     privileged: true
   labels:
     k8s-app: kubeedge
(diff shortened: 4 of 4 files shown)


cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -1004,6 +1008,12 @@ func NewUpstreamController(config *v1alpha1.EdgeController, factory k8sinformer.
 	uc.podLister = factory.Core().V1().Pods().Lister()
 	uc.configMapLister = factory.Core().V1().ConfigMaps().Lister()
 	uc.secretLister = factory.Core().V1().Secrets().Lister()
+	factory.Core().V1().PersistentVolumes()
+	uc.pvLister = factory.Core().V1().PersistentVolumes().Lister()
+	factory.Core().V1().PersistentVolumeClaims()
+	uc.pvcLister = factory.Core().V1().PersistentVolumeClaims().Lister()
+	factory.Storage().V1().VolumeAttachments()
+	uc.attachLister = factory.Storage().V1().VolumeAttachments().Lister()
```
Iceber, line 1016: It seems that you just need to call `factory...Lister()` directly.

cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -561,6 +561,10 @@ func (mh *MessageHandle) saveSuccessPoint(msg *beehiveModel.Message, info *model
 					ObjectName:       resourceName,
 				},
 			}
+			if objectSync.Spec.ObjectKind == "" {
+				klog.Errorf("Failed to init objectSync: %s, ObjectKind is empty, msg content: %v", objectSyncName, msg.GetContent())
```
fisherxu, line 565: I'm wondering which message's kind/type will be empty?

cloud/pkg/edgecontroller/controller/upstream.go
fisherxu: +1

cloud/pkg/synccontroller/objectsync.go
```diff
@@ -90,8 +86,9 @@ func sendEvents(err error, nodeName string, sync *v1alpha1.ObjectSync, resourceT
 	if CompareResourceVersion(objectResourceVersion, sync.Status.ObjectResourceVersion) > 0 {
 		// trigger the update event
 		klog.V(4).Infof("The resourceVersion: %s of %s in K8s is greater than in edgenode: %s, send the update event", objectResourceVersion, resourceType, sync.Status.ObjectResourceVersion)
-		msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.UpdateOperation, obj)
-		beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		if msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.UpdateOperation, obj); msg != nil {
+			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		}
```
fisherxu, line 91: Does the modification in this file related to the issue this pr fixed? If not, we can move it to another PR :)

cloud/pkg/edgecontroller/controller/upstream.go
fisherxu: +1

cloud/pkg/cloudhub/handler/messagehandler.go
neiba: i made a mistake, configmap object metatype is set is edgecontroller downstream: https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/edgecontroller/manager/common.go#L29

cloud/pkg/synccontroller/objectsync.go
neiba: i found related PR had been merged: https://github.com/kubeedge/kubeedge/pull/3570

cloud/pkg/cloudhub/handler/messagehandler.go
fisherxu: I'm wondering what kind of message's ObjectKind will be empty?

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -572,6 +574,9 @@ func kubeClientGet(uc *UpstreamController, namespace string, name string, queryT
 		klog.Error(err)
 		return nil, err
 	}
+	if err == nil && obj != nil {
+		util.SetMetaType(obj.(runtime.Object))
```
fisherxu, line 578: Where will use the Meta(apiversion and kind)?

cloud/pkg/synccontroller/objectsync.go
```diff
@@ -44,8 +44,13 @@ func (sctl *SyncController) manageObject(sync *v1alpha1.ObjectSync) {
 		newObject.SetNamespace(sync.Namespace)
 		newObject.SetName(sync.Spec.ObjectName)
 		newObject.SetUID(types.UID(getObjectUID(sync.Name)))
-		msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, newObject)
-		beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		if msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, newObject); msg != nil {
```
fisherxu, line 47: Will the msg be nil? Which kind of message?

fisherxu (commented):
Just some questions :)

cloud/pkg/edgecontroller/controller/upstream.go
neiba: cloudhub send msg to edge successful and then create objectsync, which will use msg resource meta
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/handler/messagehandler.go#L560

cloud/pkg/synccontroller/objectsync.go
neiba: if objectsync's spec.ObjectKind is empty, then buildEdgeControllerMessage would return nil, such as pvc's objectsync

fisherxu: Can you explain why the ObjectKind of pvc's objectsync is empty?

cloud/pkg/edgecontroller/controller/upstream.go
fisherxu: And can we get the obj from lister in L563 565 567?

cloud/pkg/synccontroller/objectsync.go
neiba: 1. edged query a pvc
2. edgecontroller upstream receive the msg
3. edgecontroller upstream get pvc from k8s api 
4. edgecontroller store pvc object to msg.content, and send msg to cloudhub
5. cloudhub send msg to edgehub
6. cloudhub get response from edge, and create pvc's objectsync
7. objectsync.spec.objectkind = util.GetMessageResourceType(msg)

step 7 msg and step 4 msg are same one, and msg.content is pvc object, which metatype is empty. So we need SetMetaType

fisherxu: OK, get it, thanks! And if the msg is empty here, is it reasonable to delete the objectsync as below?

fisherxu: And if we add the `util.SetMetaType(obj.(runtime.Object))`, maybe the msg will not be empty? :)

neiba: there are three cases:
1. objectsync's resource is deleted
2. objectsync's resource is deleted, but a same name resource created
3. objectsync.Spec.ObjectKind is empty.

case 1: program will entry line 48, send delete msg to edge and then delete objectsync
case 2: program will entry line 73, sendEvents function will send delete msg to edge and then delete objectsync
case 3: program will entry line 50, delete objectsync directly

maybe msg will not be empty, add a check just for comfort. ：）

cloud/pkg/synccontroller/objectsync.go
```diff
@@ -77,8 +82,9 @@ func sendEvents(err error, nodeName string, sync *v1alpha1.ObjectSync, resourceT
 	if err != nil && apierrors.IsNotFound(err) {
 		//trigger the delete event
 		klog.Infof("%s: %s has been deleted in K8s, send the delete event to edge", resourceType, sync.Spec.ObjectName)
-		msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, obj)
-		beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
+		if msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, obj); msg != nil {
+			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
```
neiba, line 86: not delete objectsync when msg == nil, because:
1. not crdclient :)
2. msg won't be nil almostly, add check msg == nil to avoid cloudcore crash when beehive send a nil msg.

cloud/pkg/synccontroller/objectsync.go
fisherxu: Make sense, thanks for the clarification!

cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -561,6 +561,10 @@ func (mh *MessageHandle) saveSuccessPoint(msg *beehiveModel.Message, info *model
...
+				return
```
fisherxu, line 566: Last question maybe, with `util.SetMetaType(obj.(runtime.Object))` in kubeClientGet, I think the kind will almost never empty, right?

fisherxu (commented):
@My-Pleasure Would you like to take a look?

cloud/pkg/cloudhub/handler/messagehandler.go
neiba: yep, i think so. Dynamiccontroller informers alse already called SetMetaType in eventhandler. Cloudhub send other type msgs to edge by calling ListMessageWriteLoop, which no need to create objectsync, such as device twin, applicationResponse, user data.

fisherxu: Sounds good.

My-pleasure:
I agree with adding `util.SetMetaType` in function `kubeClientGet`. It's good.
By the way, when upstream controller gets the obj from lister, that obj doesn't need to `SetMetaType`, like ConfigMap. Only pv and pvc .etc which get by kubeClient need to `SetMetaType`. And if we want get pv from lister, we should set a pv lister in upstream controller which is not had now.

fisherxu:
> By the way, when upstream controller gets the obj from lister, that obj doesn't need to SetMetaType, like ConfigMap. Only pv and pvc .etc which get by kubeClient need to SetMetaType. And if we want get pv from lister, we should set a pv lister in upstream controller which is not had now.

Just have a check, the obj in list also has no MetaType info, so we need to set the MetaType. @My-pleasure

neiba:
> I agree with adding `util.SetMetaType` in function `kubeClientGet`. It's good. By the way, when upstream controller gets the obj from lister, that obj doesn't need to `SetMetaType`, like ConfigMap. Only pv and pvc .etc which get by kubeClient need to `SetMetaType`. And if we want get pv from lister, we should set a pv lister in upstream controller which is not had now.

fisherxu (commented):
/lgtm
/approve

fisherxu:
@neiba Could you please cherry-pick this pr to release-1.9 branch using script `hack/cherry_pick_pull.sh`? :)

Issue #3527, opened 2021-12-30
Memory consumption of autonomic Kube-API


khalid-jobs:
**What happened**:
After I open autonomic kube-api, the memory consumption of edgecore will get bigger and bigger. For example, the following edgeocre's memory consumption reach 700MB. 
![image](https://user-images.githubusercontent.com/13484086/147714821-47feb28c-eb98-40ee-8f51-6cb8ca56cbbb.png).

Here are some information I've gathered.
1. If I not open autonomic kube-api in edgecore, the edgecore's memory consumption will be normal(below 150MB)
![image](https://user-images.githubusercontent.com/13484086/147714932-4cc68d3a-8b08-4b6d-b95e-dd8b0a18eb74.png)
![image](https://user-images.githubusercontent.com/13484086/147714962-80c5c4eb-8e13-48e3-b0ae-8fe25f4b95ce.png)

2. If I open autonomic kube-api in edgecore, the edgecore's memory consumption will get bigger and bigger.
When I first open autonomic kube-api, the memory consumption:  
![image](https://user-images.githubusercontent.com/13484086/147715051-9f1d0da7-2b76-485d-846a-addfa676ee92.png)
After 22 hour
![image](https://user-images.githubusercontent.com/13484086/147715086-4c6899d1-ac86-4b3a-a6a3-f663171f052f.png)
![image](https://user-images.githubusercontent.com/13484086/147715095-70012ad4-54c4-49c6-985f-8937e0a2af98.png)

**What you expected to happen**:
After I open autonomic kube-api, the memory consumption should not be bigger and bigger.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): v .18.2
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): v1.8.2
- <details><summary>Cloud nodes Environment:</summary>

  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>
- <details><summary>Edge nodes Environment:</summary>

  - edgecore version (e.g. `edgecore --version`):
  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>

[...]

PR #3605, opened 2022-02-10
fix memory leak in edgesite kube-api

Shelley-BaoYue:
/kind bug

Fixes https://github.com/kubeedge/kubeedge/issues/3527

/cc @fisherxu
/assign @fisherxu


fisherxu (commented):
/lgtm
/approve

PR #3615, opened 2022-02-14
Fix Message's String method & minor code cleanup

### LINKED ISSUE #3611: Message not printed correctly
**What happened**:

When I was experimenting the code in `staging/src/github.com/kubeedge/beehive/test`, I notcied that the message was printed as `{{feebdb70-511f-411f-ace9-cdddb65ffd85  1644800569157  false } {sourcemodule   insert test}`.

However, according to the `String` method defined on the `Message` type, it should printed as `MessageID: 2d05baa2-2b45-406c-bd4e-e5cec0e149c5 ParentID:  Source: sourcemodule Destination:  Resource: test Operation: insert`.

Compile and run `staging/src/github.com/kubeedge/beehive/test/cmd/main.go`

Commented code in `staging/src/github.com/kubeedge/beehive/test/modules/test_destination.go`

image

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

algobot76:
**What type of PR is this?**

/kind bug
/kind cleanup

Fixes #3611 

NONE

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
@@ -253,13 +253,20 @@ func (msg *Message) GetDestination() string {
 }
 
 // String the content that you want to send
-func (msg *Message) String() string {
+func (msg Message) String() string {
 	var buffer bytes.Buffer
 	buffer.WriteString("MessageID: " + msg.GetID())
-	buffer.WriteString(" ParentID: " + msg.GetParentID())
-	buffer.WriteString(" Source: " + msg.GetSource())
-	buffer.WriteString(" Destination: " + msg.GetDestination())
-	buffer.WriteString(" Resource: " + msg.GetResource())
-	buffer.WriteString(" Operation: " + msg.GetOperation())
+	buffer.WriteString(" ParentID: " + handleEmptyString(msg.GetParentID()))
+	buffer.WriteString(" Source: " + handleEmptyString(msg.GetSource()))
+	buffer.WriteString(" Destination: " + handleEmptyString(msg.GetDestination()))
+	buffer.WriteString(" Resource: " + handleEmptyString(msg.GetResource()))
+	buffer.WriteString(" Operation: " + handleEmptyString(msg.GetOperation()))
 	return buffer.String()
+}
+
+func handleEmptyString(s string) string {
+	if len(s) == 0 {
+		return "\"\""
+	}
+	return s
 }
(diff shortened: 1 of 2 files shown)


staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
```diff
@@ -253,13 +253,20 @@ func (msg *Message) GetDestination() string {
 }
 
 // String the content that you want to send
-func (msg *Message) String() string {
+func (msg Message) String() string {
```
fisherxu, line 256: Remove the pointer is not reasonable :) We should use the pointer.

staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
```diff
@@ -253,13 +253,20 @@ func (msg *Message) GetDestination() string {
...
+	buffer.WriteString(" Source: " + handleEmptyString(msg.GetSource()))
+	buffer.WriteString(" Destination: " + handleEmptyString(msg.GetDestination()))
+	buffer.WriteString(" Resource: " + handleEmptyString(msg.GetResource()))
+	buffer.WriteString(" Operation: " + handleEmptyString(msg.GetOperation()))
 	return buffer.String()
 }
+
+func handleEmptyString(s string) string {
+	if len(s) == 0 {
+		return "\"\""
```
fisherxu, line 269: If the field is empty, I think we no need to set it to `""`, because the un-empty feild also has no the `""`.

And how about just printed it as json format:

```
buffer, _ := json.Marshal(msg)
return string(buffer)
```

staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
algobot76: There are two reasons why I prefer the value receiver:

1. You can invoke the `Stirng` method on both `msg` and `&msg`. If you define the method on pointer, you can only invoke it on `&msg`.
2. The value receiver indicates that `msg` is immutable, i.e. we won't change the `msg` passed by the caller.

staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
algobot76: The purpose of setting an empty string to `""` is to make logging look better :)
Instead of this:
image
we can now get this:
image

---
Instead of using `json.Marshal`, we could use `fmt.Sprintf`:

```go
return fmt.Sprintf("MessageID: %s ParentID: %s Source: %s Destination: %s Resource: %s Operation: %s",
		msg.GetID(), msg.GetParentID(), msg.GetSource(), msg.GetDestination(), msg.GetResource(), msg.GetOperation())
```

fisherxu:
cc @zc2638

zc2638:
Can some unit tests be added to verify what the final output should look like?
@algobot76 
/remove-kind bug

staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
gy95: i guess we can just use %q to format string, this will add quotation marks to a string :)

PR #3626, opened 2022-02-25
Update build-tools to support arm arch

fisherxu:

**What type of PR is this?**
/kind feature

Build build-tools base image for arm and x86 arch.

Fixes #

```release-note
NONE
```


fisherxu:
cc @TianTianBigWang @zc2638 @zhu733756 PTAL :)

build/docker/build-tools.arm.dockerfile
```diff
@@ -0,0 +1,14 @@
+# As a basic image for building various components of KubeEdge
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install -y wget apt-utils \
```
zhu733756, line 3: Correct me if I am wrong. Do we need this logic like if-else to combine these Dockerfiles? I think one Dockerfile would make sense.

fisherxu: I'm ot sure whether the dockerfile can aware of the cpu arch or pass any parameter when building?

zhu733756: > pass any parameter when building

Do we implement `ARG arch=<control parameter>` above the `FROM`. Then we can use this parameter to determine what the cmdline to execute.

TianTianBigWang: yes, like #3629 `edgecore.dockerfile`, we can use `ARG` above the `FROM`

TianTianBigWang: to specify the base image architecture for the build

fisherxu: Updated, ptal :)

build/docker/build-tools/build-tools.dockerfile
```diff
@@ -0,0 +1,22 @@
+# As a basic image for building various components of KubeEdge
+FROM ubuntu:18.04
+ARG ARCH=amd64
+RUN apt-get update && apt-get install -y wget \
+    vim git make gcc upx-ucl 
+
+RUN if [ "${ARCH}" = "amd64" ]; then apt-get install -y gcc-aarch64-linux-gnu libc6-dev-arm64-cross gcc-arm-linux-gnueabi libc6-dev-armel-cross ; fi
```
fisherxu, line 7: For arm arch, it's no needed to install these cross-build pkgs. Install them in amd64 is because we may use it to do the cross-build in future.

Issue #3619, opened 2022-02-18
synccontroller delete new objectsync when old objectsync's uuid not equal resource' uuid


neiba:
**What happened**:

synccontroller delete new objectsync when old objectsync's uuid not equal resource' uuid

**What you expected to happen**:

synccontroller should delete the old objectsync when old objectsync's uuid not equal resource' uuid

**How to reproduce it (as minimally and precisely as possible)**:

1. create configmap-1
2. create deployment-1 on edgenode and use configmap-1 as volume
3. disconnect edgenode network
4. delete configmap-1
5. create configmap-1
6. restart cloudcore
7. reconnect edgenode network

**Anything else we need to know?**:

https://github.com/kubeedge/kubeedge/blob/611a597f245d71377f18b9698fd258170aae2724/cloud/pkg/synccontroller/objectsync.go#L65-L73

1. sendEvents should not use object which from k8s api, and this cause the problem.
2. i am not sure we should send delete msg to edge or not, new resource may had been sent to edge and overwrite the old new, so we just need delete old objectsync directly ?

**Environment**:
- Kubernetes version (use `kubectl version`):
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`):
- <details><summary>Cloud nodes Environment:</summary>

  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>
- <details><summary>Edge nodes Environment:</summary>

  - edgecore version (e.g. `edgecore --version`):
  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>

Issue #3445, opened 2021-12-08
list-watch Not running properly on edgenode


davedavedavid:
After a service is created on the K8S, the list-watch operation of the edge node is abnormal. After an add operation, there are consecutive delete operations.

**What you expected to happen**:
Services can be added properly on edge nodes.

[...]

PR #3628, opened 2022-02-25
fix-3619: use objectsync to construct msg object when objectsync uid …

neiba:
…not equal

Signed-off-by: chengengkun <1366542563@qq.com>

**What type of PR is this?**

/kind bug

Fixes # https://github.com/kubeedge/kubeedge/issues/3619

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/synccontroller/objectsync.go
+++ b/cloud/pkg/synccontroller/objectsync.go
@@ -38,19 +38,7 @@ func (sctl *SyncController) manageObject(sync *v1alpha1.ObjectSync) {
 	//ret, err := informers.GetInformersManager().GetDynamicSharedInformerFactory().ForResource(gvr).Lister().ByNamespace(sync.Namespace).Get(sync.Spec.ObjectName)
 	ret, err := sctl.kubeclient.Resource(gvr).Namespace(sync.Namespace).Get(context.TODO(), sync.Spec.ObjectName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		// trigger the delete event
-		klog.V(4).Infof("%s: %s has been deleted in K8s, send the delete event to edge in sync loop", resourceType, sync.Spec.ObjectName)
-		newObject := &unstructured.Unstructured{}
-		newObject.SetNamespace(sync.Namespace)
-		newObject.SetName(sync.Spec.ObjectName)
-		newObject.SetUID(types.UID(getObjectUID(sync.Name)))
-		if msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, newObject); msg != nil {
-			beehiveContext.Send(commonconst.DefaultContextSendModuleName, *msg)
-		} else {
-			if err := sctl.crdclient.ReliablesyncsV1alpha1().ObjectSyncs(sync.Namespace).Delete(context.Background(), sync.Name, *metav1.NewDeleteOptions(0)); err != nil {
-				klog.Errorf("Failed to delete objectsync %s: %v", sync.Name, err)
-			}
-		}
+		sctl.gcOrphanedObjectSync(sync)
 		return
 	} else if err != nil || ret == nil {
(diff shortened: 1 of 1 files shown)


vincentgoat (approved):

vincentgoat:
cc @fisherxu @wackxu PTAL.

neiba:
https://github.com/kubeedge/kubeedge/blob/6eaaf05fcd71ecbabd323403a35ad5b19eff955c/cloud/pkg/synccontroller/objectsync.go#L79-L81
If we use unstructured.Unstructured to init object, util.SetMetaType would error with: failed to set metatype: Object 'Kind' is missing in 'unstructured object has no kind'. Maybe we just need to set object uid ? @vincentgoat

vincentgoat:
> https://github.com/kubeedge/kubeedge/blob/6eaaf05fcd71ecbabd323403a35ad5b19eff955c/cloud/pkg/synccontroller/objectsync.go#L79-L81
> 
> 
> If we use unstructured.Unstructured to init object, util.SetMetaType would error with: failed to set metatype: Object 'Kind' is missing in 'unstructured object has no kind'. Maybe we just need to set object uid ? @vincentgoat

agree

cloud/pkg/synccontroller/objectsync.go
```diff
@@ -68,6 +68,7 @@ func (sctl *SyncController) manageObject(sync *v1alpha1.ObjectSync) {
 			Group:    "",
 			Resource: resource,
 		}, sync.Spec.ObjectName)
+		object.SetUID(types.UID(syncObjUID))
```
wackxu, line 71: The change is tricky， it is strange and incomprehensible， we should update the object UID

wackxu: why not do like this?
```
	newObject := &unstructured.Unstructured{}
		if msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, newObject); msg != nil {
		}
```

neiba: Yes, i think this change is tricky too. If we send delete message and return here, then we are not need to construct NotFound error. Further more, sendEvents first argument is not need.

cloud/pkg/synccontroller/objectsync.go
```diff
@@ -64,29 +52,37 @@ func (sctl *SyncController) manageObject(sync *v1alpha1.ObjectSync) {
...
+		sctl.deleteObjectSyncWithoutResource(sync)
+		return
 	}
 
-	sendEvents(err, nodeName, sync, resourceType, object.GetResourceVersion(), object)
+	sendEvents(nodeName, sync, resourceType, object.GetResourceVersion(), object)
 }
 
-func sendEvents(err error, nodeName string, sync *v1alpha1.ObjectSync, resourceType string,
+func (sctl *SyncController) deleteObjectSyncWithoutResource(sync *v1alpha1.ObjectSync) {
```
wackxu, line 62:
```suggestion
func (sctl *SyncController) gcOrphanedObjectSync(sync *v1alpha1.ObjectSync) {
```

wackxu: please also add comment to the function
```
// gcOrphanedObjectSync try to send delelte message to the edge node
// to make sure that the resource is deleted in the edge node. After the 
// message ACK is received by `cloudHub`, the objectSync will be deleted
// directly in the `cloudHub`. But if message build failed, the objectSync 
// will be deleted directly in the `syncController`.
```

neiba: Updated, please take a look. @wackxu

vincentgoat (commented):
lgtm, Thanks for your patience! @neiba

wackxu:
releated to https://github.com/kubeedge/kubeedge/issues/3445

wackxu:
we need merge this and some user has encountered this problem. cc @fisherxu @vincentgoat

fisherxu (commented):
/lgtm
/approve

RyanZhaoXB:
Hello @wackxu , this PR may not resolve the problem in https://github.com/kubeedge/kubeedge/issues/3445. The problem still exists when user uses edgemesh for DNS resolving.

Issue #3582, opened 2022-01-21
Plans for Node Group Management and Service Scope Management


vincentgoat:
Hi guys, we have plans for achieving the feature of Pod Scheduling among node groups in phase 1, and we also make plans for achieving the feature of Service Scope within a node group in phase 2.

If you have features to achieve or to get involved in the development, please see the project form below, and you can add other todos here as you want:
**Phases 1: Node Group Management**
- [ ] Define a NodeGroup CR to indicate which nodes belong to the node group.
- [ ] Refer to Karmada policy, define a PropagationPolicy CR that specifies which node group pods should be scheduled to and how many pods should run in this node group.
- [ ] Unit/E2E Test

**Phases 2: Service Scope Management**
TODO

vincentgoat:
2022.1.24 Meeting Minutes:
1. Achieve the override in cloud core, not only override images path but also other attributes, the data comes from an independent policy.
2. Use the kubectl plug-in to new operations for node group.
3. The controller runtime version depends on the Kubeedge version update.
4. Determine the API.
5. Create a Google Doc for more details about the proposal.

vincentgoat:
proposal: #3574

Congrool:
2022.1.26 Meeting:
1. Only focus on new applications which have not been deployed.
2. Find some way to solve the contention when applying propogation policy, such as make pods with specified label/annotation in pending state until the extender finds the relative policy.
3. Modify the PropagationPolicy API to support  scheduling explicit number of pods to specified nodegroups.

benjaminhuo:
Looks like NodeGroup and ServiceScope is trying to achieve goals similar to [OpenYurt YurtAppManager](https://openyurt.io/docs/core-concepts/yurt-app-manager) and [SuperEdge ServiceGroup](https://superedge.io/docs/components/servicegroup/)  but in different ways?

Both [OpenYurt YurtAppManager](https://openyurt.io/docs/core-concepts/yurt-app-manager) and [SuperEdge ServiceGroup](https://superedge.io/docs/components/servicegroup/)  distribute workload by creating sub-deployment while KubeEdge is try to distribute pods in a single deployment to different NodeGroup?

It might work well for deployment, but is this ok for statefulset to spread pods of a single statefulset to different edge locations?  For example: 
- 2 pods for a MySQL statefulset, one is primary and another is replica
- 3 pods for an Elasticsearch statefulset, each pod holds a portion of the data

And where can I find the proposal for Service Scope?

[...]

PR #3640, opened 2022-03-03
add a proposal for EdgeApplication

vincentgoat:
issue: #3582 
doc: [https://docs.google.com/document/d/19FHzmHB-a-OSUukmicEVga6neMCeHz_ozPFxaGXqJ1I/edit#heading=h.d5vhf8rdz3fj](url)

new add apis for EdgeApplication, it is may not a final solution and needs to be optimized, we can discuss more details about this proposal when you guys are free, thanks.

/cc @fisherxu @lonelyCZ @Congrool

### DIFF
--- a/cloud/pkg/apis/edgeapplication/v1alpha1/type.go
+++ b/cloud/pkg/apis/edgeapplication/v1alpha1/type.go
@@ -1,0 +1,221 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Applied")].status`,name="Applied",type=string
+// +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
+
+// EdgeApplication defines a list of resources to be deployed on the node groups.
+type EdgeApplication struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec represents the desired behavior of EdgeApplication.
+	Spec EdgeAppSpec `json:"spec"`
+
+	// Status represents the status of PropagationStatus.
+	// +optional
+	Status EdgeAppStatus `json:"status,omitempty"`
+}
+
+// EdgeAppSpec defines the desired state of EdgeApplication.
+type EdgeAppSpec struct {
+	// ResourceTemplate represents the manifest workload to be deployed on managed node groups.
+	ResourceTemplate ResourceTemplate `json:"resourceTemplate,omitempty"`
+	// WorkloadScopes represents the scope of workload to be deployed.
+	WorkloadScopes WorkloadScope `json:"workloadScopes"`
+}
+
+// WorkloadScope represents the scope of workload to be deployed.
+type WorkloadScope struct {
+	// TargetNodeGroups represents the target node groups of workload to be deployed.
+	// +optional
+	TargetNodeGroups []TargetNodeGroups `json:"targetNodeGroups,omitempty"`
+	// TargetNodes represents the target nodes of workload to be deployed.
+	// +optional
+	TargetNodes []TargetNodes `json:"targetNodes,omitempty"`
+}
+
+// TargetNodeGroups represents the target node groups of workload to be deployed.
+type TargetNodeGroups struct {
+	// Name represents the name of target node group
+	Name string `json:"name"`
+	// Overriders offers various alternatives to represent the override rules.
+	Overriders Overriders `json:"overriders,omitempty"`
+}
+
+// TargetNodes represents the target nodes of workload to be deployed.
+type TargetNodes struct {
+	// Label of target nodes
+	Label labels.Selector `json:"label"`
+	// Overriders offers various alternatives to represent the override rules.
+	Overriders Overriders `json:"overriders"`
+}
+
+// ResourceTemplate represents the manifest workload to be deployed on managed node groups.
+type ResourceTemplate struct {
+	// Manifests represents a list of Kubernetes resources to be deployed on the managed node groups.
+	// +optional
+	Manifests []Manifest `json:"manifests,omitempty"`
+}
+
+// Overriders offers various alternatives to represent the override rules.
+//
+// If more than one alternatives exist, they will be applied with following order:
+// - ImageOverrider
+// - Plaintext
+type Overriders struct {
+	// Replicas of deployment
+	// +optional
+	Replicas int `json:"replicas,omitempty"`
+	// ImageOverrider represents the rules dedicated to handling image overrides.
+	// +optional
+	ImageOverrider []ImageOverrider `json:"imageOverrider,omitempty"`
+}
+
+// ImageOverrider represents the rules dedicated to handling image overrides.
+type ImageOverrider struct {
+	// Predicate filters images before applying the rule.
+	//
+	// Defaults to nil, in that case, the system will automatically detect image fields if the resource type is
+	// Pod, ReplicaSet, Deployment or StatefulSet by following rule:
+	//   - Pod: spec/containers/<N>/image
+	//   - ReplicaSet: spec/template/spec/containers/<N>/image
+	//   - Deployment: spec/template/spec/containers/<N>/image
+	//   - StatefulSet: spec/template/spec/containers/<N>/image
+	// In addition, all images will be processed if the resource object has more than one containers.
+	//
+	// If not nil, only images matches the filters will be processed.
+	// +optional
+	Predicate *ImagePredicate `json:"predicate,omitempty"`
+
+	// Component is part of image name.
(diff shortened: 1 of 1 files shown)


cloud/pkg/apis/edgeapplication/v1alpha1/type.go
```diff
@@ -0,0 +1,221 @@
...
+	// Manifests represents a list of Kubernetes resources to be deployed on the managed node groups.
+	// +optional
+	Manifests []Manifest `json:"manifests,omitempty"`
+}
+
+// Overriders offers various alternatives to represent the override rules.
+//
+// If more than one alternatives exist, they will be applied with following order:
+// - ImageOverrider
+// - Plaintext
```
Congrool, line 73: It seems that no Plaintext in the `Overriders` struct. So, I think there's no need to give the description of the priority.

And should we support plaintext in the later release?

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
```diff
@@ -0,0 +1,221 @@
...
+	// Name represents the name of target node group
+	Name string `json:"name"`
+	// Overriders offers various alternatives to represent the override rules.
+	Overriders Overriders `json:"overriders,omitempty"`
+}
+
+// TargetNodes represents the target nodes of workload to be deployed.
+type TargetNodes struct {
+	// Label of target nodes
+	Label labels.Selector `json:"label"`
```
Congrool, line 57: How about supporting selecting nodes with nodename?

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
```diff
@@ -0,0 +1,221 @@
...
+	runtime.RawExtension `json:",inline"`
+}
+
+// EdgeAppStatus defines the observed state of EdgeApplication.
+type EdgeAppStatus struct {
+	// Conditions contain the different condition statuses for this work.
+	// Valid condition types are:
+	// 1. Applied represents workload in EdgeApplication is applied successfully on a managed node groups.
+	// 2. Progressing represents workload in EdgeApplication is being applied on a managed node groups.
+	// 3. Available represents workload in EdgeApplication exists on the managed node groups.
```
Congrool, line 160: What's the difference between `Applied` and `Available` ?

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
```diff
@@ -0,0 +1,221 @@
...
+// ManifestStatus contains running status of a specific manifest in spec.
+type ManifestStatus struct {
+	// Identifier represents the identity of a resource linking to manifests in spec.
+	// +required
+	Identifier ResourceIdentifier `json:"identifier"`
+
+	// Status reflects running status of current manifest.
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +optional
+	Status *runtime.RawExtension `json:"status,omitempty"`
```
Congrool, line 180: Will we copy the status filed of manifests as a whole into the ManifestStatus here?

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
Congrool: Do we really need the TargetNodes field?
1. if users specify both `TargetNodeGroups` and `TargetNodes`, what will happen?
2. though we select `TargetNodes` with label, how about just creating a nodegroup for them and selecting them as in `TargetNodeGroups`?

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
vincentgoat: yes, there is no need to give the priority description here now, we can support plaintext if users have requirements in the future, and I think imageOverrider can be achieved first.

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
vincentgoat: > 
I think it's ok with the node name.

vincentgoat: > Do we really need the TargetNodes field?
> 
> 1. if users specify both `TargetNodeGroups` and `TargetNodes`, what will happen?
> 2. since we select `TargetNodes` with label, how about just creating a nodegroup for them and selecting them in `TargetNodeGroups`?

1. only support one of them at the same time, if we got two specifies then an error will return? or we only support TargetNodeGroups in EdgeApplication? what's your opinion? @fisherxu @Congrool 
2. Do you mean to create a nodegroup auto?

Congrool: > Do you mean to create a nodegroup auto?

Well, I mean users can manually create a nodegroup for these nodes. Because these nodes has already had the same label which means these nodes have the same attribute, maybe location or something else, it's reasonable to organize them as a nodegroup.

Congrool: >Well, I mean users can manually create a nodegroup for these nodes. Because these nodes has already had the same label which means these nodes have the same attribute, maybe location or something else, it's reasonable to organize them as a nodegroup.

I think there're also some problems here. If we group nodes with the attribute except the location, we cannot assume that the network between these nodes is available, which means the service scope for them makes no sense. Considering this problem, should we let the users to determine which nodegroups use the service scope and which not.

vincentgoat: > > Well, I mean users can manually create a nodegroup for these nodes. Because these nodes has already had the same label which means these nodes have the same attribute, maybe location or something else, it's reasonable to organize them as a nodegroup.
> 
> I think there're also some problems here. If we group nodes with the attribute except the location, we cannot assume that the network between these nodes is available, which means the service scope for them makes no sense. Considering this problem, should we let the users to determine which nodegroups use the service scope and which not.

yes, the network problem really exists, would it be too complex for users to configure service scope and deploy scope both, we can add constraints descriptions for nodegroup that we only support nodes in one network to be added into the same nodegroup.

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
vincentgoat: yes

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
Congrool: >yes, the network problem really exists, would it be too complex for users to configure service scope and deploy scope both, we can add constraints descriptions for nodegroup that we only support nodes in one network to be added into the same nodegroup.

Agree+1
let's return back to the previous problem. Under this cercumstance, do we need to provide the service scope feature for `TargetNodes`? And users should guarrantee the network availability between these nodes selected by label.

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
vincentgoat: Reference to Karmada's resource "work", EdgeAppApplied/EdgeAppAvailable are status type, EdgeAppApplied means EdgeApplication controller already split workload to nodegroups, EdgeAppAvailable means workload status is available after deploying to nodegroups.

cloud/pkg/apis/edgeapplication/v1alpha1/type.go
vincentgoat: > > yes, the network problem really exists, would it be too complex for users to configure service scope and deploy scope both, we can add constraints descriptions for nodegroup that we only support nodes in one network to be added into the same nodegroup.
> 
> Agree+1 let's return back to the previous problem. Under this cercumstance, do we need to provide the service scope feature for `TargetNodes`? And users should guarrantee the network availability between these nodes selected by label.

no need to provide the service scope feature for `TargetNodes`, we achieve `TargetNodes` to be compatible for users deploying resources through EdgeApplication like the normal way by k8s apis .

Congrool: >no need to provide the service scope feature for TargetNodes, we achieve TargetNodes to be compatible for users deploying resources through EdgeApplication like the normal way by k8s apis .
>since we select TargetNodes with label, how about just creating a nodegroup for them and selecting them in TargetNodeGroups?

Conclude from the discussion, here's my opinion:
1. Only use `TargetNodeGroups`
2. Add a new field to NodeGroup, temporarily calling it as`EnableServiceScope`. It works like a switch to control if we need to provide the service scope function for this nodegroup. The default value is false. Users can determine to set it to true if they can guarrantee the network availability.
3. With this method, we can simplify the logic when processing the EdgeApplication.

vincentgoat: > > no need to provide the service scope feature for TargetNodes, we achieve TargetNodes to be compatible for users deploying resources through EdgeApplication like the normal way by k8s apis .
> > since we select TargetNodes with label, how about just creating a nodegroup for them and selecting them in TargetNodeGroups?
> 
> Conclude from the discussion, here's my opinion:
> 
> 1. Only use `TargetNodeGroups`
> 2. Add a new field to NodeGroup, temporarily calling it as`EnableServiceScope`. It works like a switch to control if we need to provide the service scope function for this nodegroup. The default value is false. Users can determine to set it to true if they can guarrantee the network availability.
> 3. With this method, we can simplify the logic when processing the EdgeApplication.

1. yes we only use `TargetNodeGroups`
2. In my opinion, we assume that the network of nodes in a node group is normal and users should guarantee this, if we add this switch and set it to 'false', users can visit the service in a  cross-partition way, it is not our intention, so the effect of the new switch is not very obvious.

vincentgoat:
reopen

vincentgoat:
Replace with #3722

PR #3641, opened 2022-03-03
keadm: fix wrong directory check

zc2638:

**What type of PR is this?**

/kind bug

We should use `os.IsNotExist(err)` instead of `err == os.ErrNotExist`.

Fixes #

```release-note

```


gy95 (commented):
/lgtm
there maybe be some wrapped functions to predict whether file or dir exist or not, we can use those functions.

fisherxu (commented):
/approve

PR #3650, opened 2022-03-08
new edged with integrated kubelet

Shelley-BaoYue:
/kind feature

integrate kubelet with edged


fisherxu:
The branch is just for previewing, I'll manually merge it.

If there's any issue, just fix them in follow-up PRs.

Issue #3660, opened 2022-03-09
Move helm from build to root dir


fisherxu:
Now the helm is related in build, let's move it to root dir.

kubeedge/build/helm -> kubeedge/helm

VMois:
I think this issue can be closed.

rahul3002:
can I work on this issue

rahul3002:
i would like to contribute to this issue

rohitsinghnaruka:
@fisherxu , i think this issue should be closed 
lemme know if its open as i can work upon it ??

rohitsinghnaruka:
i would like to contribute on this issue?

fisherxu:
Thanks for your concern, but I think this issue has been fixed.. @rohitsinghnaruka

inosmeet:
Should be closed by #3692

Ankitavasudev:
Hi, I'd like to work on this. This seems straightforward - moving the helm charts from \uild/helm\ to \helm/\ in the root directory.

Before I start, I have a quick question: should I also update any CI/CD workflows or Makefiles that reference the old \uild/helm\ path? Want to make sure nothing breaks.

Thanks!

PR #3643, opened 2022-03-04
keadm: unified access to version

zc2638:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3662, opened 2022-03-09
Fix and move helm base dir

zhu733756:
**What type of PR is this?**
Fix and move helm base dir.

Fixes #https://github.com/kubeedge/kubeedge/issues/3660

```release-note
helm base dir changed
```


keadm/cmd/keadm/app/cmd/util/helminstaller.go
```diff
@@ -381,15 +381,9 @@ func (cu *KubeCloudHelmInstTool) handleProfile(profileValue string) error {
...
 		}
-		profileValueSuffix := strings.TrimPrefix(profileValue, "v")
-		// confirm it startswith "v"
-		if profileValue != profileValueSuffix {
-			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=v%s", "cloudCore.image.tag", profileValueSuffix))
-			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=v%s", "iptablesManager.image.tag", profileValueSuffix))
-		} else {
-			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "cloudCore.image.tag", profileValue))
-			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "iptablesManager.image.tag", profileValue))
-		}
```
fisherxu, line 392: v is no need now?

keadm/cmd/keadm/app/cmd/util/helminstaller.go
```diff
@@ -18,8 +18,8 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"sigs.k8s.io/yaml"
 
-	keCharts "github.com/kubeedge/kubeedge/build/helm/charts"
 	"github.com/kubeedge/kubeedge/common/constants"
+	keCharts "github.com/kubeedge/kubeedge/helm/charts"
```
fisherxu, line 22: For import, it's better to use the lowcase?

keadm/cmd/keadm/app/cmd/util/helminstaller.go
zhu733756: Yeah, this is a custom flag,  "cloudCore.image.tag" or "iptablesmgr.image.tag" is controlled by user if they have a custom image. If not given, will use the default kubeedge version.

keadm/cmd/keadm/app/cmd/util/helminstaller.go
zhu733756: Done.

keadm/cmd/keadm/app/cmd/util/helminstaller.go
fisherxu: Okey, users can set the custom value, so if we always add the `v`, it not reasonable, right?

zhu733756: > Okey, users can set the custom value, so if we always add the `v`, it not reasonable, right?

I think using v is professional, but once the user sets a custom image tag, should override our default version.

fisherxu: In k8s/kubeadm, it has a version match, if match successfully, will add  the `v`. If match failed, which should means the version is customed, will not add `v`.

I think users may set the version but ignore the `v` prefix, so deploy will failed, how about we also  have a check like k8s? @zhu733756

fisherxu: Ref: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/version.go#L152

zhu733756: Seems good.

zhu733756: But I am worried that if the user gives a tag that matched the regex but is not given a "v" prefix string,  their custom image tag would not match through this method when we return a "v" prefix.

keadm/cmd/keadm/app/cmd/util/helminstaller.go
```diff
@@ -205,17 +207,21 @@ func (cu *KubeCloudHelmInstTool) beforeRenderer(baseHelmRoot string) error {
...
-		if cu.CloudcoreImage != "" {
-			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "cloudCore.image.repository", cu.CloudcoreImage))
+
+		if cu.CloudcoreRepo != "" {
+			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "cloudCore.image.repository", cu.CloudcoreRepo))
 		}
+
 		if cu.CloudcoreTag != "" {
-			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "cloudCore.image.tag", cu.CloudcoreTag))
+			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "cloudCore.image.tag", normalizedBuildVersion(cu.CloudcoreTag)))
```
zhu733756, line 216: We suggest the custom image build with a v prefix.

fisherxu (commented):
/approve
/assign @zc2638

zc2638:
I recommend getting a unified version everywhere. #3643
And it is not recommended to specify a specific component image repo like `CloudcoreRepo`, otherwise not only the maintenance cost is high, but the user cost is also high

zc2638:
There may be other resources that need to be embedded in the future. It is recommended to unify a reasonable directory and place the helm manifest resources in the secondary directory.
@fisherxu

zhu733756:
> I recommend getting a unified version everywhere. #3643 

This pr is a little different from getting the default kubeedge version, it is actually a image tag, user can define it as what they wanted. 

> And it is not recommended to specify a specific component image repo like `CloudcoreRepo`, otherwise not only the maintenance cost is high, but the user cost is also high

It comes from the common helm values definition,  repo seems to repository, always described as "kubeedge/cloudcore", and tag means the tag of this image.  See https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L415

zhu733756:
> There may be other resources that need to be embedded in the future. It is recommended to unify a reasonable directory and place the helm manifest resources in the secondary directory. @fisherxu

> I recommend getting a unified version everywhere. #3643 

The regex should move to a common position I think.

> And it is not recommended to specify a specific component image repo like `CloudcoreRepo`, otherwise not only the maintenance cost is high, but the user cost is also high

It comes from the common helm values definition. See https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L415. It is a fast path for set flag like "--set cloudCore.image.repo=xxx".  Do we unexpose it?

zhu733756:
```
1._output/local/bin/keadm beta init --advertise-address=192.168.88.6 --set cloudCore.service.enable=true --set iptablesManager.mode=external --set cloudCore.hostNetWork=false --kubeedge-version=v1.10.0 --force
2. _output/local/bin/keadm beta init --advertise-address=192.168.88.6 --set cloudCore.service.enable=true --set iptablesManager.mode=external --set cloudCore.hostNetWork=false --profile version=1.10.0 --force
3.  _output/local/bin/keadm beta init --advertise-address=192.168.88.6 --set cloudCore.service.enable=true --set iptablesManager.mode=external --set cloudCore.hostNetWork=false --profile version=v1.10.0 --force
```

Neither one of these above three commands works well.

keadm/cmd/keadm/app/cmd/beta/manifest.go
```diff
@@ -95,62 +96,55 @@ func NewBetaManifestGenerate() *cobra.Command {
...
-	if BetaInitOptions.Namespace == "" {
-		BetaInitOptions.Namespace = constants.SystemNamespace
+		kubeedgeVersion = types.DefaultKubeEdgeVersion
+		fmt.Println("Failed to get the latest KubeEdge release version, will use default version: ", kubeedgeVersion)
 	}
 
 	common := util.Common{
-		ToolVersion: semver.MustParse(kubeVer),
-		KubeConfig:  BetaInitOptions.KubeConfig,
+		ToolVersion: semver.MustParse(kubeedgeVersion),
```
zhu733756, line 129: kubeedgeVersion would use a unified method to generate in the future.

zhu733756:
Pls move to #3692.  Emmm, I found that if the user doesn't have a remote tag v1.10.0+, the e2e test would be failed.

PR #3534, opened 2022-01-06
keadm support config images cmd

gy95:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

part of https://github.com/kubeedge/kubeedge/issues/3475

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3671, opened 2022-03-10
use dependency to pull images

gy95:
**What type of PR is this?**
/kind cleanup

use docker dependencies to pull images, instead of using `docker pull` cmd directly
fix https://github.com/kubeedge/kubeedge/pull/3534#discussion_r819323699

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/beta/config.go
+++ b/keadm/cmd/keadm/app/cmd/beta/config.go
@@ -17,18 +17,20 @@ limitations under the License.
 package beta
 
 import (
+	"context"
 	"fmt"
+	"io"
 	"strings"
 
+	dockertypes "github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
+	dockerclient "github.com/docker/docker/client"
 	"github.com/spf13/cobra"
 
 	"github.com/kubeedge/kubeedge/common/constants"
 	cmdcommon "github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
We need to support docker or other remote type runtime? cc @zc2638

gy95:
> We need to support docker or other remote type runtime? cc @zc2638

ok, will open another pr to support this.

keadm/cmd/keadm/app/cmd/beta/config.go
```diff
@@ -219,14 +228,23 @@ func GetPauseImage() string {
...
+
+	rc, err := cli.ImagePull(ctx, image, dockertypes.ImagePullOptions{})
+	if err != nil {
+		return err
+	}
+	if _, err = io.Copy(io.Discard, rc); err != nil {
+		return err
 	}
-	return fmt.Errorf("dcoker pull image failed: %v", err)
+	return rc.Close()
```
zc2638, line 249: It might be better to use `defer` here

gy95: it's in the last line of the function, so it seems duplicate to use defer to close the io :)

zc2638:
```golang
if _, err = io.Copy(io.Discard, rc); err != nil {
	return err
}
```
Here if err is returned, we should also call the close method

gy95: fixed, please look again :)

fisherxu (commented):
/approve

PR #3678, opened 2022-03-10
Fix local-up exit without error msg

fisherxu:

**What type of PR is this?**
/kind test

If the kubectl is not installed, the local-up will exit, the message as below:
```
Cleaning up...
Running kind: [kind delete cluster --name test]
Deleting cluster "test" ...
go detail version: go version go1.16 linux/amd64
go version: 1.16
checking kubectl
Cleaning up...
Running kind: [kind delete cluster --name test]
Deleting cluster "test" ...
```
The kubectl error msg is not printed. Because in golang.sh source the set -x, so check kubectl will exit.

Just remove the set-x  in golang.sh, hack-local-up.sh will set it.

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -19,10 +19,6 @@
 # KubeEdge Authors:
 # To Get Detail Version Info for KubeEdge Project
 
-set -o errexit
-set -o nounset
-set -o pipefail
-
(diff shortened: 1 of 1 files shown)


wackxu:
lgtm

PR #3679, opened 2022-03-10
fix unhanle error

vincentgoat:

**What type of PR is this?**
/kind cleanup

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -574,10 +574,13 @@ func kubeClientGet(uc *UpstreamController, namespace string, name string, queryT
 		klog.Error(err)
 		return nil, err
 	}
-	if err == nil && obj != nil {
-		util.SetMetaType(obj.(runtime.Object))
+	if err != nil {
+		return nil, err
 	}
-	return obj, err
+	if err := util.SetMetaType(obj.(runtime.Object)); err != nil {
+		return nil, err
+	}
+	return obj, nil
 }
 
 func queryInner(uc *UpstreamController, msg model.Message, queryType string) {


cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -575,7 +575,10 @@ func kubeClientGet(uc *UpstreamController, namespace string, name string, queryT
 		return nil, err
 	}
 	if err == nil && obj != nil {
-		util.SetMetaType(obj.(runtime.Object))
+		err := util.SetMetaType(obj.(runtime.Object))
+		if err != nil {
+			klog.Warningf("set MetaType error %v for obj %v, %v", err, queryType, obj.GetName())
+		}
 	}
 	return obj, err
```
zc2638, line 583: Can the upper layer process guarantee that when `err == nil`, obj must be non-nil.
By the way, there is no need to output the log here, because the log is already output after the method is called.
ref: https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/edgecontroller/controller/upstream.go#L598

```golang
object, err := kubeClientGet(uc, namespace, name, queryType, msg)
if errors.IsNotFound(err) {
	klog.Warningf("message: %s process failure, resource not found, namespace: %s, name: %s", msg.GetID(), namespace, name)
	return
}
if err != nil {
	klog.Warningf("message: %s process failure with error: %s, namespace: %s, name: %s", msg.GetID(), err, namespace, name)
	return
}
```

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -575,7 +575,10 @@ func kubeClientGet(uc *UpstreamController, namespace string, name string, queryT
 		return nil, err
 	}
 	if err == nil && obj != nil {
-		util.SetMetaType(obj.(runtime.Object))
+		err := util.SetMetaType(obj.(runtime.Object))
+		if err != nil {
+			klog.Warningf("set MetaType error %v for obj %v, %v", err, queryType, obj.GetName())
+		}
```
fisherxu, line 581: If the error in not null, we need to return err? cc @neiba

neiba: i agree with you, the upper layer process guarantee that when err == nil, obj must be non-nil.

cloud/pkg/edgecontroller/controller/upstream.go
vincentgoat: But we can't log error of function util.SetMetaType()

cloud/pkg/edgecontroller/controller/upstream.go
vincentgoat: already updated  @zc2638 @neiba @fisherxu

fisherxu (commented):
/approve
/assign @zc2638

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -574,9 +574,10 @@ func kubeClientGet(uc *UpstreamController, namespace string, name string, queryT
 		klog.Error(err)
 		return nil, err
 	}
-	if err == nil && obj != nil {
-		util.SetMetaType(obj.(runtime.Object))
+	if err != nil {
+		return obj, err
 	}
+	err = util.SetMetaType(obj.(runtime.Object))
```
zc2638, line 580: `SetMetaType` should return obj as nil on error.
If you want to ignore errors, don't assign err.

zc2638: Or obj is not nil, is there any special scenario?

vincentgoat: > Or obj is not nil, is there any special scenario?

This pr try to fix go lint error, which reports "unhandle error", so we should assign this err explicitly.

zc2638: > `SetMetaType`错误时将 obj 返回为 nil。 如果您想分配错误，请不要分配错误。

Will this bring any new problems?

vincentgoat: will not

zc2638:
```suggestion
	if err := util.SetMetaType(obj.(runtime.Object)); err != nil {
	    return nil, err
	}
```

vincentgoat: updated

zc2638:
thanks for the fixing
/lgtm

PR #3692, opened 2022-03-13
Fix and move helm base dir 

zhu733756:
**What type of PR is this?**
Fix and move helm base dir.

Fixes #https://github.com/kubeedge/kubeedge/issues/3660

```release-note
helm base dir changed
```


keadm/cmd/keadm/app/cmd/common/types.go
```diff
@@ -30,15 +30,16 @@ type InitOptions struct {
...
 	KubeConfig       string
+	KubeEdgeVersion  string
 	AdvertiseAddress string
 	Manifests        string
 	Namespace        string
-	CloudcoreImage   string
+	CloudcoreRepo    string
 	CloudcoreTag     string
-	IptablesMgrImage string
+	IptablesMgrRepo  string
```
fisherxu, line 42: These two options haven't exposed in flag?

zhu733756: Yes. We can just use --set falg to confirm it.

fisherxu: Using --set flag to set value in chart directly?

zhu733756: > Using --set flag to set value in chart directly?

Yes. The user can input their own tag or repo.

fisherxu: Ok, we need more docs. And the options here can  be removed?

zhu733756: Yes, i will fix it

zhu733756: Add ref: https://github.com/kubeedge/kubeedge/tree/master/charts/cloudcore/README.md

fisherxu (commented):
/approve
/assign @zc2638

fisherxu:
Looks good.

zc2638:
The edgemesh directory under the KubeEdge root directory has been explained, here `manifests/charts/edgemesh` should not need to be added again.

fisherxu:
> The edgemesh directory under the KubeEdge root directory has been explained, here `manifests/charts/edgemesh` should not need to be added again.

We can add edgemesh as addon here?

zc2638:
> We can add edgemesh as addon here?

As a plug-in, edgemesh is a network component supported by the KubeEdge community by default. It is a reasonable way to use it as a default supported option during installation, but we also need to ensure that it does not have to be installed.

zhu733756:
> > We can add edgemesh as addon here?
> 
> As a plug-in, edgemesh is a network component supported by the KubeEdge community by default. It is a reasonable way to use it as a default supported option during installation, but we also need to ensure that it does not have to be installed.

use "--profile version=xxx" will install cloudcore chart, use  "--profile edgemesh" will install edgemesh chart if the chart implements here. Currently, we use  "--external-helm-root" to install edgemesh.

zc2638:
Edgemesh itself should have maintenance charts. Can edgemesh be introduced like middlewares (e.g. mysql, redis)?

zc2638:
/lgtm 
Other work can be completed later.

PR #3693, opened 2022-03-13
add subresource field in application

Shelley-BaoYue:

add subresource field in application for api with subresource

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/application.go
+++ b/cloud/pkg/dynamiccontroller/application/application.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	apirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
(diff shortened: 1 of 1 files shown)


RyanZhaoXB (commented):
LGTM

cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -493,7 +509,13 @@ func (c *Center) ProcessApplication(app *Application) (interface{}, error) {
...
 			return nil, err
 		}
-		retObj, err := c.kubeclient.Resource(gvr).Namespace(ns).Update(context.TODO(), obj, *option)
+		var retObj interface{}
+		var err error
+		if app.Subresource == "" {
+			retObj, err = c.kubeclient.Resource(gvr).Namespace(ns).Update(context.TODO(), obj, *option)
+		} else {
+			retObj, err = c.kubeclient.Resource(gvr).Namespace(ns).Update(context.TODO(), obj, *option, app.Subresource)
+		}
```
fisherxu, line 518: Looks good. In K8s, for create method, if it has no `subresource`, the path can't contain {name} parameter.
And for the update method, it must contain {name} parameter. 

Ref: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/installer.go#L470

fisherxu (commented):
/lgtm
/approve

Issue #3586, opened 2022-01-24
type.**k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1.ContainerPreferredAllocationRequest: dwarf: definition of unknown kind 0


dafanshu:
**What happened**:
compile error in raspberry pi 3B+:
`ubuntu@ubuntu:~/kubeedge-1.8.2$ CGO_ENABLED=1 GOARM=7 GOOS=linux GOARCH=arm go build -o ./_out ./edge/cmd/edgecore
# github.com/kubeedge/kubeedge/edge/cmd/edgecore
type.**k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1.ContainerPreferredAllocationRequest: dwarf: definition of unknown kind 0
panic: runtime error: slice bounds out of range [40:36]

goroutine 1 [running]:
cmd/link/internal/ld.decodetypeIfaceMethodCount(0x415f80, {0x91735251, 0x24, 0x24})
        /usr/local/go/src/cmd/link/internal/ld/decodesym.go:92 +0x84
cmd/link/internal/ld.(*dwctxt).newtype(0x4c8d51c, 0x11cba7)
        /usr/local/go/src/cmd/link/internal/ld/dwarf.go:637 +0x390
cmd/link/internal/ld.(*dwctxt).defgotype(0x4c8d51c, 0x11cba7)
        /usr/local/go/src/cmd/link/internal/ld/dwarf.go:536 +0x10c
cmd/link/internal/ld.(*dwctxt).dwarfVisitFunction(0x4c8d51c, 0x75561, 0x21022a0)
        /usr/local/go/src/cmd/link/internal/ld/dwarf.go:1709 +0x524
cmd/link/internal/ld.dwarfGenerateDebugInfo(0x10ec000)
        /usr/local/go/src/cmd/link/internal/ld/dwarf.go:1875 +0x1544
cmd/link/internal/ld.Main(0x415f80, {0x4, 0x8, 0x1, 0xd, 0xe, 0x1c00000, {0x0, 0x0}, {0x295198, ...}, ...})
        /usr/local/go/src/cmd/link/internal/ld/main.go:265 +0x128c
main.main()
        /usr/local/go/src/cmd/link/main.go:69 +0xbb0
`
 
**What you expected to happen**:
just compile edgecore module

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
[...]

PR #3723, opened 2022-03-21
Use armhf as default for armv7/v6

fisherxu:

**What type of PR is this?**
/kind cleanup

Now for the armv6 and later, all support armhf, and for ubuntu since 12.04, armel will not supported.

When cross-building, arm-linux-gnueabi-gcc will build armel by default, arm-linux-gnueabihf-gcc  will build armhf by default.

And armhf has better compatibility. Armv5 or before I think it will be probably rarely used?

Ref: https://github.com/kubeedge/kubeedge/issues/3586#issuecomment-1021027805
https://github.com/golang/go/wiki/GoArm#supported-architectures  
https://github.com/golang/go/wiki/GoArm#supported-operating-systems

cc @zhu733756 @dafanshu

Fixes #https://github.com/kubeedge/kubeedge/issues/3586#issuecomment-1021027805

```release-note
NONE
```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -60,7 +60,7 @@ jobs:
       - name: Install dependences
         run: |
           sudo apt-get update
-          sudo apt-get install -y upx-ucl gcc-aarch64-linux-gnu libc6-dev-arm64-cross gcc-arm-linux-gnueabi libc6-dev-armel-cross
+          sudo apt-get install -y upx-ucl gcc-aarch64-linux-gnu libc6-dev-arm64-cross gcc-arm-linux-gnueabihf
 
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


gy95:
https://github.com/kubeedge/kubeedge/blob/749bd34f3076a3a89c95298381161a9335f07488/build/docker/build-tools/build-tools.dockerfile#L7
Is this also need fixing?

fisherxu:
> https://github.com/kubeedge/kubeedge/blob/749bd34f3076a3a89c95298381161a9335f07488/build/docker/build-tools/build-tools.dockerfile#L7
> 
> 
> Is this also need fixing?

Done, though it still haven't used

gy95:
lgtm 
cc @chendave  PTAL

chendave:
/approve
thanks!

PR #3737, opened 2022-03-25
add custom config option for EdgeCore config file

TianTianBigWang:

**What type of PR is this?**
/kind feature

use for `keadm beta join` command configurate the image repository from flag, and generate edgeCore config file

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/joinbeta.go
+++ b/keadm/cmd/keadm/app/cmd/edge/joinbeta.go
@@ -181,7 +181,11 @@ func createEdgeConfigFiles(opt *common.JoinOptions) error {
 	}
 	if edgeCoreConfig == nil {
 		klog.Infoln("The configuration does not exist or the parsing fails, and the default configuration is generated")
-		edgeCoreConfig = v1alpha1.NewDefaultEdgeCoreConfig()
+		cfgOpt := &v1alpha1.ConfigOptions{
+			ImageRepository: opt.ImageRepository,
+			KubeEdgeVersion: opt.KubeEdgeVersion,
+		}
+		edgeCoreConfig = v1alpha1.NewDefaultEdgeCoreConfig(cfgOpt)
 	}
 
 	edgeCoreConfig.Modules.EdgeHub.WebSocket.Server = opt.CloudCoreIPPort
--- a/edge/cmd/edgecore/app/options/options.go
+++ b/edge/cmd/edgecore/app/options/options.go
@@ -54,7 +54,7 @@ func (o *EdgeCoreOptions) Validate() []error {
 }
 
 func (o *EdgeCoreOptions) Config() (*v1alpha1.EdgeCoreConfig, error) {
-	cfg := v1alpha1.NewDefaultEdgeCoreConfig()
+	cfg := v1alpha1.NewDefaultEdgeCoreConfig(nil)
 	if err := cfg.Parse(o.ConfigFile); err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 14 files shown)


fisherxu (commented):
I'm thinking whether we should add the config file to keadm? How do you think? @zhu733756 @zc2638

zhu733756:
> I'm thinking whether we should add the config file to keadm? How do you think? @zhu733756 @zc2638

What kind of the configuration?  The current keadm already has the function of generating edgecore.yaml.

zc2638:
> I'm thinking whether we should add the config file to keadm? How do you think? @zhu733756 @zc2638

The problem solved by this PR does not affect whether `keadm` needs a separate configuration file. 
@TianTianBigWang can open an issue to discuss this and refine it?

TianTianBigWang:
> I'm thinking whether we should add the config file to keadm? How do you think? @zhu733756 @zc2638

This pr is only used to process the `edgeCore` configuration content after specifying a custom image. 

If we need to add a configuration file to `keadm`, we can discuss it later.

keadm/cmd/keadm/app/cmd/edge/joinbeta.go
```diff
@@ -181,7 +181,11 @@ func createEdgeConfigFiles(opt *common.JoinOptions) error {
 	}
 	if edgeCoreConfig == nil {
 		klog.Infoln("The configuration does not exist or the parsing fails, and the default configuration is generated")
-		edgeCoreConfig = v1alpha1.NewDefaultEdgeCoreConfig()
+		cfgOpt := &v1alpha1.ConfigOptions{
+			ImageRepository: opt.ImageRepository,
+			KubeEdgeVersion: opt.KubeEdgeVersion,
+		}
+		edgeCoreConfig = v1alpha1.NewDefaultEdgeCoreConfig(cfgOpt)
```
fisherxu, line 188: Why not modify the `edgeCoreConfig` at here with `opt.ImageRepository` and `opt.KubeEdgeVersion` directly?

fisherxu:
> > I'm thinking whether we should add the config file to keadm? How do you think? @zhu733756 @zc2638
> 
> The problem solved by this PR does not affect whether `keadm` needs a separate configuration file. @TianTianBigWang can open an issue to discuss this and refine it?

Sorry, I think I have made some mistakes :)

keadm/cmd/keadm/app/cmd/edge/joinbeta.go
TianTianBigWang: I think if we need to add other parameters to modify the configuration file in the future, it is more convenient to use a structure directly

fisherxu: In futrue, if we need to add config field for edgecore, add it to edgecore's config directly. And if add the options for keadm, the config file for keadm is a better way, then converting the keadm config to edgecore. How do you think?

zhu733756: > In futrue, if we need to add config field for edgecore, add it to edgecore's config directly. And if add the options for keadm, the config file for keadm is a better way, then converting the keadm config to edgecore. How do you think?

And also supports converting the edgecore.yaml to keadm.yaml is cool.

TianTianBigWang: That is a good idea.
But I don't think it should affect this PR. 
For example, we might discuss the contents of the keadm configuration file, default parameters, etc. 
This should be a process.

PR #3742, opened 2022-03-26
add ErrStatus in api-server application

Shelley-BaoYue:

**What type of PR is this?**
/kind feature

add ErrStatus param in list-watch response application，direct pass k8s api-server error to edge,  instead of set `InternalError`  for all errors.

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/application.go
+++ b/cloud/pkg/dynamiccontroller/application/application.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 	"time"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


cloud/pkg/dynamiccontroller/application/application.go
```diff
@@ -313,14 +315,19 @@ func (a *Agent) Apply(app *Application) error {
...
 		return errors.New(app.Reason)
 	case Approved:
 		return nil
 	case InApplying:
 		//continue
 	}
 	app.Wait()
+	if app.getStatus() == Rejected {
+		return &app.Error
+	}
```
fisherxu, line 330: Why repeating for `Rejected` here?

fisherxu (commented):
Looks good overall.

cloud/pkg/dynamiccontroller/application/application.go
Shelley-BaoYue: > Why repeating for `Rejected` here?

wait for applying status finish here, if app response status is Rejected, return k8s error directly.

fisherxu (commented):
/lgtm
/approve

PR #3743, opened 2022-03-27
fix residual terminating pods problem after edge node reconnected

### LINKED ISSUE #3736: When an edge node goes offline, many Terminating pods are generated
**What happened**:
1、When an edge node goes offline, many Terminating pods are generated。
2、Terminating Pods are not removed while edge node is online
3、EdgeCloud crash after a long time when edgemesh is enabled.

1、Do not produce many Terminating Pods when an edge node is offline.
2、Terminating Pods are removed while edge node is online

1、Deploy the POD on the specified edge node by  nodeName。
2、Disconnect the network or power off the edge nodes
3、Wait 10 minutes
 
**Environment**:
- Kubernetes version (use `kubectl version`): 1.20.6
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): 1.9.2

  - Hardware configuration (e.g. `lscpu`): i5-9400
  - OS (e.g. `cat /etc/os-release`): Ubuntu 18.04.6 LTS
  - Kernel (e.g. `uname -a`): 5.4.0-56-generic
  - Go version (e.g. `go version`): 1.16.6
  - Others:

  </details>

  - edgecore version (e.g. `edgecore --version`): 1.9.2
  - Hardware configuration (e.g. `lscpu`): i5-9400
  - OS (e.g. `cat /etc/os-release`): 18.04.6 LTS
  - Kernel (e.g. `uname -a`): 5.4.0-56-generic
  - Go version (e.g. `go version`): 1.16.6
  - Others:

  </details>

vincentgoat:

**What type of PR is this?**
/kind bug

fix residual terminating pods problem after edge node reconnected

1. As the function describe said at L150 https://github.com/kubernetes/client-go/blob/master/util/workqueue/queue.go
You must call Done with item when you have finished processing it.
Otherwise, the key will always exists in processing set and can't process by nodeQueue.Get() again. 

2. Delete the `nodeConn` concurrently will also cause terminating pods that can't be deleted.

Fixes #3736

cc @fisherxu 

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -470,11 +470,8 @@ func (mh *MessageHandle) MessageWriteLoop(info *model.HubInfo, stopServe chan Ex
 		if err != nil {
 			klog.Errorf("Failed to send event to node: %s, affected event: %s, err: %s",
 				info.NodeID, dumpMessageMetadata(copyMsg), err.Error())
+			nodeQueue.Done(key)
 			nodeQueue.Add(key.(string))
-			if strings.Contains(err.Error(), "use of closed network connection") {
-				mh.nodeConns.Delete(info.NodeID)
-				continue
-			}
 			time.Sleep(time.Second * 2)
 		}
 


wackxu:
@vincentgoat the e2e test is failed, PTAL

vincentgoat:
> @vincentgoat the e2e test is failed, PTAL

Done

fisherxu:
cc @My-pleasure

My-pleasure:
> cc @My-pleasure

It's good that adding a `Done` before `Add`.

vincentgoat:
There is no need to delete nodeConns of the edge node here. For the scenario, killing edgecore process and reconnect immediately, then deleting the pod of the edge node, the pod will be `Terminating`.
The reason is that after the edge node reconnects and registers again, nodeConns is reloaded, meanwhile `mh.sendMsg` will retry to send and it will take about 20s, when retry timeout happend, the nodeConns of the edge node will be deleted here and we can't send message to edge node without nodeConns .

So removing `mh.nodeConns.Delete(info.NodeID)` here.

cc @My-pleasure

My-pleasure:
> There is no need to delete nodeConns of the edge node here. For the scenario, killing edgecore process and reconnect immediately, then deleting the pod of the edge node, the pod will be `Terminating`. The reason is that after the edge node reconnects and registers again, nodeConns is reloaded, meanwhile `mh.sendMsg` will retry to send and it will take about 20s, when retry timeout happend, the nodeConns of the edge node will be deleted here and we can't send message to edge node without nodeConns .
> 
> So removing `mh.nodeConns.Delete(info.NodeID)` here.
> 
> cc @My-pleasure

When there is a error which is `use of closed network connection`, we delete the nodeConn to stop send msg to this edgenode and to avoid print too many err logs. So maybe we can take out the `nodeConn` and compare with the old `conn` before deleting `nodeConn`. Like following:
```
newConn, _ := mh.nodeConns.Load(info.NodeID)
if reflect.DeepEqual(newConn, conn) {
	// delete
}
```

vincentgoat:
> > There is no need to delete nodeConns of the edge node here. For the scenario, killing edgecore process and reconnect immediately, then deleting the pod of the edge node, the pod will be `Terminating`. The reason is that after the edge node reconnects and registers again, nodeConns is reloaded, meanwhile `mh.sendMsg` will retry to send and it will take about 20s, when retry timeout happend, the nodeConns of the edge node will be deleted here and we can't send message to edge node without nodeConns .
> > So removing `mh.nodeConns.Delete(info.NodeID)` here.
> > cc @My-pleasure
> 
> When there is a error which is `use of closed network connection`, we delete the nodeConn to stop send msg to this edgenode and to avoid print too many err logs. So maybe we can take out the `nodeConn` and compare with the old `conn` before deleting `nodeConn`. Like following:
> 
> ```
> newConn, _ := mh.nodeConns.Load(info.NodeID)
> if reflect.DeepEqual(newConn, conn) {
> 	// delete
> }
> ```

when the node is disconnected, `nodeConn` will be deleted after keepalive timer timeout (default 30s), maybe we should add rateLimit queue here to avoid printing too many err logs, suggest not to concurrent delete the `nodeConn`.

My-pleasure:
> > > There is no need to delete nodeConns of the edge node here. For the scenario, killing edgecore process and reconnect immediately, then deleting the pod of the edge node, the pod will be `Terminating`. The reason is that after the edge node reconnects and registers again, nodeConns is reloaded, meanwhile `mh.sendMsg` will retry to send and it will take about 20s, when retry timeout happend, the nodeConns of the edge node will be deleted here and we can't send message to edge node without nodeConns .
> > > So removing `mh.nodeConns.Delete(info.NodeID)` here.
> > > cc @My-pleasure
> > 
> > 
> > When there is a error which is `use of closed network connection`, we delete the nodeConn to stop send msg to this edgenode and to avoid print too many err logs. So maybe we can take out the `nodeConn` and compare with the old `conn` before deleting `nodeConn`. Like following:
> > ```
> > newConn, _ := mh.nodeConns.Load(info.NodeID)
> > if reflect.DeepEqual(newConn, conn) {
> > 	// delete
> > }
> > ```
> 
> when the node is disconnected, `nodeConn` will be deleted after keepalive timer timeout (default 30s), maybe we should add rateLimit queue here to avoid printing too many err logs, suggest not to concurrent delete the `nodeConn`.

1. It's indeed not a good idea to concurrent delete the `nodeConn`.
2. If adding a rateLimit queue for the key, the msg may not be sent in time when the network is restored.
3. The `mh.nodeConns.Delete(info.NodeID)` here is not only to avoid printing too many err logs, but also to stop meaningless sending msg when error `use of closed connection` occurs.

If we remove `mh.nodeConns.Delete(info.NodeID)` here, and then we should think about how to deal with the error `use of closed connection` occurs, right?. Can we just reply on the `keepalive` to clean up invalid connection? 
cc @fisherxu

vincentgoat:
cc @fisherxu PTAL, this issue was triggered frequently.

PR #3744, opened 2022-03-28
refactor resource manager

wackxu:
**What type of PR is this?**

kind cleanup

Fixes #

there are so many event manager for different resource but the implement
 is almost same, so refactor resource manager to reduce the duplicate code. 

```release-note

```
/assign @fisherxu


cloud/pkg/edgecontroller/manager/node.go
```diff
@@ -1,30 +0,0 @@
...
-}
-
-// Events return the channel save events from watch nodes change
-func (nm *NodesManager) Events() chan watch.Event {
-	return nm.events
-}
-
-// NewNodesManager create NodesManager by kube clientset and namespace
-func NewNodesManager(si cache.SharedIndexInformer) (*NodesManager, error) {
-	events := make(chan watch.Event)
```
gy95, line 25: Is len of this event buffer equal to 0 set by design? Or just a mistake. If the former, we should take it  consideration

cloud/pkg/edgecontroller/manager/generic_event_handler.go
```diff
@@ -8,17 +8,12 @@ import (
...
 
-// Manager define the interface of a Manager, configmapManager and podManager implement it
-type Manager interface {
-	Events() chan watch.Event
-}
-
-// CommonResourceEventHandler can be used by configmapManager and podManager
-type CommonResourceEventHandler struct {
+// genericResourceEventHandler is a generic Event Handler for kubernetes resource
+type genericResourceEventHandler struct {
```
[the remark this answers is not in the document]
wackxu: done

cloud/pkg/edgecontroller/manager/node.go
wackxu: I think it is just a mistake, for node resource we do not have some special logic.

cloud/pkg/common/eventmanager/generic_event_handler.go
```diff
@@ -0,0 +1,67 @@
...
+	if !ok {
+		klog.Warningf("unknown type: %T, ignore", obj)
+		return
+	}
+	// All obj from client has been removed the information of apiversion/kind called MetaType,
+	// it is fatal to decode the obj as unstructured.Unstructure or unstructured.UnstructureList at edge.
+	err := util.SetMetaType(eventObj)
+	if err != nil {
+		klog.Warningf("failed to set meta type :%v", err)
+	}
```
gy95, line 44: sorry, Handler in cloud/pkg/devicecontroller/manager/common.go doesn't SetMetaType, and SetMetaType function has a `TODO: support crd`, we need consider more.
https://github.com/kubeedge/kubeedge/blob/8516280fa6efd9d84c7c39be9b3973e389dabf4f/cloud/pkg/devicecontroller/manager/common.go#L19-L26

wackxu: revert the changes in the device controller and will fix the TODO in the follow pr, PTAL

wackxu:
@gy95 @fisherxu PTAL

gy95 (approved):
/lgtm

PR #3747, opened 2022-03-29
first stage: refactor message layer 

wackxu:
**What type of PR is this?**

kind cleanup

1、There are three messagelayer implement in `deviceController`、 `edgecontroller`、 `dynamiccontroller`,  all of them are the same implement except the parameter, in this pr, we move the messagelayer to the common package

2、the util.go in  `deviceController`、 `edgecontroller`、 `dynamiccontroller`  also have many repeated method implementations， we also move them to the common package

3、remove the `controllerContext`  config from cloudcore componment config, because these parameter is not configred by user and the receive module and response module is consistent and will not be changed by end user

```release-note

```


cloud/pkg/common/messagelayer/context.go
```diff
@@ -0,0 +1,97 @@
+/*
+Copyright 2022 The KubeEdge Authors.
```
fisherxu, line 2: For the changed path files, do we should remain the year which means the time to implement them?

wackxu: the context.go in old directory do no have Copyright, and I have check that there are some other file do not have Copyright, I think we should add some check for the Copyright

fisherxu (commented):
/lgtm
/approve

PR #3767, opened 2022-04-06
add sync mechanism during sync device model and device

gy95:
**What type of PR is this?**

fix TODO flag, add sync mechanism during sync device model and device
```
// Wait for adding all device model
// TODO need to think about sync
```

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"reflect"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


gy95:
> If kubeedge has run, user creates the devicemodel and device, it's also needed to create devicemodel first, then create device? cc @RyanZhaoXB Like pods use secret/cm, they should not have the order restrictions

it seems so,  there're some code validating whether devicemodel exist, like: https://github.com/kubeedge/kubeedge/blob/8516280fa6efd9d84c7c39be9b3973e389dabf4f/cloud/pkg/devicecontroller/controller/downstream.go#L228-L232
What do you think? @[RyanZhaoXB](https://github.com/RyanZhaoXB)

PR #3768, opened 2022-04-06
compile inside container by default

gy95:
**What type of PR is this?**

enable compiling inside container by default

Fixes #

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ COMPONENTS=cloud \
 .EXPORT_ALL_VARIABLES:
 OUT_DIR ?= _output/local
 
+BUILD_WITH_CONTAINER?=true
 RUN = hack/make-rules/build_with_container.sh
 
 define ALL_HELP_INFO
@@ -43,10 +44,10 @@ define ALL_HELP_INFO
 #     use the debugging tools like delve. When GOLDFLAGS is unspecified, it defaults to "-s -w" which strips
 #     debug information, see https://golang.org/cmd/link for other flags.
 #
-#     Set the environment variable or arg BUILD_WITH_CONTAINER to build inside container,
-#     only docker and make are required in this mode.
-#       1) make all WHAT=cloudcore BUILD_WITH_CONTAINER=true
-#       2) export BUILD_WITH_CONTAINER=true && make all WHAT=cloudcore
+#     By default we build inside container, only docker and make are required in this mode.
+#     Set the environment variable or arg BUILD_WITH_CONTAINER to false to build with local environment. You need to install all dependencies to make building work.
+#       1) make all WHAT=cloudcore BUILD_WITH_CONTAINER=false
+#       2) export BUILD_WITH_CONTAINER=false && make all WHAT=cloudcore
 
 endef
 .PHONY: all
--- a/hack/make-rules/build_with_container.sh
+++ b/hack/make-rules/build_with_container.sh
@@ -23,7 +23,13 @@ set -o pipefail
 KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 MOUNTPATH="${MOUNTPATH:-/kubeedge}"
 KUBEEDGE_BUILD_IMAGE=${KUBEEDGE_BUILD_IMAGE:-"kubeedge/build-tools"}
+DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}"
 
 set -x
-docker run --rm -v ${KUBEEDGE_ROOT}:${MOUNTPATH} -w ${MOUNTPATH} ${KUBEEDGE_BUILD_IMAGE} "$@"
+docker run --rm -u "${UID}:${DOCKER_GID}" \
+    --init \
+    --sig-proxy=false \
+    -e XDG_CACHE_HOME=/tmp/.cache \
+    -v ${KUBEEDGE_ROOT}:${MOUNTPATH} \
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -43,10 +44,10 @@ define ALL_HELP_INFO
 #     use the debugging tools like delve. When GOLDFLAGS is unspecified, it defaults to "-s -w" which strips
 #     debug information, see https://golang.org/cmd/link for other flags.
 #
-#     Set the environment variable or arg BUILD_WITH_CONTAINER to build inside container,
-#     only docker and make are required in this mode.
-#       1) make all WHAT=cloudcore BUILD_WITH_CONTAINER=true
-#       2) export BUILD_WITH_CONTAINER=true && make all WHAT=cloudcore
+#     By default we build inside container, only docker and make are required in this mode.
+#     Set the environment variable or arg BUILD_WITH_CONTAINER to anything but true to build using local golang environtment.
```
fisherxu, line 48:
```suggestion
#     Set the environment variable or arg BUILD_WITH_CONTAINER to false to build with local environtment. You need to install all dependencies to make building work.
```
```suggestion
#     Set the environment variable or arg BUILD_WITH_CONTAINER to anything but true to build using local golang environtment.
```

hack/lib/golang.sh
```diff
@@ -342,6 +342,7 @@ kubeedge::golang::small_build_place_binaries() {
   read -r ldflags <<< "$(kubeedge::version::ldflags)"
 
   mkdir -p ${KUBEEDGE_OUTPUT_BINPATH}
+  sudo chmod -R 777 ${KUBEEDGE_ROOT}/_output
```
fisherxu, line 345: Looks like this is hard code for path, and do we need to set to `777`?

fisherxu, line 345: Looks like this is hard code for path, and do we need to set to `777`?

hack/local-up-kubeedge.sh
```diff
@@ -97,6 +97,9 @@ function build_edgecore {
 function start_cloudcore {
   CLOUD_CONFIGFILE=${KUBEEDGE_ROOT}/_output/local/bin/cloudcore.yaml
   CLOUD_BIN=${KUBEEDGE_ROOT}/_output/local/bin/cloudcore
+
+  sudo chmod -R 777 ${KUBEEDGE_ROOT}/_output
```
fisherxu, line 101: ditto

fisherxu, line 101: ditto

hack/lib/golang.sh
gy95: yes, CI runs checks with username `runner`, and when building inside the container just with the below command
https://github.com/kubeedge/kubeedge/blob/697b8be4139ad86e08ad4c457210acfbc21e79f9/hack/make-rules/build_with_container.sh#L27-L29
the -v flag bind mount a volume, will assign the _output/local/bin directory to the root permission, and the user `runner` cannot access these directory.

gy95: not know how to assign the _output/local/bin directory to the user `runner` permission, by test add flag `--user  $(id -u)` to this docker run command, it didn't work well.

fisherxu: Not sure if it will work for adding `sudo` for `make smallbuild`?

hack/make-rules/build_with_container.sh
```diff
@@ -23,7 +23,13 @@ set -o pipefail
...
+DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}"
 
 set -x
-docker run --rm -v ${KUBEEDGE_ROOT}:${MOUNTPATH} -w ${MOUNTPATH} ${KUBEEDGE_BUILD_IMAGE} "$@"
+docker run --rm -u "${UID}:${DOCKER_GID}" \
+    --init \
+    --sig-proxy=false \
+    -e XDG_CACHE_HOME=/tmp/.cache \
+    -v ${KUBEEDGE_ROOT}:${MOUNTPATH} \
+    -w ${MOUNTPATH} ${KUBEEDGE_BUILD_IMAGE} "$@"
```
gy95, line 34: -u: specify the current user to process the _output/local/bin directory permissions, or processes in containers will run as default user `root` and the directory will be assigned to the default root permission. FYI: https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf
--init --sig-proxy=false:  process signals when pressing `ctrl+C` command
-e XDG_CACHE_HOME=/tmp/.cache: fix the error: `failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied`

fisherxu (commented):
/lgtm
/approve

hack/make-rules/build_with_container.sh
```diff
@@ -23,7 +23,13 @@ set -o pipefail
 KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 MOUNTPATH="${MOUNTPATH:-/kubeedge}"
 KUBEEDGE_BUILD_IMAGE=${KUBEEDGE_BUILD_IMAGE:-"kubeedge/build-tools"}
+DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}"
 
 set -x
-docker run --rm -v ${KUBEEDGE_ROOT}:${MOUNTPATH} -w ${MOUNTPATH} ${KUBEEDGE_BUILD_IMAGE} "$@"
+docker run --rm -u "${UID}:${DOCKER_GID}" \
+    --init \
+    --sig-proxy=false \
```
fisherxu, line 31: Should be true here?

PR #3770, opened 2022-04-06
fix actualValue meta not set

TianTianBigWang:

**What type of PR is this?**
/kind bug

Fixes #

```release-note

```


edge/pkg/devicetwin/dttype/types_helper.go
```diff
@@ -81,6 +80,7 @@ func DeviceTwinToMsgTwin(deviceTwins []dtclient.DeviceTwin) map[string]*MsgTwin
 			actualValue := &TwinValue{Value: &actual}
 			if twin.ActualMeta != "" {
 				json.Unmarshal([]byte(twin.ActualMeta), &actualMeta)
+				actualValue.Metadata = &actualMeta
```
fisherxu, line 83: Looks good, any bugs will caused? Can we add some tests? :)

TianTianBigWang: Seems to be no problem in our environment

PR #3779, opened 2022-04-07
add msg resource err log

TianTianBigWang:

**What type of PR is this?**
/kind feature

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -184,7 +184,7 @@ func (uc *UpstreamController) dispatchMessage() {
 
 		resourceType, err := messagelayer.GetResourceType(msg)
 		if err != nil {
-			klog.Warningf("parse message: %s resource type with error: %s", msg.GetID(), err)
+			klog.Warningf("parse message: %s resource type with error, message resource: %s, err: %v", msg.GetID(), msg.GetResource(), err)
 			continue
 		}
 


cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -184,7 +184,7 @@ func (uc *UpstreamController) dispatchMessage() {
 
 		resourceType, err := messagelayer.GetResourceType(msg)
 		if err != nil {
-			klog.Warningf("parse message: %s resource type with error: %s", msg.GetID(), err)
+			klog.Warningf("parse message: %s resource type with error, message resource: %s, err: %s", msg.GetID(), msg.GetResource(), err)
```
wackxu, line 187: use %v for err is better

wackxu, line 187: use %v for err is better

TianTianBigWang: It looks like it doesn't need to because only `fmt error` are returned.

https://github.com/kubeedge/kubeedge/blob/9e7791f7ae567d73dabc94a9d4a7f1183dd8e120/cloud/pkg/common/messagelayer/util.go#L90-L98

fisherxu: If use `%s`, err.Error() is better, or we can use %v :)

TianTianBigWang: ok, `%v` is use.

gy95 (approved):
/lgtm

fisherxu (commented):
/approve

PR #3780, opened 2022-04-07
Remove gopath in makefile

fisherxu:

**What type of PR is this?**
/kind test

When building with container, the stdout will still print the log:
```
hack/make-rules/build_with_container.sh hack/make-rules/build.sh cloudcore
make: go: Command not found
+ docker run --rm -u 0:998 --init --sig-proxy=false -e XDG_CACHE_HOME=/tmp/.cache -v /root/codes/src/github.com/kubeedge/kubeedge:/kubeedge -w /kubeedge kubeedge/build-tools hack/make-rules/build.sh cloudcore
```

`make: go: Command not found` should not be printed. GOPATH is not used now.

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@ DESTDIR?=
 USR_DIR?=/usr/local
 INSTALL_DIR?=${DESTDIR}${USR_DIR}
 INSTALL_BIN_DIR?=${INSTALL_DIR}/bin
-GOPATH?=$(shell go env GOPATH)
 
 # make all builds both cloud and edge binaries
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,7 +2,6 @@ DESTDIR?=
 USR_DIR?=/usr/local
 INSTALL_DIR?=${DESTDIR}${USR_DIR}
 INSTALL_BIN_DIR?=${INSTALL_DIR}/bin
-GOPATH?=$(shell go env GOPATH)
```
gy95, line 5: There're also some other blocks using this environment, we can find more GOPATH related docs or codes by searching `$GOPATH` or `GOPATH` in kubeedge files.

fisherxu: OK, just have a double check, for shell script, GOPATH is a env value, it will exists once golang installed. Let's keep the env check, and just remove in Makefile.

PR #3788, opened 2022-04-12
support make release inside container

gy95:
**What type of PR is this?**

support make release inside container

Fixes #

```release-note

```

### DIFF
--- a/hack/make-rules/release.sh
+++ b/hack/make-rules/release.sh
@@ -33,11 +33,14 @@ function release() {
   local -a targets=()
   local VERSION=""
   local ARCH="amd64"
+  local arm_version=""
 
   for arg in "$@"; do
     if [[ "${arg}" == GOARM7 ]]; then
       ARCH="arm"
+      arm_version="GOARM7"
     elif [[ "${arg}" == GOARM8 ]]; then
+      arm_version="GOARM8"
       ARCH="arm64"
     else
       targets+=("${arg}")
@@ -57,7 +60,7 @@ function release() {
         if [ "${ARCH}" == "amd64" ]; then
           hack/make-rules/build.sh edgesite-server edgesite-agent
         else
-          hack/make-rules/crossbuild.sh edgesite-server edgesite-agent ${ARM_VERSION}
+          hack/make-rules/crossbuild.sh edgesite-server edgesite-agent ${arm_version}
         fi
 
         build_edgesite_release $VERSION $ARCH
@@ -66,7 +69,7 @@ function release() {
         if [ "${ARCH}" == "amd64" ]; then
           hack/make-rules/build.sh keadm
         else
-          hack/make-rules/crossbuild.sh keadm ${ARM_VERSION}
+          hack/make-rules/crossbuild.sh keadm ${arm_version}
         fi
 
         build_keadm_release $VERSION $ARCH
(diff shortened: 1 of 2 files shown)


hack/make-rules/build_with_container.sh
```diff
@@ -24,11 +24,14 @@ KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 MOUNTPATH="${MOUNTPATH:-/kubeedge}"
 KUBEEDGE_BUILD_IMAGE=${KUBEEDGE_BUILD_IMAGE:-"kubeedge/build-tools"}
 DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}"
+# please edit ENV_LIST to support more env variables
+ENV_LIST="${ENV_LIST:-ARM_VERSION}"
```
fisherxu, line 28:
```suggestion
ENV_LIST=${ENV_LIST:-"ARM_VERSION"}
```
This will be more readable?

hack/make-rules/release.sh
```diff
@@ -57,7 +60,7 @@ function release() {
         if [ "${ARCH}" == "amd64" ]; then
           hack/make-rules/build.sh edgesite-server edgesite-agent
         else
-          hack/make-rules/crossbuild.sh edgesite-server edgesite-agent ${ARM_VERSION}
```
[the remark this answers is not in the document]
fisherxu: The WHAT can be passed into container？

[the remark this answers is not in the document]
gy95:
```
# make release WHAT=keadm ARM_VERSION=GOARM7
hack/make-rules/build_with_container.sh hack/make-rules/release.sh keadm GOARM7
go detail version: go version go1.17 linux/amd64
go version: 1.17
cross building github.com/kubeedge/kubeedge/keadm/cmd/keadm GOARM7
...........
```
The WHAT and ARM_VERSION can be passed into container as format of value `keadm` or `GOARM7`directly, but not as the env format `WHAT=keadm` or `ARM_VERSION=GOARM7`

hack/make-rules/build_with_container.sh
gy95: if we want to pass in two environment or more, the before format can specify the needed double quotation, just like `env | grep "ARM_VERSION\|ANOTHER_ENV"` but not `env | grep ARM_VERSION\|ANOTHER_ENV` without quotations, the latter can not get right env.

fisherxu: Just have a check, seems have no difference?
```
➜  ~ ENV_LIST="${ENV_LIST:-GVM_ROOT\|GVM_VERSION}"
➜  ~ env | grep ${ENV_LIST}                       
GVM_ROOT=/root/.gvm
GVM_VERSION=1.0.22
➜  ~ ENV_LIST=${ENV_LIST:-"GVM_ROOT\|GVM_VERSION"}
➜  ~ env | grep ${ENV_LIST}                       
GVM_ROOT=/root/.gvm
GVM_VERSION=1.0.22
```
I think here ARM_VERSION is used as a string instead of env value, so with quotation seems more easy to understand :)

gy95: yes, we must ensure the value is quoted string, but when users set ENV_LIST by themselves, they may forget the quotation just like `export ENV_LIST=ARM_VERSION\|TEST`, so make release may fail. the former adds quotation at the outermost layer, so it can ensure the ENV_LIST environment value is a quoted string.

fisherxu: OK, and seems if you use `ARCH` in https://github.com/kubeedge/kubeedge/blob/master/hack/make-rules/release.sh#L60, the env value `ARM_VERSION` will not needed.

gy95: fixed

PR #3790, opened 2022-04-12
Fixed wrong format placeholder and nil warning on check

jpanda-cn:
**What type of PR is this?**

Fixed wrong format placeholder and nil warning on check

1. If this is your first time, please read our contributor guidelines: https://github.com/kubeedge/kubeedge/blob/master/CONTRIBUTING.md
2. Ensure you have added or ran the appropriate tests for your PR
3. Ignore the return value of the Close method

/kind test

### DIFF
--- a/edge/pkg/edgehub/common/http/http_test.go
+++ b/edge/pkg/edgehub/common/http/http_test.go
@@ -50,12 +50,12 @@ func TestNewHttpClient(t *testing.T) {
 func TestNewHTTPSClient(t *testing.T) {
 	err := util.GenerateTestCertificate(Path, BaseName, BaseName)
 	if err != nil {
-		t.Errorf("Error in generating fake certificates: %w", err)
+		t.Errorf("Error in generating fake certificates: %v", err)
 		return
 	}
 	certificate, err := tls.LoadX509KeyPair(CertFile, KeyFile)
 	if err != nil {
-		t.Errorf("Error in loading key pair: %w", err)
+		t.Errorf("Error in loading key pair: %v", err)
 		return
 	}
 	type args struct {
@@ -117,12 +117,12 @@ func TestNewHTTPSClient(t *testing.T) {
 func TestNewHTTPClientWithCA(t *testing.T) {
 	err := util.GenerateTestCertificate(Path, BaseName, BaseName)
 	if err != nil {
-		t.Errorf("Error in generating fake certificates: %w", err)
+		t.Errorf("Error in generating fake certificates: %v", err)
 		return
 	}
 	capem, err := os.ReadFile(CertFile)
 	if err != nil {
-		t.Errorf("Error in loading Cert file: %w", err)
+		t.Errorf("Error in loading Cert file: %v", err)
 		return
 	}
 	certificate := tls.Certificate{}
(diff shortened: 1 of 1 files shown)


edge/pkg/edgehub/common/http/http_test.go
```diff
@@ -258,6 +258,10 @@ func TestBuildRequest(t *testing.T) {
 				t.Errorf("BuildRequest() error = %v, expectedError = %v", err, tt.wantErr)
 				return
 			}
+			if got == nil {
+				// will never execute, but can explicitly prevent nil warnings when checking
+				return
+			}
```
TianTianBigWang, line 264: seem that `BuildRequest` method will not return nil `got` when `err != nil`

jpanda-cn: Yes, I looked at the code for the BuildRequest and was sure that the added code was never executed, but doing so explicitly organizes editing tools and lint's nil warnings.

I can remove this part of the code if needed

fisherxu (commented):
%w seems have no differece with %v? cc @gy95

gy95:
> %w seems have no differece with %v? cc @gy95

generally, %w verb is a synonym for %v.

TianTianBigWang:
> > %w seems have no differece with %v? cc @gy95
> 
> yes, generally, %w verb is a synonym for %v.

i got the tip `This verb can be used only in 'fmt.Errorf' calls (%w) ` in the code.
is it related?

gy95:
> 

yes, that's it :)

edge/pkg/edgehub/common/http/http_test.go
zc2638: In theory, if it is not for special circumstances, according to the specification, when `err == nil`, the returned value must be a valid value.

jpanda-cn: Okay, I've removed that part of the code

gy95:
thank you @jpanda-cn  and please squash your commits to one, and sign off your commit to pass the DCO :)

fisherxu:
> 

Thanks for clarification, and I'm wondering if it will be bug or print error if we use %w in t.Errorf? cc @TianTianBigWang @gy95

gy95:
```
root@ecs-8f95:~/go/src# cat main.go 
package main

import(
	"fmt"
)
func main() {
	err := fmt.Errorf("failed to find key")
	fmt.Printf("%v\n", err)
	fmt.Printf("%w\n", err)
}

root@ecs-8f95:~/go/src# go run main.go 
failed to find key
%!w(*errors.errorString=&{failed to find key})
```
yes, when using %w in print format, the message is of low readability, so does `testing.Errorf`

fisherxu:
> ```
> root@ecs-8f95:~/go/src# cat main.go 
> package main
> 
> import(
> 	"fmt"
> )
> func main() {
> 	err := fmt.Errorf("failed to find key")
> 	fmt.Printf("%v\n", err)
> 	fmt.Printf("%w\n", err)
> }
> 
> root@ecs-8f95:~/go/src# go run main.go 
> failed to find key
> %!w(*errors.errorString=&{failed to find key})
> ```
> 
> yes, when using %w in print format, the message is of low readability, so does `testing.Errorf`

Ok, we can check whether other logs should not use `%w` :)
/lgtm
/approve

PR #3792, opened 2022-04-13
fix endpointslices to kind EndpointSlice

vincentgoat:

**What type of PR is this?**
/kind bug

image
It is not the correct type in meta db when we list&watch endpointslices resource.

Fixes #

```release-note

```


fisherxu (commented):
/approve

PR #3808, opened 2022-04-21
add domain support for certgen.sh

lwabish:
**What type of PR is this?**
/kind bug

the original script failed when providing a domain name.

Fixes #

```release-note

```

### DIFF
--- a/build/tools/certgen.sh
+++ b/build/tools/certgen.sh
@@ -71,9 +71,10 @@ stream() {
     readonly K8SCA_FILE=${K8SCA_FILE:-/etc/kubernetes/pki/ca.crt}
     readonly K8SCA_KEY_FILE=${K8SCA_KEY_FILE:-/etc/kubernetes/pki/ca.key}
 
-    if [ -z ${CLOUDCOREIPS} ]; then
-        echo "You must set CLOUDCOREIPS Env,The environment variable is set to specify the IP addresses of all cloudcore"
-        echo "If there are more than one IP need to be separated with space."
+    if [ -z "${CLOUDCOREIPS}" ] && [ -z "${CLOUDCORE_DOMAINS}" ]; then
+        echo "You must set at least one of CLOUDCOREIPS or CLOUDCORE_DOMAINS Env.These environment
+variables are set to specify the IP addresses or domains of all cloudcore, respectively."
+        echo "If there are more than one IP or domain, you need to separate them with a space within a single env."
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


fisherxu:
@lwabish Why do you build the certs manually instead of using the certs generted by cloudcore?

fisherxu (commented):
@lwabish Why do you build the certs manually instead of using the certs generted by cloudcore?

lwabish:
> @lwabish Why do you build the certs manually instead of using the certs generted by cloudcore?

I was following the [Enable kubectl logs Feature](https://kubeedge.io/en/docs/setup/keadm/#enable-kubectl-logs-feature)steps after install cloudcore and init my edge node. @fisherxu

fisherxu:
Ok, it should only use the stream func.

lwabish:
> Ok, it should only use the stream func.

yes,indeed. So should I leave the genCert() funciton as it was?

lwabish:
> 

oh I see.
In this case, I deployed cloudcore in k8s,so I need to use cloudcore.kubeedge to generate the stream cert so that apiserver can locate cloudcore, is this right?

ramezanius (approved):

fisherxu:
> In this case, I deployed cloudcore in k8s,so I need to use cloudcore.kubeedge to generate the stream cert so that apiserver can locate cloudcore, is this right?

Yes, if the cloudocre domain used for edgeside, the domain name maybe a public domain. If apiserver connects to the cloudcore, it may not have to use that public domin, domain name incluster should be also fine, like `svc.ns.cluster.local`.

Anyway, set the domain name in stream's certs is reasonable, and can we set another env value like `CLOUDCORE_DOMAINS`? And the docs also need to be updated :) @lwabish

fisherxu (commented):
Looks good, and could you please squash the 4 commits? Thanks.

lwabish:
previous commits squashed. 
is the "fetch upstream commit" necessary to be squashed too?

fisherxu:
> previous commits squashed. is the "fetch upstream commit" necessary to be squashed too?

Yes, fetch upstream commit is no needed here :)

lwabish:
> > previous commits squashed. is the "fetch upstream commit" necessary to be squashed too?
> 
> Yes, fetch upstream commit is no needed here :)

done

fisherxu:
Thanks, and would you like to update the docs? Ref: https://github.com/kubeedge/website/blob/master/content/en/docs/setup/keadm.md#enable-kubectl-logs-feature

fisherxu (commented):
/lgtm
/approve

lwabish:
> 

sure, i'll work on it later.
I also found that the stream certs of cloudcore deployed using helm was generated within the helm chart,  running `certgen.sh stream` is not needed in this case.  So maybe I could also try improving the doc about stream certs.

lwabish:
> Thanks, and would you like to update the docs? Ref: https://github.com/kubeedge/website/blob/master/content/en/docs/setup/keadm.md#enable-kubectl-logs-feature

sure, i'll work on it later.
I also found that the stream certs of cloudcore deployed using helm was generated within the helm chart, running certgen.sh stream is not needed in this case. So maybe I could also try improving the doc about stream certs.

PR #3814, opened 2022-04-26
fix unbound variable GOPATH

gy95:
**What type of PR is this?**
/kind cleanup

When running `make verify` cmd, there're two errors:
```
hack/generate-groups.sh: line 44: GOPATH: unbound variable
hack/generate-crds.sh: line 60: GOPATH: unbound variable
```
if GOPATH is not set, use `go env GOPATH` to set it

Fixes #

```release-note

```

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -20,6 +20,8 @@ set -o pipefail
 
 KUBEEDGE_ROOT=$(unset CDPATH && cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)
 
+export GOPATH="${GOPATH:-$(go env GOPATH)}"
+
 ${KUBEEDGE_ROOT}/hack/generate-groups.sh "deepcopy,client,informer,lister" \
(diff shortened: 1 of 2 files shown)


hack/update-codegen.sh
```diff
@@ -20,6 +20,8 @@ set -o pipefail
 
 KUBEEDGE_ROOT=$(unset CDPATH && cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)
 
+export GOPATH="${GOPATH:-$(go env GOPATH)}"
+
```
gy95, line 24: deepcopy-gen tool and others in https://github.com/kubeedge/kubeedge/blob/master/hack/generate-groups.sh need GOPATH env variable, by official doc
`--output-base string                   Output base; defaults to $GOPATH/src/ or ./ if $GOPATH is not set. (default "./")`
So now we must put the kubeedge codes under $GOPATH/src/github.com/kubeedge/kubeedge dir, or the generated codes will be output to another dir.

fisherxu (commented):
/lgtm
/approve

PR #3821, opened 2022-04-28
bugfix: rule.Status variable update

cl2017:
Signed-off-by: chenlin.liu <chenlin.liu@daocloud.io>

**What type of PR is this?**
/kind bug

bugfix: rule.Status variable update

Fixes #

```release-note
NONE
```


fisherxu (commented):
Thanks for the fixing @cl2017 And we also have some e2e tests for the rules, could you please have a look and try add test case for this?

cl2017:
> Thanks for the fixing @cl2017 And we also have some e2e tests for the rules, could you please have a look and try add test case for this?

ok, I'll take a look and add test case, but I'm not familiar with e2e tests yet, so I need to spend some time learning it. Can I submit another PR later?

fisherxu (commented):
/lgtm
/approve

PR #3822, opened 2022-04-29
add edge node upgrade proposal

gy95:
**What type of PR is this?**
/kind documentation

Fixes #

```release-note

```


TianTianBigWang:
Is there some error case consideration for node upgrades? 

For example, what happens when the power is cut off and restarted, and what happens when the upgrade is halfway through power off and restart

docs/proposals/edge-node-upgrade.md
```diff
@@ -0,0 +1,273 @@
...
+These are not separate controllers as such but named here for clarity.
+The job of the downstream controller is to synchronize the Upgrade updates from the cloud to the edge node.
+The job of the upstream controller is the reverse.
+
+
+## Synchronizing Upgrade
+
+The below illustrations describe the flow of events that would occur when Upgrade property values are updated from the cloud/edge.
+
+- Users create/modify the Upgrade CR directly using `kubectl`.
```
wackxu, line 49: I want to know what field would be allowed to update  after we create the CR?

wackxu: And if the CR already completed, can we edit the cr?

wackxu: The CR looks like a `job` , we should consider all above

docs/proposals/edge-node-upgrade.md
```diff
@@ -0,0 +1,273 @@
...
+- Upgrade Controller Downstream receive events from K8s APIServer, and then store it in local cache
+- Upgrade Controller Downstream checks whether the upgrade is completed(ReUpgrade is false and all or partial node upgrade is completed),
+  if so, we don't need to send upgrade message to edge node
+- Upgrade Controller Downstream call K8s api to get node list using NodeNames or LabelSelector specified in CR, and
```
wackxu, line 53: why not user informer to get the node list?

docs/proposals/edge-node-upgrade.md
```diff
@@ -0,0 +1,273 @@
...
+  filter out nodes that don't meet upgrade requirements(1. edge node already on the desired version. 2. not edge node,
```
wackxu, line 54: if the edge node is in upgrade state, we should not send the message again

docs/proposals/edge-node-upgrade.md
```diff
@@ -0,0 +1,273 @@
...
+  without label {"node-role.kubernetes.io/edge": ""} 3. remove duplicated nodes)
+- Upgrade Controller Downstream send Upgrade message to each edge node. And new a Goroutine to process timeout
+  if we don't receive edge node upgrade response, if not, downstream will construct a timeout response message and send it to
+  Upstream to let Upstream update Upgrade status. And patch update Upgrade status state to Upgrading state.
```
wackxu, line 58: This sentence is a bit awkward

docs/proposals/edge-node-upgrade.md
```diff
@@ -0,0 +1,273 @@
...
+  And mark edge node unschedulable. If the ReUpgrade flag is true, we will patch update it from true to false, to avoid always reupgrading.
```
wackxu, line 59: waht does ReUpgrade mean?  It means that when the edge node upgrade failed, it can be retried?

docs/proposals/edge-node-upgrade.md
```diff
@@ -0,0 +1,273 @@
...
+- CloudHub send upgrade request to EdgeHub
+- EdgeHub receive upgrade message, and upgrade sub module will process upgrade message.
+- EdgeHub Upgrade module validates the upgrade message(checks UpgradeID is null, edgecore is already on the version)
```
wackxu, line 62: what does this mean? `edgecore is already on the version`

docs/proposals/edge-node-upgrade.md
```diff
@@ -0,0 +1,273 @@
...
+- Upgrade Controller Downstream checks whether the upgrade is completed(ReUpgrade is false and all or partial node upgrade is completed),
+  if so, we don't need to send upgrade message to edge node
+- Upgrade Controller Downstream call K8s api to get node list using NodeNames or LabelSelector specified in CR, and
+  filter out nodes that don't meet upgrade requirements(1. edge node already on the desired version. 2. not edge node,
+  without label {"node-role.kubernetes.io/edge": ""} 3. remove duplicated nodes)
+- Upgrade Controller Downstream send Upgrade message to each edge node. And new a Goroutine to process timeout
+  if we don't receive edge node upgrade response, if not, downstream will construct a timeout response message and send it to
+  Upstream to let Upstream update Upgrade status. And patch update Upgrade status state to Upgrading state.
+  And mark edge node unschedulable. If the ReUpgrade flag is true, we will patch update it from true to false, to avoid always reupgrading.
+- CloudHub send upgrade request to EdgeHub
```
wackxu, line 60: the upgrade message will need ack to make sure that we already send the message successfully

docs/proposals/edge-node-upgrade.md
```diff
@@ -0,0 +1,273 @@
...
+### CRD API Group and Version
+The `Upgrade` CRD will be cluster-scoped.
+The tables below summarize the group, kind and API version details for the CRDs.
+
+* Upgrade
+
+| Field                 | Description             |
+|-----------------------|-------------------------|
+|Group                  | upgrade.kubeedge.io     |
+|APIVersion             | v1alpha2                |
```
wackxu, line 103: why the init version is not v1alpha1?

gy95: already fixed to v1alpha1 in source codes, will also update the proposal

docs/proposals/edge-node-upgrade.md
gy95: filter the case that edgecore is already on the desired upgrade version, we don't need to run `keadm upgrade` cmd to do upgrade operation

docs/proposals/edge-node-upgrade.md
gy95: yes, that's one case. And if we want to use the existing Upgrade CR to upgrade node to another version, we can set ReUpgrade to true and modify the version, so that will trigger another upgrade operation

docs/proposals/edge-node-upgrade.md
gy95: will fix.

docs/proposals/edge-node-upgrade.md
gy95: add Upgrade related webhook to do these restriction

PR #3826, opened 2022-05-06
fix Remove container failed

gy95:
**What type of PR is this?**

If we run `docker ps -a | grep installation-package `, there're lots of containers that not removed.
and error log is 
```
"Remove container failed" err="Error response from daemon: You cannot remove a running container 20b7a40dcc023ec7add24b2fed306d4e0643e1d2bd7087130506f0bf0e41d8ac. 
Stop the container before attempting removal or force remove" containerID="20b7a40dcc023ec7add24b2fed306d4e0643e1d2bd7087130506f0bf0e41d8ac"
```
So we need to wait the container stop, and then we can remove it

Fixes #

```release-note

```


vincentgoat (commented):
lgtm

fisherxu (commented):
/approve
/lgtm
We need to ensure the container exited before removing it.  Let's pick it to 1.10 and 1.9 branch @gy95

PR #3834, opened 2022-05-09
fix HostAliases not working if pod network is set to hostNetwork

wackxu:
**What type of PR is this?**

Fixes  https://github.com/kubeedge/kubeedge/issues/3812

```
apiVersion: v1
kind: Pod
metadata:
  name: hostaliases-pod
spec:
  hostNetwork: true
  restartPolicy: Never
  hostAliases:
  - ip: "127.0.0.1"
    hostnames:
    - "foo.local"
    - "bar.local"
  - ip: "10.1.2.3"
    hostnames:
    - "foo.remote"
    - "bar.remote"
  containers:
  - name: cat-hosts
    image: busybox:1.28
    command:
    - sleep
    args:
    - "99999999"

```

image

```release-note

```


vincentgoat (commented):
LGTM

PR #3836, opened 2022-05-09
remove default KubeConfig for cloudcore

wackxu:
**What type of PR is this?**

Cloudcore uses containerized deployment for now and use the in cluster kubeconfig in pod that was generated by serviceaccount token by default, so we  do not need set default value for the kubeconfig. otherwise  we need set empty value explicitly for the kubeconfig when we uses containerized deployment

image

```release-note

```

### DIFF
--- a/pkg/apis/componentconfig/cloudcore/v1alpha1/default.go
+++ b/pkg/apis/componentconfig/cloudcore/v1alpha1/default.go
@@ -38,11 +38,9 @@ func NewDefaultCloudCoreConfig() *CloudCoreConfig {
 			TunnelPort: constants.ServerPort,
 		},
 		KubeAPIConfig: &KubeAPIConfig{
-			Master:      "",
 			ContentType: constants.DefaultKubeContentType,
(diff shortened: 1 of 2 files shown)


pkg/apis/componentconfig/cloudcore/v1alpha1/default.go
```diff
@@ -38,11 +38,9 @@ func NewDefaultCloudCoreConfig() *CloudCoreConfig {
 			TunnelPort: constants.ServerPort,
 		},
 		KubeAPIConfig: &KubeAPIConfig{
-			Master:      "",
 			ContentType: constants.DefaultKubeContentType,
 			QPS:         constants.DefaultKubeQPS,
 			Burst:       constants.DefaultKubeBurst,
-			KubeConfig:  constants.DefaultKubeConfig,
```
fisherxu, line 45: Master and KubeConfig below L174 should also be removed

fisherxu (commented):
/lgtm
/approve

PR #3845, opened 2022-05-13
dynamiccontroller: fix selector match return false when selector is e…

neiba:
…mpty

Signed-off-by: chengengkun <1366542563@qq.com>

**What type of PR is this?**

/kind bug

If we curl http://127.0.0.1:10550/api/v1/namespaces/default/pods?watch=true, we want to watch all pod in default namespace, but no pod was watched. Because dynamiccontroller listener selector's labels is empty, and can not match any pod.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/selector.go
+++ b/cloud/pkg/dynamiccontroller/application/selector.go
@@ -45,7 +45,7 @@ func (lf *LabelFieldSelector) String() string {
 }
 
 func (lf *LabelFieldSelector) Match(set labels.Set, set2 fields.Set) bool {
-	return lf.Labels().Matches(set) && lf.Fields().Matches(set2)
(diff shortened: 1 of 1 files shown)


cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -385,7 +385,7 @@ func (mh *MessageHandle) ListMessageWriteLoop(info *model.HubInfo, stopServe cha
 
 		obj, exist, _ := nodeListStore.GetByKey(key.(string))
 		if !exist {
-			klog.Errorf("nodeListStore for node %s doesn't exist", info.NodeID)
+			klog.Errorf("message with key %s doesn't exist", key.(string))
```
fisherxu, line 388: object with key %s doesn't exist?

neiba: nodeListStore store message waiting to send to edge node.

neiba:
@wackxu please have a review

PR #3859, opened 2022-05-21
refactor function: copy files from image to host

### LINKED ISSUE #3828: Refactor keadm functions about copying files from container to host
Now `keadm beta join` command pull images and copy edgecore binary from it, and support both `docker` and `remote` ContainerRuntime. But both two  `CopyResources` seems confusing, now it only support copy from container to the same path of the host, such as copy /usr/local/bin/edgecore to  /usr/local/bin/edgecore , the src and dest path must be the same.
the input argument is like 
```
files := map[string]string{
    "src_path": "temp_bind_path",
}
```
key is container path, but value is a temp path to be binded, and look into the codes, it only copy the file to the hostPath(just is the same to the container path)
https://github.com/kubeedge/kubeedge/blob/4be67e0cc8db71324d47b1eaf5f976f23fb24572/keadm/cmd/keadm/app/cmd/util/image.go#L150
https://github.com/kubeedge/kubeedge/blob/4be67e0cc8db71324d47b1eaf5f976f23fb24572/keadm/cmd/keadm/app/cmd/util/image.go#L217

To make it more readable, we should refactor it and make it input parameter like 
```
files := map[string]string{
    "src_path": "dest_path",
}
```
It seems that more meet the common logic, that we want to copy src_path file in container to the dest_path in host

/help
**Why is this needed**:

gy95:
**What type of PR is this?**

Fixes #3828 

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/image.go
+++ b/keadm/cmd/keadm/app/cmd/edge/image.go
@@ -41,13 +41,10 @@ func request(opt *common.JoinOptions, step *common.Step) error {
 	}
 
 	step.Printf("Copy resources from the image to the management directory")
-	dirs := map[string]string{
-		util.KubeEdgePath: filepath.Join(util.KubeEdgeTmpPath, "data"),
-	}
 	files := map[string]string{
-		filepath.Join(util.KubeEdgeUsrBinPath, util.KubeEdgeBinaryName): filepath.Join(util.KubeEdgeTmpPath, "bin", util.KubeEdgeBinaryName),
+		filepath.Join(util.KubeEdgeUsrBinPath, util.KubeEdgeBinaryName): filepath.Join(util.KubeEdgeUsrBinPath, util.KubeEdgeBinaryName),
 	}
-	if err := runtime.CopyResources(imageSet.Get(image.EdgeCore), dirs, files); err != nil {
+	if err := runtime.CopyResources(imageSet.Get(image.EdgeCore), files); err != nil {
 		return fmt.Errorf("copy resources failed: %v", err)
 	}
 
--- a/keadm/cmd/keadm/app/cmd/util/image.go
+++ b/keadm/cmd/keadm/app/cmd/util/image.go
@@ -39,7 +39,7 @@ import (
 
 type ContainerRuntime interface {
 	PullImages(images []string) error
-	CopyResources(edgeImage string, dirs map[string]string, files map[string]string) error
+	CopyResources(edgeImage string, files map[string]string) error
 	RunMQTT(mqttImage string) error
 }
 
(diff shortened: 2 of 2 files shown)


gy95:
now we pass in the container path `ContainerPath` and host path `HostPath`, and this function will bind host path `HostPath` to container path `"/tmp/${HOSTPATH}"`, and then container will exec cp file from `ContainerPath` to bindContainerPath "/tmp/${HOSTPATH}"
cc @zc2638

zc2638:
This PR looks good, maybe we can merge files and dirs as well?

gy95:
> This PR looks great, maybe we can merge files and dirs as well?

because now we haven't used dirs, we can remove dir parameter
But if we merge dirs and files, it's our duty to check the passed in arguments whether it is dir or it is file, to generate copy cmd(cp or cp -r)?
So better distinguish it?

gy95:
now we start a container and run `cp` cmd to accomplish copy files from container, 
maybe we can ref what `docker cp` does to accomplish this function: https://github.com/docker/cli/blob/14962747e4af2c992bca55008c27387cd2268620/cli/command/container/cp.go#L74
But not know how to copy files from container to host in containerd like what docker did

zc2638:
> 现在我们启动一个容器并运行`cp`cmd 来完成从容器中复制文件， 也许我们可以参考什么`docker cp`来完成这个功能：[https://github.com/docker/cli/blob/14962747e4af2c992bca55008c27387cd2268620/cli/command/container/cp。 go#L74](https://github.com/docker/cli/blob/14962747e4af2c992bca55008c27387cd2268620/cli/command/container/cp.go#L74) 但不知道如何将文件从容器复制到容器中的主机，就像 docker 所做的那样

You could try it, but this approach handles a lot of parts that docker volume maps to handle, but docker volume already does.

zc2638:
> because now we haven't used dirs, we can remove dir parameter But if we merge dirs and files, it's our duty to check the passed in arguments whether it is dir or it is file, to generate copy cmd(cp or cp -r)? So better distinguish it?

In theory, there is no need to distinguish whether the mapping is a directory or a file. We only need to copy the resources in the container to the host. When using it, the program must know whether they are directories or files, otherwise an error will occur.

gy95:
> > because now we haven't used dirs, we can remove dir parameter But if we merge dirs and files, it's our duty to check the passed in arguments whether it is dir or it is file, to generate copy cmd(cp or cp -r)? So better distinguish it?
> 
> In theory, there is no need to distinguish whether the mapping is a directory or a file. We only need to copy the resources in the container to the host. When using it, the program must know whether they are directories or files, otherwise an error will occur.

no, if we want to copy a file, we just need `cp xxx xxx`, but if we want to copy a dir, we need to append a **_-r_** flag, like `cp -r xxx xxx`

zc2638:
> no, if we want to copy a file, we just need `cp xxx xxx`, but if we want to copy a dir, we need to append a **_-r_** flag, like `cp -r xxx xxx`

Whether it's a file or a directory, you can use `cp -r` and it will work as expected.

gy95:
PTAL again @zc2638

keadm/cmd/keadm/app/cmd/edge/image.go
```diff
@@ -41,13 +41,10 @@ func request(opt *common.JoinOptions, step *common.Step) error {
...
-		util.KubeEdgePath: filepath.Join(util.KubeEdgeTmpPath, "data"),
-	}
 	files := map[string]string{
-		filepath.Join(util.KubeEdgeUsrBinPath, util.KubeEdgeBinaryName): filepath.Join(util.KubeEdgeTmpPath, "bin", util.KubeEdgeBinaryName),
+		filepath.Join(util.KubeEdgeUsrBinPath, util.KubeEdgeBinaryName): filepath.Join(util.KubeEdgeUsrBinPath, util.KubeEdgeBinaryName),
 	}
-	if err := runtime.CopyResources(imageSet.Get(image.EdgeCore), dirs, files); err != nil {
+	if err := runtime.CopyResources(imageSet.Get(image.EdgeCore), files); err != nil {
 		return fmt.Errorf("copy resources failed: %v", err)
 	}
```
zc2638, line 49: The dir resource copy was deleted, should we keep it?

zc2638: `dirs` resource mapping moved to `files`?

gy95: yes

gy95: for now, we don't copy dir, so the dirs input parameter can be removed totally.

zc2638: I mean, is it necessary to write the original mapping in `dirs` into `files`?
At present, the method has supported both the file path and the dir path in files.
Or maybe the original dir mapping is not needed anymore?

gy95: oh, i see, I'll remove the dir parameter :)

gy95: @zc2638 remove dir parameter totally, we can extend copy dir functions once we need it, now we only use to copy files, so not support copy dirs. PTAL

gy95:
cc @fisherxu PTAL :)

gy95:
cc @fisherxu

gy95:
cc @fisherxu PTAL

fisherxu (commented):
/approve

PR #3862, opened 2022-05-23
add rate limiter for edgehub

wackxu:
**What type of PR is this?**

/kind feature

Fixes #

add rate limiter for edgehub

```release-note
Add rate limiter for edgehub to allow uses to set rate when send message to cloudHub. Users could edit the config `MessageQPS` and `MessageBurst` field in `edgeHub` config, the default QPS is 30 and default burst is 60.
```

### DIFF
--- a/pkg/apis/componentconfig/edgecore/v1alpha1/types.go
+++ b/pkg/apis/componentconfig/edgecore/v1alpha1/types.go
@@ -246,6 +246,12 @@ type EdgeHub struct {
 	// Heartbeat indicates heart beat (second)
 	// default 15
 	Heartbeat int32 `json:"heartbeat,omitempty"`
+	// MessageQPS is the QPS to allow while send message to cloudHub.
+	// DefaultQPS: 30
+	MessageQPS int32 `json:"messageQPS,omitempty"`
+	// MessageBurst is the burst to allow while send message to cloudHub.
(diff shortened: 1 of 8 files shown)


pkg/apis/componentconfig/edgecore/v1alpha1/types.go
```diff
@@ -246,6 +246,12 @@ type EdgeHub struct {
 	// Heartbeat indicates heart beat (second)
 	// default 15
 	Heartbeat int32 `json:"heartbeat,omitempty"`
+	// MessageQPS is the QPS to allow while send message to cloudHub.
+	// DefaultQPS: 10
+	MessageQPS int32 `json:"messageQPS,omitempty"`
+	// MessageBurst is the burst to allow while send message to cloudHub.
+	// DefaultBurst: 20
```
gy95, line 253: comment is not equal to default constants: should be modified to 30 60

wackxu: Thanks for you review, I will update it, add rate limiter for edgehub may break some user edge node when user upgrade edgecore, so the default value is important, so I will set the default value big enough and do some test

fisherxu (commented):
/lgtm
/approve

PR #3864, opened 2022-05-24
update the path

my-git9:
/kind cleanup

the pkg path `pkg/apis/meta/v1alpha1/types.go` seems moved to `pkg/apis/componentconfig/meta/v1alpha1/types.go`


wackxu:
lgtm

PR #3866, opened 2022-05-24
remove invalid url

my-git9:
/kind cleanup

The image seems removed in commit: https://github.com/kubeedge/kubeedge/commit/f3678bc02426ab38d724a864f9bd8ece1b046a79

Maybe we should remove the corresponding reference or restore the previous image.


fisherxu (commented):
/lgtm
/approve

PR #3867, opened 2022-05-24
add goproxy for china mainland

my-git9:
/kind feature

Add the ability to set a proxy for mainland China.


fisherxu (commented):
Thanks for the contributions, and you'd better to set it in your own environment, instead of  writing in script :) @my-git9 
/close

PR #3868, opened 2022-05-25
once is an unused value

cl2017:
Signed-off-by: chenlin.liu <chenlin.liu@daocloud.io>

**What type of PR is this?**
/kind bug

once is an unused value, add once.Do function

Fixes #

```release-note
NONE
```


gy95:
lgtm

wackxu:
lgtm

zc2638:
It is not necessary for the once of the `context` and the once of the `channel` to exist at the same time. Some refactoring may be required here.
ref: https://github.com/kubeedge/kubeedge/blob/master/staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go#L41

cl2017:
If it is not necessary for the once, we can remove unused once variable?

PR #3881, opened 2022-05-27
fix build_with_container.sh

algobot76:
**What type of PR is this?**

/kind bug

I can't run `make` with the current build script on mac as `docker` doesn't exist in `/etc/group`. So I tweak the `build_with_container.sh` script to make it compatible with mac.

n/a

n/a

```release-note
NONE
```

### DIFF
--- a/hack/make-rules/build_with_container.sh
+++ b/hack/make-rules/build_with_container.sh
@@ -23,11 +23,12 @@ set -o pipefail
 KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 MOUNTPATH="${MOUNTPATH:-/kubeedge}"
 KUBEEDGE_BUILD_IMAGE=${KUBEEDGE_BUILD_IMAGE:-"kubeedge/build-tools"}
-DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}"
+UID_GID="$UID"
+DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}" && UID_GID="${UID}:${DOCKER_GID}"
 CONTAINER_RUN_OPTIONS="${CONTAINER_RUN_OPTIONS:--it}"
 
 echo "start building inside container"
-docker run --rm ${CONTAINER_RUN_OPTIONS} -u "${UID}:${DOCKER_GID}" \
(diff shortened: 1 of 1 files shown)


hack/make-rules/build_with_container.sh
```diff
@@ -23,11 +23,12 @@ set -o pipefail
 KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 MOUNTPATH="${MOUNTPATH:-/kubeedge}"
 KUBEEDGE_BUILD_IMAGE=${KUBEEDGE_BUILD_IMAGE:-"kubeedge/build-tools"}
-DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}"
+USERNAME="$UID"
+DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}" && USERNAME="${UID}:${DOCKER_GID}"
```
gy95, line 27: what's the difference before and after this fix? Isn't that exactly the same? :)

algobot76: To prevent the script from terminating with exit code 1 when it can't find the gid of docker.

gy95: on mac, there's no file `/etc/group`, and then `grep '^docker:' /etc/group | cut -f3 -d:)` cmd will fail. And by running `$?` cmd, we will find the exit code is not 0.
And on mac, by appending `&& USERNAME="${UID}:${DOCKER_GID}"`, the command will run successfully and the exit code will be 0
That's why this patch will fix on mac terminating with exit code 1

hack/make-rules/build_with_container.sh
```diff
@@ -23,11 +23,12 @@ set -o pipefail
 KUBEEDGE_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 MOUNTPATH="${MOUNTPATH:-/kubeedge}"
 KUBEEDGE_BUILD_IMAGE=${KUBEEDGE_BUILD_IMAGE:-"kubeedge/build-tools"}
-DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}"
+UID_GID="$UID"
```
gy95, line 26: UID="$UID" ?

algobot76: image

By default, we use uid as the default parameter for `-u`. Then append the docker's gid to the uid in the form of `uid:gid` if it is present.

PR #3907, opened 2022-06-09
update crds to newer ones

snstaberah:
crds in this file use apiextensions.k8s.io/v1beta1,which has been removed from kubernetes 1.22; use newer generated crds instead of the old ones;

**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
update crds in this file to adapt k8s after 1.22；
 apiextensions.k8s.io/v1 is available in k8s 1.16；
**Which issue(s) this PR fixes**:

Fixes #
3bf38f9923d0eeec60d0a8a0db6c4f4

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #3910, opened 2022-06-13
remove out of date crds

snstaberah:
crds of device v1alpha1 is out of date,remove them to avoid
misunderstanding

**What type of PR is this?**

Fixes #
https://github.com/kubeedge/kubeedge/pull/3907#issuecomment-1153569822

```release-note

```


snstaberah:
> maybe some versions <=v1.3 cannot install successfully with keadm, even though we not support those releases anymore, is this OK? https://github.com/kubeedge/kubeedge/blob/release-1.3/keadm/cmd/keadm/app/cmd/util/k8sinstaller.go
> 
> https://github.com/kubeedge/kubeedge/blob/2a64a950a791d8ff62f4540a02f1110e3e80d0da/keadm/cmd/keadm/app/cmd/util/k8sinstaller.go#L120-L135

thanks，i think this is a hardcore coding ：） ，maybe i can optimize this in next pr？

fisherxu:
>  the common method to support old version crds is write a conversion webhook to update them 

Sounds good for compatibility. Could you please open an issue to track it?
And for 1.3 release, it's too old.. and have been out of maintenance period. So no need to keep it.

snstaberah:
> > the common method to support old version crds is write a conversion webhook to update them
> 
> Sounds good for compatibility. Could you please open an issue to track it? And for 1.3 release, it's too old.. and have been out of maintenance period. So no need to keep it.

ok，I have opened a new issue here ：
https://github.com/kubeedge/kubeedge/issues/3913#issue-1270361473

PR #3911, opened 2022-06-13
fix type confusion in csi driver

AdamKorcz:
Signed-off-by: AdamKorcz <Adam@adalogics.com>

**What type of PR is this?**
/kind bug

This PR fixes a type confusion in the csi driver. The return value of result.GetContent() can be nil which would result in: `panic: interface conversion: interface {} is nil, not string`.

The issue was found by the `FuzzextractMessage` fuzzer (https://github.com/cncf/cncf-fuzzing/blob/main/projects/kubeedge/csidriver_fuzzer.go#L20)

@vincentgoat @kevin-wangzefeng @gy95 

Fixes #

```release-note

```


cloud/pkg/csidriver/controllerserver.go
```diff
@@ -108,7 +108,11 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 
 	klog.V(4).Infof("create volume result: %v", result)
-	data := result.GetContent().(string)
+	data, ok := result.GetContent().(string)
+	if !ok {
+		klog.Errorf("malformed content was received: %s", data)
+		return nil, errors.New(data)
```
gy95, line 114: LGTM, but if the value type is not a string, data here will be an empty string? it's better to add more info about data.

gy95: How about this? @AdamKorcz :)
```
data, ok := result.GetContent().(string)
if !ok {
	klog.Errorf("content is not string type: %v", result.GetContent())
	return nil, fmt.Errorf("content type %T is not string", result.GetContent())
}
```

fisherxu: Agree :) But seems that the returned error and klog printed error are duplicated, we can fix it in follow up PRs.

vincentgoat (approved):
/lgtm

gy95:
LGTM thanks a lot :)

fisherxu (commented):
/lgtm
/approve

Issue #3760, opened 2022-03-31
Umbrella issue for Device Management Interface(DMI)


RyanZhaoXB:
hello guys. we have make a proposal for Device Management Interface (i.e. DMI), which integrates the device management interface and optimizes the device management capability of KubeEdge.  

designing doc for en: [DMI doc](https://docs.google.com/document/d/143y_ZqkDzUt95J6S99Wc9cJHx8k3mRy18a6T_4ZqaO8/edit#heading=h.1plg3hjvj5c3)

desining doc for zh: [DMI文档](https://shimo.im/docs/pmkxQg2bxyTlYjAN)

we can discuss more details about this proposal and welcome to make suggestion for DMI, thanks.

**What would you like to be added/modified**:
we want to change the device management module in KubeEdge to build a cloud native device twin management platform which covers device management, device operation and device data. and we want to make DMI the standard for device management in cloud-edge collaboration, just like CRI and CSI.

**Why is this needed**:
for the current way, KubeEdge can only get the device connected via mapper and synchronize the device twin, which can not meet the users' requirements for device management. 

Proposal link: #3914

Task List:
1. Add dmi proto check to ci

feng-crazy:
设备管理是基于 K8S的CRD的，K8S的apiserver，没有批量创建CRD的接口，如果我要建几万个设备的话，就会很慢，我们的场景是这样的，我们做了一个创建设备模板和实例的界面，界面上一个批量创建设备的功能，是根据导入excel解析设备，来创建的设备实例，所以要循环调用创建CRD的接口，这个过程非常的长。

RyanZhaoXB:
> 设备管理是基于 K8S的CRD的，K8S的apiserver，没有批量创建CRD的接口，如果我要建几万个设备的话，就会很慢，我们的场景是这样的，我们做了一个创建设备模板和实例的界面，界面上一个批量创建设备的功能，是根据导入excel解析设备，来创建的设备实例，所以要循环调用创建CRD的接口，这个过程非常的长。

[...]

PR #3915, opened 2022-06-14
add DMI apis: interface and definition

RyanZhaoXB:
**What type of PR is this?**
/kind feature

for the current way, KubeEdge can only get the device connected via mapper and synchronize the device twin, which can not meet the users' requirements for device management. in this pr, we add the apis of DMI including service interface and struct definition. it also update the CRD of the device instance and device model. 

Fixes #https://github.com/kubeedge/kubeedge/issues/3760

we will implement these apis later in the next submission. 

```release-note
NONE
```


gy95:
need run `hack/update-codegen.sh` and `hack/generate-crds.sh` to update generated codes and crds

pkg/apis/devices/v1alpha2/device_model_types.go
```diff
@@ -25,6 +25,25 @@ import (
...
+	// Required: The device command url.
+	Url string `json:"url"`
+	// Required: The device command method.
+	Method string `json:"method"`
+	// Required: The device command status code.
+	StatusCode []string `json:"status_code"`
+	// +optional
+	Parameters []string `json:"parameters, omitempty"`
+	// +optional
+	Response []interface{} `json:"response, omitempty"`
```
TianTianBigWang, line 46: maybe just `interface{}` or `[]byte` ?

RyanZhaoXB: ok. i will make it interface{}

RyanZhaoXB:
> need run `hack/update-codegen.sh` and `hack/generate-crds.sh` to update generated codes and crds

i have make it in the latest commit

TianTianBigWang:
lgtm

PR #3917, opened 2022-06-14
fix sendresp stuck occasionally when sendsync receive select exited

wackxu:

**What type of PR is this?**

Fixes #

in `SendSync ` , we send messages synchronously.  After we send a message, we will create a new channel and store it in the map according to the message id. Then, we use select to wait for the response message. If it exceeds the specified time, just return directly. 

in `SendResp`, when we got a response message, we will find the resonse channel according to the message parentID, and then we insert the message to the response channel.

There is an extreme corner case，when `SendSync` select receive signal from the timeout channel, but not execute the return statement， at the same time， in `SendSync` , we already got the lock and a response message is inserting to the response channel, but  no goroutine is ready for receive the message in `SendSync `. because the response channleis an  unbuffered channel, so this will blocked. and the lock can not release then the whole process is stuck.

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go
@@ -152,7 +152,12 @@ func (ctx *Context) SendResp(message model.Message) {
 	ctx.anonChsLock.RLock()
 	defer ctx.anonChsLock.RUnlock()
 	if channel, exist := ctx.anonChannels[anonName]; exist {
-		channel <- message
+		select {
+		case channel <- message:
+		default:
+			klog.Warningf("no goroutine is ready for receive the message from "+
+				"unbuffered response channel, discard this resp message for %s", message.GetParentID())
+		}
 		return
 	}
 


gy95:
LGTM
and maybe need to add some more descriptions

gy95:
lgtm

vincentgoat (approved):
lgtm

zc2638:
Will this cause message loss?

wackxu:
> Will this cause message loss?

yes，in this case, the resp message will be lost and the request will be retry in the edge.

zc2638:
The `metaManager` module calls sendToEdged in many places, and there is no retry. If the message is lost, will it have an impact?

wackxu:
> The `metaManager` module calls sendToEdged in many places, and there is no retry. If the message is lost, will it have an impact?

The sendResp is send response to the message that is send by method `sendSync`, which means it need to synchronously wait for the result of sending a message.  and we can see that if it received the resp timeout, it will return an err in `sendSync` , and in most cases, the message will be rely on edged coordination mechanism to retry sending. and in some case if sendSync err it will just log the err. 

image

In this pr, we fixed an extreme case of sendsync and sendResp that will cause the channel and lock block. in this case the upstream message can not be send to the cloud and whole edge node will not work.

wackxu:
the extreme corner case
```
There is an extreme corner case，when SendSync select receive signal from the timeout channel, but not execute the return statement， at the same time， in SendSync , we already got the lock and a response message is inserting to the response channel, but no goroutine is ready for receive the message in SendSync . because the response channleis an unbuffered channel, so this will blocked. and the lock can not release then the whole process is stuck.
```

zc2638:
Thanks for your patience in explaining.
/lgtm

fisherxu (commented):
/approve

PR #3918, opened 2022-06-14
remove duplicate sending msg to full channel

gy95:
**What type of PR is this?**

This block is confusing
When we write data failed, we don't need to write again, IMOH, we can remote the second write operation.
Before https://github.com/kubeedge/kubeedge/pull/2664 , we use `select` to try to write data to channel again, but I think it's not needed.

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go
@@ -173,7 +173,6 @@ func (ctx *Context) SendToGroup(moduleType string, message model.Message) {
 		case ch <- message:
 		default:
 			klog.Warningf("The module %s message channel is full, message: %+v", module, message)
-			ch <- message
 		}
 	}
 	if channelList := ctx.getTypeChannel(moduleType); channelList != nil {


vincentgoat (approved):
LGTM

staging/src/github.com/kubeedge/beehive/pkg/core/channel/context_channel.go
```diff
@@ -173,7 +173,6 @@ func (ctx *Context) SendToGroup(moduleType string, message model.Message) {
 		case ch <- message:
 		default:
 			klog.Warningf("The module %s message channel is full, message: %+v", module, message)
-			ch <- message
```
fisherxu, line 176: If the channel is full, should we wait to write or discard? If just discard, the message maybe lost? If wait to write, whether the order of the message can be ensured between multi goroutine?

gy95: yes, i think we can just drop the messages when the message channel is full.

fisherxu: Need to check the design model or where use the sendToGroup, whether we should discard :)

Issue #3919, opened 2022-06-14
create mapper automatically when a device instance is created


RyanZhaoXB:
**What would you like to be added/modified**:
i want to modify the cloudcore to create a mapper deployment automatically when a device is created.

**Why is this needed**:
when a user create a device, he should have created a mapper deployment on the specific edge node in advance. now we improve the user experience. when a user create a device with mapper image filled in the device annotation with the key "image", the cloudcore will make a deployment name in the form of mapper-{node_name}-{protocol_name}, and will query if this deployment existed. if the mapper deployment has not been created, the cloudcore will create one with the mapper image of device automatically. and the configmap will also be mounted automatically by volume. 

thus the user can create his device just through one action and there is no need to check whether the mapper has been created or create a mapper manually. and if the user do not want to generate the mapper automatically, he should just not fill the image in the device annotation.

the device instance should be just as follows if the user wants the cloudcore to create mapper automatically.
```
apiVersion: devices.kubeedge.io/v1alpha2
kind: Device
metadata:
  name: counter
  labels:
    description: 'counter'
  annotations:
    image: 'kubeedge/kubeedge-pi-counter:v1.0.0'
spec:
  protocol: 
    customizedProtocol:
      protocolName: counter
  deviceModelRef:
    name: counter-model
  nodeSelector:
    nodeSelectorTerms:
    - matchExpressions:
      - key: ''
        operator: In
        values:
        - ecs-edgenode
```

PR #3920, opened 2022-06-14
create mapper automatically when a device instance is created

RyanZhaoXB:
Signed-off-by: RyanZhaoXB <zhaoran11@huawei.com>

**What type of PR is this?**

/kind feature

when a user create a device, he should have created a mapper deployment on the specific edge node in advance. now we improve the user experience in this pr. when a user create a device with mapper image filled in the device annotation with the key "image", the cloudcore will make a deployment name in the form of mapper-{node_name}-{protocol_name}, and will query if this deployment existed. if the mapper deployment has not been created, the cloudcore will create one with the mapper image of device automatically. and the configmap will also be mounted automatically by volume. and if the user do not want to generate the mapper automatically, he should just not fill the image in the device annotation.

the device instance should be just as follows if the user wants the cloudcore to create mapper automatically. 
```
apiVersion: devices.kubeedge.io/v1alpha2
kind: Device
metadata:
  name: counter
  labels:
    description: 'counter'
  annotations:
    image: 'kubeedge/kubeedge-pi-counter:v1.0.0'
spec:
  protocol: 
    customizedProtocol:
      protocolName: counter
  deviceModelRef:
    name: counter-model
  nodeSelector:
    nodeSelectorTerms:
    - matchExpressions:
      - key: ''
        operator: In
        values:
        - ecs-edgenode
```

Fixes #https://github.com/kubeedge/kubeedge/issues/3919

when the DMI is developed, the automatical creation of the data service of the mapper should also be implemented in the same way.  

```release-note
NONE
```


cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -167,6 +175,119 @@ func (dc *DownstreamController) syncDevice() {
...
+}
+
+// generateMapperDeployment generate the deployment for a mapper
+func generateMapperDeployment(mapperName, namespace, nodeName, mapperImage string) appv1.Deployment {
+	appLabelInfo := map[string]string{
+		App: mapperName,
+	}
+	nodeSelectorList := make(map[string]string)
+	nodeSelectorList[EdgeLabel] = ""
+	nodeSelectorList[NodeHostNameLabel] = nodeName
```
gy95, line 250: I think we can not use this label to enable k8s to select the exact `nodeName` node to be scheduled?

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -167,6 +175,119 @@ func (dc *DownstreamController) syncDevice() {
...
+		},
+		Spec: appv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: appLabelInfo,
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: appLabelInfo,
+				},
+				Spec: v1.PodSpec{
```
gy95, line 265: PodSpec has a field `NodeName`, we can set this field with the input parameter `nodeName`?

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -167,6 +175,119 @@ func (dc *DownstreamController) syncDevice() {
 	}
 }
 
+// addDeviceMapper is function to add deviceMapper deployment
+func (dc *DownstreamController) addDeviceMapper(device *v1alpha2.Device, nodename string) {
+	protocol, err := getProtocolNameOfDevice(device)
```
gy95, line 180: add a pre check to check whether the device has image annotation just like below?
```
annotations:
    image: 'kubeedge/kubeedge-pi-counter:v1.0.0'
```
If not, we can just return and do nothing, no need to continue the next processing.

gy95 (commented):
LGTM
I think it's OK. :)

wackxu:
why not use daemonset to deploy the device mapper? If we have many device and edge node, then there maybe many deployment

cloud/pkg/devicecontroller/controller/downstream.go
RyanZhaoXB: OK. i have make it field of NodeName instead of NodeSelector.

RyanZhaoXB:
> 



> why not use daemonset to deploy the device mapper? If we have many device and edge node, then there maybe many deployment

i think we should not use daemonset here, at least not now. the deployment mode of mapper is highly related to the scenario of users. not every node can process the same business of the same protocol. 

and for the current ability of KubeEdge mapper, every device is bounded to a specific node and every pod of mapper can only get the device information through a configmap with name of `device-profile-config-{node_name}`. daemonset can not make it work. maybe we can improve it in the implementation of DMI.

fisherxu:
Some aspects need to consider :)
1. Whether every user need to auto create the mapper deployment? Or we can add some flag to control it.
2. If user want to update the mapper, how to do it, if update the deployment directly, the image in annotation will not be updated.
3. If the deployment is auto cteated, who will delete it?

RyanZhaoXB:
> Some aspects need to consider :)
> 
> 1. Whether every user need to auto create the mapper deployment? Or we can add some flag to control it.
> 2. If user want to update the mapper, how to do it, if update the deployment directly, the image in annotation will not be updated.
> 3. If the deployment is auto cteated, who will delete it?

let me explain it.
> 1. Whether every user need to auto create the mapper deployment? Or we can add some flag to control it.
not every user need to create the mapper deployment. the field of image in annotation of device is the flag. if the user does not fill image in the annotation of device, cloudcore will not create mapper automatically. everything run as before.

> 2. If user want to update the mapper, how to do it, if update the deployment directly, the image in annotation will not be updated.
the user can just udpate the existed mapper deployment directly because cloudcore will not create a new mapper if an old one exists. 

> 3. If the deployment is auto cteated, who will delete it?
the user will delete it now. when a use delete a device, we do not know if this user does not want to use the mapper any more or he just want to delete this device only. it depends on the user's opinion. thus KubeEdge does not delete mapper for use. for the further consideration, KubeEdge can delete the mapper when all devices of a single protocol on the specific node are all deleted. this capability can be added in the next version.

fisherxu:
The relationship of device instance, mapper and device needs to to clarified, one mapper can map to multi device instances and multi devices? One device instance should be map to one device?
In the current implementation, if I remember correctly, one mapper only map to one device instance and device? cc @RyanZhaoXB

RyanZhaoXB:
> The relationship of device instance, mapper and device needs to to clarified, one mapper can map to multi device instances and multi devices? One device instance should be map to one device? In the current implementation, if I remember correctly, one mapper only map to one device instance and device? cc @RyanZhaoXB

In the current implementation, one device instance maps to one device and one mapper can manage multi devices. the mapper will manage a list of devices which are managed by this mapper with the same protocol. the information of the devices are imported to the mappers through a configmap. this configmap should carry all devices bounded to the same node. all the mappers deployed on this node can get this configmap and find the devices on its own list of devices. @fisherxu

Issue #3607, opened 2022-02-10
Upgrade K8s dependency to 1.22


fisherxu:
**What would you like to be added/modified**:
K8s has released v1.22.6, we need to upgrade to v1.22

PR #3924, opened 2022-06-15
feat: use tls.CipherSuites() in cloudcore stream server to resolve CV…

Abirdcfly:
…E-2016-2183

**What type of PR is this?**
/kind bug

Our customer found this issue (CVE-2016-2183 in cloudcore stream port, default is 10003)  when using a third-party security scanner, but according to go upstream https://github.com/golang/go/issues/41476, it doesn't seem to be a problem with golang itself.

But such a fix has no side effect on kubeedge. And either `docker run --rm -ti drwetter/testssl.sh -W cloucoreIP:10003` or `nmap -sV --script ssl-enum-ciphers -p 10003 cloucoreIP` way both consider it a security issue. So I think it makes sense to fix it.

Fixes #

```release-note
use tls.CipherSuites() in cloudcore stream port, this disable some insecure cipher like TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA or TLS_RSA_WITH_3DES_EDE_CBC_SHA 
```

### DIFF
--- a/cloud/pkg/cloudstream/streamserver.go
+++ b/cloud/pkg/cloudstream/streamserver.go
@@ -306,15 +306,20 @@ func (s *StreamServer) Start() {
 		return
 	}
 	pool.AppendCertsFromPEM(data)
-
+	secureCiphers := tls.CipherSuites()
+	selectedCipherSuites := make([]uint16, 0, len(secureCiphers))
+	for _, s := range secureCiphers {
+		selectedCipherSuites = append(selectedCipherSuites, s.ID)
+	}
 	streamServer := &http.Server{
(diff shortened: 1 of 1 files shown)


Abirdcfly:
Kubernetes 1.23 start to use go 1.17 ref https://github.com/kubernetes/kubernetes/pull/103692
And KubeEdge HEAD (master) relies on kubernetes 1.22.6.

IMO, we can freeze this commit for now until we rely on kubernetes 1.23 😂 (BTW, The last update of kubernetes dependency version is https://github.com/kubeedge/kubeedge/issues/3607) 

/hold

koulq:
The question is on which version was released, I encountered this side

PR #3925, opened 2022-06-16
always upload e2e logs

gy95:
**What type of PR is this?**

we should always upload e2e logs, whether the e2e is successful, or is failed, or is canceled due to a timeout

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -171,6 +171,7 @@ jobs:
       - run: make e2e
 
       - uses: actions/upload-artifact@v3
+        if: always()
         with:
           name: e2e-test-logs
           path: |


.github/workflows/main.yaml
```diff
@@ -171,6 +171,7 @@ jobs:
       - run: make e2e
 
       - uses: actions/upload-artifact@v3
+        if: always()
```
fisherxu, line 174: It will not upload by default if running successed?

gy95: no, it will upload logs always,  regardless of success, failure, or cancellation.

gy95: without this `if: always()` flag, will only upload logs when the last step is successful

fisherxu: Oh, so if without `if: always()`, when `make e2e` run failed, this upload-artifact wil be skipped?

gy95: yes
ref: https://github.com/kubeedge/kubeedge/actions/runs/2501087906

fisherxu (commented):
/lgtm
/approve

PR #3937, opened 2022-06-17
handle error and clean up

falser101:

**What type of PR is this?**
/kind cleanup

error handle

Fixes #

```release-note
None
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/helm/installer.go
+++ b/keadm/cmd/keadm/app/cmd/helm/installer.go
@@ -35,7 +35,7 @@ const (
 	DefaultProfileFilename = "version.yaml"
 	DefaultHelmValuesPath  = "values.yaml"
 
-	DefaultHelmTimeout = time.Duration(60 * time.Second)
+	DefaultHelmTimeout = 60 * time.Second
 
 	DefaultHelmInstall  = true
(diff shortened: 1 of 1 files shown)


wackxu:
lgtm

keadm/cmd/keadm/app/cmd/helm/installer.go
```diff
@@ -277,7 +277,10 @@ func (cu *KubeCloudHelmInstTool) runHelmManifest(r *Renderer, stdout io.Writer)
 	}
 	buf.WriteString(manifests)
```
wackxu, line 278: should we handle the err return by buf.WriteString?

falser101: `func (b *Buffer) WriteString(s string) (n int, err error) {
	b.lastRead = opInvalid
	m, ok := b.tryGrowByReslice(len(s))
	if !ok {
		m = b.grow(len(s))
	}
	return copy(b.buf[m:], s), nil
}`

// WriteString appends the contents of s to the buffer, growing the buffer as
// needed. The return value n is the length of s; err is always nil. If the
// buffer becomes too large, WriteString will panic with ErrTooLarge.
I think we do not need handle

PR #3950, opened 2022-06-20
remove stale performance tests and stub for cloudcore and edgecore

wackxu:

**What type of PR is this?**

/kind cleanup

remove stale performance tests and stub for cloudcore and edgecore

1、delete the stub of cloudcore and edgecore, the stub is used for testing, but the stub is too simple, and it is very different from the actual logic of cloudcore and edgeocre, and cannot reflect the real situation of edgecore and cloudcore when test.

2、delete the performance test in `tests/performance`.  first the performance is use the old stub and old config and the code doesn't work now, second, There is a big gap between the results of the stub test and the real environment, and it does not have reference significance， third, the test case in the performance is coverd by https://github.com/kubeedge/kubeedge/blob/master/build/edgemark/edgemark_setup_guide.md, we can use edgemark and clusterload to test the case in the performance, see the latest performance report https://kubeedge.io/zh/blog/scalability-test-report/

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #3957, opened 2022-06-21
add controllerManager to manifest

gy95:
**What type of PR is this?**

Usage:
Install v1.10.0 kubeedge and controllermanager, by default, we don't install controller, so we should use flag `--set controllerManager.enable=true` to install it, and due to v1.10.0 controller-manager not exist, we use flag  ` --set controllerManager.image.repository=guoyao95/controller-manager` to use our private image
```
_output/local/bin/keadm init --advertise-address=10.0.0.115 --profile version=v1.10.0 --set controllerManager.enable=true --set controllerManager.image.repository=guoyao95/controller-manager
```

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/helm/installer.go
+++ b/keadm/cmd/keadm/app/cmd/helm/installer.go
@@ -383,9 +383,15 @@ func (cu *KubeCloudHelmInstTool) handleProfile(profileValue string) error {
 			profileValue = "v" + profileValue
 		}
 
-		cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "cloudCore.image.tag", profileValue))
-		cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "iptablesManager.image.tag", profileValue))
-		cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "controllerManager.image.tag", profileValue))
+		if !SetsContainSubstring(cu.Sets, "cloudCore.image.tag") {
+			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "cloudCore.image.tag", profileValue))
+		}
+		if !SetsContainSubstring(cu.Sets, "iptablesManager.image.tag") {
+			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "iptablesManager.image.tag", profileValue))
+		}
+		if !SetsContainSubstring(cu.Sets, "controllerManager.image.tag") {
+			cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "controllerManager.image.tag", profileValue))
+		}
 
 	case IptablesMgrProfileKey:
 		if profileValue == "" {
@@ -410,6 +416,16 @@ func (cu *KubeCloudHelmInstTool) handleProfile(profileValue string) error {
 	return nil
 }
 
+// SetsContainSubstring checks whether a substring contains in string set
+func SetsContainSubstring(sets []string, sub string) bool {
+	for _, v := range sets {
+		if strings.Contains(v, sub) {
+			return true
+		}
+	}
+	return false
+}
+
 func (cu *KubeCloudHelmInstTool) rebuildFlagVals() error {
 	unDuplicatedStore := make(map[string]string)
 
@@ -423,6 +439,9 @@ func (cu *KubeCloudHelmInstTool) rebuildFlagVals() error {
 
 		unDuplicatedStore[p[0]] = p[1]
 	}
+
+	// clear Sets to avoid append on the exist slice
+	cu.Sets = []string{}
 
 	for k, v := range unDuplicatedStore {
 		cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", k, v))
--- a/keadm/cmd/keadm/app/cmd/helm/installer.go
+++ b/keadm/cmd/keadm/app/cmd/helm/installer.go
@@ -41,16 +41,17 @@ const (
 	DefaultHelmWait     = true
 	DefaultHelmCreateNs = true
 
-	VersionProfileKey       = "version"
-	IptablesMgrProfileKey   = "iptablesmgr"
-	ExternalIptablesMgrMode = "external"
-	InternalIptablesMgrMode = "internal"
-	EdgemeshProfileKey      = "edgemesh"
+	VersionProfileKey           = "version"
+	IptablesMgrProfileKey       = "iptablesmgr"
+	ControllerManagerProfileKey = "controllermanager"
+	ExternalIptablesMgrMode     = "external"
+	InternalIptablesMgrMode     = "internal"
+	EdgemeshProfileKey          = "edgemesh"
 )
 
 var (
 	ErrListProfiles = errors.New("can not list profiles")
-	ValidProfiles   = map[string]bool{VersionProfileKey: true, IptablesMgrProfileKey: true}
+	ValidProfiles   = map[string]bool{VersionProfileKey: true, IptablesMgrProfileKey: true, ControllerManagerProfileKey: true}
 )
 
 // KubeCloudHelmInstTool embeds Common struct
@@ -231,7 +232,7 @@ func (cu *KubeCloudHelmInstTool) buildRenderer(baseHelmRoot string) (*Renderer, 
 	var subDir string
 	if cu.existsProfile && cu.isInnerProfile() {
 		switch cu.ProfileKey {
-		case VersionProfileKey, IptablesMgrProfileKey:
+		case VersionProfileKey, IptablesMgrProfileKey, ControllerManagerProfileKey:
 			componentName = CloudCoreHelmComponent
 			subDir = fmt.Sprintf("%s/%s", ChartsSubDir, CloudCoreHelmComponent)
 		// we can implement edgemesh here later.
@@ -356,6 +357,7 @@ func (cu *KubeCloudHelmInstTool) checkProfile(baseHelmRoot string) error {
 
 	// iptalesmgr is also an valid profile key.
 	validProfiles[IptablesMgrProfileKey] = true
+	validProfiles[ControllerManagerProfileKey] = true
 	if ok := validProfiles[cu.ProfileKey]; !ok {
 		validKeys := make([]string, len(validProfiles))
 		for k := range validProfiles {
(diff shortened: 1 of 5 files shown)


gy95:
cc @zhu733756   PTAL :)
I'm not familiar with helm, worried about missing something

manifests/charts/cloudcore/templates/deployment_controllermanager.yaml
```diff
@@ -0,0 +1,38 @@
...
+kind: Deployment
+metadata:
+  {{- with .Values.controllerManager.labels }}
+  labels: {{- toYaml . | nindent 4 }}
+  {{- end }}
+  name: kubeedge-controller-manager
+  namespace: kubeedge
+spec:
+  selector:
+    {{- with .Values.controllerManager.labels }}
```
zhu733756, line 12: do we need annotations?

keadm/cmd/keadm/app/cmd/helm/installer.go
```diff
@@ -231,7 +232,7 @@ func (cu *KubeCloudHelmInstTool) buildRenderer(baseHelmRoot string) (*Renderer,
 	var subDir string
 	if cu.existsProfile && cu.isInnerProfile() {
 		switch cu.ProfileKey {
-		case VersionProfileKey, IptablesMgrProfileKey:
+		case VersionProfileKey, IptablesMgrProfileKey, ControllerManagerProfileKey:
```
gy95, line 235: Is this right? @zhu733756

manifests/charts/cloudcore/values.yaml
```diff
@@ -99,3 +99,32 @@ iptablesManager:
     requests:
       cpu: 100m
       memory: 25Mi
+
+controllerManager:
+  enable: false
+  hostNetWork: true
+  image:
```
zhu733756, line 106: Is this component needed host network? I'm not familiar with it.

keadm/cmd/keadm/app/cmd/helm/installer.go
```diff
@@ -397,6 +399,10 @@ func (cu *KubeCloudHelmInstTool) handleProfile(profileValue string) error {
 		}
 
 		return fmt.Errorf("the given mode of iptablesmgr %s is not supported, only support internal or external", profileValue)
+
+	case ControllerManagerProfileKey:
+		cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "controllerManager.image.tag", currentVersion))
```
zhu733756, line 404: the handling logic of the image tag should be consistent?

zhu733756:
Do we need a svc to post some api to this component?

manifests/charts/cloudcore/values.yaml
gy95: no, these deployment files are inherit from dir https://github.com/kubeedge/kubeedge/tree/master/build/controllermanager

keadm/cmd/keadm/app/cmd/helm/installer.go
zhu733756: > Is this right? @zhu733756

Yes

gy95:
> Do we need a svc to post some api to this component?

maybe not, I don't find svc in dir https://github.com/kubeedge/kubeedge/tree/master/build/controllermanager
WDYT @vincentgoat @Congrool

manifests/charts/cloudcore/templates/deployment_controllermanager.yaml
gy95: added. please help check again :)

manifests/charts/cloudcore/values.yaml
wackxu: it does not need host network

zhu733756:
lgtm

Congrool:
I'm not sure if we cloud set the image tag through `--set controllerManager.image.tag=v1.10.0-beta.0-264-g315c619`. I tried this , but the image tag for the controller-manager was still `v1.10.0`.

gy95:
maybe this is due to `--profile version=v1.10.0 ` overwrite the controllermanager version, can just remove `--profile version=v1.10.0 ` and try again

gy95:
@Congrool  the second commit fix this bug, it's due to `cu.Sets` contains duplicated string, If we specify `controllerManager.image.tag` in cmd flag, we don't need to contruct it again with `--profile version=v1.10.0`
cc @zhu733756   please help review the second commit whether is OK. :)

zhu733756:
thanks @gy95 

/lgtm

Congrool:
lgtm

keadm/cmd/keadm/app/cmd/helm/installer.go
```diff
@@ -381,8 +383,15 @@ func (cu *KubeCloudHelmInstTool) handleProfile(profileValue string) error {
 			profileValue = "v" + profileValue
 		}
 
-		cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "cloudCore.image.tag", profileValue))
-		cu.Sets = append(cu.Sets, fmt.Sprintf("%s=%s", "iptablesManager.image.tag", profileValue))
+		if !SetsContainSubstring(cu.Sets, "cloudCore.image.tag") {
```
wackxu, line 386: the behavior is changed vs before, it is right `!SetsContainSubstring`?

manifests/charts/cloudcore/templates/rbac_controllermanager.yaml
```diff
@@ -0,0 +1,55 @@
...
+rules:
+  - apiGroups: ["apps.kubeedge.io"]
+    resources: ["nodegroups", "nodegroups/status"]
+    verbs: ["*"]
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["list", "watch", "patch", "get"]
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["list", "watch", "get", "delete"]
```
wackxu, line 45: do we need delete pod?

Congrool: Oh, actually it does not delete the pod. I'll revise it.

keadm/cmd/keadm/app/cmd/helm/installer.go
gy95: yes, without this, cu.Sets will contain multiple duplicated `cloudCore.image.tag`, one is here `--profile version=v1.10.0`, and will also contain one if we specify in cmd line `--set cloudCore.image.tag=v1.10.0`

wackxu: What is the priority of these two parameters？if user build their own cloudcore image with a different tag, how to  achieve this?

manifests/charts/cloudcore/templates/deployment_controllermanager.yaml
```diff
@@ -0,0 +1,36 @@
...
+        - name: controller-manager
+          image: {{ .Values.controllerManager.image.repository }}:{{ .Values.controllerManager.image.tag }}
+          imagePullPolicy: {{ .Values.controllerManager.image.pullPolicy }}
+      restartPolicy: Always
+      serviceAccountName: controller-manager
+      {{- with .Values.controllerManager.affinity }}
+      affinity: {{ toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.controllerManager.tolerations }}
+      tolerations: {{ toYaml . | nindent 8 }}
```
wackxu, line 34: suggest add imagePullSecrets to allow user config it

keadm/cmd/keadm/app/cmd/helm/installer.go
gy95: --set has a higher priority, If we don't use `--set` to set a image tag, we can use `--profile version=v1.10.0` to specify one uniform image tag(this tag will apply for all images, including cloudcore, iptables-manager, and controller-manager)

for example, if we want to use v1.10.0 cloudcore image and v1.11.0 controller-manager, we can use the below commd
```
_output/local/bin/keadm init --advertise-address=10.0.0.115 --profile version=v1.10.0 --set controllerManager.enable=true --set controllerManager.image.tag=v1.11.0
```

wackxu: got it, thanks for your explain

manifests/charts/cloudcore/templates/rbac_controllermanager.yaml
Congrool: Sorry for my mistake, the authorization for deleting pod is needed for nodegroup controller. I'll explain the scenario:

Considering that we have an eapp deployed in the cluster which having pods running on `node-a` and `node-b`, and `node-a` `node-b` are belonging to `nodegroup-hangzhou`, when we remove `node-a` from `nodegroup-hangzhou`, we should evict the pods of eapp running on `node-a`, and then create new instances on `node-b` or other nodes in `nodegroup-hangzhou`.   The eviction needs to delete the pod.

wackxu: I think we should not do this is nodegroup controller, In edgeapplication, we manage resouces such as deployment， statefulset and so on, we donot manage pods directly. when user delete a pod,  he should do this carefully as below
https://kubernetes.io/zh-cn/docs/tasks/administer-cluster/safely-drain-node/
when he drain the node, all pod on this node will be evict. and the k8s kube-controller will watch this and then create new pod for the deployment or statefulset.

wackxu: there will be conflict when we mange pod vs k8s controller manager

manifests/charts/cloudcore/templates/deployment_controllermanager.yaml
wackxu: I think we should also add priority for all manager component

manifests/charts/cloudcore/templates/rbac_controllermanager.yaml
Congrool: You're right when deleting a node from cluster. But other scenarios we should take into consideration. Let's continue with the above example:  
Situation 1: `node-a` is not deleted from cluster, but just be an orphan node, which means it does not belong to any nodegroup.
Situation 2: `node-a` is moved from `nodegroup-hangzhou` to `nodegroup-beijing`.

The original idea of eviciting pods on node is to handle the above 2 situations, but the situation you mentioned above seems also be included. I need to figure out if there's competation between nodegroup controller and k8s controller manager when deleting nodes.

gy95:
cc @fisherxu

gy95:
cc @fisherxu  PTAL

fisherxu (commented):
/lgtm
/approve

PR #3972, opened 2022-06-23
update deprecate dependency

vincentgoat:

**What type of PR is this?**
/kind cleanup

3 projects in KubeEdges dependency tree were found to be deprecated or unmaintained.
- github.com/golang/protobuf
- github.com/gorilla/mux (see https://github.com/gorilla/mux/issues/659) 
- github.com/paypal/gatt (see https://github.com/paypal/gatt/issues/75). gatt is only used for testing.

2nd and 3rd issues are solved in this PR.
As for now, github.com/golang/protobuf is remain used by github.com/container-storage-interface/spec/lib/go/csi, this might need a deeper analysis on how to solve. /help

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/servers/httpserver/server.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/server.go
@@ -29,8 +29,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/emicklei/go-restful"
 	"github.com/golang-jwt/jwt"
-	"github.com/gorilla/mux"
 	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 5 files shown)


gy95:
Do you mean to remove or replace the two packages? there're still some other codes using the two packages. this patch don't remove these packages totally

vincentgoat:
> Do you mean to remove or replace the two packages? there're still some other codes using the two packages. this patch don't remove these packages totally

Yes, replace totally. Which are the other codes?

gy95:
For example:
https://github.com/kubeedge/kubeedge/blob/f4420cd3a354d8a8f9ef02fd8d195ca2d0eeab3a/edge/test/cloudhub/stub.go#L8
IMHO, if we've already removed dependency package totally, we need run `make verify-vendor` and `make verify-vendor-licenses` to update vendor

cloud/pkg/cloudhub/servers/httpserver/server.go
```diff
@@ -29,8 +29,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/emicklei/go-restful"
 	"github.com/golang-jwt/jwt"
-	"github.com/gorilla/mux"
```
fisherxu, line 33: Is there other part using `github.com/gorilla/mux` in KubeEdge?

fisherxu:
> IMHO, if we've already removed dependency package totally, we need run make verify-vendor and make verify-vendor-licenses to update vendor

Maybe other vender will still use it.. @gy95 @vincentgoat

cloud/pkg/cloudhub/servers/httpserver/server.go
gy95: only edge/test/cloudhub/stub.go
https://github.com/kubeedge/kubeedge/blob/f4420cd3a354d8a8f9ef02fd8d195ca2d0eeab3a/edge/test/cloudhub/stub.go#L8

And github.com/docker/distribution still use it

fisherxu (commented):
/lgtm
/approve

PR #3973, opened 2022-06-23
synccontroller: fix loop

xuxiawei:
**What type of PR is this?**

/kind bug

`checkObjectSync` returns true when the resource does not exist, and the error is nil, so we can skip it here.

Fixes #

```release-note

```


vincentgoat (approved):
lgtm

fisherxu (commented):
/lgtm
/approve

PR #3974, opened 2022-06-23
bump klog to 2.70

Abirdcfly:

**What type of PR is this?**
/kind cleanup

Like https://github.com/kubernetes/kubernetes/pull/110724, after bump klog to 2.70 `klog.Fatal` no longer dumps the backtrace of all goroutines.

Fixes #

But note that k8s was upgraded from 2.60, while we upgraded from 2.8.
We need more testing than just ci.

```release-note
NONE
```


gy95:
@Abirdcfly  please make ci happy :) 
And we've modified all `klog.Fatal` to `klog.Exit` in https://github.com/kubeedge/kubeedge/pull/3256 , and we may need to revert from `klog.Exit` to `klog.Fatal` 
If `klog.Exit`and `klog.Fatal` are the same, we may not need to do this. 
And what's the difference between `klog.Exit`and `klog.Fatal` when we use `klog@2.70`?

Abirdcfly:
> we've modified all klog.Fatal to klog.Exit in https://github.com/kubeedge/kubeedge/pull/3256

Sorry I didn't notice this pr before, I think this approach is also workable.
It's just that we need to make sure that none of the subsequent commits, should not use `klog.Fatalf` and `klog.Fatal`, this is hard....😂

for example:
https://github.com/kubeedge/kubeedge/blob/33281fc8c74faf537bcf0485d188898d57202041/cloud/cmd/controllermanager/app/controllermanager.go#L42-L52

> we may need to revert from klog.Exit to klog.Fatal
I think both ways are ok, revert or not.

gy95:
ok, let's focus on bumping klog to 2.70 in this PR, we can open another PR to do other things :)

gy95:
hi @Abirdcfly  please make ci happy

PR #3976, opened 2022-06-23
Possible type confusions

vincentgoat:

**What type of PR is this?**
/kind cleanup

Several possible type confusions were discovered in the KubeEdge code base. These should be avoided by checking for an error when type casting.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/router/listener/http.go
+++ b/cloud/pkg/router/listener/http.go
@@ -78,7 +78,11 @@ func (rh *RestHandler) RemoveListener(key interface{}) {
 func (rh *RestHandler) matchedPath(uri string) (string, bool) {
 	var candidateRes string
 	rh.handlers.Range(func(key, value interface{}) bool {
-		pathReg := key.(string)
+		pathReg, ok := key.(string)
+		if !ok {
+			klog.Errorf("key type %T error", key)
+			return true
+		}
 		if match := utils.IsMatch(pathReg, uri); match {
 			if candidateRes != "" && utils.RuleContains(pathReg, candidateRes) {
 				return true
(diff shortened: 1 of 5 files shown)


edge/pkg/eventbus/eventbus.go
```diff
@@ -124,12 +124,24 @@ func (eb *eventbus) pubCloudMsgToEdge() {
...
+				continue
+			}
+			topic, ok := message["topic"].(string)
+			if !ok {
+				klog.Errorf("Message topic body type is %T and not string type", message["topic"])
+				continue
+			}
+			payload, err := json.Marshal(&message)
+			if err != nil {
+				klog.Errorf("marshal message %v error: %v", topic, err)
```
gy95, line 142: print topic or message?

cloud/pkg/router/listener/http.go
```diff
@@ -78,7 +78,11 @@ func (rh *RestHandler) RemoveListener(key interface{}) {
 func (rh *RestHandler) matchedPath(uri string) (string, bool) {
 	var candidateRes string
 	rh.handlers.Range(func(key, value interface{}) bool {
-		pathReg := key.(string)
+		pathReg, ok := key.(string)
+		if !ok {
+			klog.Errorf("key type %T error", key)
+			return true
```
gy95, line 84: return false

gy95 (approved):

edge/pkg/eventbus/eventbus.go
vincentgoat: > print topic or message?

The topic is fine, avoid printing any sensitive message.

gy95: ok

cloud/pkg/router/listener/http.go
vincentgoat: The return value should be `true`, according to the Range function.
image

fisherxu (commented):
/approve

PR #3981, opened 2022-06-24
Link root changelog to ./CHANGELOG/README

snstaberah:
modify root changelog links to ./CHANGELOG/README
so we can avoid modify 2 same changelog index files in one project

**What type of PR is this?**

Fixes #

```release-note

```


gy95:
LGTM
Thanks

zc2638:
Thanks for your contribution. 
And LGTM for me.

But I have a question, we already have a README in the CHANGELOG directory, do we still need to keep the CHANGELOG.md in the root directory?
@gy95 @fisherxu

fisherxu:
> But I have a question, we already have a README in the CHANGELOG directory, do we still need to keep the CHANGELOG.md in the root directory?

Good idea, how about making the `CHANGELOG.md` in root link to README in the CHANGELOG directory as in k8s. @snstaberah @zc2638 

Ref:  https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG.md

snstaberah:
> 

i think this is a better way to avoid maintain 2 changelog index file in one project，i have modified the commit，please take a look

zc2638:
Maybe it's better to use file softlink?

fisherxu:
> Maybe it's better to use file softlink?

`softlink` as in k8s will be better :)

fisherxu:
@snstaberah Looks like it's not a softlink now, we need to run `ln -s CHANGELOG/README.md CHANGELOG.md` to create it.

snstaberah:
> @snstaberah Looks like it's not a softlink now, we need to run `ln -s CHANGELOG/README.md CHANGELOG.md` to create it.

done，please take a look again

fisherxu (commented):
/approve

cc @zc2638

PR #3994, opened 2022-06-29
fix complex parameters

vincentgoat:

**What type of PR is this?**
/kind cleanup

1. Functions always returning same value
2. The Quic and WS connections of the Viaduct package contain logic that could be removed altogether. In both cases, the switch statement can be removed. A comment should be left to explain why the functions look like they do. It could be considered whether `processControlMessage` could be removed entirely.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/membership.go
+++ b/edge/pkg/devicetwin/dtmanager/membership.go
@@ -158,7 +158,9 @@ func dealMembershipGet(context *dtcontext.DTContext, resource string, msg interf
 		return errors.New("assertion failed")
 	}
 
-	dealMembershipGetInner(context, contentData)
+	if err := dealMembershipGetInner(context, contentData); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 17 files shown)


edge/pkg/devicetwin/dtmanager/membership.go
```diff
@@ -345,11 +347,13 @@ func dealMembershipGetInner(context *dtcontext.DTContext, payload []byte) error
 	topic := dtcommon.MemETPrefix + context.NodeName + dtcommon.MemETGetResultSuffix
 	klog.Infof("Deal getting membership successful and send the result")
 
-	context.Send("",
+	err = context.Send("",
 		dtcommon.SendToEdge,
 		dtcommon.CommModule,
 		context.BuildModelMessage(modules.BusGroup, "", topic, messagepkg.OperationPublish, result))
-
+	if err != nil {
```
wackxu, line 354: suggest  `return err`

edge/pkg/edged/volume/csi/csi_plugin.go
```diff
@@ -126,7 +124,7 @@ func (h *RegistrationHandler) ValidatePlugin(pluginName string, endpoint string,
 func (h *RegistrationHandler) RegisterPlugin(pluginName string, endpoint string, versions []string) error {
 	klog.Infof(log("Register new plugin with name: %s at endpoint: %s", pluginName, endpoint))
 
-	highestSupportedVersion, err := h.validateVersions("RegisterPlugin", pluginName, endpoint, versions)
```
gy95, line 129: this file is derived from kubernetes, is this need fixing?

gy95 (approved):
/lgtm

fisherxu (commented):
/approve

PR #4000, opened 2022-06-30
fix possible nil pointer of message

vincentgoat:

**What type of PR is this?**
/kind cleanup

a nil value could be passed which could trigger a nil-pointer dereference panic. These do not necessarily impose a risk in terms of security. Even if these do not have security implications now, this might change every time changes are made to the KubeEdge source tree.
Nil-pointers can be exploitable by an attacker if triggerable by an attacker under the right conditions. Naturally, whether a possible nil-pointer dereference is triggerable by an attacker depends largely on how the function is used. In many cases it is never possible to pass nil. As a rule of thumb, we recommend checking pointers for nil at the beginning of the function body. In our analysis of pointer-handling in KubeEdge, we used the Message type as an example to demonstrate an area for improvement.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/handler/messagehandler.go
+++ b/cloud/pkg/cloudhub/handler/messagehandler.go
@@ -116,7 +116,10 @@ func (mh *MessageHandle) HandleServer(container *mux.MessageContainer, writer mu
 		klog.Errorf("Fail to serve node %s, reach node limit", nodeID)
 		return
 	}
-
+	if container.Message == nil {
+		klog.Errorf("Handle a nil message error, node : %s", nodeID)
+		return
+	}
 	klog.V(4).Infof("[cloudhub/HandlerServer] get msg from edge(%v): %+v", nodeID, container.Message)
 	if container.Message.GetOperation() == model.OpKeepalive {
 		klog.V(4).Infof("Keepalive message received from node: %s", nodeID)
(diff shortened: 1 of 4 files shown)


cloud/pkg/cloudhub/handler/messagehandler.go
```diff
@@ -388,8 +391,15 @@ func (mh *MessageHandle) ListMessageWriteLoop(info *model.HubInfo, stopServe cha
 			klog.Errorf("nodeListStore for node %s doesn't exist", info.NodeID)
 			continue
 		}
-		msg := obj.(*beehiveModel.Message)
-
+		msg, ok := obj.(*beehiveModel.Message)
+		if !ok {
+			klog.Errorf("list message is invalid for node: %s", info.NodeID)
```
wackxu, line 396: print the message type

staging/src/github.com/kubeedge/viaduct/pkg/mux/pattern.go
```diff
@@ -39,10 +39,18 @@ func (pattern *MessagePattern) Op(operation string) *MessagePattern {
...
+	if message == nil {
+		klog.Errorf("nil message can't be matched, operation: %s", pattern.operation)
+		return false
+	}
 	return strings.Compare(pattern.operation, message.GetOperation()) == 0 ||
 		strings.Compare(pattern.operation, "*") == 0
 }
 
 func (pattern *MessagePattern) Match(message *model.Message) bool {
+	if message == nil {
```
wackxu, line 51: I think  it maybe over programming defense

fisherxu (commented):
/lgtm
/approve

Issue #4001, opened 2022-06-30
Set InsecureSkipVerify to false in http request


vincentgoat:
**What would you like to be added/modified**:
Set InsecureSkipVerify to false in http request

**Why is this needed**:
“InsecureSkipVerify controls whether a client verifies the server's certificate chain and host name. If InsecureSkipVerify is true, crypto/tls accepts any certificate presented by the server and any host name in that certificate. In this mode, TLS is susceptible to machine-in-the-middle attacks unless custom verification is used. This should be used only for testing or in combination with VerifyConnection or VerifyPeerCertificate.”
Described in https://pkg.go.dev/crypto/tls 

InsecureSkipVery: true was used in the following places:

https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/eventbus/common/util/common.go#L86
https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgestream/edgestream.go#L87
https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/common/http/http.go#L38
https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/common/http/http.go#L60
https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/edgehub/clients/quicclient/quicclient.go#L62
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/router/utils/http/http.go#L38
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/router/utils/http/http.go#L45

falser101:
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/router/utils/http/http.go#L45

PR #4003, opened 2022-07-01
feat:Set InsecureSkipVerify to false in http request

falser101:

**What type of PR is this?**
/kind feature

Set InsecureSkipVerify to false in http request

Fixes # #4001

--> NONE


gy95:
maybe this is not correct, we need to pass in ca and certs

falser101:
“InsecureSkipVerify controls whether a client verifies the server's certificate chain and host name. If InsecureSkipVerify is true, crypto/tls accepts any certificate presented by the server and any host name in that certificate. In this mode, TLS is susceptible to machine-in-the-middle attacks unless custom verification is used. This should be used only for testing or in combination with VerifyConnection or VerifyPeerCertificate.”
Described in https://pkg.go.dev/crypto/tls

PR #4011, opened 2022-07-05
cleanup: Error string fmt and handle error

falser101:
Error string fmt and handle error

**What type of PR is this?**
/kind cleanup
 Error string fmt and handle error
Fixes #

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/admissioncontroller/admit_rule.go
+++ b/cloud/pkg/admissioncontroller/admit_rule.go
@@ -42,7 +42,7 @@ func admitRule(review admissionv1beta1.AdmissionReview) *admissionv1beta1.Admiss
 		return &reviewResponse
 	default:
 		err := fmt.Errorf("unsupported webhook operation %v", review.Request.Operation)
-		klog.Errorf("unsupported webhook operation %v", review.Request.Operation)
+		klog.Errorf("Unsupported webhook operation %v", review.Request.Operation)
 		return toAdmissionResponse(err)
 	}
 }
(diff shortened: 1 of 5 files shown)


cloud/pkg/admissioncontroller/admit_rule.go
```diff
@@ -41,8 +41,8 @@ func admitRule(review admissionv1beta1.AdmissionReview) *admissionv1beta1.Admiss
 		reviewResponse.Allowed = true
 		return &reviewResponse
 	default:
-		err := fmt.Errorf("Unsupported webhook operation %v", review.Request.Operation)
-		klog.Errorf("Unsupported webhook operation %v", review.Request.Operation)
+		err := fmt.Errorf("unsupported webhook operation %v", review.Request.Operation)
+		klog.Errorf("unsupported webhook operation %v", review.Request.Operation)
```
Shelley-BaoYue, line 45: I think it's better to keep the first letter capitalized in klog.

fisherxu (commented):
/lgtm
/approve

cloud/pkg/admissioncontroller/admit_rule.go
zc2638: I agree with you.
But it looks like @fisherxu doesn't think so.

zc2638: ref: https://github.com/kubeedge/kubeedge/pull/4011/files#diff-364f053b5a428af19fb62f64c0c1675599b8d5ddc316c75bc71e25fc6340e54aR36

PR #4017, opened 2022-07-06
use register-gen to generate register.go

Congrool:
**What type of PR is this?**
/kind feature

Currently, we maintain the register.go of each CRD ourself. Now, we have:
1. [register.go of apps](https://github.com/kubeedge/kubeedge/blob/master/pkg/apis/apps/v1alpha1/register.go)
2. [register.go of devices](https://github.com/kubeedge/kubeedge/blob/master/pkg/apis/devices/v1alpha2/register.go)
3. [register.go of reliablesyncs](https://github.com/kubeedge/kubeedge/blob/master/pkg/apis/reliablesyncs/v1alpha1/register.go)
4. [register.go of rules](https://github.com/kubeedge/kubeedge/blob/master/pkg/apis/rules/v1/register.go)

Manually writing these files has the possibility of introducing bugs by mistake. And it's not convenience for us to keep the pace with kubernetes when code-generator is updated in the [go.mod](https://github.com/kubeedge/kubeedge/blob/master/go.mod#L53), noticing that some register.go were written long ago and never updated when we bumped kubernetes version.

With this pr, we can use `hack/update-codegen.sh` to automatically generate register.go when we are going to develop new CRDs, and can check if existing register.go files are out of date in [Github Action](https://github.com/kubeedge/kubeedge/blob/master/.github/workflows/main.yaml#L41-L43) like what we do for deepcopy files.

Fixes #

```release-note

```


pkg/apis/devices/v1alpha2/zz_generated.register.go
```diff
@@ -0,0 +1,69 @@
...
+package v1alpha2
+
+import (
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GroupName specifies the group name used to register the objects.
+const GroupName = "devices.kubeedge.io"
```
gy95, line 28: what's GroupName generated from？how `hack/update-codegen.sh` construct this groupname with the suffix `.kubeedge.io`

Congrool: We set it in the comment of [doc.go](https://github.com/kubeedge/kubeedge/pull/4017/files#diff-21f741b214c9d2a115d1ba0a8aba09960ad3c6d6fd4f7ee261a87ffecee02d6bR22). It's originally at [pkg/apis/devices/v1alpha2/register.go](https://github.com/kubeedge/kubeedge/pull/4017/files#diff-2a5d882eaa81f850553891fa2e14085876d105e757d64322c0f491fd234a03b9).

gy95: that means we need to manually write two files: `doc.go` and `types.go` if we want to develop a new CRD?
And does every `doc.go` need to contain the below comments? I find some CRD `doc.go` include only part of the below content.
```
// Package v1alpha2 contains API Schema definitions for the devices v1alpha2 API group
// +k8s:openapi-gen=true
// +k8s:deepcopy-gen=package,register
// +k8s:conversion-gen=kubeedge/cloud/pkg/apis/devices
// +k8s:defaulter-gen=TypeMeta
// +groupName=devices.kubeedge.io
```

Congrool: Yes, `doc.go` and `types.go` are all we need to write.
In my mind, not all of these are necessary, and the following two are needed.
1. `+k8s:deepcopy-gen` is used to generate zz_generated.deepcopy.go
2. `+groupName` indicates the group name of this CRD

Something like:  
1. `+k8s:openapi-gen` is used by `openapi-gen`.
2. `+k8s:conversion-gen` is used by `conversion-gen` to generate conversion funcs.

I've not found any usage here, and just respect the old code.

gy95: ok, thanks for your explanation :)

wackxu:
LGTM

Congrool:
rebase #4125, PTAL @gy95 @wackxu @fisherxu

Congrool:
friendly ping @fisherxu , any other work I need to do?

PR #3624, opened 2022-02-22
update k8s to 1.22

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:
Fixes  https://github.com/kubeedge/kubeedge/issues/3607
Update k8s to v1.22

PR #4022, opened 2022-07-09
Fix admission not work: bump admissionregistration.k8s.io/v1beta1 to admissionregistration.k8s.io/v1

gy95:
**What type of PR is this?**
/kind bug

by k8s official docs, 
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#webhook-resources-v122
```
The admissionregistration.k8s.io/v1beta1 API version of MutatingWebhookConfiguration and ValidatingWebhookConfiguration is no longer served as of v1.22.
```
And now kubeedge depends on k8s v1.22.6(https://github.com/kubeedge/kubeedge/pull/3624), so we should upgrade admissionregistration.k8s.io/v1beta1 to admissionregistration.k8s.io/v1

when i use admission to register webhook for CRDs, there're some errors:
`failed to register the webhook with error: the server could not find the requested resource`: So I replace v1beta1 admission with v1 admission.

And for webhook, we must set both `SideEffects` and `AdmissionReviewVersions` fields, or it will report errors.

And we must use `responseAdmissionReview.SetGroupVersionKind(admissionv1.SchemeGroupVersion.WithKind("AdmissionReview"))` to set AdmissionReview GVK, or it will also report error 
```
Failed calling webhook, failing open validateupgrade.kubeedge.io: failed calling webhook "validateupgrade.kubeedge.io": received invalid webhook response: expected webhook response of admission.k8s.io/v1, Kind=AdmissionReview, got /, Kind=
```

And do some clean up work

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/admissioncontroller/admission.go
+++ b/cloud/pkg/admissioncontroller/admission.go
@@ -8,8 +8,8 @@ import (
 	"net/http"
 	"os"
 
-	admissionv1beta1 "k8s.io/api/admission/v1beta1"
-	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	admissionv1 "k8s.io/api/admission/v1"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -31,8 +31,9 @@ const (
 	ValidateDeviceModelWebhookName  = "validatedevicemodel.kubeedge.io"
 	ValidateRuleWebhookName         = "validatedrule.kubeedge.io"
 	ValidateRuleEndpointWebhookName = "validatedruleendpoint.kubeedge.io"
-	OfflineMigrationConfigName      = "mutate-offlinemigration"
-	OfflineMigrationWebhookName     = "mutateofflinemigration.kubeedge.io"
+
+	OfflineMigrationConfigName  = "mutate-offlinemigration"
+	OfflineMigrationWebhookName = "mutateofflinemigration.kubeedge.io"
 
 	AutonomyLabel = "app-offline.kubeedge.io=autonomy"
 )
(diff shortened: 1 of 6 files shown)


cloud/pkg/admissioncontroller/admission.go
```diff
@@ -144,133 +149,171 @@ func configTLS(opt *options.AdmissionOptions, restConfig *restclient.Config) (*t
...
 func (ac *AdmissionController) registerWebhooks(opt *options.AdmissionOptions, cabundle []byte) error {
-	ignorePolicy := admissionregistrationv1beta1.Ignore
-	failPolicy := admissionregistrationv1beta1.Fail
-	deviceModelCRDWebhook := admissionregistrationv1beta1.ValidatingWebhookConfiguration{
+	ignorePolicy := admissionregistrationv1.Ignore
+	failPolicy := admissionregistrationv1.Fail
+	noneSideEffect := admissionregistrationv1.SideEffectClassNone
+
+	// DeviceModel webhook
+	deviceModelCRDWebhook := admissionregistrationv1.ValidatingWebhookConfiguration{
```
wackxu, line 157: Why split one ValidatingWebhookConfiguration to multi ValidatingWebhookConfiguration

gy95: you mean keep only one `ValidatingWebhookConfiguration` and contain multi `MutatingWebhook` in that one `ValidatingWebhookConfiguration` ?

gy95: fixed, i misunderstood the config name, it's `validate-devicemodel`, i thought one config for one CRD. change the name to `validate-configure`

wackxu: yes，the name is not good, but if we just update the name,  for user who upgrade the webhook, there maybe two ValidatingWebhookConfiguration, the old one and the new one, it may execute twice for one crd

gy95: oh, will revert it and add a TODO to make it more compatible

gy95: Fixed, please help review again

gy95:
cc @fisherxu

gy95:
cc @vincentgoat @Shelley-BaoYue @wackxu  PTAL :)

vincentgoat (commented):
Haven't verified the whole function of Webhook, but the current changes look fine.

gy95:
updated description: add about when k8s not support admissionregistration.k8s.io/v1beta1

fisherxu (commented):
/lgtm
/approve

PR #4046, opened 2022-07-12
add license for cloudcore

xuxiawei:
Signed-off-by: xuxiawei <xiawei.xu@daocloud.io>

**What type of PR is this?**

Fixes #

```release-note

```


cloud/cmd/cloudcore/app/server.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2019 The KubeEdge Authors.
```
chendave, line 2: consider that this file is firstly added in 2019, this should be fine.

PR #4060, opened 2022-07-22
add readyz handler to admission

gy95:
**What type of PR is this?**

add readyz handler to admission

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/admissioncontroller/admission.go
+++ b/cloud/pkg/admissioncontroller/admission.go
@@ -19,6 +19,7 @@ import (
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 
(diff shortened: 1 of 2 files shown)


cloud/pkg/admissioncontroller/admission.go
```diff
@@ -98,6 +99,8 @@ func Run(opt *options.AdmissionOptions) error {
 	http.HandleFunc("/rules", serveRule)
 	http.HandleFunc("/ruleendpoints", serveRuleEndpoint)
 	http.HandleFunc("/offlinemigration", serveOfflineMigration)
+	http.Handle("/readyz/", http.StripPrefix("/readyz/", &healthz.Handler{}))
```
wackxu, line 102: why  we  need StripPrefix?

gy95: nothing, just ref: https://github.com/karmada-io/karmada/blob/ed9b838056ee45720abf51b2d603548e6ce922bf/cmd/webhook/app/webhook.go#L125

PR #4063, opened 2022-07-23
support auto gen certs for admission

gy95:
**What type of PR is this?**

generate ca and certs in golang code, so that we wouldn't generate certs manually using script https://github.com/kubeedge/kubeedge/blob/master/build/admission/gen-admission-secret.sh before we deploy `admission` components by following the doc https://github.com/kubeedge/kubeedge/blob/master/build/admission/README.md
this is helpful for us to deploy `admission` in container  mode using `keadm`, omit the step of manually generating a certificate
and this operation can be used by other components. Is this OK?

Fixes #

```release-note

```


cloud/pkg/admissioncontroller/admission.go
```diff
@@ -84,7 +99,19 @@ func Run(opt *options.AdmissionOptions) error {
 	controller.Client = cli
 	controller.CrdClient = vcli
 
-	caBundle, err := os.ReadFile(opt.CaCertFile)
+	// generate ca certs
+	if err := genCerts(opt); err != nil {
```
wackxu, line 103: when admission have multi replicas, there has some problems, the latter replicas will gen certs again and will cover the former replicas, so two replicas has different cabunlde and kube-apiserver will has error when access the admisson

gy95: yes, so store the ca and certs in secret, and check whether secret exsit or not, if exist, read from it, or generate it, 
just like what shell script `build/admission/gen-admission-secret.sh` did

gy95:
cc @vincentgoat @wackxu @Shelley-BaoYue  PTAL

cloud/pkg/admissioncontroller/admission.go
```diff
@@ -115,10 +136,92 @@ func Run(opt *options.AdmissionOptions) error {
...
+				Namespace: opt.AdmissionServiceNamespace,
+			},
+			Data: map[string][]byte{
+				"tls.key": keyData,
+				"tls.crt": certData,
+				"ca.crt":  caData,
+			},
+		}
+		_, err = controller.Client.CoreV1().Secrets(opt.AdmissionServiceNamespace).Create(context.TODO(), newSecret, metav1.CreateOptions{})
+		if err != nil {
```
wackxu, line 184: ignore already  exist err when multi replicas generate certificates

cloud/pkg/admissioncontroller/admission.go
```diff
@@ -115,10 +136,92 @@ func Run(opt *options.AdmissionOptions) error {
...
+	if apierrors.IsNotFound(err) {
+		// generate certificate
+		klog.Infof("Generate ca and certs...")
+		if err := genCerts(opt); err != nil {
+			return nil, fmt.Errorf("failed to generate certs: %v", err)
+		}
+
+		keyData, _ := os.ReadFile(filepath.Join(admissionPkiPath, fmt.Sprintf("%s.key", admissionPkiName)))
+		certData, _ := os.ReadFile(filepath.Join(admissionPkiPath, fmt.Sprintf("%s.crt", admissionPkiName)))
+		caData, _ := os.ReadFile(filepath.Join(caPkiPath, fmt.Sprintf("%s.crt", caPkiName)))
```
wackxu, line 169: process these err

cloud/pkg/admissioncontroller/admission.go
wackxu: lgtm

PR #4092, opened 2022-08-04
fix unhandled transaction error

elpsyr:

**What type of PR is this?**
/kind cleanup

Fix some unhandled transaction error.
If the error caused by rollback is not captured, it will be commit successfully.

Fixes #

```release-note

```


elpsyr:
the "err" should be replace with "errRollback" in klog

PR #4097, opened 2022-08-05
remove nodeLocks from MessageHandle

wackxu:

**What type of PR is this?**

/kind cleanup

for websocket connection, we already have locks to make sure that there is  only one goroutine to write to the connection on the same time, so we do not need the lock in the messageHandle, and also add lock for the quic connection

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/viaduct/pkg/conn/quic.go
+++ b/staging/src/github.com/kubeedge/viaduct/pkg/conn/quic.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/lucas-clemente/quic-go"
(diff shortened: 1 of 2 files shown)


staging/src/github.com/kubeedge/viaduct/pkg/conn/quic.go
```diff
@@ -289,6 +293,9 @@ func (conn *QuicConnection) WriteMessageAsync(msg *model.Message) error {
 	}
 	defer conn.streamManager.ReleaseStream(api.UseTypeMessage, stream)
 
+	conn.locker.Lock()
+	defer conn.locker.Unlock()
```
gy95, line 297: move to the last line? line302?

wackxu: done

gy95: LGTM 👍

wackxu:
@zc2638 Could you take a look at this, thanks

fisherxu (commented):
/approve

PR #4098, opened 2022-08-05
fix unhandled transaction error

elpsyr:

**What type of PR is this?**
/kind cleanup

Fix some unhandled transaction error.
If the error caused by rollback is not captured, it will be commit successfully.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtclient/device_db.go
+++ b/edge/pkg/devicetwin/dtclient/device_db.go
@@ -6,7 +6,7 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/common/dbm"
 )
 
-//Device the struct of device
+// Device the struct of device
 type Device struct {
 	ID          string `orm:"column(id); size(64); pk"`
 	Name        string `orm:"column(name); null; type(text)"`
@@ -15,14 +15,14 @@ type Device struct {
 	LastOnline  string `orm:"column(last_online); null; type(text)"`
 }
 
-//SaveDevice save device
+// SaveDevice save device
 func SaveDevice(doc *Device) error {
 	num, err := dbm.DBAccess.Insert(doc)
 	klog.V(4).Infof("Insert affected Num: %d, %v", num, err)
 	return err
 }
 
-//DeleteDeviceByID delete device by id
+// DeleteDeviceByID delete device by id
 func DeleteDeviceByID(id string) error {
 	num, err := dbm.DBAccess.QueryTable(DeviceTableName).Filter("id", id).Delete()
 	if err != nil {
@@ -67,13 +67,13 @@ func QueryDeviceAll() (*[]Device, error) {
 	return devices, nil
 }
 
-//DeviceUpdate the struct for updating device
+// DeviceUpdate the struct for updating device
 type DeviceUpdate struct {
 	DeviceID string
 	Cols     map[string]interface{}
 }
 
-//UpdateDeviceMulti update device  multi
+// UpdateDeviceMulti update device  multi
 func UpdateDeviceMulti(updates []DeviceUpdate) error {
 	var err error
 	for _, update := range updates {
@@ -85,7 +85,7 @@ func UpdateDeviceMulti(updates []DeviceUpdate) error {
 	return nil
 }
 
-//AddDeviceTrans the transaction of add device
+//AddDeviceTrans add device with transaction
 func AddDeviceTrans(adds []Device, addAttrs []DeviceAttr, addTwins []DeviceTwin) error {
 	var err error
 	obm := dbm.DBAccess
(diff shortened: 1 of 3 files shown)


edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -89,58 +89,98 @@ func UpdateDeviceMulti(updates []DeviceUpdate) error {
...
 	}
 
 	for _, attr := range addAttrs {
 		err = SaveDeviceAttr(&attr)
 		if err != nil {
-			obm.Rollback()
+			errRollback := obm.Rollback()
+			if errRollback != nil {
+				klog.Errorf("transaction rollback failed: %v", errRollback)
+				return errRollback
```
wackxu, line 117: I think we should return the SaveDeviceAttr err,  it is the main err that why AddDeviceTrans exit, so  ignore the rollback err is better

edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -89,58 +89,98 @@ func UpdateDeviceMulti(updates []DeviceUpdate) error {
...
+			}
 			return err
 		}
 	}
-	obm.Commit()
+	err = obm.Commit()
```
wackxu, line 180: suggest `return obm.Commit()`

edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -89,58 +89,98 @@ func UpdateDeviceMulti(updates []DeviceUpdate) error {
...
+			}
 			return err
 		}
 	}
 
 	for _, twin := range addTwins {
 		err = SaveDeviceTwin(&twin)
 		if err != nil {
-			obm.Rollback()
+			errRollback := obm.Rollback()
```
wackxu, line 126: ditto

edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -89,58 +89,98 @@ func UpdateDeviceMulti(updates []DeviceUpdate) error {
...
+	if err != nil {
+		klog.Errorf("begin transaction failed: %v", err)
+		return err
+	}
+	for _, deleteID := range deletes {
+		err = DeleteDeviceByID(deleteID)
 		if err != nil {
-			obm.Rollback()
+			errRollback := obm.Rollback()
+			if errRollback != nil {
```
wackxu, line 155: ditto

edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -89,58 +89,98 @@ func UpdateDeviceMulti(updates []DeviceUpdate) error {
...
+				return errRollback
+			}
 			return err
 		}
-		err = DeleteDeviceAttrByDeviceID(delete)
+		err = DeleteDeviceAttrByDeviceID(deleteID)
 		if err != nil {
-			obm.Rollback()
+			errRollback := obm.Rollback()
+			if errRollback != nil {
```
wackxu, line 164: ditto

edge/pkg/devicetwin/dtclient/deviceattr_db.go
```diff
@@ -100,30 +100,50 @@ func UpdateDeviceAttrMulti(updates []DeviceAttrUpdate) error {
...
+			if errRollback != nil {
+				klog.Errorf("transaction rollback failed: %v", errRollback)
+				return errRollback
+			}
 			return err
 		}
 	}
-	obm.Commit()
+	err = obm.Commit()
+	if err != nil {
```
wackxu, line 144: ditto

edge/pkg/devicetwin/dtclient/device_db.go
elpsyr: you are right，**no matter whether the rollback is successful or not**, the transaction will not be committed, so there is no need to handle the rollback error .
i will close this pr

elpsyr:
no matter whether the rollback is successful or not, the transaction will not be committed, so there is no need to handle the rollback error .

wackxu:
for begin or commit err, I think we should hanle it,  so other code lgtm

elpsyr:
yeah,i will reopen it later

elpsyr:
@wackxu I handle the error of transaction begin&commit,and leave some commit about rollback like this:
```
// we do not handle the error of rollback because:
// 1. no matter whether the rollback is successful or not,
// the transaction will not be committed.
// 2. err of SaveDevice is the main err that why AddDeviceTrans exit,
// so we should the SaveDevice err.
```
thanks  for your patient : )

elpsyr:
btw ,sometimes i add the `Signed-off-by: Author Name <authoremail@example.com> ` in commit message,but i still fail with DCO check. 
Is there anything else to pay attention to?

PR #4102, opened 2022-08-08
do not return err when PodSandbox not exist

elpsyr:

**What type of PR is this?**

1. we have no need to return err when PodSandbox not exist , just logs it. Otherwise, edged will fail to start.
（kubelet do not return err when PodSandbox not exist  in kubernetes v1.24）
4. we need to handle err of cpuManager.Start() 

Fixes #

```release-note

```


fisherxu (commented):
/lgtm
/approve

Issue #4090, opened 2022-08-03
Pod with the same name stucks on Pending status after recreation


bluven:
**What happened**:

Sometimes I created pod with the same name, but the pod will stuck on Pending phase. After executing `docker ps`, I found that there was even no containers related to the pod.

**What you expected to happen**:

Pod data should be synchronized to edge nodes after recreation.

**How to reproduce it (as minimally and precisely as possible)**:

Create a pod, delete it and then recreate it again. I use a little program to do that:
```
package main

import (
	"context"
	corev1 "k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/api/errors"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"log"
	"sigs.k8s.io/controller-runtime/pkg/client"
	"sigs.k8s.io/controller-runtime/pkg/client/config"
	"time"
)

var namespace = "default"

func main() {
	cfg, err := config.GetConfig()
	if err != nil {
		panic(err)
	}

	cli, err := client.New(cfg, client.Options{})
	if err != nil {
		panic(err)
	}

	pod := nginxPod2()

	log.Println("creating pod...")
	if err = cli.Create(context.Background(), &pod); err != nil {
		panic(err)
	}

	for {
		key := client.ObjectKey{Name: pod.Name, Namespace: pod.Namespace}
		if err = cli.Get(context.Background(), key, &pod); err != nil {
			panic(err)
		}
		log.Printf("nginx pod phase: %s", pod.Status.Phase)

		if pod.Status.Phase == corev1.PodRunning {
			break
		}

		time.Sleep(time.Second)
	}
	log.Println("nginx is running")

	log.Println("deleting nginx...")
	if err = cli.Delete(context.Background(), &pod); err != nil {
		panic(err)
	}

	for {
		key := client.ObjectKey{Name: pod.Name, Namespace: pod.Namespace}
		err = cli.Get(context.Background(), key, &pod)
		if err == nil {
			continue
		}

		if errors.IsNotFound(err) {
			break
		}

		panic(err)
	}

	log.Println("recreating pod...")
        # increase delay may avoid this problem
	time.Sleep(20 * time.Millisecond)
	pod = nginxPod2()
	if err = cli.Create(context.Background(), &pod); err != nil {
		panic(err)
	}

	for {
		key := client.ObjectKey{Name: pod.Name, Namespace: pod.Namespace}
[...]

PR #4116, opened 2022-08-15
fix delete and the recreate statefulset pod is not running bug

wackxu:

**What type of PR is this?**

/kind bug

Fixes 
https://github.com/kubeedge/kubeedge/issues/4090
https://github.com/kubeedge/kubeedge/issues/4055

Now, In the message key func, that resources with the same name in the same namespace have the same key. When a pod with the same name is deleted and then quickly created, The deletion event of the old pod will arrive first, and it will be stored in the message queue and store. Then the new pod creation event will arrive, because the message is sent asynchronously, So there is a possibility that the deletion event of the old pod is still in the store, but the creation event of the new pod has arrived.  For now we do not compare the resource UID when compare with the message in the store, and the store message is a delete message so the new resouce message can not add to the queue and lead to the resource can not send to the edge node. This usually happens when a statefulSet deletes a pod and new pod are quickly created.

In this PR, we changed the message key func to use the resouce UID, In the kubernetes the UID is globally unique，So We can avoid the problem of quickly deleting and recreate pods with the same name.

I also add e2e for the statefulset lifecycle in the edge node

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -167,10 +167,7 @@ func getMsgKey(obj interface{}) (string, error) {
 	msg := obj.(*beehiveModel.Message)
 
 	if msg.GetGroup() == edgeconst.GroupResource {
-		resourceType, _ := messagelayer.GetResourceType(*msg)
-		resourceNamespace, _ := messagelayer.GetNamespace(*msg)
-		resourceName, _ := messagelayer.GetResourceName(*msg)
-		return strings.Join([]string{resourceType, resourceNamespace, resourceName}, "/"), nil
+		return GetMessageUID(*msg)
 	}
 
 	return "", fmt.Errorf("failed to get message key")
--- a/tests/e2e/deployment/deployment_test.go
+++ b/tests/e2e/deployment/deployment_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package deployment
 
 import (
+	"context"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -24,6 +26,10 @@ import (
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
 
 	"github.com/kubeedge/kubeedge/tests/e2e/constants"
 	. "github.com/kubeedge/kubeedge/tests/e2e/testsuite"
@@ -37,6 +43,13 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
 	var UID string
 	var testTimer *utils.TestTimer
 	var testSpecReport SpecReport
+
+	var clientSet clientset.Interface
+
+	BeforeEach(func() {
+		clientSet = utils.NewKubeClient(ctx.Cfg.KubeConfigPath)
+	})
+
 	Context("Test application deployment and delete deployment using deployment spec", func() {
 		BeforeEach(func() {
 			// Get current test SpecReport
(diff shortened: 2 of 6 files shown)


tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,105 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+		})
+
+		AfterEach(func() {
+			// End test timer
+			testTimer.End()
+			// Print result
+			testTimer.PrintResult()
+
+			By(fmt.Sprintf("get StatefulSet %s", UID))
+			deployment, err := utils.GetStatefulSet(clientSet, metav1.NamespaceDefault, UID)
```
gy95, line 226: s/deployment/statefulset
and the below codes also need fixing

tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,105 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+
+			utils.PrintTestcaseNameandStatus()
+		})
+
+		It("Basic StatefulSet test", func() {
+			replica := int32(2)
+			// Generate the random string and assign as a UID
+			UID = "edge-statefulset-" + utils.GetRandomString(5)
+
+			By(fmt.Sprintf("create deployment %s", UID))
```
gy95, line 250: s/deployment/statefulset
also below need fixing

tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,105 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+				Expect(err).To(BeNil())
+
+				By(fmt.Sprintf("wait for pod %s running", fmt.Sprintf("%s-1", UID)))
+				err = wait.PollImmediate(5*time.Second, 120*time.Second, func() (bool, error) {
+					pod, err := clientSet.CoreV1().Pods("default").Get(context.TODO(), deletePodName, metav1.GetOptions{})
+					if err != nil {
+						return false, err
+					}
+					if pod.Status.Phase == corev1.PodRunning {
+						return true, nil
```
gy95, line 303: return false, nil?

tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,105 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+					}
+					return false, nil
```
gy95, line 305: return true, nil

tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,105 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+			By(fmt.Sprintf("wait for pod of deployment %s running", UID))
+			utils.WaitforPodsRunning(ctx.Cfg.KubeConfigPath, *podList, 240*time.Second)
+
+			deletePodName := fmt.Sprintf("%s-1", UID)
+			for i := 0; i < 5; i++ {
+				By(fmt.Sprintf("delete pod %s", deletePodName))
+				err = utils.DeletePod(clientSet, deletePodName, "default")
+				Expect(err).To(BeNil())
+
+				By(fmt.Sprintf("wait for pod %s running", fmt.Sprintf("%s-1", UID)))
```
gy95, line 296: wait for pod %s stopped

tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,105 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+			labelSelector := labels.SelectorFromSet(map[string]string{"app": UID})
+			podList, err := utils.ListPods(clientSet, corev1.NamespaceDefault, labelSelector, nil)
+			Expect(err).To(BeNil())
+			Expect(podList).NotTo(BeNil())
+
+			By(fmt.Sprintf("wait for pod of deployment %s running", UID))
+			utils.WaitforPodsRunning(ctx.Cfg.KubeConfigPath, *podList, 240*time.Second)
+
+			deletePodName := fmt.Sprintf("%s-1", UID)
+			for i := 0; i < 5; i++ {
```
gy95, line 291: is this right? create statefulset only once, but delete pod 5 times?

tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,105 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+			By(fmt.Sprintf("create deployment %s", UID))
+			d := utils.NewTestStatefulSet(UID, ctx.Cfg.AppImageURL[1], replica)
+			_, err := utils.CreateStatefulSet(clientSet, d)
+			Expect(err).To(BeNil())
+
+			By(fmt.Sprintf("get pod for deployment %s", UID))
+			labelSelector := labels.SelectorFromSet(map[string]string{"app": UID})
+			podList, err := utils.ListPods(clientSet, corev1.NamespaceDefault, labelSelector, nil)
+			Expect(err).To(BeNil())
+			Expect(podList).NotTo(BeNil())
```
gy95, line 285: Expect(len(podList.Items)).ShouldNot(Equal(0))

vincentgoat (commented):
lgtm

tests/e2e/deployment/deployment_test.go
wackxu: if phase is not PodRunning, we should return false to wait to become running

tests/e2e/deployment/deployment_test.go
wackxu: for statefulet, we delete the pod, the kube-controller will create it immedally, so I wait the pod running.
Without the fix, if we delete the statefulset pod, the new pod will be pending and will not run again

tests/e2e/deployment/deployment_test.go
gy95: oh, i misunderstood, create 1 statefulset and delete pod 5 times to check if statefulset create pod automatically

tests/e2e/deployment/deployment_test.go
wackxu: update to `wait for pod %s running again`

wackxu: I want the delete pod run again

tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,103 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+		})
+
+		AfterEach(func() {
+			// End test timer
+			testTimer.End()
+			// Print result
+			testTimer.PrintResult()
+
+			By(fmt.Sprintf("get StatefulSet %s", UID))
+			deployment, err := utils.GetStatefulSet(clientSet, metav1.NamespaceDefault, UID)
```
gy95, line 226: deployment->statefulset

tests/e2e/deployment/deployment_test.go
```diff
@@ -194,4 +207,103 @@ var _ = Describe("Application deployment test in E2E scenario", func() {
...
+
+			By(fmt.Sprintf("wait for pod of StatefulSet %s disappear", UID))
+			err = utils.WaitForPodsToDisappear(clientSet, metav1.NamespaceDefault, labelSelector, constants.Interval, constants.Timeout)
+			Expect(err).To(BeNil())
+
+			utils.PrintTestcaseNameandStatus()
+		})
+
+		It("Basic StatefulSet test", func() {
+			replica := int32(2)
```
gy95, line 246: it this replica designed?

gy95 (approved):

tests/e2e/deployment/deployment_test.go
wackxu: yes

wackxu: user always have multi replicas when use statefulset

gy95: /lgtm

tests/e2e/utils/pod.go
```diff
@@ -35,6 +39,42 @@ const (
...
+	return wait.PollImmediate(interval, timeout, func() (bool, error) {
+		Infof("Waiting for pod %s to disappear", label.String())
+		options := metav1.ListOptions{LabelSelector: label.String()}
+		pods, err := c.CoreV1().Pods(ns).List(context.TODO(), options)
+		if err != nil {
+			return false, err
+		}
+
+		if pods != nil && len(pods.Items) == 0 {
+			Infof("Pod %s no longer exists", label.String())
```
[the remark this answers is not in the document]
wackxu: done

wackxu:
@zc2638 Could you take a look at this, thanks

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -167,10 +167,7 @@ func getMsgKey(obj interface{}) (string, error) {
 	msg := obj.(*beehiveModel.Message)
 
 	if msg.GetGroup() == edgeconst.GroupResource {
-		resourceType, _ := messagelayer.GetResourceType(*msg)
-		resourceNamespace, _ := messagelayer.GetNamespace(*msg)
-		resourceName, _ := messagelayer.GetResourceName(*msg)
-		return strings.Join([]string{resourceType, resourceNamespace, resourceName}, "/"), nil
+		return GetMessageUID(*msg)
```
zc2638, line 170: Will the `getMsgKey` method involve the parsing of the upper-level method, if it is changed to UID

wackxu: No, we only have two place use the key func, one is when we add the message to the queue, the another is that we pass the  keyFunc `getMsgKey` to the messageStore, so the internal change will not affect  upper-level method

zc2638: thanks for your explanation.

fisherxu (commented):
/approve

PR #4119, opened 2022-08-16
bugfix sqlite 'transaction has already been committed or rolled back' 

### LINKED ISSUE #3978: meta_v2 info sync bug on edge side
809d6198c5b5d2da3454a81d02eb5e7

when the cloud frequently update the meta_v2 info on edge side , there will be a sqlite transcation error @RyanZhaoXB

Shelley-BaoYue:

**What type of PR is this?**
/kind bug

DB operation in edgecore devicetwin  use a global variables `DBAccess orm.Ormer` to begin a transaction. When this transaction commit or rollback in devicetwin operation, other db operation which use the same variables. `orm.Ormer` will fail with the err `transaction has already been committed or rolled back`.  

In this PR, we use a new `Ormer` to begin a transaction in devicetwin, avoid  affecting other db operation.

Fixes #3978

### DIFF
--- a/edge/pkg/common/dbm/db.go
+++ b/edge/pkg/common/dbm/db.go
@@ -37,3 +37,19 @@ func InitDBConfig(driverName, dbName, dataSource string) {
 		}
 	})
 }
+
+type newOrmerFunc func() orm.Ormer
+
+var DefaultOrmFunc newOrmerFunc = newOrmer
+
+func newOrmer() orm.Ormer {
+	return orm.NewOrm()
+}
+
+// RollbackTransaction rollback transaction and log err if rollback fail
+func RollbackTransaction(orm orm.Ormer) {
+	err := orm.Rollback()
+	if err != nil {
+		klog.Errorf("failed to rollback transaction, err: %v", err)
+	}
+}
--- a/edge/pkg/devicetwin/dtclient/device_db.go
+++ b/edge/pkg/devicetwin/dtclient/device_db.go
@@ -1,6 +1,7 @@
 package dtclient
 
 import (
+	"github.com/astaxie/beego/orm"
 	"k8s.io/klog/v2"
 
 	"github.com/kubeedge/kubeedge/edge/pkg/common/dbm"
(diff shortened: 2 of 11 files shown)


gy95:
i thought we can use only one orm 👀

Shelley-BaoYue:
> i thought we can use only one orm 👀

yes,  i tried to use only one orm and one transaction in one function. 😆  but too much errors include db errors are not handled in devicetwin, so transaction can not commit or rollback correctly.  The optimization work is mentioned in this issue https://github.com/kubeedge/kubeedge/issues/4126

wackxu:
> i thought we can use only one orm 👀

From the beego 1.x doc, if we want to run sql in one transaction, we must use the same ormer for the sql.

wackxu:
in beego 2.x,
```
In v2.x, there is a big big change:

The ORM instance should be stateless, so it’s now thread safe.

```

edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -87,60 +88,80 @@ func UpdateDeviceMulti(updates []DeviceUpdate) error {
...
+	}
+
 	for _, add := range adds {
-		err = SaveDevice(&add)
+		err = SaveDevice(obm, &add)
 
 		if err != nil {
 			klog.Errorf("save device failed: %v", err)
-			obm.Rollback()
+			dbm.RollbackTransaction(obm)
```
vincentgoat, line 103: A defer function to run `dbm.RollbackTransaction(obm)` will be more clean up.

edge/pkg/common/dbm/db.go
```diff
@@ -37,3 +37,19 @@ func InitDBConfig(driverName, dbName, dataSource string) {
 		}
 	})
 }
+
+type newOrmerfunc func() orm.Ormer
```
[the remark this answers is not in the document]
Shelley-BaoYue: Done.

Shelley-BaoYue:
> > i thought we can use only one orm 👀
> 
> From the beego 1.x doc, if we want to run sql in one transaction, we must use the same ormer for the sql.

Now we use the same ormer in one transaction. with this issue https://github.com/kubeedge/kubeedge/issues/4126， we will do more optimization for db transactions.

edge/pkg/devicetwin/dtclient/device_db.go
Shelley-BaoYue: Done.

fisherxu:
> 




> 
> ```
> In v2.x, there is a big big change:
> 
> The ORM instance should be stateless, so it’s now thread safe.
> ```
> 
> transaction
> 
> https://beego.vip/docs/mvc/model/transaction.md


If we create one Ormer for every func, is there any conflict or dirty write?

> in beego 2.x,

Now we still use the 1.x beego  @Shelley-BaoYue @wackxu

wackxu:
> 

yes, we should be careful when we bump beego to v2

fisherxu (commented):
/lgtm
/approve

PR #4122, opened 2022-08-17
remove GetMirrorPodByPod function

elpsyr:

**What type of PR is this?**
/kind cleanup

We don't have static pod(in kubeedge), so we don't need GetMirrorPodByPod.
If we call  GetMirrorPodByPod, only false&nil will be returned. this is a meaningless judgment.

Fixes #

```release-note

```


fisherxu (commented):
Through now we don't support static pod, but will add it back in future release. And the pr make sense now :)
/lgtm
/approve

PR #4125, opened 2022-08-18
 1.fix delete pod terminating sometimes. 2.fix  update pod status timeout. 3.fix pod condition remain ready false when node status changes from offline to online

wackxu:
**What type of PR is this?**

/kind bug

Fixes  https://github.com/kubeedge/kubeedge/issues/4172

The E2E test is failed sometimes, after debug on the master, serveral problem should be fixed.

1.  When a pod is quickly create and delete when specify nodeName, the pod can not be deleted and will always be terminating,   It also happens when the node is offline, and we create and delete the pod specify nodeName to it. In this case the statu manager does not add the pod to its podStatuses, so the batch sync will not process the pod and the pod will be unhandled. so update the batch sync of status manager to the old version code(batch sync from pod manager)

2. Pod status update timeout when the pod is deleted forced or err, so add response for this situation

3、when node disconnect from cloudcore and the pod condition type ready will be updated by kube-controller and become to fasle, 
```
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2022-09-09T07:47:31Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2022-09-09T08:31:53Z"
    status: "False"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: null
    status: "True"
    type: PodScheduled
```

after the node online, the pod condition ready remains false and never becomes to true.

when pod status is inconsistent between apiserver and edgecore,edgecoreshould forcibly send an update to reconcile the inconsistence, because edgecore should be the source of truth of pod status. So we add needsReconcile back to the status manager. 

```release-note

```


gy95 (approved):

kevin-wangzefeng:
please update the PR title

edge/pkg/metamanager/process.go
```diff
@@ -95,17 +95,6 @@ func msgDebugInfo(message *model.Message) string {
 	return fmt.Sprintf("msgID[%s] resource[%s]", message.GetID(), message.GetResource())
 }
 
-func resourceUnchanged(resType string, resKey string, content []byte) bool {
-	if resType == model.ResourceTypePodStatus {
-		dbRecord, err := dao.QueryMeta("key", resKey)
-		if err == nil && len(*dbRecord) > 0 && string(content) == (*dbRecord)[0] {
```
wackxu, line 101: for now, the pod status is updated when we detect changes between cache and apiserver, so the upstream podstatus message is infrequently，when we receive the update message, we must send it to the cloud.

gy95: the pr contains 3 bug fixes? the title description misses the first commit info

wackxu: done

fisherxu (commented):
/lgtm
/approve

RyanZhaoXB:
> I have cherry pick @RyanZhaoXB commit to fix the failed device e2e and the e2e is green now, cc @fisherxu @gy95 @vincentgoat @Shelley-BaoYue PTAL

OK. i will close https://github.com/kubeedge/kubeedge/pull/4213

PR #4138, opened 2022-08-26
When '--dry-run' flag used, print k8s resource of Cloudcore

WillardHu:
…minifest of Cloudcore

**What type of PR is this?**
/kind bug

When 'init' command use the '--dry-run' flag, keadm did not print k8s resource on the stdout, but usage is 'Print the generated k8s resources on the stdout, not actual excute. Always use in debug mode'

Fixes #

```release-note

```


zhu733756:
I think this fix cannot meet your requirement, the installation is already down by the above codes. But we support keadm beta manifest generate cmd, does it meet your requirement?

WillardHu:
> I think this fix cannot meet your requirement, the installation is already down by the above codes. But we support keadm beta manifest generate cmd, does it meet your requirement?

thanks, this will solve my problem

PR #4140, opened 2022-08-26
fix the definition of tolerations in helm values.yaml

WillardHu:

**What type of PR is this?**
/kind bug

io.k8s.api.core.v1.PodSpec.tolerations is an array type, use '--set cloudCore.tolerations[0].xxx' to set values, cause 'init' command to return an error message: "execute keadm command failed:  cannot build renderer: failed parsing --set data:unable to parse key: interface conversion: interface {} is map[string]interface {}, not []interface {}"

Fixes #

```release-note

```

### DIFF
--- a/docs/proposals/refactor-keadm-on-the-cloud.md
+++ b/docs/proposals/refactor-keadm-on-the-cloud.md
@@ -49,7 +49,7 @@ spec:
           - matchExpressions:
           - key: node-role.kubernetes.io/edge
               operator: DoesNotExist
-    tolerations: {}
+    tolerations: []
     nodeSelector: {}
     resources:
       limits:
@@ -110,7 +110,7 @@ spec:
             - matchExpressions:
               - key: node-role.kubernetes.io/edge
                   operator: DoesNotExist
-    tolerations: {}
+    tolerations: []
     nodeSelector: {}
     resources:
       limits:
--- a/manifests/charts/cloudcore/values.yaml
+++ b/manifests/charts/cloudcore/values.yaml
@@ -24,7 +24,7 @@ cloudCore:
         - matchExpressions:
           - key: node-role.kubernetes.io/edge
             operator: DoesNotExist
-  tolerations: {}
+  tolerations: []
   nodeSelector: {}
   resources:
     limits:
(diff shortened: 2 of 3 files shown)


fisherxu:
cc @zhu733756

zhu733756:
> cc @zhu733756

Thanks！ If you want to make it an array, not an object, pls also fix the related codes like https://github.com/kubeedge/kubeedge/blob/master/manifests/charts/cloudcore/templates/deployment_cloudcore.yaml#L35

WillardHu:
> > cc @zhu733756
> 
> Thanks！ If you want to make it an array, not an object, pls also fix the related codes like https://github.com/kubeedge/kubeedge/blob/master/manifests/charts/cloudcore/templates/deployment_cloudcore.yaml#L35

Before the modification, when using the 'init' command with the '--set=cloudCore.tolerations.xxx' flag, command returns error message: 'invalid type for io.k8s.api.core.v1.PodSpec.tolerations: got "map", expected "array"', but with the flag is '--set=cloudCore.tolerations[0].xxx', command returns error message:  'interface {} is map[string]interface {}, not []interface {}.'

The io.k8s.api.core.v1.PodSpec.tolerations is an array type. We don't need to modify the defined in the deployment, we just to modify the 'values.yaml' to make it work.

fisherxu (commented):
/approve
cc @zhu733756  plz take the final review

zhu733756:
/lgtm
Yes, you are right @WillardHu ， I think of range syntax with array

docs/proposals/refactor-keadm-on-the-cloud.md
```diff
@@ -49,7 +49,7 @@ spec:
           - matchExpressions:
           - key: node-role.kubernetes.io/edge
               operator: DoesNotExist
-    tolerations: {}
+    tolerations: []
     nodeSelector: {}
```
gy95, line 53: @zhu733756   one question, here nodeSelector is an object, how to specify its value in the command line during keadm installation. :)

zhu733756: Can you try this? 

image

gy95: yes, it works, thanks :)

PR #4141, opened 2022-08-26
fix force delete pod when node is offline

wackxu:

**What type of PR is this?**

/kind bug

Fixes # 

when node is offline and we force delete the pod in the node, and then the edge node is online,
the deleted pod container is always in the edge node unless we restart the edgecore.

```
	// When the edge node is offline and the pod in the node is deleted forcefully,
	// and then we make the node online, We do not have the pod full information
	// because the pod is deleted from the kube apiServer, then the syncController
	// will send a message with the pod name, namespace and UID, so we can not filter
	// pod according to the node name, if the pod exist in the podManager, it indicates
	// that the pod is belong to this edge node, so we can delete pod from the edged,
	// otherwise the pod is not belong to the node, we just return.
```

The releate E2E will be added in anoter pr

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -1320,16 +1320,26 @@ func (e *edged) handlePod(op string, content []byte) (err error) {
 		return err
 	}
 
+	// When the edge node is offline and the pod in the node is deleted forcefully,
+	// and then we make the node online, We do not have the pod full information
+	// because the pod is deleted from the kube apiServer, then the syncController
+	// will send a message with the pod name, namespace and UID, so we can not filter
+	// pod according to the node name, if the pod exist in the podManager, it indicates
+	// that the pod is belong to this edge node, so we can delete pod from the edged,
+	// otherwise the pod is not belong to the node, we just return.
+	if op == model.DeleteOperation {
+		if delPod, ok := e.podManager.GetPodByUID(pod.UID); ok {
+			e.deletePod(delPod)
+		}
+		return nil
+	}
+
 	if filterPodByNodeName(&pod, e.nodeName) {
 		switch op {
 		case model.InsertOperation:
 			e.addPod(&pod)
 		case model.UpdateOperation:
 			e.updatePod(&pod)
-		case model.DeleteOperation:
-			if delPod, ok := e.podManager.GetPodByUID(pod.UID); ok {
-				e.deletePod(delPod)
-			}
 		}
 	}
 


edge/pkg/edged/edged.go
```diff
@@ -1310,16 +1310,26 @@ func (e *edged) handlePod(op string, content []byte) (err error) {
...
+	// that the pod is belong to this edge node, so we can delete pod from the edged,
+	// otherwise the pod is not belong to the node, we just return.
+	if op == model.DeleteOperation {
+		if delPod, ok := e.podManager.GetPodByName(pod.Namespace, pod.Name); ok {
+			e.deletePod(delPod)
+		}
+		return nil
+	}
+
 	if filterPodByNodeName(&pod, e.nodeName) {
```
gy95, line 1327: you mean in this circumstance, here pod doesn't contain node info, so will not trigger a delete operation?

wackxu: yes, in this case, 
```
	if apierrors.IsNotFound(err) {
		// trigger the delete event
		klog.V(4).Infof("%s: %s has been deleted in K8s, send the delete event to edge in sync loop", resourceType, sync.Spec.ObjectName)
		newObject := &unstructured.Unstructured{}
		if msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.DeleteOperation, newObject); msg != nil {
		} else {
			if err := sctl.crdclient.ReliablesyncsV1alpha1().ObjectSyncs(sync.Namespace).Delete(context.Background(), sync.Name, *metav1.NewDeleteOptions(0)); err != nil {
				klog.Errorf("Failed to delete objectsync %s: %v", sync.Name, err)
			}
		}
		return
	}
```

wackxu:
Steps to reproduce the problem

1、create pod in the edge node and wait the pod is running
2、Simulation node is offline, we can use `chaosmesh` or iptables
3、when the node is offline, force delete the pod
`kubectl delete pod xx --force --grace-period=0`
4、make the node online
5、check the container of the pod is running and is not deleted.

edge/pkg/edged/edged.go
wackxu: for most case we can get the node info from the pod

Shelley-BaoYue:
In other normal deletion scenarios，will this modification cause some problem？

wackxu:
> the pod is not belong to the node, we just return.

No, if the pod is belong to this node, we could find it from the pod manager, and the pod will be deleted from edged.

If  the pod is not belong to the node,  we could not find the info from the pod manager, so the delete message will not be processed and return directlly.

edge/pkg/edged/edged.go
gy95: /lgtm

fisherxu (commented):
/lgtm
And if setting the nodename for pod in synccontroller, the issue can also be fixed? :) If fixed in edged, do the new kubelet still need this fixing?

Shelley-BaoYue:
> /lgtm And if setting the nodename for pod in synccontroller, the issue can also be fixed? :) If fixed in edged, do the new kubelet still need this fixing?

The new Kubelet also needs. I will rebase and add this fixing.

wackxu:
> /lgtm And if setting the nodename for pod in synccontroller, the issue can also be fixed? :) If fixed in edged, do the new kubelet still need this fixing?

Set nodeName in the synccontroller can also solve the problem. but we need to sense whether the  kind is a pod or not, and then set the name of the node

fisherxu:
PR need rebase :)

wackxu:
> PR need rebase :)

done

fisherxu (commented):
/lgtm
/approve

fisherxu:
Please cherry-pick the PR to release1.9~1.11 :)

PR #4149, opened 2022-08-30
simplify unnecessary groupMap

Congrool:
**What type of PR is this?**

/kind cleanup

No code will use the key of groupMap now, we can simplify it to groupList. Additionally, we can avoid to pubConnectInfo twice to metaGroup when detecting the re-connection between cloud and edge.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -17,11 +17,10 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/config"
 )
 
-var groupMap = map[string]string{
-	"resource": modules.MetaGroup,
-	"twin":     modules.TwinGroup,
-	"func":     modules.MetaGroup,
-	"user":     modules.BusGroup,
+var groupList = []string{
+	modules.MetaGroup,
+	modules.TwinGroup,
+	modules.BusGroup,
 }
 
 var (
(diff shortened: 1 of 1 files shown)


edge/pkg/edgehub/process.go
```diff
@@ -17,11 +17,10 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/config"
 )
 
-var groupMap = map[string]string{
-	"resource": modules.MetaGroup,
-	"twin":     modules.TwinGroup,
-	"func":     modules.MetaGroup,
```
gy95, line 23: it seems that group `func` is never used, can be deleted directly? 
cc @wackxu

Congrool:
BTW, it seems that EventBus and SeriviceBus do not care about connection state and never handle `messagepkg.ResourceTypeNodeConnectio`. So can we just send it to twin group and meta group?

edge/pkg/edgehub/process.go
wackxu: yes, we can delete it since we do not have edge function

Congrool:
friendly ping @gy95 , PTAL

gy95:
cc @vincentgoat  PTAL

vincentgoat:
> BTW, it seems that EventBus and SeriviceBus do not care about connection state and never handle msg of type `messagepkg.ResourceTypeNodeConnection`. So can we just send it to twin group and meta group?

agree.
/lgtm

PR #4151, opened 2022-08-31
use module group name instead of src name as argument when `SendToGroup`

Congrool:
Signed-off-by: Congrool <chpzhangyifei@zju.edu.cn>

**What type of PR is this?**

/kind cleanup

For `beehiveContext.SendToGroup`, we'd better use common module group name. If using `model.SrcEdgeController` which indicates the `source` name, it's not easy to read.

Additionally, `model.SrcEdgeController` and `model.SrcDeviceController` seem nowhere to be used as `source` of message. So we can remove these redundant constants.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudhub/channelq/channelq.go
+++ b/cloud/pkg/cloudhub/channelq/channelq.go
@@ -316,9 +316,9 @@ func (q *ChannelMessageQueue) Publish(msg *beehiveModel.Message) error {
 	case application.MetaServerSource:
 		beehiveContext.Send(modules.DynamicControllerModuleName, *msg)
 	case model.ResTwin:
-		beehiveContext.SendToGroup(model.SrcDeviceController, *msg)
+		beehiveContext.SendToGroup(modules.DeviceControllerModuleGroup, *msg)
 	default:
-		beehiveContext.SendToGroup(model.SrcEdgeController, *msg)
+		beehiveContext.SendToGroup(modules.EdgeControllerGroupName, *msg)
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


vincentgoat (commented):
lgtm

fisherxu (commented):
/lgtm
/cc @zc2638

cloud/pkg/cloudhub/channelq/channelq.go
```diff
@@ -316,9 +316,9 @@ func (q *ChannelMessageQueue) Publish(msg *beehiveModel.Message) error {
 	case application.MetaServerSource:
 		beehiveContext.Send(modules.DynamicControllerModuleName, *msg)
 	case model.ResTwin:
-		beehiveContext.SendToGroup(model.SrcDeviceController, *msg)
+		beehiveContext.SendToGroup(modules.DeviceControllerModuleGroup, *msg)
 	default:
-		beehiveContext.SendToGroup(model.SrcEdgeController, *msg)
+		beehiveContext.SendToGroup(modules.EdgeControllerGroupName, *msg)
```
zc2638, line 321: Is it possible to keep having the same naming format.
e.g.
`DeviceControllerModuleGroup` and `EdgeControllerModuleGroup`
e.g.
`DeviceControllerGroupName` and `EdgeControllerGroupName`

zc2638: Thanks a lot, it will be better if you have time to fix it.

PR #4160, opened 2022-09-05
fix: Ignore the cache timestamp of the MachineInfo Metrics

tonyzaizai:
Signed-off-by: tl <tonyzaizai@kubesphere.io>

**What type of PR is this?**

 /kind bug

   Avoid adding timestamp to machine metrics when they're cached.
   Please refer to the following K8s PRs for more details:
        link: https://github.com/kubernetes/kubernetes/pull/95210#issuecomment-726143798
        link: https://github.com/kubernetes/kubernetes/pull/97006
   Analysis of causes :
       1. The cadvisor collects machine metrics when it launches and then caches them for later scrape.
           It is assumed that as long as the machine is not restarted, the hardware will not change and so do the machine metrics.
       2. This timestamp of the machine metrics should be set when they're scraped, not when they're cached.
       3. If using the timestamp when the machine metrics are cached, the timestamp when these metrics are scraped will be outdated 
          which may cause issues like 'out of order sample' when sending these metrics to prometheus.

Fixes # 

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -543,6 +543,17 @@ func newEdged(enable bool) (*edged, error) {
 	if ed.machineInfo, err = ed.newMachineInfo(); err != nil {
 		return nil, err
 	}
+	// Avoid adding timestamp to machine metrics when they're cached
+	// Please refer to the following K8s PRs for more details:
+	//    link: https://github.com/kubernetes/kubernetes/pull/95210#issuecomment-726143798
+	//    link: https://github.com/kubernetes/kubernetes/pull/97006
+	// Analysis of causes :
+	//   1. The cadvisor collects machine metrics when it launches and then caches them for later scrape.
+	//      It is assumed that as long as the machine is not restarted, the hardware will not change and so do the machine metrics.
+	//   2. This timestamp of the machine metrics should be set when they're scraped, not when they're cached.
+	//   3. If using the timestamp when the machine metrics are cached, the timestamp when these metrics are scraped will be outdated
+	//      which may cause issues like 'out of order sample' when sending these metrics to prometheus.
+	ed.machineInfo.Timestamp = time.Time{}
 
 	if ed.containerRuntimeName == kubetypes.RemoteContainerRuntime {
 		// create a log manager


benjaminhuo:
This fixes an edge container prometheus remote write issue when scraping metrics from:
`<edge-node-ip>:10350/metrics/cadvisor`

edge/pkg/edged/edged.go
```diff
@@ -543,6 +543,17 @@ func newEdged(enable bool) (*edged, error) {
...
+	// Avoid adding timestamp to machine metrics when they're cached
+	// Please refer to the following K8s PRs for more details:
+	//    link: https://github.com/kubernetes/kubernetes/pull/95210#issuecomment-726143798
+	//    link: https://github.com/kubernetes/kubernetes/pull/97006
+	// Analysis of causes :
+	//   1. The cadvisor collects machine metrics when it launches and then caches them for later scrape.
+	//      It is assumed that as long as the machine is not restarted, the hardware will not change and so do the machine metrics.
+	//   2. This timestamp of the machine metrics should be set when they're scraped, not when they're cached.
+	//   3. If using the timestamp when the machine metrics are cached, the timestamp when these metrics are scraped will be outdated
+	//      which may cause issues like 'out of order sample' when sending these metrics to prometheus.
```
fisherxu, line 555: @tonyzaizai @benjaminhuo Thanks for the clarification, a small question, if we set Timestamp to zero, where will set the Timestamp for `metrics are scraped`? :)

benjaminhuo: Usually the agent such as prometheus agent or fluentbit will set the timestamp to current time if the timestamp is zero: 
- https://github.com/fluent/fluent-bit/blob/master/plugins/in_prometheus_scrape/prom_scrape.c#L113 
- https://github.com/calyptia/cmetrics/blob/master/src/cmt_decode_prometheus.c#L278 

K8s fixed this in the same way:
- https://github.com/google/cadvisor/issues/2740
- https://github.com/kubernetes/kubernetes/pull/97692

KubeEdge includes part of the k8s fix 
- https://github.com/kubeedge/kubeedge/blob/master/vendor/k8s.io/kubernetes/pkg/kubelet/server/server.go#L384

The other part is in kubelet which is edged for kubeedge, and that part is not included:
- https://github.com/kubernetes/kubernetes/pull/97692/files#diff-67dd9e8f3ee257072765326cb4f242852554a2c0753563fa51e292c0a63a7b94R566

benjaminhuo: cadvisor will send metrics with or without timestamp when the metrics are requested/collected: https://github.com/google/cadvisor/blob/65fa5b44d3840d465a5c004a0f6722e0fd540137/metrics/prometheus_machine.go#L240

fisherxu: Thanks for your explanation :)

fisherxu (commented):
/lgtm
/approve

fisherxu:
@tonyzaizai Could you pleasae cherry-pick th patch to release-1.11 and release-1.10?

tonyzaizai:
> @tonyzaizai Could you pleasae cherry-pick th patch to release-1.11 and release-1.10?

ok

benjaminhuo:
> @tonyzaizai Could you pleasae cherry-pick th patch to release-1.11 and release-1.10?

@fisherxu There is no patch release for v1.9.x?

fisherxu:
Sorry for the delay, it's needed to pick to 1.9, I'll help to cherry-pick.

PR #4162, opened 2022-09-06
Fix e2e bug

gy95:
**What type of PR is this?**

Fix e2e is still successful when there are e2e failures.

Fixes #

```release-note

```

### DIFF
--- a/tests/e2e/scripts/fast_test.sh
+++ b/tests/e2e/scripts/fast_test.sh
@@ -36,14 +36,22 @@ then
     --image-url=nginx \
     --kube-master="https://$MASTER_IP:6443" \
     --kubeconfig=$KUBECONFIG \
-    --test.v \
-    2>&1 | tee -a /tmp/testcase.log
+    --test.v
+    GINKGO_TESTING_RESULT=$?
 else
    ginkgo -v ./$compilemodule/$compilemodule.test -- \
     --image-url=nginx \
     --image-url=nginx \
     --kube-master="https://$MASTER_IP:6443" \
     --kubeconfig=$KUBECONFIG \
-    --test.v \
-   2>&1 | tee -a  /tmp/testcase.log
+    --test.v
+   GINKGO_TESTING_RESULT=$?
+fi
+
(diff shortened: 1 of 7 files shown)


gy95:
statefulset e2e errors:
```
1. an empty namespace may not be set during creation
2. StatefulSet.apps \"edge-statefulset-7q1ld\" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string(nil): `selector` does not match template `labels`]
```

gy95:
cc @wackxu

wackxu (approved):
Thanks for you fix, LGTM

tests/e2e/scripts/fast_test.sh
```diff
@@ -36,14 +36,22 @@ then
     --image-url=nginx \
     --kube-master="https://$MASTER_IP:6443" \
     --kubeconfig=$KUBECONFIG \
-    --test.v \
-    2>&1 | tee -a /tmp/testcase.log
```
fisherxu, line 40: Is it possible to still store the log in this file for debugging?

gy95: can use `> /tmp/testcase.log` to replace `tee -a `, but i think this is duplicated. the log will be stored in action logs directly, we can look at action logs.

gy95: but if we use `>`, i'm not quite sure whether the logs will be printed on action std out

gy95: by test, if we use `>`, ginkgo will not print logs to the action stdout
cannot think a proper way to store the log in the file and use the ginkgo exit code meanwhile, so just logs to the action stdout directly

fisherxu: If users run e2e in local host, the log file should be still needed, we can fix the e2e bug first, then explore a way to store the log file.

fisherxu: Open an issue to track it #4174 @gy95

fisherxu (commented):
/lgtm
/approve

PR #4187, opened 2022-09-16
fix keadm_deprecated_e2e error

gy95:
**What type of PR is this?**
/kind bug

In `keadm deprecated e2e`, we will run 2 versions test: one is based on the latest source codes, the other is based on the latest official release. So we must cleanup all related things of the first e2e test before we run the second e2e test. And in `keadm deprecated e2e`, we'll use systemd to start edgecore, so we must cleanup edgecore.service also

Fixes #

```release-note

```


gy95:
cc @wackxu @fisherxu 
ptal

wackxu:
we need merge this to solve the e2e failed asap

gy95:
by test many times, it indeed works well...

PR #4196, opened 2022-09-19
fix delete and create same name pod failed occasionally

wackxu:

**What type of PR is this?**

/kind bug

Fixes #

in PR https://github.com/kubeedge/kubeedge/pull/4116, we fixed delete and the recreate statefulset pod is not running bug. But in some extreme case, there is a chance that the new pod info is deleted by the old pod info use the podUID,  if we use  podManager.GetPodByName in the sync for pod, it will be failed

image

```
E0919 04:19:01.491325   22146 edged.go:1021] worker [2] handle pod addition item [edge-statefulset-rgqyc-1] failed with not found error.
E0919 04:19:01.491335   22146 edged.go:1025] worker [2] handle pod addition item [edge-statefulset-rgqyc-1] failed: sync pod failed: failed to "StartContainer" for "nginx" with CreateContainerConfigError: "cannot find volume \"kube-api-access-8r2rz\" to mount into container \"nginx\"", re-add to queue
I0919 04:19:01.497013   22146 edged.go:957] sync loop ignore event: [ContainerDied], with pod [5f66b9c3-6142-4f16-b52a-3d553c72eb3f] not found
I0919 04:19:01.497036   22146 edged.go:957] sync loop ignore event: [ContainerStarted], with pod [5f66b9c3-6142-4f16-b52a-3d553c72eb3f] not found
I0919 04:19:02.511745   22146 edged.go:957] sync loop ignore event: [ContainerDied], with pod [5f66b9c3-6142-4f16-b52a-3d553c72eb3f] not found
```

we must not use podManager.GetPodByName, in kubernetes GetPodByName is only used for the log/exec API, and in the inner logic of kubelet, the GetPodByUID is used.

in a pod create and delete process, the pod info can be filled again to the podManager if we use GetPodByUID to sync the pod  and the GetPodByName can work for the log/exec. 

```release-note

```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -356,7 +356,7 @@ func (uc *UpstreamController) updatePodStatus() {
 			case model.UpdateOperation:
 				for _, podStatus := range podStatuses {
 					getPod, err := uc.kubeClient.CoreV1().Pods(namespace).Get(context.Background(), podStatus.Name, metaV1.GetOptions{})
-					if errors.IsNotFound(err) {
+					if (err == nil && getPod.UID != podStatus.UID) || errors.IsNotFound(err) {
 						klog.Warningf("message: %s, pod not found, namespace: %s, name: %s", msg.GetID(), namespace, podStatus.Name)
 
 						// Send request to delete this pod on edge side
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -881,11 +881,7 @@ func (e *edged) syncLoopIteration(plegCh <-chan *pleg.PodLifecycleEvent, houseke
 					}
 				}
 				klog.Infof("Will restart pod [%s]", pod.Name)
-				key := types.NamespacedName{
-					Namespace: pod.Namespace,
-					Name:      pod.Name,
-				}
-				e.podAdditionQueue.Add(key.String())
+				e.podAdditionQueue.Add(pod.UID)
 			}
 
 		case update := <-e.readinessManager.Updates():
@@ -899,11 +895,7 @@ func (e *edged) syncLoopIteration(plegCh <-chan *pleg.PodLifecycleEvent, houseke
 				break
 			}
 
-			key := types.NamespacedName{
-				Namespace: pod.Namespace,
-				Name:      pod.Name,
-			}
-			e.podAdditionQueue.Add(key.String())
+			e.podAdditionQueue.Add(pod.UID)
 
 		case update := <-e.startupManager.Updates():
 			started := update.Result == proberesults.Success
@@ -916,11 +908,7 @@ func (e *edged) syncLoopIteration(plegCh <-chan *pleg.PodLifecycleEvent, houseke
 				break
 			}
 
-			key := types.NamespacedName{
-				Namespace: pod.Namespace,
-				Name:      pod.Name,
-			}
-			e.podAdditionQueue.Add(key.String())
+			e.podAdditionQueue.Add(pod.UID)
 
 		case plegEvent := <-plegCh:
 			if pod, ok := e.podManager.GetPodByUID(plegEvent.ID); ok {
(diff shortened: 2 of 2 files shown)


edge/pkg/edged/edged.go
```diff
@@ -1088,14 +1058,13 @@ func (e *edged) consumePodAddition(namespacedName *types.NamespacedName) error {
...
 
-	podUID := pod.GetUID()
 	t, ok := e.podLastSyncTime.Load(podUID)
 	if !ok {
 		t = time.Time{}
 	}
 	curPodStatus, err := e.podCache.GetNewerThan(podUID, t.(time.Time))
 	if err != nil {
-		return fmt.Errorf("pod %s cache newer failed: %v", podName, err)
+		return fmt.Errorf("pod %s cache newer failed: %v", pod.Name, err)
```
gy95, line 1067: we print Name and UID meanwhile, is this need unifying?

edge/pkg/edged/edged.go
```diff
@@ -1046,18 +1014,21 @@ func (e *edged) podRemoveWorkerRun(consumers int) {
...
 				}
-				namespacedName := NewNamespacedNameFromString(item.(string))
-				podName := namespacedName.Name
-				klog.Infof("consumer: [%d], worker get removed pod [%s]\n", i, podName)
-				err := e.consumePodDeletion(&namespacedName)
+
+				podUID := item.(types.UID)
+
+				//namespacedName := NewNamespacedNameFromString(item.(string))
+				//podName := namespacedName.Name
```
gy95, line 1021: remove?

wackxu: done

edge/pkg/edged/edged.go
wackxu: use format.Pod(pod) that contains pod name namespace and uid

edge/pkg/edged/edged.go
```diff
@@ -1001,28 +969,28 @@ func (e *edged) podAddWorkerRun(consumers int) {
 					klog.Errorf("consumer: [%d], worker addition queue is shutting down!", i)
 					return
 				}
-				namespacedName := NewNamespacedNameFromString(item.(string))
-				podName := namespacedName.Name
-				klog.Infof("worker [%d] get pod addition item [%s]", i, podName)
-				backOffKey := fmt.Sprintf("pod_addition_worker_%s", podName)
+
+				podUID := item.(types.UID)
```
vincentgoat, line 973: It is better to check if `type assert` is ok to avoid a panic error.

wackxu: It is a inner queue and we do not set other types data, so it is ok for now as before

gy95:
can try rebase to the master, maybe related to merged DMI

wackxu:
the device releated e2e is always failed, we should fix it cc @RyanZhaoXB 

```
 TestCase:E2E_UPDATE_DEVICE_1: Update device instance for LED device
```

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -356,7 +356,7 @@ func (uc *UpstreamController) updatePodStatus() {
 			case model.UpdateOperation:
 				for _, podStatus := range podStatuses {
 					getPod, err := uc.kubeClient.CoreV1().Pods(namespace).Get(context.Background(), podStatus.Name, metaV1.GetOptions{})
-					if errors.IsNotFound(err) {
+					if (err == nil && getPod.UID != podStatus.UID) || errors.IsNotFound(err) {
```
wackxu, line 359: in kubeedge, the message is process asynchronously，when the old pod is deleted from k8s, the message may exists and if the same name pod is created, it will incorrectly updating the state of  new pod

fisherxu:
We can merge it then pick to other releases, if the new edged merged in, will this pr still work?

fisherxu:
cc @wackxu @Shelley-BaoYue

wackxu:
> We can merge it then pick to other releases, if the new edged merged in, will this pr still work?

it is only for old version edgecore, if new edged merged, the fix will be deleted.

fisherxu (commented):
/lgtm
/approve

PR #4104, opened 2022-08-08
Verify token when list/watch and add autonomy switch

vincentgoat:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
1. Change `MetaServer` to HTTPS server.
2. Verify bearer token when client list/watch from `MetaServer`.
3. Add an autonomy switch for security when edgenode is offline, avoiding listing/watching all resources.
4. The default support kubeconfig.insecure-skip-tls-verify = true when listing/watching from `MetaServer`.
image

**Which issue(s) this PR fixes**:

Fixes #3805 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

Please refer to #4108

```release-note

```

PR #4226, opened 2022-09-23
add featuregate for authorization

vincentgoat:

**What type of PR is this?**
/kind feature

Addition of PR: https://github.com/kubeedge/kubeedge/pull/4104
add featuregate for authorization

Featuregate of `requireAuthorization` default is false as it is in the alpha stage, follow below to enable this gate.
edgecore.yaml:
```
apiVersion: edgecore.config.kubeedge.io/v1alpha1
database:
  aliasName: xxx
  ...
featureGates:
  requireAuthorization: true
kind: EdgeCore
modules:
  ...
```
cloudcore.yaml:
```
commonConfig:
  tunnelPort: xxx
kind: CloudCore
featureGates:
  requireAuthorization: true
```

Fixes #

```release-note

```

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -19,13 +19,18 @@ func init() {
 }
 
 const (
-	// ProcessStatusSync supports synchronization between process message channel statuses
-	ProcessStatusSync featuregate.Feature = "ProcessStatusSync"
+	// RequireAuthorization supports application access authorization from edge sides.
+	// It will determine whether app can acquire meta data from kube-apiserver (if node is online) or from local host db (when node is offline)
+	// without authorization. When node is offline and this value set to true, meta data can't be retrieved from meta server
+	// because authorization offline is not achieved as of now.
+	// alpha: v1.12
+	// owner: @vincentgoat
+	RequireAuthorization featuregate.Feature = "requireAuthorization"
 )
 
 // defaultFeatureGates consists of all known Kubeedge-specific feature keys.
 // To add a new feature, define a key for it above and add it here. The features will be
 // available throughout Kubeedge binaries.
 var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	ProcessStatusSync: {Default: false, PreRelease: featuregate.Alpha},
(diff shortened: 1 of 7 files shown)


pkg/features/features.go
```diff
@@ -19,13 +19,14 @@ func init() {
...
+	// alpha: v1.12
+	RequireAuthorization featuregate.Feature = "requireAuthorization"
 )
 
 // defaultFeatureGates consists of all known Kubeedge-specific feature keys.
 // To add a new feature, define a key for it above and add it here. The features will be
 // available throughout Kubeedge binaries.
 var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
-	ProcessStatusSync: {Default: false, PreRelease: featuregate.Alpha},
+	RequireAuthorization: {Default: true, PreRelease: featuregate.Alpha},
```
wackxu, line 31: this is strange, it is a alpha feature but the default value is true

pkg/features/features.go
```diff
@@ -19,13 +19,14 @@ func init() {
 }
 
 const (
-	// ProcessStatusSync supports synchronization between process message channel statuses
-	ProcessStatusSync featuregate.Feature = "ProcessStatusSync"
+	// RequireAuthorization Supports application access authorization from edge sides
+	// alpha: v1.12
+	RequireAuthorization featuregate.Feature = "requireAuthorization"
```
wackxu, line 24:
```suggestion
	// owner: @vincentgoat
	// alpha: v1.12

	// RequireAuthorization Supports application access authorization from edge sides
```

wackxu: the comment is less accurate，please add more detail

pkg/features/features.go
fisherxu: Agreed, default value should be false? If users want to use it, just set it in config file

vincentgoat: done

pkg/features/features.go
vincentgoat: done

fisherxu (commented):
/lgtm

PR #4229, opened 2022-09-24
remove not used kubeedge-version flag in keadm init command

gy95:
**What type of PR is this?**
/kind cleanup

remove not used kubeedge-version flag in keadm init command
this flag is never used, we can remove it(now we use --profile version=v1.10.1 to set image version)

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cloud/init.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/init.go
@@ -86,9 +86,6 @@ func newInitOptions() *types.InitOptions {
 }
 
 func addInitOtherFlags(cmd *cobra.Command, initOpts *types.InitOptions) {
-	cmd.Flags().StringVar(&initOpts.KubeEdgeVersion, types.KubeEdgeVersion, initOpts.KubeEdgeVersion,
-		"Use this key to set the default image tag")
-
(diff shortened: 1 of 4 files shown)


fisherxu (commented):
`kubeedgeversion` should be a better name than `profile version`? cc @zhu733756

gy95:
https://github.com/kubeedge/kubeedge/blob/821009caab97eb0f69f82b47463e8fdf62d5f9a0/keadm/cmd/keadm/app/cmd/cloud/init.go#L111-L114
now both `--profile` and `--set` are used to set deploy parameters, we may do some work to unify this.

zhu733756:
> `kubeedgeversion` should be a better name than `profile version`? cc @zhu733756

Yes

zhu733756:
> https://github.com/kubeedge/kubeedge/blob/821009caab97eb0f69f82b47463e8fdf62d5f9a0/keadm/cmd/keadm/app/cmd/cloud/init.go#L111-L114
> 
> 
> now both `--profile` and `--set` are used to set deploy parameters, we may do some more work to unify this.

Agree.  I think "--profile" includes the defined profiles such as external iptablesmgr and so on, "--set" is the function of helm. They do the same thing.

PR #4231, opened 2022-09-26
bugfix grep nodeReady in local-up shell

Shelley-BaoYue:
**What type of PR is this?**
/kind bug

bugfix `grep Ready` in local-up-kubeedge shell. Use -w can judge node is Ready or Not more precisely. 

Sometimes node status will be `UnKnow` or `NotReady`, it's need to regexp match to `Ready`.

### DIFF
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -255,7 +255,7 @@ if [[ "${ENABLE_DAEMON}" = false ]]; then
 else
     while true; do
         sleep 3
-        kubectl get nodes | grep edge-node | grep -q Ready && break
+        kubectl get nodes | grep edge-node | grep -q -w Ready && break
     done
     kubectl label node edge-node disktype=test
 fi


hack/local-up-kubeedge.sh
```diff
@@ -255,7 +255,7 @@ if [[ "${ENABLE_DAEMON}" = false ]]; then
 else
     while true; do
         sleep 3
-        kubectl get nodes | grep edge-node | grep -q Ready && break
+        kubectl get nodes | grep edge-node | grep -q ^Ready$ && break
```
fisherxu, line 258:
```suggestion
        kubectl get nodes | grep edge-node | grep -q -w Ready && break
```
Just had a double check, `-w` can also do regexp match :)

Shelley-BaoYue: > Just had a double check, `-w` can also do regexp match :)

yes. `-w` is better here.

fisherxu (commented):
/lgtm
/approve

fisherxu:
please cherry-pick to release-1.12 :) @Shelley-BaoYue

PR #4244, opened 2022-09-27
bugfix delete statefulset pod 

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

When delete statefulset pod, cloud will send create new pod and delete old pod request to edge, which these two request have same resource. Because the processing order on edge is not fully guaranteed, a situation may arise: the pod record that create new pod request insert in DB may be deleted by delete old pod request. So when we delete statefulset, new pod can not running. 

In this PR, when we delete pod recored in DB, we will compare the pod UID. If not equal, don't allow to delete this record.

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -1,11 +1,13 @@
 package metamanager
 
 import (
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
 
 	authenticationv1 "k8s.io/api/authentication/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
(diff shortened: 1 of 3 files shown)


edge/pkg/metamanager/process.go
```diff
@@ -247,40 +249,28 @@ func (m *metaManager) processDelete(message model.Message) {
...
 	if resType == model.ResourceTypePod {
-		podStatusKey := strings.Replace(message.GetResource(),
-			constants.ResourceSep+model.ResourceTypePod+constants.ResourceSep,
-			constants.ResourceSep+model.ResourceTypePodStatus+constants.ResourceSep, 1)
-		err = dao.DeleteMetaByKey(podStatusKey)
+		err = processDeletePodDB(message)
 		if err != nil {
-			klog.Errorf("delete meta failed, %s", msgDebugInfo(&message))
-			feedbackError(err, "Error to delete meta to DB", message)
+			klog.Errorf("delete pod meta failed, %s， err:%v", msgDebugInfo(&message), err)
```
fisherxu, line 261:
```suggestion
			klog.Errorf("delete pod meta failed, %s, err:%v", msgDebugInfo(&message), err)
```

edge/pkg/metamanager/process.go
```diff
@@ -293,6 +283,64 @@ func (m *metaManager) processDelete(message model.Message) {
...
+		return err
+	}
+
+	err = json.Unmarshal(msgContent, &msgPod)
+	if err != nil {
+		return err
+	}
+
+	if podDB.UID != msgPod.UID {
+		klog.Info("pod UID is not equal to pod stored in DB, don't need to delete pod DB")
```
fisherxu, line 316:
```suggestion
		klog.Warn("pod UID is not equal to pod stored in DB, don't need to delete pod DB")
```

fisherxu (commented):
/lgtm
/approve

PR #4245, opened 2022-09-27
use kubelet default NodeStatusUpdateFrequency and NodeStatusReportFre…

wackxu:
…quency

**What type of PR is this?**

Fixes #

If we set NodeStatusUpdateFrequency default to 10s, the NodeStatusReportFrequency will also be the same with the NodeStatusUpdateFrequency,  In big scale, this may cause the api server pressure, so we can keep the default with kubelet

```release-note

```

### DIFF
--- a/pkg/apis/componentconfig/edgecore/v1alpha2/default.go
+++ b/pkg/apis/componentconfig/edgecore/v1alpha2/default.go
@@ -40,7 +40,6 @@ func NewDefaultEdgeCoreConfig() *EdgeCoreConfig {
 
 	in := kubeletconfigv1beta1.KubeletConfiguration{}
 	in.ContentType = "application/json"
-	in.NodeStatusUpdateFrequency = metav1.Duration{Duration: constants.DefaultNodeStatusUpdateFrequency}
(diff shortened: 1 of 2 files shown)


pkg/apis/componentconfig/edgecore/v1alpha2/default.go
```diff
@@ -40,7 +40,6 @@ func NewDefaultEdgeCoreConfig() *EdgeCoreConfig {
 
 	in := kubeletconfigv1beta1.KubeletConfiguration{}
 	in.ContentType = "application/json"
-	in.NodeStatusUpdateFrequency = metav1.Duration{Duration: constants.DefaultNodeStatusUpdateFrequency}
```
fisherxu, line 43: The const `NodeStatusUpdateFrequency` can also be removed？

wackxu: done

fisherxu (commented):
/lgtm
/approve

fisherxu:
Need to cherry-pick to release-1.12 :)

PR #4246, opened 2022-09-27
restore device integration test

Shelley-BaoYue:
**What type of PR is this?**

/kind test

deployment integration test is not suitable in new edged which Kubelet integrated, so comment out deployment test temporarily here. And device integration test need start edgecore by itself.


fisherxu (commented):
/lgtm
/approve

PR #4259, opened 2022-09-27
update changelog for 1.12.0, add Important Steps for new Edged

Shelley-BaoYue:
**What type of PR is this?**

/kind documentation

add Important Steps for new edged, for new edged should update the configuration if edged in EdgeCore


fisherxu (commented):
/lgtm
/approve

PR #4267, opened 2022-09-28
Error reported but no log output

JiaweiGithub:

**What type of PR is this?**

 /kind cleanup

Error reported but no log output

Fixes #

```release-note

```


edge/pkg/devicetwin/process.go
```diff
@@ -151,7 +151,8 @@ func SyncDeviceFromSqlite(context *dtcontext.DTContext, deviceID string) error {
 		return err
 	}
 	if len(*devices) <= 0 {
-		return errors.New("Not found device from db")
+		klog.Errorf("not found device from db")
+		return err
```
fisherxu, line 155: The err seems incorrect

JiaweiGithub: You are right, already updated

PR #4282, opened 2022-10-08
make it easier to restart cloudcore in local-up-kubeedge.sh

vincentgoat:

**What type of PR is this?**
/kind cleanup

Currently, we got the error `Both of ca and caKey should be specified!` when we restart cloudcore process.

The ca files and cert files of Cloudcore and Edgecore are in the same directory when we debug in local-up-kubeedge.sh, and only the rootCA.crt exists, the rootCA.key doesn't exist. It will cause Cloudcore can't restart without removing the ca files and cert files.
So we can separate the directory path of Cloudcore and Edgecore.

Fixes #

```release-note

```


vincentgoat:
cc @fisherxu @gy95

gy95:
lgtm

fisherxu (commented):
/approve
/lgtm

PR #4296, opened 2022-10-12
move common application to pkg/metaserver and agent to metamanager

wackxu:

**What type of PR is this?**

/kind cleanup

It is strange that the cloudcore and edgecore code refer to each other, It makes the code hard to read, In this PR, we move the common application and agent to pkg/metaserver. In the follow PR,  I will rename the file and dir name in the cloudcore.

Fixes #

```release-note

```

### DIFF
--- a/pkg/metaserver/application.go
+++ b/pkg/metaserver/application.go
@@ -1,0 +1,251 @@
+/*
+Copyright 2021 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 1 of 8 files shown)


vincentgoat (commented):
lgtm

pkg/metaserver/application.go
```diff
@@ -0,0 +1,234 @@
+/*
+Copyright 2022 The KubeEdge Authors.
```
kevin-wangzefeng, line 2:
```suggestion
Copyright 2021 The KubeEdge Authors.
```

Should be 2021 instead of 2022 here.
The file content was originally written in 2021 and the general rule is that the year to include in a copyright notice is the year of first publication of the work. 

Ref: https://www.copyrightlaws.com/copyright-symbol-notice-year/

> WHICH YEAR SHOULD YOU USE IN YOUR COPYRIGHT NOTICE?
> The general rule is that the year to include in a copyright notice is the year of first publication of the work. First publication is when the work is made available to the public without restriction. This includes selling a book or leaving free copies of it in a public place or posting content on your blog. Publication does not include, for example, circulating copies internally to co-workers or sending copies of a book to a publisher.
> 
> For compilations or derivative works which incorporate previously published content, the year of first publication of the compilation or derivative work is sufficient.

wackxu: done

fisherxu (commented):
/lgtm
/approve

Issue #4297, opened 2022-10-12
Some errors happened when installing kubeedge-1.12.0


bluven:
**What happened**:

I wanna give it a try on kubedge-v1.12.0, but there are some problems happened.

1. cloudcore verison is still v1.11.0 even though I specified 1.12.0
```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "2"
    meta.helm.sh/release-name: cloudcore
    meta.helm.sh/release-namespace: kubeedge
  creationTimestamp: "2022-10-12T03:47:23Z"
  generation: 2
  labels:
    app.kubernetes.io/managed-by: Helm
    k8s-app: kubeedge
    kubeedge: cloudcore
  name: cloudcore
  namespace: kubeedge
  resourceVersion: "13541"
  uid: cee3f6c2-ded7-463f-8bdf-deaf7f0facae
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      k8s-app: kubeedge
      kubeedge: cloudcore
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        k8s-app: kubeedge
        kubeedge: cloudcore
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/edge
                operator: DoesNotExist
      containers:
      - image: kubeedge/cloudcore:v1.11.1
        imagePullPolicy: IfNotPresent
        name: cloudcore
        ports:
        - containerPort: 10000
          hostPort: 10000
          name: cloudhub
          protocol: TCP
        - containerPort: 10001
          hostPort: 10001
          name: cloudhub-quic
          protocol: TCP
        - containerPort: 10002
          hostPort: 10002
          name: cloudhub-https
          protocol: TCP
        - containerPort: 10003
          hostPort: 10003
          name: cloudstream
          protocol: TCP
        - containerPort: 10004
          hostPort: 10004
          name: tunnelport
          protocol: TCP
        resources:
          limits:
            cpu: 200m
            memory: 1Gi
          requests:
            cpu: 100m
            memory: 512Mi
        securityContext:
          privileged: true
[...]

PR #4299, opened 2022-10-12
keadm reset supports remote runtime remove mqtt container

gy95:
**What type of PR is this?**

support remote runtime in `keadm reset` cmd, and also Fixes #https://github.com/kubeedge/kubeedge/issues/4297#issuecomment-1275667908

```release-note

```


keadm/cmd/keadm/app/cmd/common/types.go
```diff
@@ -87,8 +87,10 @@ type CollectOptions struct {
 }
 
 type ResetOptions struct {
-	Kubeconfig string
-	Force      bool
+	Kubeconfig  string
+	Force       bool
+	RuntimeType string
+	Endpoint    string
```
fisherxu, line 93: Is it possible to wrap some common global flags as in kubeadm, so all the commands can use them instead of defining for each one :) @gy95

keadm/cmd/keadm/app/cmd/util/image.go
```diff
@@ -312,6 +340,31 @@ func (runtime *CRIRuntime) RunMQTT(mqttImage string) error {
 	return runtime.RuntimeService.StartContainer(containerID)
 }
 
+func (runtime *CRIRuntime) RemoveMQTT() error {
+	sandboxFilter := &runtimeapi.PodSandboxFilter{
```
fisherxu, line 344: Another question, is it needed to use cri for create a container in edgenode? Or remote runtime, we can only use the cri. cc @zc2638

zc2638: Sorry, I may not have understood what you meant.
What you're trying to say is that whatever "remote runtime" is, just consider using CRI, not docker?

fisherxu: I mean there's no need to use cri for creating mqtt container. Whether can we just use  the oci api to create?

zc2638: OCI includes runtime spec, image spec, and distribution spec, but I can't find the corresponding interface in opencontainers. 
Can I see the available associated interfaces?

fisherxu (commented):
/lgtm
/approve

Issue #4327, opened 2022-10-21
Umbrella issue for test cases enhancement


vincentgoat:
## Description
We summarized some task items as follows for KubeEdge code quality testing, please comment with one or more items' names if you are interested to join.

### Unit Test
- [x] integrated Codecov into CI/CD -- @gy95 
#### Complete UT cases in the following directory to improve the coverage to at least 80%.
- [ ] kubeedge/cloud/pkg/admissioncontroller @gy95 
- [ ] kubeedge/cloud/pkg/cloudhub @wackxu 
  - kubeedge/cloud/pkg/cloudhub/servers/httpserver/updateconfig_test.go https://github.com/kubeedge/kubeedge/pull/4461 
- [ ] kubeedge/cloud/pkg/cloudstream @wackxu 
- [ ] kubeedge/cloud/pkg/csidriver
  - kubeedge/cloud/pkg/csidriver/uds_test.go https://github.com/kubeedge/kubeedge/pull/4446
- [ ] kubeedge/cloud/pkg/devicecontroller @RyanZhaoXB 
  - kubeedge/cloud/pkg/devicecontroller/manager/devicemodel_test.go https://github.com/kubeedge/kubeedge/pull/4467 
- [ ] kubeedge/cloud/pkg/dynamiccontroller @wackxu 
- [ ] kubeedge/cloud/pkg/edgecontroller @vincentgoat 
- [ ] kubeedge/cloud/pkg/nodeupgraderjobcontroller @gy95 
- [ ] kubeedge/cloud/pkg/router @gy95 
- [x] kubeedge/cloud/pkg/synccontroller @vincentgoat 
- [ ] kubeedge/edge/devicetwin @ksankeerth 
- [x] kubeedge/edge/pkg/edgehub -- @ksankeerth 
- [ ] kubeedge/edge/pkg/eventbus @Nageshbansal 
- [ ] kubeedge/edge/pkg/servicebus @gy95 
- [ ] kubeedge/edge/pkg/edgestream @fisherxu 
- [ ] kubeedge/edge/pkg/metamanager @Shelley-BaoYue 
  - kubeedge/edge/pkg/metamanager/client/nodestatus_test.go https://github.com/kubeedge/kubeedge/pull/4502 
- [ ] kubeedge/edge/pkg/edged @Shelley-BaoYue 
- [ ] kubeedge/keadm/cmd/keadm/app @gy95 
[...]

PR #4328, opened 2022-10-23
[UT] pkg/image enhancement

my-git9:
**What type of PR is this?**
/kind test
/kind cleanup

part of: https://github.com/kubeedge/kubeedge/issues/4327

[UT] pkg/image enhancement

```release-note

```

### DIFF
--- a/pkg/image/image_test.go
+++ b/pkg/image/image_test.go
@@ -72,3 +72,154 @@ func TestEdgeSet(t *testing.T) {
 		})
 	}
 }
+
+func TestCloudSet(t *testing.T) {
+	type args struct {
+		imageRepository string
+		version         string
+	}
+	tests := []struct {
+		name string
+		args args
+		want Set
+	}{
+		{
+			name: "repo nil, ver not nil",
+			args: args{
+				imageRepository: "",
+				version:         "v1.12.0",
+			},
+			want: Set{
+				CloudAdmission:         "kubeedge/admission:v1.12.0",
+				CloudCloudcore:         "kubeedge/cloudcore:v1.12.0",
+				CloudIptablesManager:   "kubeedge/iptables-manager:v1.12.0",
+				CloudControllerManager: "kubeedge/controller-manager:v1.12.0",
+			},
+		},
+		{
+			name: "repo nil, ver nil",
+			args: args{
+				imageRepository: "",
+				version:         "",
+			},
+			want: Set{
+				CloudAdmission:         "kubeedge/admission",
+				CloudCloudcore:         "kubeedge/cloudcore",
+				CloudIptablesManager:   "kubeedge/iptables-manager",
+				CloudControllerManager: "kubeedge/controller-manager",
+			},
+		},
+		{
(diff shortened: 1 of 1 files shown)


pkg/image/image_test.go
```diff
@@ -72,3 +72,154 @@ func TestEdgeSet(t *testing.T) {
...
+func TestSet_Get(t *testing.T) {
+	tests := []struct {
+		name string
+		s    Set
+		args string
+		want string
+	}{
+		{
+			name: "get ver",
+			s:    Set{"name": "foo", "ver": "bar"},
```
zc2638, line 157: Thanks for your test submission.
It is recommended to give some meaningful test content, otherwise for the test, the actual effect of the method may not be reflected.

pkg/image/image_test.go
```diff
@@ -72,3 +72,154 @@ func TestEdgeSet(t *testing.T) {
...
+		s    Set
+		args string
+		want string
+	}{
+		{
+			name: "get image version",
+			s:    Set{"kubeedge-test/cloudcore": "1.12.0"},
+			args: "kubeedge-test/cloudcore",
+			want: "1.12.0",
+		},
```
zc2638, line 160:
```suggestion
		{
			name: "get cloudcore image",
			s:    Set{"cloudcore": "kubeedge-test/cloudcore:1.12.0"},
			args: "cloudcore",
			want: "kubeedge-test/cloudcore:1.12.0",
		},
```
This case needs to be modified, everything else is ok.

my-git9: Adjusted, thanks

fisherxu (commented):
/approve

pkg/image/image_test.go
```diff
@@ -72,3 +72,154 @@ func TestEdgeSet(t *testing.T) {
...
+func TestSet_List(t *testing.T) {
+	tests := []struct {
+		name string
+		s    Set
+		want []string
+	}{
+		{
+			name: "test list",
+			s:    Set{"cloudcore": "kubeedge-test/cloudcore:v1.12.0", "admission": "kubeedge-test/admission:v1.12.0", "controller-manager": "kubeedge-test/controller-manager:v1.12.0", "iptables-manager": "kubeedge-test/iptables-manager:v1.12.0"},
+			want: []string{"kubeedge-test/cloudcore:v1.12.0", "kubeedge-test/admission:v1.12.0", "kubeedge-test/controller-manager:v1.12.0", "kubeedge-test/iptables-manager:v1.12.0"},
```
gy95, line 180: https://pkg.go.dev/reflect#DeepEqual
Array values are deeply equal when their corresponding elements are deeply equal.

ref: https://go.dev/blog/maps
When iterating over a map with a range loop, the iteration order is not specified and is not guaranteed to be the same from one iteration to the next. If you require a stable iteration order you must maintain a separate data structure that specifies that order. This example uses a separate sorted slice of keys to print a map[int]string in key order:

So this test is not right?
@my-git9

my-git9: I'm sorry for that

PR #4330, opened 2022-10-23
change namespace value to Release.Namespace

my-git9:
**What type of PR is this?**
/kind cleanup

The value of namespace should be written as `Release.namespace` as elsewhere, not a fixed value

Fixes #

```release-note

```

### DIFF
--- a/manifests/charts/cloudcore/templates/deployment_controllermanager.yaml
+++ b/manifests/charts/cloudcore/templates/deployment_controllermanager.yaml
@@ -9,7 +9,7 @@ metadata:
   annotations: {{- toYaml . | nindent 4 }}
   {{- end }}
   name: kubeedge-controller-manager
-  namespace: kubeedge
+  namespace: {{ .Release.Namespace }}
 spec:
   selector:
     {{- with .Values.controllerManager.labels }}
--- a/manifests/charts/cloudcore/templates/rbac_controllermanager.yaml
+++ b/manifests/charts/cloudcore/templates/rbac_controllermanager.yaml
@@ -23,7 +23,7 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: controller-manager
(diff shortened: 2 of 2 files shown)


wackxu:
lgtm

manifests/charts/cloudcore/templates/deployment_controllermanager.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
   annotations: {{- toYaml . | nindent 4 }}
   {{- end }}
   name: kubeedge-controller-manager
-  namespace: kubeedge
+  namespace: {{ .Release.Namespace }}
```
gy95, line 12: i can not find where Release.Namespace come from?

my-git9: It comes from the namespace where the chart was released to.

gy95: can you point it out in source codes? thanks

my-git9: Is the document ok? https://helm.sh/docs/topics/charts/#predefined-values
image

gy95: lgtm
/cc @zhu733756 @zc2638

zc2638:
I have a question. Other resources are missing the specified namespace. 
need to add them?
@my-git9

my-git9:
> I have a question. Other resources are missing the specified namespace. need to add them? @my-git9

Not specifying under `meta.namespace` has no effect, but the namespace of serviceaccout specified in rbac's rolebinding and clusterrolebinding must be specified. In other places about rbac, it is already `{{ .Release.Namespace }}`

PR #4345, opened 2022-10-26
edge node attestation via cert common name

vincentgoat:

**What type of PR is this?**
/feature

CloudHub will attest edge node cert via cert common name (format as: kubeedge.io:node:{node_name}).
The attestation result is `edge-node-1` can not register with the cert of `edge-node-2`. Only corresponding nodes and certificates can be authenticated.

Adding featureGates `nodeAttestation` to determine if attesting node cert while registering. The default value is false means won't execute attesting in Alpha phase, it will set to `true` by default in the beta phase.
Examples:
```
featureGates:
 nodeAttestation: true
```

Fixes #

**Does this PR introduce a user-facing change?**: **YES**
Check the details above.

```release-note

```


vincentgoat:
/hold 
dependences error, need to dump K8s version and golang version.

vincentgoat:
/unhold
Bump quic version to v0.24.0

vincentgoat:
Updates:
As of now KubeEdge uses quic-go v0.10.1 to run both quic client and server.
There is a compatibility issue of quic version when trying to upgrade quic version, thinking of the case that deployments that have existing deployments of quic-go client and server that cannot be updated all at the same time.

quic-go client version v0.10.1 <-----> quic-go server version v0.10.1: 
can't validate the client tls certificate hostname on the server side because quic-go did not support.

quic-go client version v0.10.1 <-----> v0.11.0 <= quic-go server version v0.11.0 < v0.18.0: can't run quic-go server because the init function in the file internal/handshake/unsafe.go will validate qtls and tls, and the validation results were failed when building with >= golang 1.16

quic-go client v0.10.1 <-----> quic-go version >= v0.18.1: 
InvalidVersionNegotiationPacket error when negotiating, the format of the Version Negotiation Packet is not compatible.

PR #4346, opened 2022-10-26
add healthcheck and metrics for cloudcore

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

add healthcheck and metrics for cloudcore, include NodeCountMetric, NodeCountDiffRateMetric and ClusterMasterConnectionMetric.

### DIFF
--- a/cloud/pkg/cloudhub/metric/server.go
+++ b/cloud/pkg/cloudhub/metric/server.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2022 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metric
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+
+	"github.com/gorilla/mux"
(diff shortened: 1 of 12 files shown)


cloud/pkg/cloudhub/metric/server.go
```diff
@@ -0,0 +1,95 @@
...
+	router.HandleFunc(config.HealthPatten, healthHandler)
+
+	cert, err := tls.X509KeyPair(hubconfig.Config.Cert, hubconfig.Config.Key)
+	if err != nil {
+		klog.Warningf("generate key pair failed with error: %s", err)
+	}
+
+	tlsConfig := &tls.Config{
+		Certificates:       []tls.Certificate{cert},
+		InsecureSkipVerify: insecureSkipVerify,
```
vincentgoat, line 69: InsecureSkipVerify controls whether a client verifies the server's certificate chain and host name.
As for an HTTPS server, there is no need to set this value.

cloud/pkg/cloudhub/metric/server.go
```diff
@@ -0,0 +1,95 @@
...
+func StartMetricServer(config *cloudcoreConfig.CommonConfig, insecureSkipVerify bool) {
+	address := fmt.Sprintf("%s:%d", config.Host, config.MetricPort)
+
+	router := mux.NewRouter()
+	router.HandleFunc(config.MetricPatten, metricHandler)
+	router.HandleFunc(config.HealthPatten, healthHandler)
+
+	cert, err := tls.X509KeyPair(hubconfig.Config.Cert, hubconfig.Config.Key)
+	if err != nil {
+		klog.Warningf("generate key pair failed with error: %s", err)
```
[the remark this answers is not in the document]
Shelley-BaoYue: > Return here if an error happens?

done.

cloud/pkg/cloudhub/metric/server.go
Shelley-BaoYue: > InsecureSkipVerify controls whether a client verifies the server's certificate chain and host name. As for an HTTPS server, there is no need to set this value.

done

PR #4356, opened 2022-10-31
keadm: support --namespace flag to install kubeedge in other namespace

my-git9:
Signed-off-by: xin.li <xin.li@daocloud.io>

**What type of PR is this?**

keadm: support --namespace flag to install kubeedge in other namespace

Fixes #

```release-note
keadm: support --namespace flag to install kubeedge in other namespace
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cloud/init.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/init.go
@@ -101,6 +101,9 @@ func addInitOtherFlags(cmd *cobra.Command, initOpts *types.InitOptions) {
 	cmd.Flags().StringVarP(&initOpts.Manifests, types.Files, "f", initOpts.Manifests,
 		"Allow appending file directories of k8s resources to keadm, separated by commas")
 
+	cmd.Flags().StringVarP(&initOpts.Namespace, types.Namespace, "n", constants.SystemNamespace,
+		"Use this key to set the namespace")
+
 	cmd.Flags().BoolVarP(&initOpts.DryRun, types.DryRun, "d", initOpts.DryRun,
 		"Print the generated k8s resources on the stdout, not actual excute. Always use in debug mode")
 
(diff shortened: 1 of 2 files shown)


my-git9:
The test result:
```shell
$ keadm init -n kt version=v1.20.0
Kubernetes version verification passed, KubeEdge installation will start...
CLOUDCORE started
=========CHART DETAILS=======
NAME: cloudcore
LAST DEPLOYED: Mon Oct 31 14:36:14 2022
NAMESPACE: kt
STATUS: deployed
REVISION: 1
$
$ kubectl get po -n kt
NAME                         READY   STATUS    RESTARTS   AGE
cloudcore-86df766bdb-j56rp   1/1     Running   0          43m
```

keadm/cmd/keadm/app/cmd/cloud/init.go
```diff
@@ -148,7 +150,7 @@ func AddInit2ToolsList(toolList map[string]types.ToolsInstaller, initOpts *types
 		AdvertiseAddress: initOpts.AdvertiseAddress,
 		KubeEdgeVersion:  initOpts.KubeEdgeVersion,
 		Manifests:        initOpts.Manifests,
-		Namespace:        constants.SystemNamespace,
+		Namespace:        initOpts.Namespace,
```
wackxu, line 153: There are many places where  `constants.SystemNamespace` is used, It's difficult to eliminate these uses

my-git9: But I feel that `namespace` can be better selected by users themselves

my-git9: We can gradually optimize this starting from the installation. What do you think？

my-git9:
There are many places where constants.SystemNamespace is used, I will submit an issue to discuss whether to modify it.

PR #4359, opened 2022-10-31
support gloal image Registry for chart

my-git9:
**What type of PR is this?**
/kind feature

support gloal image Registry for chart

Fixes #

The test result:
```shell
$ helm upgrade --install kubeedge ./ --dry-run |grep image
        image: docker.io/kubeedge/cloudcore:v1.12.0
        imagePullPolicy: IfNotPresent
$ helm upgrade --set global.imageRegistry=10.0.0.1 --install kubeedge ./ --dry-run |grep image
        image: 10.0.0.1/kubeedge/cloudcore:v1.12.0
        imagePullPolicy: IfNotPresent
```

```release-note
chart: support gloal image Registry for chart 
```

### DIFF
--- a/manifests/charts/cloudcore/values.yaml
+++ b/manifests/charts/cloudcore/values.yaml
@@ -3,10 +3,14 @@
 # Declare variables to be passed into your templates.
 appVersion: "1.12.0"
 
+global:
+  imageRegistry: ""
+
 cloudCore:
   replicaCount: 1
   hostNetWork: "true"
   image:
+    registry: docker.io
     repository: "kubeedge/cloudcore"
     tag: "v1.12.0"
     pullPolicy: "IfNotPresent"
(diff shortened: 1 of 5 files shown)


manifests/charts/cloudcore/values.yaml
```diff
@@ -3,10 +3,14 @@
...
 appVersion: "1.12.0"
 
+global:
+  imageRegistry: ""
+
 cloudCore:
   replicaCount: 1
   hostNetWork: "true"
   image:
+    registry: docker.io
```
gy95, line 13: if we specify this two registry both, what's the expected behavior?

my-git9: If the images come from different registry (which is rare), we do not need to configure global.
 If they come from the same registry, you can configure global. 
Global first

zhu733756: I agree with @gy95, we had better simplify the codes to avoid bringing confusion.

my-git9: ok, I will close it

PR #4362, opened 2022-10-31
improve  example output about kubeedge

my-git9:
**What type of PR is this?**
improve  example output about kubeedge
/kind cleanup

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cloud/init.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/init.go
@@ -40,7 +40,7 @@ If not installed, please install the Kubernetes first.
 keadm init
 - This command will render and install the Charts for Kubeedge cloud component
 
-keadm init --advertise-address=127.0.0.1 --profile version=v1.9.0 --kube-config=/root/.kube/config
+keadm init --advertise-address=127.0.0.1 --profile version=v%s --kube-config=/root/.kube/config
   - kube-config is the absolute path of kubeconfig which used to secure connectivity between cloudcore and kube-apiserver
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/cloud/init.go
```diff
@@ -57,7 +57,7 @@ func NewCloudInit() *cobra.Command {
 		Use:     "init",
 		Short:   "Bootstraps cloud component. Checks and install (if required) the pre-requisites.",
 		Long:    cloudInitLongDescription,
-		Example: cloudInitExample,
+		Example: fmt.Sprintf(cloudInitExample, types.DefaultKubeEdgeVersion),
```
gy95, line 60: this is not correct for now. `types.DefaultKubeEdgeVersion` is 1.12.0 but without prefix `v`
Now we distiguish `v1.12.0`(with prefix `v`) and `1.12.0` (without prefix `v`) in installation, some errors happen when we mix them up.
we should optimize it, so that installation is always successful during installation both with and without prefix.
@my-git9

gy95: you can try it if you have interest :)

my-git9: The test is ok:
image

gy95: yes, it is.
I mean there're some other codes using kubeedge-version like keadm join/deprecated init/deprecated join etc. these code can be optimized.

zhu733756:
LGTM

keadm/cmd/keadm/app/cmd/cloud/init.go
my-git9: OK, I can try it later, thanks

fisherxu (commented):
/approve

PR #4366, opened 2022-11-01
[fix] format var and sort import

donghui12:
**What type of PR is this?**
/kind bug

fix var name and sort import

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/eventbus/common/util/common.go
+++ b/edge/pkg/eventbus/common/util/common.go
@@ -16,8 +16,8 @@ import (
 var (
 	// TokenWaitTime to wait
 	TokenWaitTime = 120 * time.Second
-	// Loop connect to wait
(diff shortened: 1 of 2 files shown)


edge/pkg/servicebus/servicebus.go
```diff
@@ -10,10 +10,10 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/astaxie/beego/orm"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/klog/v2"
 
+	"github.com/astaxie/beego/orm"
```
Shelley-BaoYue, line 16: "github.com/astaxie/beego/orm"  should be placed at the second part which is for third-party library.  The third part is for kubeedge itself.

gy95:
please fix DCO @donghui12

PR #4368, opened 2022-11-02
fix force delete pod when node offline

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

fix the same bug with this pr #4141 since v1.12 version

when node is offline and we force delete the pod in the node, and then the edge node is online,
the deleted pod container is always in the edge node unless we restart the edgecore.


gy95:
not stale, continue?

fisherxu (commented):
/lgtm
/approve

PR #4373, opened 2022-11-03
donot trigger ci when only docs changed

gy95:
**What type of PR is this?**
/kind cleanup

donot trigger ci when only docs changed

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,20 @@
 name: CIFuzz
-on: [pull_request]
+on: 
+  push:
+    branches:
+      - master
+    tags:
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - '**/OWNERS'
+      - '**/MAINTAINERS'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - '**/OWNERS'
+      - '**/MAINTAINERS'
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -2,8 +2,18 @@ name: FOSSA
 on:
   push:
     branches: [master]
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - '**/OWNERS'
+      - '**/MAINTAINERS'
   pull_request:
     branches: [master]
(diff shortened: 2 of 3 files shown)


.github/workflows/cifuzz.yml
```diff
@@ -1,5 +1,14 @@
 name: CIFuzz
-on: [pull_request]
+on: 
+  push:
+    branches:
+      - master
+    tags:
+    paths-ignore:
+      - '**.md'
```
vincentgoat, line 8: Adding this one?
- 'docs/**'
- 'hack/**'
- 'CHANGELOG/**'

gy95: files need ignoring include the below files.
docs/**(include some images) **.md OWNERS MAINTAINERS
I think this is all.

.github/workflows/cifuzz.yml
```diff
@@ -1,5 +1,20 @@
...
+      - 'docs/**'
+      - 'OWNERS'
+      - 'MAINTAINERS'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
```
fisherxu, line 15: Small question, `**` is the grammer in github action? :) @gy95

.github/workflows/cifuzz.yml
fisherxu: > 'hack/**'

There are many scripts under hack dir, they need to run the ci check. :)

fisherxu: There are many scripts under hack, they need to be checked. :)

gy95: yes, hack dir should be checked

.github/workflows/cifuzz.yml
gy95: can match more string than only one `*`
ref: https://docs.github.com/cn/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

gy95: /hold need some fix

fisherxu: Is `**/OWNERS` needed?

gy95: /hold cancel
fixed

fisherxu (commented):
/lgtm
/approve

PR #4375, opened 2022-11-03
add generic informer

wackxu:

**What type of PR is this?**

/kind feature

Fixes #

For now, we use dynamic informer in the dynamic controller and use built-in informer in the edgecontroller, So If user use the list-watch and enable dynamic controller, there will be two informer and for mainly used k8s built-in resources, there maybe two informers for one resources and it will increase the memory usage for cloudcore, especially in large scale scenarios.

And also in the synccontroller, we have a period sync for objectsync, and for every objectsync, we need access the kube-apiserver to make sure the resource exists or not, and this increase the burden of kube-apiserver, so we change to get the resource from the informer cache to reduce the call to the apiserver

```release-note

```


cloud/pkg/common/informers/informers.go
```diff
@@ -17,74 +17,148 @@ limitations under the License.
...
+	GetKubeInformerFactory() k8sinformer.SharedInformerFactory
+
+	// GetKubeEdgeInformerFactory return KubeEdge CRD resources InformerFactory
+	GetKubeEdgeInformerFactory() crdinformers.SharedInformerFactory
+
+	// GetDynamicInformerFactory return third-party CRD resources InformerFactory
+	GetDynamicInformerFactory() dynamicinformer.DynamicSharedInformerFactory
+
+	// KubeEdgeCustomInformer return custom informer
 	KubeEdgeCustomInformer
```
gy95, line 80: can delete KubeEdgeCustomInfomer in another PR?

gy95 (approved):
can merge this patch first, and do another refactor

cloud/pkg/dynamiccontroller/application/eventhandler.go
```diff
@@ -87,10 +86,10 @@ type CommonResourceEventHandler struct {
 	listenersLock sync.RWMutex
 	messageLayer  messagelayer.MessageLayer
 	gvr           schema.GroupVersionResource
-	informer      informers.GenericInformer
+	informer      *genericinformers.InformerPair
```
gy95, line 89: InformerPairs is only used here, can be refactored

cloud/pkg/common/informers/informers.go
```diff
@@ -17,74 +17,148 @@ limitations under the License.
...
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/common/constants"
+	edgescheme "github.com/kubeedge/kubeedge/pkg/client/clientset/versioned/scheme"
 	crdinformers "github.com/kubeedge/kubeedge/pkg/client/informers/externalversions"
 )
 
+var (
+	// KubeScheme is built-in resources scheme for converting group,
+	// version, and kind information to and from Go schemas
+	KubeScheme = scheme.Scheme
```
gy95, line 46: not expost this value.

gy95 (approved):
can do this in another PR.

cloud/pkg/common/informers/informers.go
wackxu: done

cloud/pkg/common/informers/informers.go
wackxu: For now, we can not delete this KubeEdgeCustomInfomer since some customer informer need filter the resource

cloud/pkg/dynamiccontroller/application/eventhandler.go
wackxu: I don't quite understand what you mean, can you elaborate?

wackxu: updated, PTAL

fisherxu (commented):
/lgtm
/approve

PR #4376, opened 2022-11-04
auto adjust cloudcore config

wackxu:

**What type of PR is this?**

/kind feature

Fixes #

implementation of https://github.com/kubeedge/kubeedge/pull/4553

```release-note
NodeLimit is a maximum number of edge node that can connect to the single CloudCore instance. You should take this parameter seriously, because this parameter is closely related to the number of goroutines for message processing.
The edgecontroller config will be auto adjust according to the nodeLimit config
```

### DIFF
--- a/cloud/cmd/cloudcore/app/server.go
+++ b/cloud/cmd/cloudcore/app/server.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/component-base/cli/globalflag"
 	"k8s.io/component-base/term"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
 
 	"github.com/kubeedge/beehive/pkg/core"
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
@@ -101,6 +102,10 @@ kubernetes controller which manages devices so that the device metadata/status d
 			}
 
 			config.CommonConfig.TunnelPort = *tunnelport
+
+			if changed := v1alpha1.AdjustCloudCoreConfig(config); changed {
+				updateCloudCoreConfigMap(config)
+			}
 
 			gis := informers.GetInformersManager()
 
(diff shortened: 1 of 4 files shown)


common/constants/default.go
```diff
@@ -75,26 +77,27 @@ const (
...
-	DefaultQueryConfigMapWorkers             = 4
-	DefaultQuerySecretWorkers                = 4
+	DefaultQueryConfigMapWorkers             = 100
+	DefaultQuerySecretWorkers                = 100
 	DefaultQueryPersistentVolumeWorkers      = 4
 	DefaultQueryPersistentVolumeClaimWorkers = 4
 	DefaultQueryVolumeAttachmentWorkers      = 4
-	DefaultCreateNodeWorkers                 = 4
+	DefaultCreateNodeWorkers                 = 100
 	DefaultPatchNodeWorkers                  = 4
```
gy95, line 92: remove Node related values?

cloud/cmd/cloudcore/app/server.go
```diff
@@ -102,6 +103,10 @@ kubernetes controller which manages devices so that the device metadata/status d
 
 			config.CommonConfig.TunnelPort = *tunnelport
 
+			if changed := v1alpha1.AdjustCloudCoreConfig(config); changed {
+				updateCloudCoreConfigMap(config)
```
gy95, line 107: also write back to local configure file? move this operation to line84, before ValidateCloudCoreConfiguration?

common/constants/default.go
wackxu: done

gy95:
link the proposal :)

cloud/cmd/cloudcore/app/server.go
wackxu: the local configure file that mouted by the configmap will be updated automatically. and the autoadjust configuration is compliant with validation requirements

vincentgoat:
Is there any metrics data to support this adjustment?

wackxu:
> Is there any metrics data to support this adjustment?

The calculation formula is based on previous large-scale test experience and We can make adjustments based on subsequent user feedback

common/constants/default.go
```diff
@@ -79,26 +81,24 @@ const (
 	DefaultKubeNamespace           = v1.NamespaceAll
 	DefaultKubeQPS                 = 100.0
 	DefaultKubeBurst               = 200
+	DefaultNodeLimit               = 500
```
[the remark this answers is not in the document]
wackxu: We calculate the number of workers according to nodeLimit，I think nodeLimit default value is too big, because there are few users with this node size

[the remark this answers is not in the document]
fisherxu: Okey.

fisherxu (commented):
/lgtm
/approve

PR #4377, opened 2022-11-04
remove pod cache

wackxu:

**What type of PR is this?**

/kind feature

Fixes #
We don't need to cache pod data, in large-scale scenarios, cloudcore takes up too much memory, so we remove pod cache
and add filter for eventhandler

```release-note

```

### DIFF
--- a/cloud/pkg/edgecontroller/manager/common.go
+++ b/cloud/pkg/edgecontroller/manager/common.go
@@ -15,7 +15,8 @@ type Manager interface {
 
 // CommonResourceEventHandler can be used by configmapManager and podManager
 type CommonResourceEventHandler struct {
-	events chan watch.Event
+	events      chan watch.Event
+	eventFilter EventFilter
 }
 
 func (c *CommonResourceEventHandler) obj2Event(t watch.EventType, obj interface{}) {
@@ -36,20 +37,29 @@ func (c *CommonResourceEventHandler) obj2Event(t watch.EventType, obj interface{
 
 // OnAdd handle Add event
 func (c *CommonResourceEventHandler) OnAdd(obj interface{}) {
+	if c.eventFilter != nil && !c.eventFilter.Create(obj) {
+		return
+	}
 	c.obj2Event(watch.Added, obj)
 }
 
 // OnUpdate handle Update event
 func (c *CommonResourceEventHandler) OnUpdate(oldObj, newObj interface{}) {
+	if c.eventFilter != nil && !c.eventFilter.Update(oldObj, newObj) {
+		return
+	}
 	c.obj2Event(watch.Modified, newObj)
 }
 
 // OnDelete handle Delete event
 func (c *CommonResourceEventHandler) OnDelete(obj interface{}) {
+	if c.eventFilter != nil && !c.eventFilter.Delete(obj) {
+		return
+	}
 	c.obj2Event(watch.Deleted, obj)
 }
 
 // NewCommonResourceEventHandler create CommonResourceEventHandler used by configmapManager and podManager
-func NewCommonResourceEventHandler(events chan watch.Event) *CommonResourceEventHandler {
-	return &CommonResourceEventHandler{events: events}
+func NewCommonResourceEventHandler(events chan watch.Event, filter EventFilter) *CommonResourceEventHandler {
+	return &CommonResourceEventHandler{events: events, eventFilter: filter}
 }
--- a/cloud/pkg/edgecontroller/manager/event_filter.go
+++ b/cloud/pkg/edgecontroller/manager/event_filter.go
@@ -1,0 +1,29 @@
+/*
+Copyright 2022 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package manager
+
+// Filter filter the event before enqueuing the key
+type EventFilter interface {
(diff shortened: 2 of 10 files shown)


cloud/pkg/edgecontroller/manager/common.go
```diff
@@ -8,14 +8,17 @@ import (
...
 // Manager define the interface of a Manager, configmapManager and podManager implement it
 type Manager interface {
 	Events() chan watch.Event
 }
 
 // CommonResourceEventHandler can be used by configmapManager and podManager
 type CommonResourceEventHandler struct {
-	events chan watch.Event
+	events      chan watch.Event
+	eventFilter EventFilterFunc
```
gy95, line 21: changed to updateFilter? now it's only for update.
And I think whether we can also add `AddFilter` `DeleteFilter`, so that we can also filter some on add or delete, like filter out Not edge node related resources

wackxu:
```
// Filter filter the event before enqueuing the key
type EventFilter interface {
	// Create returns true if the Create event should be processed
	Create(obj interface{}) bool

	// Delete returns true if the Delete event should be processed
	Delete(obj interface{}) bool

	// Update returns true if the Update event should be processed
	Update(oldObj, newObj interface{}) bool
}
```

wackxu: done

cloud/pkg/edgecontroller/manager/event_filter.go
```diff
@@ -0,0 +1,29 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package manager
+
+// Filter filter the event before enqueuing the key
+type EventFilter interface {
```
fisherxu, line 20: Looks good, and is it possible to use the `FilteringResourceEventHandler` in client-go? Ref: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/controller.go#L255

gy95: K8s do this like 
https://github.com/kubernetes/kubernetes/blob/8e48df135318026d5f8a972a96a2ff665889568a/pkg/scheduler/eventhandlers.go#L258-L285
do we need to use k8s structure directly? or just use store resource to channel

wackxu: No，`FilteringResourceEventHandler ` does not fit our use case, Only one object can be filtered. But our use case in most scenes is to filter the updat event to compare the old and new object.

wackxu: > K8s do this like https://github.com/kubernetes/kubernetes/blob/8e48df135318026d5f8a972a96a2ff665889568a/pkg/scheduler/eventhandlers.go#L258-L285 do we need to use k8s structure directly? or just use store resource to channel


In k8s, most of the usage methods are as follows, first compare old obj and new obj, if it passes the filter, then join the queue

```
// When a pod is updated, figure out what job/s manage it and wake them up.
// If the labels of the pod have changed we need to awaken both the old
// and new job. old and cur must be *v1.Pod types.
func (jm *Controller) updatePod(old, cur interface{}) {
	curPod := cur.(*v1.Pod)
	oldPod := old.(*v1.Pod)
	if curPod.ResourceVersion == oldPod.ResourceVersion {
		// Periodic resync will send update events for all known pods.
		// Two different versions of the same pod will always have different RVs.
		return
	}
	if curPod.DeletionTimestamp != nil {
		// when a pod is deleted gracefully it's deletion timestamp is first modified to reflect a grace period,
		// and after such time has passed, the kubelet actually deletes it from the store. We receive an update
		// for modification of the deletion timestamp and expect an job to create more pods asap, not wait
		// until the kubelet actually deletes the pod.
		jm.deletePod(curPod)
		return
	}

	// the only time we want the backoff to kick-in, is when the pod failed
	immediate := curPod.Status.Phase != v1.PodFailed

	curControllerRef := metav1.GetControllerOf(curPod)
	oldControllerRef := metav1.GetControllerOf(oldPod)
	controllerRefChanged := !reflect.DeepEqual(curControllerRef, oldControllerRef)
	if controllerRefChanged && oldControllerRef != nil {
		// The ControllerRef was changed. Sync the old controller, if any.
		if job := jm.resolveControllerRef(oldPod.Namespace, oldControllerRef); job != nil {
			jm.enqueueController(job, immediate)
		}
	}

	// If it has a ControllerRef, that's all that matters.
	if curControllerRef != nil {
		job := jm.resolveControllerRef(curPod.Namespace, curControllerRef)
		if job == nil {
			return
		}
		jm.enqueueController(job, immediate)
		return
	}

	// Otherwise, it's an orphan. If anything changed, sync matching controllers
	// to see if anyone wants to adopt it now.
	labelChanged := !reflect.DeepEqual(curPod.Labels, oldPod.Labels)
	if labelChanged || controllerRefChanged {
		for _, job := range jm.getPodJobs(curPod) {
			jm.enqueueController(job, immediate)
		}
	}
}
```

cloud/pkg/edgecontroller/manager/pod.go
```diff
@@ -2,36 +2,25 @@ package manager
...
-
-	// pods, key is UID, value is *v1.Pod
-	pods sync.Map
 }
 
-func (pm *PodManager) isPodUpdated(old *CachePod, new *v1.Pod) bool {
+func isPodUpdated(old *v1.Pod, new *v1.Pod) bool {
 	// does not care fields
 	old.ObjectMeta.ResourceVersion = new.ObjectMeta.ResourceVersion
 	old.ObjectMeta.Generation = new.ObjectMeta.Generation
```
gy95, line 26: from k8s description, we cannot modify the resource from event, seems modifying the old value is not right

wackxu: update the parameter to struct type and we will not update the filed in the cache

vincentgoat (commented):
/lgtm

fisherxu (commented):
/approve

PR #4382, opened 2022-11-07
set NodeStatusImage not to be reported

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

Set the default value of NodeStatusMaxImage zero, for we don't need to report node image info, which may affect message transmission performance. 

If you need to report this info, you can set the value of NodeStatusMaxImage in edgecore.yaml.


fisherxu (commented):
/approve

Issue #4388, opened 2022-11-08
keadm didn't install the right version of cloudcore even though I specified --kubeedge-version flag


James-JLC:
keadm has been updated to version 1.12.1, see below
`$keadm version`
`$version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.1", GitCommit:"919ad5378eaca3cb0c666c22a19db01261cbc9a6", GitTreeState:"clean", BuildDate:"2022-11-02T08:22:56Z", GoVersion:"go1.17", Compiler:"gc", Platform:"linux/amd64"}`

I tried install cloudcore of v1.12.1 with the following command
`keadm init --advertise-address=xxx.xxx.xxx.xxx --kubeedge-version=1.12.1 --kube-config=/home/anyone/.kube/config`

But v1.12.0 was installed successfully.

```W1108 10:43:39.788375       1 validation.go:154] TLSTunnelPrivateKeyFile does not exist in /etc/kubeedge/certs/server.key, will load from secret
W1108 10:43:39.788435       1 validation.go:157] TLSTunnelCertFile does not exist in /etc/kubeedge/certs/server.crt, will load from secret
W1108 10:43:39.788447       1 validation.go:160] TLSTunnelCAFile does not exist in /etc/kubeedge/ca/rootCA.crt, will load from secret
I1108 10:43:39.788468       1 server.go:92] Version: v1.12.0
W1108 10:43:39.788488       1 client_config.go:615] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
```

gy95:
you can try 
`keadm init --advertise-address=xxx.xxx.xxx.xxx --profile version=v1.12.1 --kube-config=/home/anyone/.kube/config`
with `--profile version=v1.12.1` but not `--kubeedge-version=1.12.1`

James-JLC:
> you can try `keadm init --advertise-address=xxx.xxx.xxx.xxx --profile version=v1.12.1 --kube-config=/home/anyone/.kube/config` with `--profile version=v1.12.1` but not `--kubeedge-version=1.12.1`

Yes, I know. But it's not the right way to do that because we have kubeedge-version flag.

James-JLC:
And if `--profile version=v1.12.1` is recommended, then we should remove `--kubeedge-version` flag, because it's confused and useless.

James-JLC:
[...]

PR #4389, opened 2022-11-08
Install the right version of KubeEdge when user specified --kubeedge-version flag

James-JLC:
It's related to [#4388](https://github.com/kubeedge/kubeedge/issues/4388)


vincentgoat:
cc @gy95

James-JLC:
This PR will break keadm --profile function. So close it now.

PR #4393, opened 2022-11-09
Fix the issues mentioned in bug #4388

### LINKED ISSUE #4388: keadm didn't install the right version of cloudcore even though I specified --kubeedge-version flag

James-JLC:
Right now keadm try to get the latest version of cloudcore using url `https://kubeedge.io/latestversion` no matter which version is specified with `--kubeedge-version` flag. And I don't think --profile version=1.12.1 is the right way to install the version of cloudcore .

And when --profile iptablesmgr=external is set, keadm sets the wrong version of cloudcore to 1.12.0 instead of v1.12.0.

This PR is to fix #4388

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cloud/init.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/init.go
@@ -121,22 +121,24 @@ func addForceOptionsFlags(cmd *cobra.Command, initOpts *types.InitOptions) {
 //AddInit2ToolsList reads the flagData (containing val and default val) and join options to fill the list of tools.
 func AddInit2ToolsList(toolList map[string]types.ToolsInstaller, initOpts *types.InitOptions) error {
 	var latestVersion string
-	var kubeedgeVersion string
-	for i := 0; i < util.RetryTimes; i++ {
-		version, err := util.GetLatestVersion()
-		if err != nil {
-			fmt.Println("Failed to get the latest KubeEdge release version, error: ", err)
-			continue
+	kubeedgeVersion := initOpts.KubeEdgeVersion
+	if kubeedgeVersion == "" {
+		for i := 0; i < util.RetryTimes; i++ {
+			version, err := util.GetLatestVersion()
+			if err != nil {
+				fmt.Println("Failed to get the latest KubeEdge release version, error: ", err)
+				continue
+			}
+			if len(version) > 0 {
+				kubeedgeVersion = strings.TrimPrefix(version, "v")
+				latestVersion = version
+				break
+			}
 		}
-		if len(version) > 0 {
-			kubeedgeVersion = strings.TrimPrefix(version, "v")
-			latestVersion = version
-			break
+		if len(latestVersion) == 0 {
+			kubeedgeVersion = types.DefaultKubeEdgeVersion
+			fmt.Println("Failed to get the latest KubeEdge release version, will use default version: ", kubeedgeVersion)
 		}
-	}
-	if len(latestVersion) == 0 {
-		kubeedgeVersion = types.DefaultKubeEdgeVersion
-		fmt.Println("Failed to get the latest KubeEdge release version, will use default version: ", kubeedgeVersion)
 	}
 
 	common := util.Common{
--- a/keadm/cmd/keadm/app/cmd/helm/installer.go
+++ b/keadm/cmd/keadm/app/cmd/helm/installer.go
@@ -365,6 +365,9 @@ func (cu *KubeCloudHelmInstTool) checkProfile(baseHelmRoot string) error {
 func (cu *KubeCloudHelmInstTool) handleProfile(profileValue string) error {
 	// the current version
 	currentVersion := cu.ToolVersion.String()
+	if !strings.HasPrefix(currentVersion, "v") {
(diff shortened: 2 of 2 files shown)


gy95 (approved):
Now we have two flags to specify the kubeedge version `--kubeedge-version=1.12.0` and `--profile version=1.12.0`, we should keep the only one left, and this may need some work.
But this patch is great, we can merge it first, and then do some another optimization.
cc @zhu733756

keadm/cmd/keadm/app/cmd/cloud/init.go
```diff
@@ -121,23 +121,25 @@ func addForceOptionsFlags(cmd *cobra.Command, initOpts *types.InitOptions) {
...
 func AddInit2ToolsList(toolList map[string]types.ToolsInstaller, initOpts *types.InitOptions) error {
 	var latestVersion string
-	var kubeedgeVersion string
-	for i := 0; i < util.RetryTimes; i++ {
-		version, err := util.GetLatestVersion()
-		if err != nil {
-			fmt.Println("Failed to get the latest KubeEdge release version, error: ", err)
-			continue
+	kubeedgeVersion := initOpts.KubeEdgeVersion
+	if kubeedgeVersion == "" {
```
gy95, line 125: LGTM
this is right, if we've already specified the version, there's no need to check the latest release version

keadm/cmd/keadm/app/cmd/helm/installer.go
```diff
@@ -365,6 +365,9 @@ func (cu *KubeCloudHelmInstTool) checkProfile(baseHelmRoot string) error {
 func (cu *KubeCloudHelmInstTool) handleProfile(profileValue string) error {
 	// the current version
 	currentVersion := cu.ToolVersion.String()
+	if !strings.HasPrefix(currentVersion, "v") {
+		currentVersion = "v" + currentVersion
+	}
```
gy95, line 370: Now the `--kubeedge-version=v1.12.0` flag has a higher priority than `--profile version=v1.11.0`?
In other words, which version we will use if we use both `--kubeedge-version=v1.12.0` and `--profile version=v1.11.0`? 1.12 or 1.11? @James-JLC  :)

James-JLC: As I mentioned int #4388, we should not provide the option for uses to set kubeedge version using --profile flag, because we already have --kubeedge-version. And I don't know the historical reason --profile version=v1.xx.x flag is provided as an option, so I didn't make the changes here. But I do think it should be removed. As you said some optimizations need to be done right now or later.

zhu733756:
LGTM

BTW, --profile is used to decide to read which helm value.yaml.

keadm/cmd/keadm/app/cmd/helm/installer.go
duanhaowei: > As I mentioned in #4388, we should not provide the option for users to set kubeedge version using --profile flag, because we already have --kubeedge-version. And I don't know the historical reason --profile version=v1.xx.x flag is provided as an option, so I didn't make the changes here. But I do think it should be removed. As you said some optimizations need to be done right now or later.

there is not '--profile version' patamter show in kubeedge.io document . developer will prefer ' --kubeedge-version'

James-JLC: The option is displayed in both [kubeedge.io document](https://kubeedge.io/en/docs/setup/keadm/#keadm-beta-init) and the output of command `keadm init`.

gy95: /lgtm

duanhaowei: > profile version'

emmmm , my mistake . 
"you can make your custom values file here, and add flags like --profile version=v1.9.0 --set key=value to use this profile."

PR #4400, opened 2022-11-11
remove NodeUpgradeJob controller local map cache

gy95:
**What type of PR is this?**
remove NodeUpgradeJob controller local map cache
use informer to cache Upgrade data, don't cache data in local map, instead use informer onAdd/onUpgrade to filter data directly

Fixes #

```release-note

```


cloud/pkg/nodeupgradejobcontroller/manager/common.go
```diff
@@ -17,46 +17,61 @@ limitations under the License.
...
 import (
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/watch"
-	"k8s.io/klog/v2"
+	"k8s.io/client-go/tools/cache"
 )
 
-// Manager define the interface of a Manager, NodeUpgradeJob Manager implement it
-type Manager interface {
-	Events() chan watch.Event
```
gy95, line 27: Do we need use channel to cache events, or just use AddFunc/UpdateFunc/DeleteFunc to process event directly

PR #4404, opened 2022-11-13
replaced io/ioutil with "os / io" package.

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

**What type of PR is this?**
/kind cleanup

As of Go >=1.16, this function "ioutil.ReadFile" simply calls os.ReadFile.
We should clean all io/util package according the url: https://go.dev/doc/go1.16#ioutil


gy95:
welcome
/lgtm

PR #4405, opened 2022-11-13
Keep the flag info capital letter

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

**What type of PR is this?**
/kind documentation

We should let the flag info keep the same style.


yanggangtony:
Sorry for that mis commit.
It just the vendor , So maybe fix it in the third package.

PR #4419, opened 2022-11-17
support config path in keadm join

gy95:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/join.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join.go
@@ -60,7 +60,7 @@ keadm join --cloudcore-ipport=10.20.30.40:10000 --edgenode-name=testing123 --kub
 `
 )
 
-var edgeCoreConfig *v1alpha2.EdgeCoreConfig
+var edgeCoreConfig v1alpha2.EdgeCoreConfig
 
 func NewEdgeJoin() *cobra.Command {
 	joinOptions := newOption()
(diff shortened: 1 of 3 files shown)


keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -214,20 +218,19 @@ func createDirs() error {
 func createEdgeConfigFiles(opt *common.JoinOptions) error {
 	var edgeCoreConfig *v1alpha2.EdgeCoreConfig
```
vincentgoat, line 219:
```suggestion
	var edgeCoreConfig v1alpha2.EdgeCoreConfig
```

keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -288,6 +291,7 @@ func createEdgeConfigFiles(opt *common.JoinOptions) error {
 	if errs := validation.ValidateEdgeCoreConfiguration(edgeCoreConfig); len(errs) > 0 {
 		return errors.New(pkgutil.SpliceErrors(errs.ToAggregate().Errors()))
 	}
+	configFilePath := filepath.Join(util.KubeEdgePath, "config/edgecore.yaml")
```
vincentgoat, line 294: Would it be better to update the file in the user config path and make this file the component edgecore configuration file?
It will be tricky if there are two config files.

keadm/cmd/keadm/app/cmd/edge/join.go
gy95: function NewDefaultEdgeCoreConfig will return a pointer, so here use the pointer but not struct

keadm/cmd/keadm/app/cmd/edge/join.go
gy95: fixed

gy95:
@wackxu  PTAL

keadm/cmd/keadm/app/cmd/edge/join.go
vincentgoat: It can cause a nil pointer issue if pass a pointer parameter to function `Unmarshal` at line 235. We should new a variable passed to function `Unmarshal`.
FYI: https://adalogics.com/blog/fuzzing-istio-cve-CVE-2022-23635

keadm/cmd/keadm/app/cmd/edge/join.go
vincentgoat: Will component `edgecore` startup with the specified config file?

gy95: we need to add option `--config /path/to/config.yaml` to edgecore start command  like `edgecore --config /root/config.yaml`

PR #4421, opened 2022-11-17
Extended EdgeApplication to select NodeGroups by label

### LINKED ISSUE #3959: Add ability for EdgeApplications to target NodeGroups by label

I'd like to add the ability to use NodeGroups not only by name, but also by label. 

**Why is this needed**:
Using the scenario below it is possible to handle an EdgeApplication for different (dynamic created/deleted) NodeGroups.

In the example below there are two NodeGroups: blue and green containing blue and green nodes accordingly.
Both NodeGroups share the same label `env: dev`.
The EdgeApplication applies to all NodeGroups that match the label `env: dev`. If a new matching NodeGroups appears, it is automatically taken into account and vice versa if the NodeGroup is being removed.  It shall be possible to either use the `name` or `matchLabels` fields in `targetNodeGroups` for each entry independently.

```
apiVersion: groupmanagement.kubeedge.io/v1alpha1
kind: NodeGroup
metadata:
  name: blue
  labels:
    env: dev
spec:
  matchLabels:
    env: blue
---
apiVersion: groupmanagement.kubeedge.io/v1alpha1
kind: NodeGroup
metadata:
  name: green
  labels:
    env: dev
spec:
  matchLabels:
    env: green
---
apiVersion: groupmanagement.kubeedge.io/v1alpha1
kind: EdgeApplication
metadata:
  name: nginx-app 
spec:
  workloadTemplate:
    manifests:
    - apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: nginx
    spec:
      selector:
        matchLabels:
          app: nginx
      template:
        metadata:
          labels:
            app: nginx
        spec:
          containers:
          - name: nginx
            image: nginx:latest
          ports:
          - containerPort: 80
  workloadScopes:
    targetNodeGroups:
      - matchLabels: 
            env: dev
        overriders:
          replicas: 2
          imageOverrider:
            - component: "registry"
              operator: "replace"
              value: "dev.registry.io"
```

batthebee:
**What type of PR is this?**

/kind feature
/kind api-change

Motivation:

We have many independently operated nodes on which the same software stack, which consists of deployments and services, is to be rolled out. It is important that the corresponding services are only accessible locally on the devices.

This is currently realized by giving each individual node its own nodegroup and manually entering all nodegroups in the edge application. Accordingly, we always have a manual step as soon as a new nodegroup is added.

Using a label selector, we can label the new nodegroup and automatically receive the software stack without manual entry into the edgeapplication.

Fixes #3959

```release-note
Extended EdgeApplication to be able to select NodeGroups by label
```

### DIFF
--- a/cloud/pkg/controllermanager/edgeapplication/utils/utils.go
+++ b/cloud/pkg/controllermanager/edgeapplication/utils/utils.go
@@ -1,15 +1,19 @@
 package utils
 
 import (
+	"context"
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 
+	"github.com/imdario/mergo"
 	"github.com/kubeedge/kubeedge/cloud/pkg/controllermanager/edgeapplication/overridemanager"
 	appsv1alpha1 "github.com/kubeedge/kubeedge/pkg/apis/apps/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type ResourceInfo struct {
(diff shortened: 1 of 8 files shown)


fisherxu:
cc @vincentgoat @Congrool  ptal :)

cloud/pkg/controllermanager/edgeapplication/utils/utils.go
```diff
@@ -32,16 +36,62 @@ type TemplateInfo struct {
...
-		copied := edgeApp.Spec.WorkloadScope.TargetNodeGroups[index].Overriders.DeepCopy()
-		infos = append(infos, overridemanager.OverriderInfo{
-			TargetNodeGroup: edgeApp.Spec.WorkloadScope.TargetNodeGroups[index].Name,
-			Overriders:      copied,
-		})
+		name := edgeApp.Spec.WorkloadScope.TargetNodeGroups[index].Name
+		overriders := edgeApp.Spec.WorkloadScope.TargetNodeGroups[index].Overriders.DeepCopy()
+		if nodegroup, exist := sorted[name]; exist {
+			if err := mergo.Merge(&nodegroup.Overriders, overriders); err != nil {
+				klog.Errorf("failed to merge overriders of nodegroup %s, %v", name, err)
```
Congrool, line 79: We may also need to provide error info about which edgeapp it failed at.

Congrool:
I originally thought you would add `matchFields` to targetNodeGroups, like
```yaml
  workloadScopes:
    targetNodeGroups:
      - matchLabels: 
            env: dev
        overriders:
          replicas: 2
          imageOverrider:
            - component: "registry"
              operator: "replace"
              value: "dev.registry.io"
```
It may encounter the problems as we discussed in #3959. So I prefer the current implementation in this pr. It also looks reasonable when using `targetNodeGroupSelector` to deploy the application and uses `targetNodeGroups` to specify the different settings for individual nodes, because `targetNodeGroupSelector` still respect overriders set in `targetNodeGroups` as we can see in the code.

So I think we can just remove the overrider field of `targetNodeGroupSelector` since it's not implemented and also give discriptions in api comment about the cooperation between `targetNodeGroupSeletor` and `targetNodeGroups` to make users know about how to uses two fields to override and deploy application.

PR #4425, opened 2022-11-19
Fix e2e variable GOPATH not set

gy95:
**What type of PR is this?**

I try to run `make e2e` on my physical machine, but it reports an error like below.
```
# make e2e
tests/e2e/scripts/execute.sh
/root/go/src/github.com/kubeedge/kubeedge
ginkgo is not found, install first
cp: cannot stat '/bin/ginkgo': No such file or directory
Makefile:293: recipe for target 'e2e' failed
make: *** [e2e] Error 1
```
For go, GOPATH is not set as an environment variable, so it's better to get it using `go env GOPATH`

Fixes #

```release-note

```

### DIFF
--- a/tests/e2e/scripts/execute.sh
+++ b/tests/e2e/scripts/execute.sh
@@ -22,6 +22,8 @@ cd $workdir
 curpath=$PWD
 echo $PWD
 
+GOPATH=${GOPATH:-$(go env GOPATH)}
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
/approve

cc @vincentgoat  @wackxu

PR #4430, opened 2022-11-21
stream server painc if close channel wrongly

vincentgoat:

**What type of PR is this?**
/kind bug

Fix 2 panic errors:
- [Error 1]: close of closed channel 
One of the place will call the close functions: https://github.com/kubeedge/kubeedge/blob/5d18a3c348531b60da34e5dc32da3e595800c22c/cloud/pkg/cloudstream/session.go#L55
Another at: https://github.com/kubeedge/kubeedge/blob/5d18a3c348531b60da34e5dc32da3e595800c22c/cloud/pkg/cloudstream/session.go#L98

- [Error 2]: send on closed channel
Send data to channel https://github.com/kubeedge/kubeedge/blob/master/pkg/stream/edgedexecconnection.go#L39 after channel was closed https://github.com/kubeedge/kubeedge/blob/master/pkg/stream/edgedexecconnection.go#L98

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/cloudstream/containerexec_connection.go
+++ b/cloud/pkg/cloudstream/containerexec_connection.go
@@ -22,6 +22,7 @@ type ContainerExecConnection struct {
 	Conn         net.Conn
 	session      *Session
 	edgePeerStop chan struct{}
+	closeChan    chan bool
 }
 
 func (c *ContainerExecConnection) String() string {
@@ -41,10 +42,15 @@ func (c *ContainerExecConnection) GetMessageID() uint64 {
 }
 
 func (c *ContainerExecConnection) SetEdgePeerDone() {
-	close(c.edgePeerStop)
+	select {
+	case <-c.closeChan:
+		return
+	case c.EdgePeerDone() <- struct{}{}:
+		klog.V(6).Infof("success send channel deleting connection with messageID %v", c.MessageID)
+	}
 }
 
-func (c *ContainerExecConnection) EdgePeerDone() <-chan struct{} {
+func (c *ContainerExecConnection) EdgePeerDone() chan struct{} {
 	return c.edgePeerStop
 }
 
@@ -74,6 +80,7 @@ func (c *ContainerExecConnection) SendConnection() (stream.EdgedConnection, erro
 
 func (c *ContainerExecConnection) Serve() error {
 	defer func() {
+		close(c.closeChan)
 		klog.V(6).Infof("%s stop successfully", c.String())
 	}()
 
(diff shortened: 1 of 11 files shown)


cloud/pkg/cloudstream/containerexec_connection.go
```diff
@@ -21,6 +21,7 @@ type ContainerExecConnection struct {
 	Conn         net.Conn
 	session      *Session
 	edgePeerStop chan struct{}
+	closeChan    chan bool
```
wackxu, line 24: We can use sync.Once to make sure the edgePeerStop channel is closed only once

pkg/stream/edgedconnection.go
```diff
@@ -8,5 +8,7 @@ type EdgedConnection interface {
 	Serve(tunnel SafeWriteTunneler) error
 	CacheTunnelMessage(msg *Message)
 	GetMessageID() uint64
+	CloseReadChannel()
+	CleanChannel()
```
wackxu, line 12: Why do we need CleanChannel function？

pkg/stream/edgedexecconnection.go
```diff
@@ -39,49 +41,82 @@ func (e *EdgedExecConnection) CacheTunnelMessage(msg *Message) {
 	e.ReadChan <- msg
```
wackxu, line 41: from your fix, I do not know how to prevent write to the closed readChan

wackxu: can we add a flag to indicate the channel is closed then just return

cloud/pkg/cloudstream/containerexec_connection.go
vincentgoat: According to the channel closing principle, one general principle of using Go channels is don't close a channel from the receiver side and don't close a channel if the channel has multiple concurrent senders.
It will be more graceful to close a channel from the receiver side.

pkg/stream/edgedexecconnection.go
vincentgoat: This would be a way.
This is the way I took it: M receivers, N senders, any one of them say "let's end the game" by notifying a moderator to close an additional signal channel.
FIY: https://go101.org/article/channel-closing.html

pkg/stream/edgedconnection.go
vincentgoat: To make sure the stop channel won't block the sender go routine.

cloud/pkg/cloudstream/containerexec_connection.go
```diff
@@ -41,10 +42,15 @@ func (c *ContainerExecConnection) GetMessageID() uint64 {
 }
 
 func (c *ContainerExecConnection) SetEdgePeerDone() {
-	close(c.edgePeerStop)
```
fisherxu, line 44: Seems haven't found that where close the `edgePeerStop` channel?

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -136,6 +136,7 @@ func (s *StreamServer) getContainerLogs(r *restful.Request, w *restful.Response)
 		session:      session,
 		ctx:          r.Request.Context(),
 		edgePeerStop: make(chan struct{}),
```
fisherxu, line 138: Need to set the size to `2`?

cloud/pkg/cloudstream/streamserver.go
```diff
@@ -188,6 +189,7 @@ func (s *StreamServer) getMetrics(r *restful.Request, w *restful.Response) {
...
```
fisherxu, line 191: ditto

cloud/pkg/cloudstream/containerexec_connection.go
vincentgoat: The memory of channel `edgePeerStop` can be released by the automatic memory management of golang when the exec connection is down.

fisherxu: okey.

cloud/pkg/cloudstream/streamserver.go
vincentgoat: As we discussed, edgePeerStop channel of log_connnecton and metrics_connnection will not block the sender side.

PR #4427, opened 2022-11-19
update golang to 1.17

gy95:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #4426

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #4432, opened 2022-11-22
fix update-vendor-Licenses script

gy95:
**What type of PR is this?**
/kind cleanup
/kind bug

When i try to update golang to 1.17 in #4427 , i found licence check cannot be passed.
it always report error `No license could be found for github.com/godbus/dbus - aborting.`

I found that we don't use `github.com/godbus/dbus`, but we use `github.com/godbus/dbus/v5`, so `vendor/github.com/godbus/dbus` directory only contains a `v5` subdirectory, we don't need to check package `github.com/godbus/dbus` licenses, we only need to check  `github.com/godbus/dbus/v5` licenses

So update update-vendor-licenses script, ref: https://github.com/kubernetes/kubernetes/blob/master/hack/update-vendor-licenses.sh

Fixes #

```release-note

```

### DIFF
--- a/LICENSES/vendor/github.com/coreos/go-systemd/LICENSE
+++ b/LICENSES/vendor/github.com/coreos/go-systemd/LICENSE
@@ -1,9 +1,0 @@
-= vendor/github.com/coreos/go-systemd licensed under: =
-
-CoreOS Project
-Copyright 2018 CoreOS, Inc
-
-This product includes software developed at CoreOS, Inc.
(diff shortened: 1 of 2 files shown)


gy95:
cc @vincentgoat

vincentgoat (commented):
/lgtm 
We can ignore the Fuzzing error for this PR.

fisherxu (commented):
/lgtm
/approve

PR #4442, opened 2022-11-25
chart: add kubeVersion support statement

my-git9:
Add kubeVersion support statement for chart

### DIFF
--- a/manifests/charts/cloudcore/Chart.yaml
+++ b/manifests/charts/cloudcore/Chart.yaml
@@ -3,5 +3,6 @@ name: cloudcore
 version: 0.1.0
 appVersion: 1.12.0
 description: The KubeEdge cloudcore component.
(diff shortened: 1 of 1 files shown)


manifests/charts/cloudcore/Chart.yaml
```diff
@@ -3,5 +3,6 @@ name: cloudcore
 version: 0.1.0
 appVersion: 1.12.0
 description: The KubeEdge cloudcore component.
+kubeVersion: ">= 1.16.0-0 < 1.23.0-0"
```
fisherxu, line 6: Hard code is not so elegant, and is it necessary to write the k8s version here? @my-git9

my-git9: I think it is necessary to a certain extent, it can prevent uninformed users from deploying kubeedge on unsupported k8s versions

PR #4447, opened 2022-11-26
support set cloudcore config file path in keadm init

gy95:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/manifests/charts/cloudcore/templates/configmap_cloudcore.yaml
+++ b/manifests/charts/cloudcore/templates/configmap_cloudcore.yaml
@@ -6,6 +6,12 @@ metadata:
   labels: {{- toYaml . | nindent 4 }}
   {{- end }}
 data:
+  {{- if (ne .Values.cloudCore.configFile "") }}
+  {{- $files := .Files }}
+  {{- $config := .Values.cloudCore.configFile -}}
+  cloudcore.yaml: |
+      {{ $files.Get "$config" }}
+  {{ else }}
   cloudcore.yaml: |
     apiVersion: cloudcore.config.kubeedge.io/v1alpha2
     kind: CloudCore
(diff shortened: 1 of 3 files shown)


manifests/charts/cloudcore/templates/configmap_cloudcore.yaml
```diff
@@ -6,6 +6,12 @@ metadata:
   labels: {{- toYaml . | nindent 4 }}
   {{- end }}
 data:
+  {{- if (ne .Values.cloudCore.configFile "") }}
+  {{- $files := .Files }}
+  {{- $config := .Values.cloudCore.configFile -}}
+  cloudcore.yaml: |
+      {{ $files.Get "$config" }}
```
gy95, line 13: @zhu733756  hi, do you know how to access a file outside `charts` directory, e.g. /etc/kubeedge/config/cloudcore.yaml. I found files.Get can only get the content of files in charts directory.

zhu733756: I see a function like here in https://helm.sh/docs/chart_template_guide/accessing_files/  .
The page says that "Then we print each file name ({{ . }}: |-) followed by the contents of the file {{ $files.Get . }}."
Can you try it?

gy95: we cannot access the files outside chart directory , by asking in helm slack, the answer is below
`
That doesnt work by design. Imagine you give this chart to your colleage. If they dont have the file in the same place, the chart is broken. It is probabably also a security risk. Imagine some third party chart goes out and fetches your /etc/password file.
`

zhu733756: Got it. I think they are correct.

gy95:
cannot be implemented, ref the above comments
/close

PR #4452, opened 2022-11-28
fix: security docs link 404

YTGhost:
Signed-off-by: HIHIA <283304489@qq.com>

**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #4449 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #4456, opened 2022-11-30
pass checks when a PR that only changes the docs

gy95:
**What type of PR is this?**

fix https://github.com/kubeedge/kubeedge/pull/4452#issuecomment-1330042041

now when we send a PR that only changes the docs, the CI check status is always `Waiting for status to be reported`
with this patch, it will display successful status

test on my local branch: https://github.com/gy95/kubeedge/pull/16

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/main-doc.yaml
+++ b/.github/workflows/main-doc.yaml
@@ -1,0 +1,112 @@
+name: Main CI WorkFlow
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+    paths:
+      - '**.md'
+      - 'docs/**'
+      - '**/OWNERS'
+      - '**/MAINTAINERS'
+  pull_request:
+    paths:
+      - '**.md'
+      - 'docs/**'
+      - '**/OWNERS'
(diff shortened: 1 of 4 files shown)


.github/workflows/main-doc.yaml
```diff
@@ -0,0 +1,98 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 30
+    name: Verify vendor, licenses, do lint
+    env:
+      GOPATH: /home/runner/work/${{ github.repository }}
+    steps:
+      - run: 'echo "only docs modified, no need to trigger CI"'
```
fisherxu, line 28: Do we need to run lint here? Now we have run the whitenoise in lint IIRC, or later run other lint check for doc :) @gy95

gy95: good catch

fisherxu (commented):
/lgtm
/approve

Issue #4423, opened 2022-11-18
Enhancement for edge list-watch


wackxu:
- [x] Add GC for watcher in the dynamic controller
- [x] Add watcher synchronization mechanism between cloud and edge
- [x] fix 10s delay for edge request when the node is offline
- [ ] Reduce the send event when the watch is initialized
- [x] Use multi-goroutines to improve the efficiency of sending watch events in large-scale scenarios
- [ ] edge pod use inclusterconfig list-watch natively

PR #4469, opened 2022-12-05
clean up handlerCenter and SelectorListener

wackxu:

**What type of PR is this?**

/kind cleanup

Fixes # part of https://github.com/kubeedge/kubeedge/issues/4423

1. add lock for handlerCenter so that we will support parallel process watch Application
2. use Application ID as SelectorListener so that we can sync watch request between cloud and edge in the following pr

```release-note

```

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/eventhandler.go
+++ b/cloud/pkg/dynamiccontroller/application/eventhandler.go
@@ -34,7 +34,8 @@ import (
 )
 
 // HandlerCenter is used to prepare corresponding CommonResourceEventHandler for listener
-// CommonResourceEventHandler then will send to the listener the objs it is interested in, including subsequent changes (watch event)
+// CommonResourceEventHandler then will send to the listener the objs it is interested in,
+// including subsequent changes (watch event)
 type HandlerCenter interface {
 	AddListener(s *SelectorListener) error
 	DeleteListener(s *SelectorListener)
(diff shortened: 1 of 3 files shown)


wackxu:
cc @gy95 @vincentgoat @fisherxu PTAL

cloud/pkg/dynamiccontroller/application/eventhandler.go
```diff
@@ -59,23 +62,27 @@ func NewHandlerCenter(informerFactory dynamicinformer.DynamicSharedInformerFacto
...
 
 func (c *handlerCenter) ForResource(gvr schema.GroupVersionResource) *CommonResourceEventHandler {
-	var handler *CommonResourceEventHandler
-	if store, ok := c.handlers[gvr]; ok {
-		handler = store
-	} else {
-		klog.Infof("[metaserver/HandlerCenter] prepare a new resourceEventHandler(%v)", gvr)
-		handler = NewCommonResourceEventHandler(gvr, c.dynamicSharedInformerFactory, c.messageLayer)
-		c.handlers[gvr] = handler
+	if handler, ok := c.handlers[gvr]; ok {
```
gy95, line 65: just confusing about where to add lock. here not add a lock, but `ForResource` interface is also used by other modules

wackxu: good catch, updated

cloud/pkg/dynamiccontroller/application/eventhandler.go
```diff
@@ -76,6 +82,8 @@ func (c *handlerCenter) AddListener(s *SelectorListener) error {
 }
 
 func (c *handlerCenter) DeleteListener(s *SelectorListener) {
+	c.handlerLock.Lock()
+	defer c.handlerLock.Unlock()
```
vincentgoat, line 86: It will be more efficient to not use the function `defer` in a simple process function.

wackxu: done

vincentgoat (commented):
/lgtm

fisherxu (commented):
/approve

PR #4472, opened 2022-12-06
add listener manager to store listener

wackxu:

**What type of PR is this?**

/kind feature

we want to sync watch request between cloud and edge, so we cache the listener by nodeID and gvr so that we can
compare the difference between cloud and edge easilly

Fixes # part of https://github.com/kubeedge/kubeedge/issues/4423

```release-note

```

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/listener_manager.go
+++ b/cloud/pkg/dynamiccontroller/application/listener_manager.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2022 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


cloud/pkg/dynamiccontroller/application/listener_manager.go
```diff
@@ -0,0 +1,105 @@
...
+	listeners, exists = lm.listenerByGVR[listener.gvr]
+	if exists {
+		delete(listeners, listener.id)
+		if len(lm.listenerByGVR[listener.gvr]) == 0 {
+			delete(lm.listenerByGVR, listener.gvr)
+		}
+	}
+}
+
+func (lm *listenerManager) GetListenersForNode(nodeName string) map[string]*SelectorListener {
```
vincentgoat, line 83: Seems the function `GetListenersForNode` is not called except for the test file.

wackxu: this will be used in the follow PR

wackxu: please see https://github.com/kubeedge/kubeedge/pull/4484

PR #4473, opened 2022-12-06
drop nodeIP validation 

### LINKED ISSUE #4414: kubeedge v1.12.1 modify the nodeIP configuration of the edge node, and cannot report to the cloud?
In the version 1.9.2 of kubeedg, I modified the edge node's edged: nodeIP in/etc/kubeedge/config/edgecore.yaml, for example, 1.1.1.1. Then restart edgecore service。 After success, I can see in the cloud that the internal ip of kubectl get node - owide is 1.1.1.1.
However, when I upgrade the kubeedge to 1.12.1, I repeat the above operation. The nodeIP cannot be reported to the cloud, and kubelet displays an error:

> kubelet_ node_ status. go:517] "Failed to set some node status fields" err="failed to validate nodeIP: node IP: \"1.1.1.1\" not found in the host's network interfaces" node="edgenode01"。

 The network interface on the edge node does not have the IP address 1.1.1.1. When the IP addresses of my two edge nodes are in a different network environment, but the IP addresses of the LAN are the same, will the internal ip displayed on the cloud be the same? How do I distinguish them?
change the edge node's edged: nodeIP in/etc/kubeedge/config/edgecore.yaml,and restart edgecore.service
no
**Environment**:
v1.19.13
v1.12.1

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

drop nodeIP validation in Kubelet. Due to the particularity of edge scenario, the node IP of edgenode may not be the IP of host. 

Fixes #4414 

review first plz. This pr need be  merged when update the new kubeedge/kubernetes tag in next release/patch version.

### DIFF
--- a/vendor/k8s.io/kubernetes/pkg/kubelet/nodestatus/setters.go
+++ b/vendor/k8s.io/kubernetes/pkg/kubelet/nodestatus/setters.go
@@ -78,15 +78,9 @@ func NodeAddress(nodeIPs []net.IP, // typically Kubelet.nodeIPs
 
 	return func(node *v1.Node) error {
 		if nodeIPSpecified {
-			if err := validateNodeIPFunc(nodeIP); err != nil {
-				return fmt.Errorf("failed to validate nodeIP: %v", err)
-			}
 			klog.V(4).InfoS("Using node IP", "IP", nodeIP.String())
 		}
 		if secondaryNodeIPSpecified {
-			if err := validateNodeIPFunc(secondaryNodeIP); err != nil {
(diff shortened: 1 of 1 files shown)


bumblebeeK:
Absolutely agreed，the lightweight server doesn't have the public ip address in network interfaces.

wackxu (approved):
LGTM

vincentgoat:
Please close this PR here as https://github.com/kubeedge/kubernetes/pull/4 has already been merged.

PR #4482, opened 2022-12-08
support specifying kubeedge-version both with and without prefix v

gy95:
**What type of PR is this?**

support specifying kubeedge-version both with and without prefix v

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/deprecated/init.go
+++ b/keadm/cmd/keadm/app/cmd/deprecated/init.go
@@ -130,6 +130,12 @@ func Add2CloudToolsList(toolList map[string]types.ToolsInstaller, flagData map[s
 			fmt.Println("Failed to get the latest KubeEdge release version, will use default version: ", kubeVer)
 		}
 	}
+
+	// If users specified the kubeedge-version with prefix `v`, remove it
+	// So we support both format: 1.12.0 without prefix `v`, and v1.12.0 with prefix `v`
+	// remove prefix v
+	kubeVer = strings.TrimPrefix(kubeVer, "v")
+
 	common := util.Common{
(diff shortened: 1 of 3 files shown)


keadm/cmd/keadm/app/cmd/deprecated/init.go
```diff
@@ -130,6 +130,13 @@ func Add2CloudToolsList(toolList map[string]types.ToolsInstaller, flagData map[s
 			fmt.Println("Failed to get the latest KubeEdge release version, will use default version: ", kubeVer)
 		}
 	}
+
+	// If users specified the kubeedge-version with prefix `v`, remove it
+	// So we support both format: 1.12.0 without prefix `v`, and v1.12.0 with prefix `v`
+	if strings.HasPrefix(kubeVer, "v") {
```
vincentgoat, line 136: strings.TrimPrefix will call the function strings.HasPrefix and it's no need to repeatedly call it here.

keadm/cmd/keadm/app/cmd/deprecated/join.go
```diff
@@ -115,6 +115,13 @@ func Add2EdgeToolsList(toolList map[string]types.ToolsInstaller, flagData map[st
 			kubeVer = types.DefaultKubeEdgeVersion
 		}
 	}
+
+	// If users specified the kubeedge-version with prefix `v`, remove it
+	// So we support both format: 1.12.0 without prefix `v`, and v1.12.0 with prefix `v`
+	if strings.HasPrefix(kubeVer, "v") {
```
vincentgoat, line 121: ditto

gy95: fixed

PR #4488, opened 2022-12-10
do not save edgecore token on disk

gy95:
**What type of PR is this?**

Fixes #

```release-note
do not save edgecore token on disk
```

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -13,13 +13,16 @@ import (
 	"k8s.io/component-base/cli/globalflag"
 	"k8s.io/component-base/term"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
 
 	"github.com/kubeedge/beehive/pkg/core"
+	"github.com/kubeedge/kubeedge/common/constants"
 	"github.com/kubeedge/kubeedge/edge/cmd/edgecore/app/options"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/dbm"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin"
 	"github.com/kubeedge/kubeedge/edge/pkg/edged"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub"
+	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/certificate"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgestream"
 	"github.com/kubeedge/kubeedge/edge/pkg/eventbus"
 	"github.com/kubeedge/kubeedge/edge/pkg/metamanager"
@@ -30,6 +33,7 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/features"
 	"github.com/kubeedge/kubeedge/pkg/util"
 	"github.com/kubeedge/kubeedge/pkg/util/flag"
+	utilvalidation "github.com/kubeedge/kubeedge/pkg/util/validation"
 	"github.com/kubeedge/kubeedge/pkg/version"
 )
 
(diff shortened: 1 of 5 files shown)


wackxu:
I think we should not store the token in any file, this token is only used when the edge node first join, for example

```
keadm join --token=xxxx ...
```

so the token could only happen in the command, and edgecore will use the certificate to access the cloudcore

so we can just remove the token filed from the edgecore config

gy95:
just for some backward compatibility, e.g.  use v1.13 keadm  to install v1.12 edgecore, we should keep `token` for some releases and then delete it.

edge/cmd/edgecore/app/server.go
```diff
@@ -62,6 +65,31 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
...
+
+					// cleanup token
+					if err := cleanupToken(*config, opts.ConfigFile); err != nil {
+						klog.Exit(err)
+					}
+				}()
+			}
+
+			// Get token from bootstrap file if it exists
+			bootstrapFile := "/etc/kubeedge/bootstrap-edgecore.conf"
```
wackxu, line 84: set this variable as a constant

edge/cmd/edgecore/app/server.go
```diff
@@ -62,6 +65,31 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
...
+			if utilvalidation.FileIsExist(bootstrapFile) {
```
wackxu, line 85: if the file does not exist, we should throw an err and exitt

gy95: if edgecore restart, bootstrap file also not exist, we should not throw an error

gy95: add a pre check

edge/cmd/edgecore/app/server.go
gy95: fixed

kevin-wangzefeng:
security improvement
/cc @vincentgoat

kevin-wangzefeng (commented):
There's a failure of CI fuzzing check, please take a look, thanks

keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -232,6 +270,9 @@ func createEdgeConfigFiles(opt *common.JoinOptions) error {
 	}
 
 	edgeCoreConfig.Modules.EdgeHub.WebSocket.Server = opt.CloudCoreIPPort
+	// TODO: remove this after release 1.14
+	// this is for keeping backward compatibility
+	// don't save token in configuration edgecore.yaml
```
fisherxu, line 275: Let's also add a `depracated` comment in token field, then rermove it in future release.

gy95: if we run edgecore diretly, i think we can allow users to fill in token in edgecore.yaml, or users should write this token to /etc/kubeedge/bootstrap-edgecore.conf, this is not easy use

fisherxu: If user run edgecore diretly, they should prepare the certs by themselfs. And token field in config file is not safe, we must remove it in future. cc @vincentgoat How do you think?

vincentgoat: Agree. The token should be constrained into the file /etc/kubeedge/bootstrap-edgecore.conf for temp use.

keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -403,3 +432,15 @@ func runEdgeCore() error {
...
+func createBootstrapFile(opt *common.JoinOptions) error {
+	bootstrapFile := constants.BootstrapFile
+	_, err := os.Create(bootstrapFile)
+	if err != nil {
+		return err
+	}
+
+	// write token to bootstrap-edgecore.conf file
+	token := []byte(opt.Token)
+	return os.WriteFile(bootstrapFile, token, 0666)
```
vincentgoat, line 445: `0640` is the proper permission for the file.

gy95: fixed

pkg/apis/componentconfig/edgecore/v1alpha2/types.go
```diff
@@ -297,6 +297,7 @@ type EdgeHub struct {
 	// Optional if quic is configured
 	WebSocket *EdgeHubWebSocket `json:"websocket,omitempty"`
 	// Token indicates the priority of joining the cluster for the edge
+	// Deprecated: will be removed after release 1.14, will not be saved in configuration file
```
fisherxu, line 300:
```suggestion
	// Deprecated: will be removed in future release, will not be saved in configuration file
```
Need more discussion for removing in which one release :)

gy95: yes, fixed

fisherxu (commented):
/appove
/lgtm

PR #4490, opened 2022-12-12
fix 10s delay for edge request when the node is offline

wackxu:
**What type of PR is this?**

/kind bug

Fixes # part of #4423 

fix 10s delay for edge request when the node is offline

when the node if offline, the request for metaserver will generate the application and apply the application to the cloudcore, 
in the apply func, we use the sendSync function and will wait max time 10s to get the result

```
resp, err := beehiveContext.SendSync(edgemodule.EdgeHubModuleName, *msg, 10*time.Second)
```
but due to the network is lost between edgecore and cloudcore and the apply will spend 10s every time when process the request and this will affect code processing, such as edgemesh, will have a 10s deplay for get and list request when the node is offline. So if the node is offline, we do not generate the application since it is useless

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -15,7 +15,6 @@ import (
 	cloudmodules "github.com/kubeedge/kubeedge/cloud/pkg/common/modules"
 	"github.com/kubeedge/kubeedge/common/constants"
 	connect "github.com/kubeedge/kubeedge/edge/pkg/common/cloudconnection"
-	messagepkg "github.com/kubeedge/kubeedge/edge/pkg/common/message"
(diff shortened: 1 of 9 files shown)


gy95:
LGTM

edge/pkg/metamanager/process.go
```diff
@@ -485,8 +470,6 @@ func (m *metaManager) process(message model.Message) {
 		m.processQuery(message)
 	case model.ResponseOperation:
 		m.processResponse(message)
-	case messagepkg.OperationNodeConnection:
```
vincentgoat, line 488: Also can remove the code on the sender side?

wackxu: done

vincentgoat (commented):
/lgtm

PR #4498, opened 2022-12-14
bump containerd to v1.5.16

vincentgoat:

**What type of PR is this?**
/kind cleanup

no changes in code, but containerd v1.5.16 is a security release,
so updating, to prevent scanners from marking the dependency to have a vulnerability.

From what I found, it is better to bump docker/docker version to solve this cve, but as of now, docker's latest version does not release this fix. So we may replace containerd version as a workaround.
```
$ go mod why -m github.com/containerd/containerd
# github.com/containerd/containerd
github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/util
github.com/docker/docker/client
github.com/containerd/containerd/platforms
```

Fixes # https://github.com/kubeedge/kubeedge/issues/4499

```release-note

```


fisherxu (commented):
/lgtm
/approve

PR #4506, opened 2022-12-16
Use multiple goroutines to send watch events

wackxu:
**What type of PR is this?**

In the old code implementation, we use a goroutine to send watch eventsg to each listener, which has low processing efficiency and high latency, and if a listener processes slowly, other listeners will be affected. In the new implementation, Every listener use its input channel and one goroutine to send watch event and it will improve processing efficiency

Fixes # part of #4423

```release-note

```

### DIFF
--- a/cloud/pkg/common/messagelayer/context.go
+++ b/cloud/pkg/common/messagelayer/context.go
@@ -18,6 +18,7 @@ package messagelayer
 
 import (
 	"strings"
+	"sync"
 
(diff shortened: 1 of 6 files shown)


wackxu:
This can be reviewed
/assign @fisherxu @vincentgoat @gy95

wackxu:
This should be merged since we have add to the CHANGELOG cc @vincentgoat @fisherxu @gy95

PR #4514, opened 2022-12-22
ignore kube-proxy environment checking

vincentgoat:

**What type of PR is this?**
/kind cleanup

If we launch an edge node and then restart it, an error occurs:
```
failed to check the running environment: kube-proxy should not running on edge node when running edgecore
```
because pod `kube-proxy` will be launched on every node if there is a deamonset in the K8s cluster.

So, it would be more kind to log a warning rather than exit the edgecore process.

Fixes #

```release-note

```

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -154,7 +154,8 @@ func environmentCheck() error {
 		case "kubelet": // if kubelet is running, return error
 			return errors.New("kubelet should not running on edge node when running edgecore")
 		case "kube-proxy": // if kube-proxy is running, return error
-			return errors.New("kube-proxy should not running on edge node when running edgecore")
+			klog.Warningf("kube-proxy is running on edge node, please make sure the kube-proxy is running in the right mode")
+			return nil
 		}
 	}
 


edge/cmd/edgecore/app/server.go
```diff
@@ -154,7 +154,8 @@ func environmentCheck() error {
 		case "kubelet": // if kubelet is running, return error
 			return errors.New("kubelet should not running on edge node when running edgecore")
 		case "kube-proxy": // if kube-proxy is running, return error
-			return errors.New("kube-proxy should not running on edge node when running edgecore")
```
wackxu, line 157: I think when user use edgemesh in the edge node, we had better not run kube-proxy

edge/cmd/edgecore/app/server.go
```diff
@@ -154,7 +154,8 @@ func environmentCheck() error {
...
+			klog.Warningf("kube-proxy is running on edge node, please make sure the kube-proxy is running in the right mode")
```
wackxu, line 157:
```suggestion
			klog.Warningf("kube-proxy is running on edge node. If you plan to use edgemesh, please do not schedule kube-proxy pod to the edge node")
```

edge/cmd/edgecore/app/server.go
fisherxu: Yes, now if users deploy edgemesh on edgenode, it should not install kube-proxy on edge, and now we can't support kube-proxy run normally by default for edge list-watch.

vincentgoat: It's one of the scenes. This check method may not be smooth and we can declare it in the Edgemesh guidebook instead.

PR #4524, opened 2022-12-29
fix edged print too much err log

wackxu:

**What type of PR is this?**

/kind cleanup

Fixes #

fix edged print too much err log

for asynchronous message, the edged only process pod or volume resource, so when wen enable list-watch at edge, there are some resource that edged will do not process so we do not send the resource type that the edged is not processed.

20efa6114f02cbe9a68c2b9ad3bd783

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -21,7 +21,7 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator"
 )
 
-//Constants to check metamanager processes
+// Constants to check metamanager processes
 const (
 	OK = "OK"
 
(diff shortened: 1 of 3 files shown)


edge/pkg/metamanager/process.go
```diff
@@ -38,18 +38,25 @@ func feedbackError(err error, info string, request model.Message) {
...
-	} else {
+		return
+	}
+
+	if resType == model.ResourceTypePod || resType == constants.CSIResourceTypeVolume {
 		beehiveContext.Send(modules.EdgedModuleName, *message)
+		return
 	}
+
+	klog.Warningf("the resType is not pod or volume, do not send resType %s to edged", resType)
```
vincentgoat, line 59: Will this also bring out too much log?

wackxu: done, update to log v4

vincentgoat (commented):
/lgtm

fisherxu (commented):
UT seems always failed

QianQianowo:
The message from dynamiccontroller is sent to edged from metamanager. Why we should do this? These messages are alaways requested by the container running on edged through metaserver. Edged maybe not need to process them.

PR #4527, opened 2022-12-29
update sandboxImage to kubeedge/pause:3.6

gy95:
**What type of PR is this?**

keep kubeedge/pause version the same with k8s vendor sandbox image tag
https://github.com/kubeedge/kubeedge/blob/0de307ffae3694c8b45c510ddae0f9b8aef5f4c2/vendor/k8s.io/kubernetes/cmd/kubelet/app/options/container_runtime.go#L28-L32

Fixes #

```release-note

```

### DIFF
--- a/common/constants/default.go
+++ b/common/constants/default.go
@@ -42,14 +42,16 @@ const (
 	DefaultMqttKeyFile  = "/etc/kubeedge/certs/server.key"
 
 	// Edged
-	DefaultKubeletConfig               = "/etc/kubeedge/config/kubeconfig"
-	DefaultDockerAddress               = "unix:///var/run/docker.sock"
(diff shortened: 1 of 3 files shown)


wackxu:
Do we  upload the image kubeedge/pause:3.5 to dockerhub?

gy95:
yes, and if we bump k8s dependency to v1.23, we should also update kubeedge/pause:3.6 to dockerhub

wackxu:
> yes, and if we bump k8s dependency to v1.23, we should also update kubeedge/pause:3.6 to dockerhub

In this version, we will upgrade the k8s dependency to v1.23, so we can update to kubeedge/pause:3.6 directly

fisherxu (commented):
/lgtm
/approve

PR #4530, opened 2022-12-30
fix comfigmap/secret optional error

vincentgoat:

**What type of PR is this?**
/kind bug

Optional configmap/secret in a pod is allowed. It won't matter if the optional configmap/secret is not found when a pod is launched.
```
  volumes:
  - name: cm-volume
    projected:
      defaultMode: 420
      sources:
      - configMap:
          name: cm-test-opt
          optional: true
```

Fixes #

```release-note

```


wackxu:
I  do not know  why we update like this?  now for edged we only proxy the request to cloud and response to the edged.

vincentgoat:
> I do not know why we update like this? now for edged we only proxy the request to cloud and response to the edged.

If configmap/secret resource did not exist in etcd, then the `not found` error should respond immediately.

fisherxu (commented):
/lgtm
/approve

PR #4540, opened 2023-01-06
use NodePort as default cloudcore service type

gy95:
**What type of PR is this?**

now we use ClusterIP as cloudcore service type, this patch change it to NodePort

service type is defined by .Values.cloudCore.service.type .Values.cloudCore.hostNetWork
https://github.com/kubeedge/kubeedge/blob/2f33fccbb999af56794a08fcd2fea98010ac9975/manifests/charts/cloudcore/templates/service_cloudcore.yaml#L12-L17

the default hostNetwork is set to "true"
https://github.com/kubeedge/kubeedge/blob/2f33fccbb999af56794a08fcd2fea98010ac9975/manifests/charts/cloudcore/values.yaml#L6-L8

the default servicetype is set to NodePort
https://github.com/kubeedge/kubeedge/blob/2f33fccbb999af56794a08fcd2fea98010ac9975/manifests/charts/cloudcore/values.yaml#L58-L60

So we can change .Values.cloudCore.hostNetWork from "true" to false, so that it uses NodePort as cloudcore serviceType

Fixes #

```release-note

```

### DIFF
--- a/manifests/profiles/version.yaml
+++ b/manifests/profiles/version.yaml
@@ -5,7 +5,7 @@ appVersion: "1.12.0"
 
 cloudCore:
   replicaCount: 1
-  hostNetWork: "true"
+  hostNetWork: false
   image:
     repository: "kubeedge/cloudcore"
     tag: "v1.12.0"
@@ -52,7 +52,7 @@ cloudCore:
     router:
       enable: "false"
   service:
-    enable: "false"
+    enable: true
     type: "NodePort"
     cloudhubNodePort: "30000"
     cloudhubQuicNodePort: "30001"
--- a/tests/e2e/scripts/keadm_e2e.sh
+++ b/tests/e2e/scripts/keadm_e2e.sh
@@ -59,7 +59,7 @@ function start_kubeedge() {
   export KUBECONFIG=$HOME/.kube/config
   docker run --rm kubeedge/installation-package:$IMAGE_TAG cat /usr/local/bin/keadm > /usr/local/bin/keadm && chmod +x /usr/local/bin/keadm
 
-  /usr/local/bin/keadm init --advertise-address=$MASTER_IP --profile version=$KUBEEDGE_VERSION --set cloudCore.service.enable=false --kube-config=$KUBECONFIG --force
+  /usr/local/bin/keadm init --advertise-address=$MASTER_IP --profile version=$KUBEEDGE_VERSION --kube-config=$KUBECONFIG --force
   
   # ensure tokensecret is generated
   while true; do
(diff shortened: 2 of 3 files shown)


vincentgoat:
@cc20110101 
This PR would solve the cloudcore service type `clusterIP` issue.

gy95:
this seems not work, i change hostNetWork from true to false, but cloudcore service also uses ClusterIP as its service type, it should be NodePort.
It seems keadm do not recognize the difference between true and false, and treat them the same. i guess the two bool values are treated as string values.
@zhu733756  for help :)

gy95:
oh no, i forgot to modify manifests/profiles/version.yaml 😞  now it changed to NodePort

wackxu:
I think this is a break change, the default port for edge node to access is changed. If users use helm upgrade the cloudcore helm release, the edge node then can not connect to the cloudcore

wackxu:
Also, if we decide to update this, you must update the install in the website

tests/e2e/scripts/keadm_e2e.sh
```diff
@@ -70,7 +70,7 @@ function start_kubeedge() {
   cd $KUBEEDGE_ROOT
   export TOKEN=$(sudo /usr/local/bin/keadm gettoken --kube-config=$KUBECONFIG)
   sudo systemctl set-environment CHECK_EDGECORE_ENVIRONMENT="false"
-  sudo -E CHECK_EDGECORE_ENVIRONMENT="false" /usr/local/bin/keadm join --token=$TOKEN --cloudcore-ipport=$MASTER_IP:10000 --edgenode-name=edge-node --kubeedge-version=$KUBEEDGE_VERSION
+  sudo -E CHECK_EDGECORE_ENVIRONMENT="false" /usr/local/bin/keadm join --token=$TOKEN --certport 30002 --cloudcore-ipport=$MASTER_IP:30000 --edgenode-name=edge-node --kubeedge-version=$KUBEEDGE_VERSION
```
wackxu, line 73: if we need update the keadm_deprecated_e2e.sh?

gy95: no, deprecated_e2e install cloudcore in binary progress mode, but not in k8s container mode.

zhu733756:
> oh no, i forgot to modify manifests/profiles/version.yaml 😞 now it changed to NodePort

Yes,  the former configuration "HostNetwork: true" would not use service template.  If we use false, we should enable the service.

manifests/profiles/version.yaml
```diff
@@ -5,7 +5,7 @@ appVersion: "1.12.0"
 
 cloudCore:
   replicaCount: 1
-  hostNetWork: "true"
+  hostNetWork: false
   image:
     repository: "kubeedge/cloudcore"
     tag: "v1.12.0"
```
zhu733756, line 11: Line#55 in this file should be "  enable: "true"" ?

gy95: yes

fisherxu (commented):
Seems this patch is changing from `hostNetwork` to `NodePort`? cc @gy95 @zhu733756

gy95:
yes, before this patch, cloudcore pod use hostNetwork=true, and cloudcore serviceType is ClusterIP
after this, cloudcore pod use hostNetwork=false, and cloudcore serviceType is NodePort

fisherxu:
The access `port` in cloudcore will be changed after this patch merged, it will be better write in docs, user can set it bu themselves. And if we want to change to `NodePort`, it's better to have  more discussion and notification later. @gy95

gy95:
> 

sure yes

PR #4548, opened 2023-01-10
new keadm support v1alpha1 edge config  

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

After v1.12.0 released, the edgecore config has changed to v1alpha2 version. We cannot use the new keadm to install the earlier version.  With this pr, the new keadm both support creating v1alpha1 and v1alpha2 edge config.

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/join.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/blang/semver"
 	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
 	kubetypes "k8s.io/kubernetes/pkg/kubelet/types"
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -212,6 +214,12 @@ func createDirs() error {
 }
 
 func createEdgeConfigFiles(opt *common.JoinOptions) error {
+	// Determines whether the kubeEdgeVersion is earlier than v1.12.0
+	// If so, we need to create edgeconfig with v1alpha1 version
+	if res, err := util.CompareVersion(opt.KubeEdgeVersion, "v1.12.0"); err == nil && res == -1 {
```
gy95, line 219: maybe just use github.com/blang/semver version.compare

Shelley-BaoYue: > maybe just use github.com/blang/semver version.compare

good idea. has fixed.

wackxu:
Since we are about to release version 1.13, I think the strategy for supporting older versions needs to be finalized，for example, in kubernetes, 

```
kubeadm can be used with Kubernetes components that are the same version as kubeadm or one version older. The Kubernetes version can be specified to kubeadm by using the --kubernetes-version flag of kubeadm init or the [ClusterConfiguration.kubernetesVersion](https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/) field when using --config. This option will control the versions of kube-apiserver, kube-controller-manager, kube-scheduler and kube-proxy.

Example:

kubeadm is at 1.26
kubernetesVersion must be at 1.26 or 1.25
```

fisherxu (commented):
/lgtm
/approve

PR #4578, opened 2023-02-01
bugfix container terminated when edgecore restart

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

When edgecore restart, edged need to query pod from metamanager and set into pod update channel.  In 1.13 version, edged dropped these pod, cause containerd termineted when edgecore restart.

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -300,8 +300,6 @@ func (e *edged) handlePodListFromMetaManager(content []byte, updatesChan chan<- 
 	}
 
 	var pods []*v1.Pod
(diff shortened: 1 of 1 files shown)


Shelley-BaoYue:
cc @vincentgoat  @wackxu

wackxu:
/lgtm
need cherry-pick to v1.13

vincentgoat (commented):
/lgtm

huapox:
+1
--
self fixed with the same code area

fisherxu (commented):
/lgtm
/approve

PR #4585, opened 2023-02-06
fix cni issue for containerd/cni in localup

fisherxu:
Signed-off-by: fisherxu <fisherxu1@gmail.com>

**What type of PR is this?**
/kind bug

When I use localup script to install the cluste, the edgenode will get  notReady, since there's no cni netconf file in node. So it's need to create a netconf file if it doesn't exist.

Fixes #

cc @gy95 @vincentgoat 

```release-note
NONE
```

### DIFF
--- a/hack/lib/install.sh
+++ b/hack/lib/install.sh
@@ -98,11 +98,56 @@ verify_docker_installed(){
 function install_cni_plugins() {
   # install CNI plugins if not exist
   if [ ! -f "/opt/cni/bin/loopback" ]; then
-    echo -e "install CNI plugins..."
+    echo -e "installing CNI plugins..."
     mkdir -p /opt/cni/bin
-    wget https://github.com/containernetworking/plugins/releases/download/v1.1.1/cni-plugins-linux-amd64-v1.1.1.tgz
-    tar Cxzvf /opt/cni/bin cni-plugins-linux-amd64-v1.1.1.tgz
+    wget $CNI_DOWNLOAD_ADDR
+    tar Cxzvf /opt/cni/bin ${CNI_DOWNLOAD_ADDR##*/}
+    # remove installation package after the installation is complete
+    rm -rf ${KUBEEDGE_ROOT}/${CNI_DOWNLOAD_ADDR##*/}
   else
     echo "CNI plugins already installed and no need to install"
+  fi
+
+  # create CNI netconf file if not exist
+  CNI_DIR="/etc/cni/net.d"
+  mkdir -p /etc/cni/net.d
+  if [ "`ls -A $CNI_DIR`" = "" ]; then
+    echo -e "creating CNI netconf file..."
+    cat > /etc/cni/net.d/10-containerd-net.conflist <<EOF
+    {
+     "cniVersion": "1.0.0",
+     "name": "containerd-net",
+     "plugins": [
+       {
+         "type": "bridge",
+         "bridge": "cni0",
+         "isGateway": true,
+         "ipMasq": true,
+         "promiscMode": true,
+         "ipam": {
+           "type": "host-local",
+           "ranges": [
+             [{
+               "subnet": "10.88.0.0/16"
+             }],
+             [{
+               "subnet": "2001:db8:4860::/64"
+             }]
+           ],
+           "routes": [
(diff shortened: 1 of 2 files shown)


hack/local-up-kubeedge.sh
```diff
@@ -263,6 +258,12 @@ check_control_plane_ready
 kubectl delete daemonset kindnet -nkube-system
 kubectl create ns kubeedge
 
+# install CNI plugins
+if [[ "${CONTAINER_RUNTIME}" = "remote" ]]; then
+  # we need to install CNI plugins only when we use remote(containerd) as edgecore container runtime
+  install_cni_plugins
+fi
```
fisherxu, line 265: Install cni when kind cluster is ready.

hack/lib/install.sh
```diff
@@ -98,11 +98,55 @@ verify_docker_installed(){
...
+           ]
+         }
+       },
+       {
+         "type": "portmap",
+         "capabilities": {"portMappings": true}
+       }
+     ]
+    }
+EOF
```
fisherxu, line 148: This file is copied from: https://kubernetes.io/docs/tasks/administer-cluster/migrating-from-dockershim/troubleshooting-cni-plugin-related-errors/

hack/lib/install.sh
```diff
@@ -98,11 +98,55 @@ verify_docker_installed(){
...
+    tar Cxzvf /opt/cni/bin ${CNI_DOWNLOAD_ADDR##*/}
+    # remove installation package after the installation is complete
+    rm -rf ${KUBEEDGE_ROOT}/${CNI_DOWNLOAD_ADDR##*/}
   else
     echo "CNI plugins already installed and no need to install"
   fi
+
+  # create CNI netconf file if not exist
+  CNI_DIR="/etc/cni/net.d"
+  if [ "`ls -A $CNI_DIR`" = "" ]; then
```
gy95, line 113: if this directory not exist, it will also NOT create netconf file, so better add a `mkdir -p /etc/cni/net.d`

gy95: maybe just check whether file `/etc/cni/net.d/10-containerd-net.conflist` exist or not

fisherxu: Done, when CNI_DIR not exist,  it will also print `ls: cannot access '/etc/cni/net.d': No such file or directory`

fisherxu: Seems this file exists in node of github action, the local-up can run in CI :)

gy95: on my local machine, this file not exists, but it seems run well 😂

hack/lib/install.sh
```diff
@@ -98,11 +98,56 @@ verify_docker_installed(){
...
   else
     echo "CNI plugins already installed and no need to install"
   fi
+
+  # create CNI netconf file if not exist
+  CNI_DIR="/etc/cni/net.d"
+  mkdir -p /etc/cni/net.d
+  if [ "`ls -A $CNI_DIR`" = "" ]; then
+    echo -e "creating CNI netconf file..."
+    cat > /etc/cni/net.d/10-containerd-net.conflist <<EOF
```
vincentgoat, line 116: One exceptional case is that cni0 already existed on the host before we install CNI plugins. CNI plugin works without this conf file.
Suggesting check if cni0 exists before installing: `ip a |grep cni0`

hack/local-up-kubeedge.sh
```diff
@@ -21,6 +21,7 @@ LOG_LEVEL=${LOG_LEVEL:-2}
 TIMEOUT=${TIMEOUT:-60}s
 PROTOCOL=${PROTOCOL:-"WebSocket"}
 CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"remote"}
+CNI_DOWNLOAD_ADDR="https://github.com/containernetworking/plugins/releases/download/v1.1.1/cni-plugins-linux-amd64-v1.1.1.tgz"
```
gy95, line 24: moved into function install_cni_plugins, function install_cni_plugins is also used by keadm_e2e and keadm_deprecated_e2e

hack/lib/install.sh
gy95: add sudo?

PR #4587, opened 2023-02-06
correct default comments for edged configuration 

Shelley-BaoYue:
**What type of PR is this?**

/kind cleanup

Correct inconsistencies between comments and defaults in edged configuration


gy95:
LGTM

PR #4612, opened 2023-02-22
deal with error in delete pod upstream msg

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

The response error when delete pod failed wasn't dealt with. In older version, err in  `edge/pkg/metamanager/client/pod.go` actually is nil. So in this pr, we set err in msg content and parse it at edge.


fisherxu (commented):
/lgtm
/approve

PR #4614, opened 2023-02-23
feedback when failed to get special resource

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

Without feedback error to edged, when failed to get special resource, edged can not get response until timeout.


Shelley-BaoYue:
cc @vincentgoat

vincentgoat (commented):
/lgtm

Shelley-BaoYue:
/close
fixed in #4802

PR #4631, opened 2023-02-28
fix delete statefulset pod error

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

When delete sts pod, the sequence of the downstream `delete` and `insert` messages cannot be guaranteed, and the pod names are the same, so we need to check if podUID equals when we delete the podDB. There are two mistakes here:
- we use `insert` instead of `insert of replace` when we process insert msg, cause `insert` fails when old pod db is not deleted.
- we use query first and then delete, it may cause the new DB be deleted.  So we place the judgment and deletion statement in one stomic sql operation statement in this pr.

### DIFF
--- a/edge/pkg/metamanager/dao/meta.go
+++ b/edge/pkg/metamanager/dao/meta.go
@@ -1,6 +1,7 @@
 package dao
 
 import (
+	"fmt"
 	"strings"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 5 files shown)


caoxuhao:
I merge this commit to my  kubeedge v1.12.2, and build it. The following Pictures show that when edgecore receive insert cmd before delete cmd, the pod status is also running. It proves this pr is really worked, and the bug is fixed.
image

image

edge/pkg/metamanager/dao/meta.go
```diff
@@ -47,6 +48,17 @@ func DeleteMetaByKey(key string) error {
 	return err
 }
 
+// DeleteMetaByKeyAndPodUID delete meta by key and podUID
+func DeleteMetaByKeyAndPodUID(key, podUID string) (int64, error) {
+	sqlStr := fmt.Sprintf("DELETE FROM meta WHERE key = '%s' and value LIKE '%%%s%%'", key, podUID)
+	res, err := dbm.DBAccess.Raw(sqlStr).Exec()
+	if err != nil {
+		klog.Errorf("delete pod by key %s and podUID failed, err: %v", key, podUID)
```
TianTianBigWang, line 56: seem that log is wrong, maybe like this?
`
klog.Errorf("delete pod by key %s and podUID %s failed, err: %v", key, podUID, err)
`

Shelley-BaoYue: > seem that log is wrong, maybe like this? `klog.Errorf("delete pod by key %s and podUID %s failed, err: %v", key, podUID, err)`

yes， I have fixed it.

Shelley-BaoYue:
> I merge this commit to my kubeedge v1.12.2, and build it. The following Pictures show that when edgecore receive insert cmd before delete cmd, the pod status is also running. It proves this pr is really worked, and the bug is fixed. image image

thanks.

fisherxu (commented):
/approve

PR #4637, opened 2023-03-03
Optimize bool value judgment

panguicai008:
**What type of PR is this?**

/kind cleanup

```release-note
NONE
```

### DIFF
--- a/cloud/test/integration/controllermanager/controllermanager_test.go
+++ b/cloud/test/integration/controllermanager/controllermanager_test.go
@@ -1452,8 +1452,8 @@ func isOwnedByEdgeApp(ownerReferences []metav1.OwnerReference, edgeApp *appsv1al
 	for _, o := range ownerReferences {
 		if o.APIVersion == appsv1alpha1.GroupVersion.String() &&
 			o.Kind == "EdgeApplication" &&
-			*o.BlockOwnerDeletion == true &&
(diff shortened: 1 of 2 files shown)


cloud/test/integration/controllermanager/controllermanager_test.go
```diff
@@ -1452,8 +1452,8 @@ func isOwnedByEdgeApp(ownerReferences []metav1.OwnerReference, edgeApp *appsv1al
 	for _, o := range ownerReferences {
 		if o.APIVersion == appsv1alpha1.GroupVersion.String() &&
 			o.Kind == "EdgeApplication" &&
-			*o.BlockOwnerDeletion == true &&
-			*o.Controller == true &&
+			*o.BlockOwnerDeletion &&
+			*o.Controller &&
```
zc2638, line 1456: Is it possible that `o.BlockOwnerDeletion` or `o. Controller` is nil, which will cause a null pointer panic?

panguicai008: o.BlockOwnerDeletion and o. Controller have specific values，for example 
- apiVersion: apps.kubeedge.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: EdgeApplication
    name: sample

PR #4656, opened 2023-03-10
function returns a more accurate error message

WillardHu:
…error message

**What type of PR is this?**
/kind cleanup

In the processDeletePodDB(..) func, if `dao.QueryMeta(..)` or `dao.DeleteMetaByKey(..)` returns an error, there's no way to know where err was returned

Fixes #

```release-note

```


vincentgoat:
It would be better if UT could be added correspondingly.

WillardHu:
> It would be better if UT could be added correspondingly.
Done

PR #4670, opened 2023-03-15
allow to use more special letter  in device twin value

QianQianowo:
**What type of PR is this?**
/kind feature

We can use more special letter, such as % + , = @ # : / ^ ~ ? & ! *, in device twin value. And we need to validate the device twin  value which have  integer and float-pointing types.  

Fixes #

```release-note

```


Shelley-BaoYue:
The old devices will affected with more stricter validation?

QianQianowo:
> 

>

QianQianowo:
> The old devices will affected with more stricter validation?

This pr is to add more special letters such as % + , = @ # : / ^ ~ ? & ! *. Those special letters on the old devices will not affected.

PR #4671, opened 2023-03-15
bugfix metaserver enable

QianQianowo:
**What type of PR is this?**
/kind bug

The `Enable` parameter in the configure of metaserver has not been used.  The imitator module may still work even if the `Enable` parameter set `False`.

Fixes #

```release-note

```


Shelley-BaoYue:
Actually, I have considered this pr before, but I am wondering if there is a more elegant way to solve it.  With this pr merged, we need to do this judgement for each message, whether metaserver is enable or not.

PR #4674, opened 2023-03-15
bugfix device status updated when timestamp exists in twin

QianQianowo:
**What type of PR is this?**
/kind bug

The device status will update too frequently when timestamp exists in the `Metadata` of device twin properties. We only need to check whether the Desired Value for the device twin property is updated.

Fixes #

```release-note

```


WillardHu:
This fix will improve system performance, but is there ambiguity if the customer cloud applications watch to the device status for time-series data?
e.g.: We get the temperature per hour [13, 13, 12, 12, 12, 11, 11, 11]，but after modification we will get [13,12,11]

PR #4676, opened 2023-03-15
bugfix get bad anonName when send twin message to cloud

QianQianowo:
**What type of PR is this?**
/kind bug

The `beehiveContext.Send` method is used in device twin module to send twin message to cloud before. But the response message is send by edgehub module using method `beehiveContext.SendResp`. It may cause an error and the response message will be lost.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/communicate.go
+++ b/edge/pkg/devicetwin/dtmanager/communicate.go
@@ -23,17 +23,21 @@ var (
 	ActionCallBack map[string]CallBack
 )
 
+const (
+	syncMsgRespTimeout = time.Minute
+)
+
 func init() {
 	initActionCallBack()
 }
 
-//CommWorker deal app response event
+// CommWorker deal app response event
 type CommWorker struct {
 	Worker
 	Group string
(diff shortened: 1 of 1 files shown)


RyanZhaoXB:
please fix UT

TianTianBigWang:
why use `beehiveContext.SendSync` rather than `beehiveContext.SendResp`?

fisherxu (commented):
/sig device-iot

QianQianowo:
> 

`beehiveContext.SendSync` means send a synchronized beehive message and wait for the response. The reply message will be returned through `beehiveContext.SendResp`. `beehiveContext.SendResp` is always used to send a reply message to the synchronized messages.

QianQianowo:
> why use `beehiveContext.SendSync` rather than `beehiveContext.SendResp`?

beehiveContext.SendSync means send a synchronized beehive message and wait for the response. The reply message will be returned through beehiveContext.SendResp. beehiveContext.SendResp is always used to send a reply message to the synchronized messages.

Issue #4684, opened 2023-03-20
success to report an illegal device twin data to edgecore for the first time


RyanZhaoXB:
**What happened**:
When my device report an illegal data to edgecore **for the first time**, the edgecore will not validate the data and still report it to cloudcore.

**What you expected to happen**:
Validate the data and do not send it to cloudcore when the data is illegal for the first time.

**How to reproduce it (as minimally and precisely as possible)**:
create a device in KubeEdge and use simulator to report an illegal device twin reported value. and get device twin reported value from k8s apiserver.

PR #4686, opened 2023-03-20
fix the bug of report illegal devicetwin value for the first time

RyanZhaoXB:
**What type of PR is this?**

/kind bug

update isTwinValueDiff to fix the bug of report illegal devicetwin value for the first time.  fix this bug and enhance the security of KubeEdge Device Management.

Fixes  https://github.com/kubeedge/kubeedge/issues/4684

NONE

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/twin.go
+++ b/edge/pkg/devicetwin/dtmanager/twin.go
@@ -494,16 +494,18 @@ func isTwinValueDiff(twin *dttype.MsgTwin, msgTwin *dttype.MsgTwin, dealType int
 	twinValue := twin.Expected
 	msgTwinValue := msgTwin.Expected
 
+	if twinValue != nil {
+		hasTwin = true
+	}
 	if dealType == DealActual {
 		twinValue = twin.Actual
 		msgTwinValue = msgTwin.Actual
-	}
-	if twinValue != nil {
 		hasTwin = true
 	}
 	if msgTwinValue != nil {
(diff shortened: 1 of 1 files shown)


TianTianBigWang:
I feel that the error has nothing to do with the judgment conditions. 

If `twinValue` is not `nil`, then whether it is before or after `dealType == DealActual`,it can set `hasTwin=true` ?

RyanZhaoXB:
> I feel that the error has nothing to do with the judgment conditions.
> 
> If `twinValue` is not `nil`, then whether it is before or after `dealType == DealActual`,it can set `hasTwin=true` ?

You are right. if `dealType==DealActual`, `hasTwin` should always be `true`. But for the origin code, `twinValue` will be set to `nil` because `twin.Actual` is nil before the mapper report `actual value` for the first time. Thus `hasTwin` will be set to false and the verification will be skipped.

TianTianBigWang:
> > I feel that the error has nothing to do with the judgment conditions.
> > If `twinValue` is not `nil`, then whether it is before or after `dealType == DealActual`,it can set `hasTwin=true` ?
> 
> You are right. if `dealType==DealActual`, `hasTwin` should always be `true`. But for the origin code, `twinValue` will be set to `nil` because `twin.Actual` is nil before the mapper report `actual value` for the first time. Thus `hasTwin` will be set to false and the verification will be skipped.

OK, I see what you mean. I suggest adding a comment to the code.😊

PR #4696, opened 2023-03-23
update pod db when patch successfully

### LINKED ISSUE #4687: The status update of the pod on the edge node failed.
When I use a script to concurrently connect 50 edge nodes to the cloud, there is a phenomenon of frequent switching between the Ready and NotReady states of the nodes (I suspect that this is due to network blocking causing loss of heartbeat requests). At this time, the DeamonSet will start pods on newly connected edge nodes, and when the node status stabilizes, the PodReady and ContainersReady in the pod's Condition are often set to false and not updated. I have read the source code and have some questions. The PatchPodStatus() method in vendor/k8s.io/kubernetes/pkg/util/pod/pod.go will compare the old and new podStatus, and then only patch the changed parts in podStatus. At the same time, the processPatch() method in edge/pkg/metamanager/process.go will first update the patch request content to the database and then send it to the cloud. The problem now is, when encountering unstable network conditions when connecting to edge nodes as described above, will there be a situation where the patch update to the database is successful but the sendToCloud fails? Also, because kubelet only patches the changed content, will it cause some pod statuses to never be updated?

**Environment**:
- Kubernetes version (use `kubectl version`):1.20.10
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`):cloudcore:1.12.1   edgecore:1.12.2

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

update pod meta in edge database when patch pod successfully,  because edged query pod request will read from edge db.  

Fixes #4687

### DIFF
--- a/edge/pkg/metamanager/client/pod.go
+++ b/edge/pkg/metamanager/client/pod.go
@@ -4,16 +4,19 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog/v2"
 
 	"github.com/kubeedge/beehive/pkg/core/model"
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
/lgtm
/approve

Issue #4701, opened 2023-03-27
edge node get Ready and NotReady repeatedly after the mapper get offline


RyanZhaoXB:
**What happened**:

we have installed an KubeEdge cluster,  deployed a mapper on one edge node and created a device instance. the mapper reported the device twin of the device instance to KubeEdge continuously. suddenly the mapper was offline and after that the edge node started to get `Ready` and `NotReady` repeatedly. 

**What you expected to happen**:
the edge node keeps to get `Ready`.

**How to reproduce it (as minimally and precisely as possible)**:
deploy a mapper and create a device. the mapper keeps to report the device twin for a while. then delete the deployment of the mapper to stop reporting device twin. 

**Anything else we need to know?**:

we have checked the log of edgecore. we found a log of logs like this:
``` 
W0217 16:11:08.145257   21830 context_channel.go:180] The module websocket message channel is full, message: {Header:{ID:c5918410-7207-4acd-aa6a-e5446e053aee ParentID: Timestamp:1676621409138 ResourceVersion: Sync:true MessageType:} Router:{Source:edged Destination: Group:meta Operation:patch Resource:default/nodepatch/edge-line-01} Content:{"status":{"$setElementOrder/conditions":[{"type":"MemoryPressure"},{"type":"DiskPressure"},{"type":"PIDPressure"},{"type":"Ready"}],"conditions":[{"lastHeartbeatTime":"2023-02-17T08:10:09Z","lastTransitionTime":"2023-02-17T08:10:09Z","message":"kubelet has sufficient memory available","reason":"KubeletHasSufficientMemory","status":"False","type":"MemoryPressure"},{"lastHeartbeatTime":"2023-02-17T08:10:09Z","lastTransitionTime":"2023-02-17T08:10:09Z","message":"kubelet has no disk pressure","reason":"KubeletHasNoDiskPressure","status":"False","type":"DiskPressure"},{"lastHeartbeatTime":"2023-02-17T08:10:09Z","lastTransitionTime":"2023-02-17T08:10:09Z","message":"kubelet has sufficient PID available","reason":"KubeletHasSufficientPID","status":"False","type":"PIDPressure"},{"lastHeartbeatTime":"2023-02-17T08:10:09Z","lastTransitionTime":"2023-02-17T08:10:09Z","message":"edge is posting ready status","reason":"EdgeReady","status":"True","type":"Ready"}]}}}
```
[...]

PR #4702, opened 2023-03-27
[bugfix]edgehub process send device twin msg to twingroup directly

### LINKED ISSUE #4700: Memory leaks caused by devicetwin module when the device mapper is deployed
**What happened**:
Now when I try to deploy deivce mapper pods ，I find  that the edgecore process is consuming more and more memory , so I suspect a memory leak caused by improper devicetwin module code.

Just deploy the device mapper in any kubeedge environment.

After I try to understand the device twin code and edgehub code , It is basically confirmed that the problem is caused by the error of confusing the relationship between synchronous asynchro and response when the two modules send acknowledgement messages using the beehave framework.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.21.8
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): v1.21.1

  - OS (e.g. `cat /etc/os-release`): NAME="Ubuntu" VERSION="18.04.5 LTS (Bionic Beaver)"
  - Kernel (e.g. `uname -a`):Linux controller01 4.15.0-191-generic #202-Ubuntu SMP Thu Aug 4 01:49:29 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
  - Go version (e.g. `go version`):go version go1.18 linux/amd64
  - Others:

  </details>

  - edgecore version (e.g. `edgecore --version`):v1.21.1
  - OS (e.g. `cat /etc/os-release`): NAME="Ubuntu" VERSION="18.04.5 LTS (Bionic Beaver)"
  - Kernel (e.g. `uname -a`):Linux edge-line-01 4.19.90+ #1 SMP Sat Jan 29 00:19:13 CST 2022 aarch64 aarch64 aarch64 GNU/Linux
  - Go version (e.g. `go version`):go version go1.18.5 linux/arm64
  - Others: This problem also occurs on the edge of euleros systems.

  </details>

RyanZhaoXB:
edgehub process send twin msg to twingroup directly to avoid the response of device report be handled by beehiveContext.SendResp()

**What type of PR is this?**
/kind bug

Here is the flow chart of the related step of device management. After the `devicetwin` send a message to cloud, the `upstreamController` of `devicecontroller` will handle this message and send a response message back to edge. The response message deliverd to edgehub will be handled by `isSyncResponse()` and will never be sent to `devicetwin` in the current code which will cause that the `dealConfirm()` function will never be triggered and the `ConfirmMap` will never be cleaned correctly. Further more, every 60s after `devicetwin` does not receive any message, the `checkConfirm()` function will range the `ConfirmMap` and send all the history messages to cloud, which will cause many problem.
设备消息回复流程

In this PR, I intercept the response message of device report at `edgehub` and send it to `devicetwin` directly, which will fix the problem temporarily.

Fixes #4700 #4437 #4701

This bug is related to the reliability transmission mechanism of device messages which could cause many related problems of device. In this patch version, we will use a temporay fix to solve this problem. The code related to device message transmission will be reconstructed in the next offical version, which will completely fix this problem.

### DIFF
--- a/edge/pkg/devicetwin/process.go
+++ b/edge/pkg/devicetwin/process.go
@@ -52,8 +52,10 @@ func (dt *DeviceTwin) distributeMsg(m interface{}) error {
 		klog.Infof("Send msg to the %s module in twin", dtcommon.CommModule)
 		confirmMsg := dttype.DTMessage{Msg: model.NewMessage(message.Msg.GetParentID()), Action: dtcommon.Confirm}
 		if err := dt.DTContexts.CommTo(dtcommon.CommModule, &confirmMsg); err != nil {
+			klog.Errorf("fail to send msg %s to CommModule with err: %+v", confirmMsg.Msg.Header.ID, err)
 			return err
 		}
+		return nil
 	}
 	if !classifyMsg(&message) {
 		return errors.New("not found action")
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -73,6 +73,14 @@ func (*defaultHandler) Process(message *model.Message, clientHub clients.Adapter
 		md = modules.BusGroup
 	}
 
+	// TODO: just for a temporary fix.
+	// The code related to device twin message transmission will be reconstructed
+	//  by using sendSync function instead of send function.
+	if group == messagepkg.TwinGroupName {
+		beehiveContext.SendToGroup(md, *message)
+		return nil
+	}
+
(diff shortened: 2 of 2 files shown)


wackxu (commented):
suggest add TODO for the temp fix

RyanZhaoXB:
> suggest add TODO for the temp fix

OK, I will add TODO and recommit it.

cl2017 (commented):
/lgtm

fisherxu (commented):
/lgtm
/approve

PR #4712, opened 2023-03-31
replace get process lib

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

replace get process lib with `shirou/gopsutil/v3`, because lib `mitchellh/go-ps` hasn't been maintained for a long time.


wackxu (approved):
LGTM

PR #4727, opened 2023-04-11
update UT of devicetwin to reduce the duplication

RyanZhaoXB:
**What type of PR is this?**

/kind test

There is much duplication for the code of devicetwin UT. This PR modifies the related code and reduce the duplication.

### DIFF
--- a/edge/pkg/devicetwin/dtcontext/dtcontext_test.go
+++ b/edge/pkg/devicetwin/dtcontext/dtcontext_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package dtcontext
 
 import (
-	"encoding/json"
 	"errors"
 	"reflect"
 	"sync"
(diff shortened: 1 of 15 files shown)


edge/pkg/devicetwin/dtcontext/dtcontext_test.go
```diff
@@ -27,6 +26,7 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/common/modules"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dtcommon"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dttype"
+	devicetest "github.com/kubeedge/kubeedge/edge/test/device_test"
```
wackxu, line 29: We'd better not import the edge/test package from edge/pkg, the edge/test package is used for integration test, mabe we can put the code in devicetwin/testutil directory

RyanZhaoXB: OK

RyanZhaoXB: I have move the code to devicetwin/testutil

PR #4728, opened 2023-04-11
update type of Device from devicetwin/dttype/types

RyanZhaoXB:
**What type of PR is this?**

/kind cleanup

make all the device-related types referenced from `devicetwin/dttype/types` and remove the types defined in `devicecontroller/types` which can reduce the duplication of the code.


cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -40,6 +40,7 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/pkg/devicecontroller/constants"
 	"github.com/kubeedge/kubeedge/cloud/pkg/devicecontroller/manager"
 	"github.com/kubeedge/kubeedge/cloud/pkg/devicecontroller/types"
+	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dttype"
```
wackxu, line 43: I think it is strange that the cloudcore code import edgecore code, I think we can put the common type in the pkg/ directory

RyanZhaoXB: I have moved the edge/pkg/devicetwin/dttype to pkg/common/dttype

PR #4753, opened 2023-04-25
simplified regexp and fix grammar error

cl2017:
**What type of PR is this?**
/kind cleanup

fix grammar error

Fixes #

Please have a look at regex changes, the editor show "\" is redundant, I tried to delete it, and didn't make a difference. Whether we should delete it?

```release-note
NONE
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -46,7 +46,7 @@ import (
 )
 
 var (
-	kubeReleaseRegex = regexp.MustCompile(`^v?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)([-0-9a-zA-Z_\.+]*)?$`)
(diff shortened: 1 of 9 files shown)


keadm/cmd/keadm/app/cmd/util/common.go
```diff
@@ -46,7 +46,7 @@ import (
 )
 
 var (
-	kubeReleaseRegex = regexp.MustCompile(`^v?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)([-0-9a-zA-Z_\.+]*)?$`)
```
fisherxu, line 49: Why delete the `\` here?  We just import it from kubernetes, and it's also changed in K8s, ref to this patch: https://github.com/kubernetes/kubernetes/pull/112241    @cl2017

cl2017: ok, I had changed it from k8s

cl2017:
@fisherxu I've already changed it, please have a look

Iceber:
lgtm, But it needs to be organized for commits

PR #4771, opened 2023-05-18
Fixed the issue that the Resource field in the message cloudcore sent to edgecore was not same when the targetUrl started with /

hellocn9:
Fixed the issue that the Resource field in the message cloudcore sent to edgecore was not same when the targetUrl started with /

**What type of PR is this?**
/kind bug

In cloudcore, the RegisterListener and UnregisterListener methods of servicebus use different methods for splicing TargetURLs, resulting in inconsistent splicing results when the TargetURL starts with /, and different Resource fields in the final message sent to edgecore.

fix


hellocn9:
@sids-b @kevin-wangzefeng please review, thanks

hellocn9:
@kevin-wangzefeng @sids-b  Can you review this pr? Or what shoud I do?

PR #4785, opened 2023-05-25
wait for cache sync

vincentgoat:
**What type of PR is this?**
/kind bug

informerPair.Informer.HasSynced() is false in the cloudcore rebooting stage, the object (especially crd) not found error will occur, and then the function `manageObject` in the syncController will delete the edge object incorrectly.

Fixes #

```release-note

```


wackxu (approved):
good catch, we should cherry-pick it to the 1.13
/lgtm

Issue #4766, opened 2023-05-12
Is there any planned support for Kubernetes 1.24 or 1.25 in the near future?


lszxyz:

Shelley-BaoYue:
Yes, we plan to support for Kubernetes 1.24 in the upcoming version. Pr: #4789  It is still in the final debugging phase.

yjqg6666:
The most recent supported version is 1.23 [which is EOL at 2023-02-28](https://kubernetes.io/releases/patch-releases/#non-active-branch-history).

yjqg6666:
Any release schedule for the new version coming?

PR #4789, opened 2023-05-26
Bump k8s  to1.24

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

bump Kubernetes dependence version to 1.24

#4766


yjqg6666:
Can we run kubeedge on k8s 1.24 if merged your branch into the current master and do a self-building ?

Shelley-BaoYue:
> Can we run kubeedge on k8s 1.24 if merged your branch into the current master and do a self-building ?

It should work. Feel free to communicate here if you meet any questions.

yjqg6666:
> It should work. Feel free to communicate here if you meet any questions.
Cool, thanks.

Issue #4763, opened 2023-05-10
[OSPP] Migration Solution for Edge Devices among Multi-nodes based on KubeEdge Device Management Interface(DMI)


RyanZhaoXB:
### project name:

Migration Solution for Edge Devices among Multi-nodes based on KubeEdge Device Management Interface(DMI)

### project description:

KubeEdge provides cloud-native lifecycle management for edge smart devices. In the existing KubeEdge DMI framework, each edge device is bound to a specified edge node. When the edge node is faulty, the edge device is disconnected from KubeEdge. This project plans to propose a multi-node edge device migration solution based on KubeEdge DMI framework. A single device can be specified to a group of edge nodes for delivery, and one of the edge nodes is selected for connection. When the connected edge node is faulty, the edge device automatically connects to other edge nodes in the group.

### Problems to be Solved:

In the current device management solution of KubeEdge, edge devices are bound to edge nodes one by one. When creating edge devices, you need to specify edge nodes for successful delivery. Once an edge node is faulty and offline, the edge device also goes offline. Therefore, it is difficult to implement HA access for edge devices.

### Application Scenario:

This topic focuses on the scenario where devices and nodes are not directly connected by cables. If devices are directly connected by cables, manual operation is required during switchover, in which scenario, automatic switchover is meaningless. In this topic, there are two types of indirect connection scenarios:

1. Use wireless access protocols, such as Bluetooth or Wi-Fi.
2. A wired LAN access protocol is used. For example, devices and nodes are connected to routers and are selected based on IP addresses. A practical scenario, for example, is network cameras management.

### Skill Requirements:

Linux, Golang, Kubernetes

##### Main Target

Manage edge devices like pods.

![设备迁移](https://github.com/kubeedge/kubeedge/assets/100741994/6e34ad1f-5351-4280-8b4f-05280077433a)

##### Specify multiple nodes to deliver edge devices.

* [Basic requirement] Update the Device CRD and DeviceController modules in CloudCore so that users can specify one or a group of edge nodes for deliver  edge devices.

* [Advanced requirement] A group of edge nodes can be specified in more ways than a simple list. For example, node group predicates mechanism is implemented by using node affinity and taint.

##### Multiple Edge Device Selection Policies

[...]

PR #4803, opened 2023-05-30
device node selector

cl2017:
**What type of PR is this?**
/kind feature

Before, we use device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0] to bind node, this way is equivocating, because it not really implement 'NodeSelector' feature just like k8s. This is an issue left over from history.

This PR added device.Spec.NodeName to mean this device onto what node, and compatible with previous usage.

Fixes https://github.com/kubeedge/kubeedge/issues/4868

```release-note

```

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -149,16 +149,16 @@ func (dc *DownstreamController) syncDevice() {
 
 // addToConfigMap adds device in the configmap
 func (dc *DownstreamController) addToConfigMap(device *v1alpha2.Device) {
-	configMap, ok := dc.configMapManager.ConfigMap.Load(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0])
+	configMap, ok := dc.configMapManager.ConfigMap.Load(getDeviceNode(device))
 	if !ok {
 		nodeConfigMap := &v1.ConfigMap{}
-		nodeConfigMap.Name = constants.DeviceProfileConfigPrefix + device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0]
+		nodeConfigMap.Name = constants.DeviceProfileConfigPrefix + getDeviceNode(device)
 		nodeConfigMap.Namespace = device.Namespace
 		nodeConfigMap.Data = make(map[string]string)
 		// TODO: how to handle 2 device of multiple namespaces bind to same node ?
 		dc.addDeviceProfile(device, nodeConfigMap)
 		// store new config map
-		dc.configMapManager.ConfigMap.Store(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0], nodeConfigMap)
+		dc.configMapManager.ConfigMap.Store(getDeviceNode(device), nodeConfigMap)
 
 		if _, err := dc.kubeClient.CoreV1().ConfigMaps(device.Namespace).Get(context.Background(), nodeConfigMap.Name, metav1.GetOptions{}); err != nil {
 			if apierrors.IsNotFound(err) {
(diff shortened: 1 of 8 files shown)


pkg/apis/devices/v1alpha2/device_instance_types.go
```diff
@@ -42,6 +42,11 @@ type DeviceSpec struct {
 	// Refer to k8s.io/kubernetes/pkg/apis/core NodeSelector for more details
 	// +optional
 	NodeSelector *v1.NodeSelector `json:"nodeSelector,omitempty"`
+	// NodeName is a request to schedule this device onto a specific node. If it is non-empty,
+	// the scheduler simply schedules this device onto that node, assuming that it fits
+	// resource requirements.
+	// +optional
+	NodeName string `json:"nodeName,omitempty"`
```
RyanZhaoXB, line 49: I think we need 2 `NodeName`. One is in the field of `Spec` and it means users specify the node for device deployment, which could be allowed to be empty. The other one is in the field of `Status` and it means the device is scheduled to the node which could not be empty.

RyanZhaoXB, line 49: I think we need 2 `NodeName`. One is in the field of `Spec` and it means users specify the node for device deployment, which could be allowed to be empty. The other one is in the field of `Status` and it means the device is scheduled to the node which could not be empty.

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -831,6 +831,22 @@ func deleteDeviceModelAndVisitors(deviceModel *v1alpha2.DeviceModel, deviceProfi
...
 }
 
+// getDeviceNode get the Node bound to this device
+func getDeviceNode(device *v1alpha2.Device) string {
+	nodeID := device.Spec.NodeName
+	if len(nodeID) == 0 {
+		// The useage of device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0] is an issue left over from history.
+		// It not really implement 'NodeSelector' feature just like k8s, this problem mentioned at https://github.com/kubeedge/kubeedge/issues/4763
+
+		// TODO: implement NodeSelector feature. When complete, device.Spec.NodeName field means this device onto what node, and device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values[0] can be deleted at here.
```
RyanZhaoXB, line 841: we can delete the field of `NodeSelector` directly and it will be proposed in https://github.com/kubeedge/kubeedge/issues/4763

RyanZhaoXB:
need to fix the CI.
```
hack/../build/crds is out of date. Please run hack/generate-crds.sh
```

pkg/apis/devices/v1alpha2/device_instance_types.go
cl2017: Yes, NodeName in Status will done at https://github.com/kubeedge/kubeedge/pull/4845, can we finish spec.nodeName here first?

cloud/pkg/devicecontroller/controller/downstream.go
cl2017: Delete it directly may cause compatibility issues, and I prefer to do this after #4763 completed, what do you think?

cl2017:
> need to fix the CI.
> 
> ```
> hack/../build/crds is out of date. Please run hack/generate-crds.sh
> ```

ok, done

cloud/pkg/devicecontroller/controller/downstream.go
fisherxu: +1, must keep compatible with prefer version.

pkg/apis/devices/v1alpha2/device_instance_types.go
fisherxu: Sounds good.

cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -367,12 +367,12 @@ func addDeviceInstanceAndProtocol(device *v1alpha2.Device, deviceProfile *types.
 // deviceAdded creates a device, adds in deviceManagers map, send a message to edge node if node selector is present.
 func (dc *DownstreamController) deviceAdded(device *v1alpha2.Device) {
 	dc.deviceManager.Device.Store(device.Name, device)
-	if len(device.Spec.NodeSelector.NodeSelectorTerms) != 0 && len(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions) != 0 && len(device.Spec.NodeSelector.NodeSelectorTerms[0].MatchExpressions[0].Values) != 0 {
+	if nodeID := getDeviceNode(device); len(nodeID) != 0 {
```
fisherxu, line 370: Let's use `nodeName` instead of `nodeID` in all of this PR,  `nodeID` may be confused with `node UUID`

cl2017: ok, done

cl2017:
> LGTM, and we need to add e2e test for `NodeName` in this PR, and the `NodeSelector` test cases can be refactored later :)

I've modified the device e2e test, using nodeName instead of nodeSelector directly, because it looks more accurate, is that ok?

PR #4807, opened 2023-06-01
fix helm default values

WillardHu:
**What type of PR is this?**

/kind bug

1. Upgrade cloudcore default image tag to "v1.13.0" (controllermanager and iptables-manager no "v1.13.0" tag in dockerhub).
2. If the property `cloudCore.hostNetWork` is a string value,  The expression `{{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}` is always `false` in service_cloudcore.yaml, this means service will not be a NodePort type.
3. Set all enable fields to bool type.

Fixes #

```release-note

```

### DIFF
--- a/manifests/charts/cloudcore/values.yaml
+++ b/manifests/charts/cloudcore/values.yaml
@@ -1,14 +1,14 @@
 # Default values for kubeedge.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
-appVersion: "1.12.0"
+appVersion: "1.13.0"
 
 cloudCore:
   replicaCount: 1
-  hostNetWork: "true"
+  hostNetWork: true
   image:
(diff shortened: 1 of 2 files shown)


zhu733756:
> /assign @zhu733756

Yes

thanks for fixing
/lgtm

zhu733756:
> > /assign @zhu733756
> 
> Yes
> 
> thanks for fixing /lgtm

Pls also check version.yaml

WillardHu:
> > > /assign @zhu733756
> > 
> > 
> > Yes
> > thanks for fixing /lgtm
> 
> Pls also check version.yaml

Done, thanks

fisherxu (commented):
/approve

PR #4813, opened 2023-06-06
[bugfix] add the right to deleting configmaps in rbac of cloudcore to support device-profile

### LINKED ISSUE #4812: fail to delete configmap of device-profile when deleting all the devices
**What happened**:
When all the devices in KubeEdge have been deleted, the configmap of device-profile still existed and it still kept the information of one device. Check the log of  cloudcore and it shows:
cloudcore-log

the configmap of device-profile is deleted correctly

create a device in KubeEdge and then delete it.

configmaps "device-profile-config-xxldev1" is forbidden: User "system:serviceaccount:kubeedge:cloudcore" cannot delete resource "configmaps" in API group "" in the namespace "kubeedge"

It seems that we should config the rbac of cloudcore and give it the right to delete configmap in https://github.com/kubeedge/kubeedge/blob/0fc4c4d125b6a84a6982028bc581d71fcf7f9123/manifests/charts/cloudcore/templates/rbac_cloudcore.yaml#L20

RyanZhaoXB:
add the right to deleting configmaps in rbac of cloudcore to support devicecontroller to delete configmaps of device-profile

**What type of PR is this?**

/kind bug

Fixes #4812

### DIFF
--- a/build/cloud/03-clusterrole.yaml
+++ b/build/cloud/03-clusterrole.yaml
@@ -16,7 +16,7 @@ rules:
   resources: ["nodes", "nodes/status", "pods/status"]
   verbs: ["patch"]
 - apiGroups: [""]
-  resources: ["pods"]
+  resources: ["pods", "configmaps"]
   verbs: ["delete"]
 - apiGroups: ["coordination.k8s.io"]
   resources: ["leases"]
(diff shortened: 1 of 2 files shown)


kevin-wangzefeng:
/lgtm
/approve
Thanks

PR #4814, opened 2023-06-06
 fix mqtt container exited: save mqtt meta in sqlite

### LINKED ISSUE #4783: /bin will be deleted when use cri runtime
**What happened**:
when use keadm join with cri runtime(eg isulad)，/bin will be deleted  after seconds edgcore started. @Shelley-BaoYue  @RyanZhaoXB @vincentgoat 

/bin or other system dir will not be deleted

install isulad, and join with
```shell
keadm join xxx --runtimetype remot --remote-runtime-endpoint unix:///var/run/isulad.sock
```
after some time, /bin will be deleted

we have checked out this issue with bpftrace.
And have some idead in [keadm-mqtt-bug.md](https://github.com/xingleigao/kubeedge/blob/keadm-mqtt-bug/docs/proposals/keadm-mqtt-bug.md)

**Environment**:
- Kubernetes version (use `kubectl version`): any
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): >=v1.9.5

  - OS (e.g. `cat /etc/os-release`): centos 7

  </details>

  - edgecore version (e.g. `edgecore --version`): >=v1.9.5
  - OS (e.g. `cat /etc/os-release`):centos 7

  </details>

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

A temporary solution to solve the exited mqtt container when edgecore start. 

Problem identification: when using cri runtime, mqtt container started by kubelet cri function. When edgecore start, edged(lite-kubelet) list-watch all pods in clusters which do not contains mqtt pod, therefore kubelet will remove mqtt container.

Solution: Save mqtt pod data in sqlite before edged start. 

_**Note： This is only a temporary solution. In the next version, we hope to support static pod to solve this problem.**_

Fixes #4783


common/constants/default.go
```diff
@@ -166,4 +166,6 @@ const (
 
 	EdgeNodeRoleKey   = "node-role.kubernetes.io/edge"
 	EdgeNodeRoleValue = ""
+
+	DeafultMosquittoContianerName = "mqtt-kubeedge"
```
fisherxu, line 170:
```suggestion
	DeafultMosquittoContainerName = "mqtt-kubeedge"
```

fisherxu: Seems haven't used ?

pkg/apis/componentconfig/edgecore/v1alpha2/types.go
```diff
@@ -920,6 +920,9 @@ type EventBus struct {
 	// skip checking other EventBus configs.
 	// default true
 	Enable bool `json:"enable"`
+	// Enable indicates whether start mqtt container when edgecore start
+	// default true
+	EnableMqttContainer bool `json:"enableMqttContainer"`
```
fisherxu, line 925: As a workaround solution, it's not needed to add new api field, can we assume that if evetbus module is enabled and mode is `external`, then we  need to save the `mqtt container`, and start it.

common/constants/default.go
Shelley-BaoYue: > Seems haven't used ?

message resource and name used it, for example:
image

pkg/apis/componentconfig/edgecore/v1alpha2/types.go
Shelley-BaoYue: > As a workaround solution, it's not needed to add new api field, can we assume that if evetbus module is enabled and mode is `external`, then we need to save the `mqtt container`, and start it.

I hope the --with-mqtt config in keadm can take effect, maybe I can set mode=internal when --with-mqtt=false?

fisherxu: There're two scenarios users deploy edgecore:

1. Users don't use keadm to deploy edgecore, when they enable `eventbus`, they will deploy the mqtt by self
2. Users use keadm to deploy edgecore, when they set --with-mqtt=true, we need to deploy mqtt, or no need to deploy

If we can set a environment value in edgecore, like `WITH-MQTT`, default set as false, edgecore will not deploy mqtt, if uses use keadm to deploy, and set --with-mqtt=true, set the `WITH-MQTT` to true, and deploy mqtt with edgecore.

common/constants/default.go
fisherxu: OK, so please just the typo

pkg/apis/componentconfig/edgecore/v1alpha2/types.go
Shelley-BaoYue: > There're two scenarios users deploy edgecore:
> 
> 1. Users don't use keadm to deploy edgecore, when they enable `eventbus`, they will deploy the mqtt by self
> 2. Users use keadm to deploy edgecore, when they set --with-mqtt=true, we need to deploy mqtt, or no need to deploy
> 
> If we can set a environment value in edgecore, like `WITH-MQTT`, default set as false, edgecore will not deploy mqtt, if uses use keadm to deploy, and set --with-mqtt=true, set the `WITH-MQTT` to true, and deploy mqtt with edgecore.

Fixed. PTAL  @fisherxu  

image

fisherxu (commented):
/lgtm
/approve

PR #4839, opened 2023-07-05
fix the container of mqtt can't be deleted when exec keadm reset

Onion-of-dreamed:
**What type of PR is this?**
/kind bug

the container of mqtt can't be deleted when exec keadm reset, It turned out to be the way to filter containers using the label "ancestor=eclipse-mosquitto:1.6.15", but i found it can't find the currect container of mqtt, use the container name is a better way to handle this action

```release-note
None
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/image.go
+++ b/keadm/cmd/keadm/app/cmd/util/image.go
@@ -154,7 +154,7 @@ func (runtime *DockerRuntime) RemoveMQTT() error {
 		All: true,
 	}
 	options.Filters = filters.NewArgs()
-	options.Filters.Add("ancestor", constants.DefaultMosquittoImage)
+	options.Filters.Add("name", image.EdgeMQTT)
 
 	mqttContainers, err := runtime.Client.ContainerList(runtime.ctx, options)
 	if err != nil {


keadm/cmd/keadm/app/cmd/util/image.go
```diff
@@ -152,7 +152,7 @@ func (runtime *DockerRuntime) RemoveMQTT() error {
 		All: true,
 	}
 	options.Filters = filters.NewArgs()
-	options.Filters.Add("ancestor", constants.DefaultMosquittoImage)
+	options.Filters.Add("name", image.EdgeMQTT)
```
fisherxu, line 155: Where is the key `name` and value `image.EdgeMQTT` set? :)

Onion-of-dreamed: i don't find the ancetor be setted any where, so i use the container name to handle it, the container name is setted when the mqtt container run, be like https://github.com/kubeedge/kubeedge/blob/d0d3c8ac461b2d1882907f4743b565e052789494/keadm/cmd/keadm/app/cmd/util/image.go#L143 and https://github.com/kubeedge/kubeedge/blob/d0d3c8ac461b2d1882907f4743b565e052789494/keadm/cmd/keadm/app/cmd/util/image.go#L377

PR #4841, opened 2023-07-07
add runtime validation in keadm

Shelley-BaoYue:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**

Add runtime validation in keadm. Return a more explicit error message.

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/join.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join.go
@@ -253,6 +253,9 @@ func createEdgeConfigFiles(opt *common.JoinOptions) error {
 	if v.Major <= 1 && v.Minor < 12 {
 		return createV1alpha1EdgeConfigFiles(opt)
 	}
+	if v.Major >= 1 && v.Minor >= 14 && opt.RuntimeType != constants.DefaultRuntimeType {
+		return fmt.Errorf("since KubeEdge v1.14, runtime type only supports `remote`")
+	}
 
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -253,6 +253,9 @@ func createEdgeConfigFiles(opt *common.JoinOptions) error {
 	if v.Major <= 1 && v.Minor < 12 {
 		return createV1alpha1EdgeConfigFiles(opt)
 	}
+	if v.Major >= 1 && v.Minor >= 14 && opt.RuntimeType != constants.DefaultRuntimeType {
+		return fmt.Errorf("since KubeEdge v1.14, runtime type only supports `remote`")
```
fisherxu, line 257: If users want to use Docker, the runtime type still set to `remote`, then edged connect to `cri-dockerd`?

Shelley-BaoYue: > If users want to use Docker, the runtime type still set to `remote`, then edged connect to `cri-dockerd`?

set runtimeType to remote and set `remoteRuntimeEndpoint` to `var/run/cri-dockerd.sock`, edged will connect to `cri-dockerd`.  See more detail in #4843

fisherxu: Okey :)

fisherxu (commented):
/lgtm
/approve

PR #4864, opened 2023-07-21
perf on helm chart of cloudcore

### LINKED ISSUE #4862: cloudcore deployed by helm chart fails to run because of conflict bind addresss with calico-node
**What happened**:
I am deploying cloudcore with helm chart.It fails to run because of hostNetwork mode is enabled and 9091 port is already used by calico-node.
```
root@node3:~# ss -tlnp|grep 9091
LISTEN 0      4096               *:9091             *:*    users:(("calico-node",pid=6497,fd=13))
```

As is shown in the code below, monitoring server is fixed to 127.0.0.1:9091
https://github.com/kubeedge/kubeedge/blob/d0d3c8ac461b2d1882907f4743b565e052789494/pkg/apis/componentconfig/cloudcore/v1alpha1/default.go#L40

one click deploy using helm chart without to much hacking. 

deploy cloudcore with helm chart from master branch and my modification of values.yaml is as follow:
```yaml
cloudcore:
  cloudCore:
    modules:
      cloudHub:
        advertiseAddress:
          - "127.0.0.1"
        dnsNames:
          - "xxxxx.com"
  iptablesManager:
    mode: "external"
```
and the kubernetes cluster is using calico as cni plugin
I wonder if it is necessary to use hostNetwork and priviledged mode when deploying cloudcore?Can I expose it with its service?
**Environment**:
- Kubernetes version (use `kubectl version`): v1.24.10+rke2r1
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): 1.13

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

lwabish:
**What type of PR is this?**
/kind feature

1. LoadBalancer type service is not supported refers to the design of chart now.
2. Service nodePort is mixed with hostNetwork value by some custom logic, which is not flexible , I think.

Fixes #4862

```release-note

```

### DIFF
--- a/manifests/charts/cloudcore/templates/service_cloudcore.yaml
+++ b/manifests/charts/cloudcore/templates/service_cloudcore.yaml
@@ -10,39 +10,35 @@ metadata:
   {{- end }}
   name: cloudcore
 spec:
-  {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
   type: {{ .Values.cloudCore.service.type }}
-  {{- else }}
-  type: ClusterIP
-  {{- end }}
   ports:
   - port: 10000
     targetPort: 10000
-    {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
+    {{- if eq .Values.cloudCore.service.type "NodePort" }}
     nodePort: {{ .Values.cloudCore.service.cloudhubNodePort }}
     {{- end }}
     name: cloudhub
   - port: 10001
     targetPort: 10001
-    {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
+    {{- if eq .Values.cloudCore.service.type "NodePort" }}
     nodePort: {{ .Values.cloudCore.service.cloudhubQuicNodePort }}
     {{- end }}
     name: cloudhub-quic
   - port: 10002
     targetPort: 10002
-    {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
+    {{- if eq .Values.cloudCore.service.type "NodePort" }}
     nodePort: {{ .Values.cloudCore.service.cloudhubHttpsNodePort }}
     {{- end }}
     name: cloudhub-https
   - port: 10003
     targetPort: 10003
-    {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
+    {{- if eq .Values.cloudCore.service.type "NodePort" }}
     nodePort: {{ .Values.cloudCore.service.cloudstreamNodePort }}
     {{- end }}
     name: cloudstream
   - port: 10004
     targetPort: 10004
-    {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
+    {{- if eq .Values.cloudCore.service.type "NodePort" }}
     nodePort: {{ .Values.cloudCore.service.tunnelNodePort }}
     {{- end }}
     name: tunnelport


zhu733756:
> /cc @zhu733756 ptal :)

Sorry for the late comments. Will response quickly

manifests/charts/cloudcore/values.yaml
```diff
@@ -1,8 +1,3 @@
-# Default values for kubeedge.
-# This is a YAML-formatted file.
-# Declare variables to be passed into your templates.
-appVersion: "1.13.0"
-
 cloudCore:
   replicaCount: 1
   hostNetWork: true
```
zhu733756, line 3: if this flag still keep, we still need to a choice?  I remember hostNotWork has a different working principle. Pls help to check it, thanks!

manifests/charts/cloudcore/templates/service_cloudcore.yaml
```diff
@@ -10,39 +10,35 @@ metadata:
...
-  {{- else }}
-  type: ClusterIP
-  {{- end }}
   ports:
   - port: 10000
     targetPort: 10000
-    {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
+    {{- if eq .Values.cloudCore.service.type "NodePort" }}
     nodePort: {{ .Values.cloudCore.service.cloudhubNodePort }}
     {{- end }}
```
zhu733756, line 19: Oh, the hostNetWork mode will not use nodePort field. Am I right?

zhu733756: cc  @lwabish

lwabish: Yes, you are right.But the point is, is it better to expose these fields directly to chart users?
For example,nodePort field is needed by LoadBalancer type service. 
As we can see now, the logics here in our chart can't cover this situation, which is also the motivation of this PR.
So in my opinion, it would be much more flexible to minimize our logic in charts unless necessary.

manifests/charts/cloudcore/values.yaml
lwabish: I am not sure why hostnetwork is here, maybe it's for some special cases?
But I think it's okay to keep it here and make it possible for chart users to customize the values easily.
Of course, It would be better to add some comments here to avoid confusion, as I had before when seeing this.

manifests/charts/cloudcore/templates/service_cloudcore.yaml
zhu733756: have you tried hostNetWork: false to satisfy your senses？

manifests/charts/cloudcore/values.yaml
zhu733756: I think the older version is using the hostNetwork mode  on the bare meta vm.

manifests/charts/cloudcore/templates/service_cloudcore.yaml
lwabish: Just tried using LoadBalancer type service to expose cloudcore tcp and https ports to get an edgecore joined successfully.
According to the reply of issue #4862 ,hostnetwork and priviledged is not necessary for cloudcore. I dont't know why they are exposed, maybe for some special use cases.

zhu733756: > Just tried using LoadBalancer type service to expose cloudcore tcp and https ports to get an edgecore joined successfully. According to the reply of issue #4862 ,hostnetwork and priviledged is not necessary for cloudcore. I dont't know why they are exposed, maybe for some special use cases.

CC @fisherxu maybe NodePort as the default choice is more suitable?

WillardHu: When stream is enabled (logs/exec) and IptablesManager mode is 'internal', The Cloudcore hostNetwork should be true.

lwabish: Got it! I think the bottleneck is the current logic doesn't support Loadbalancer type service.

PR #4929, opened 2023-08-11
Bump k8s 1.26.7

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

Update Kubernetes dependency version to v1.26.7


fisherxu (commented):
/approve
The fuzzing test is not required, and once the PR merged, I will set the K8s verion in github action.

fisherxu:
/assign @WillardHu 
For final review :)

WillardHu:
/lgtm

The historical code has some functions in the pkg package that use `exit` to handle error. In the future, we can see if we can use `panic` instead. The `exit` operation is non-interruptible and cannot be unit tested.

fisherxu:
> The historical code has some functions in the pkg package that use exit to handle error. In the future, we can see if we can use panic instead. The exit operation is non-interruptible and cannot be unit tested.

Can you open an issue to track it? @WillardHu

WillardHu:
> > The historical code has some functions in the pkg package that use exit to handle error. In the future, we can see if we can use panic instead. The exit operation is non-interruptible and cannot be unit tested.
> 
> Can you open an issue to track it? @WillardHu

OK 
https://github.com/kubeedge/kubeedge/issues/5012

PR #4950, opened 2023-08-22
revert build tool ubuntu version to 18.04

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

revert build tool ubuntu version to 18.04. If using Ubuntu 22.04 to build, we may encounter similar errors below when running on lower versions of Ubuntu machines.

image

PR #4953, opened 2023-08-22
update kubeedge build tool image 1.19.12-ke2

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

update kubeedge build tool image 1.19.12-ke2 built based on ubuntu 18.04, resolve the problem described in #4950


fisherxu (commented):
/lgtm
/approve

PR #4962, opened 2023-08-29
Support log level in kubemark

q77190858:
What type of PR is this?
/kind bug

Support log level in kubemark
Which issue(s) this PR fixes:
NONE

NONE

NONE

### DIFF
--- a/edge/pkg/devicetwin/process.go
+++ b/edge/pkg/devicetwin/process.go
@@ -52,8 +52,10 @@ func (dt *DeviceTwin) distributeMsg(m interface{}) error {
 		klog.Infof("Send msg to the %s module in twin", dtcommon.CommModule)
 		confirmMsg := dttype.DTMessage{Msg: model.NewMessage(message.Msg.GetParentID()), Action: dtcommon.Confirm}
 		if err := dt.DTContexts.CommTo(dtcommon.CommModule, &confirmMsg); err != nil {
+			klog.Errorf("fail to send msg %s to CommModule with err: %+v", confirmMsg.Msg.Header.ID, err)
 			return err
 		}
+		return nil
 	}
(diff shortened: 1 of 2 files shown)


build/edgemark/hollow_edge_node_template.yaml
```diff
@@ -22,8 +22,6 @@ spec:
             - --name=$(NODE_NAME)
             - --http-server=https://{{server}}:10002
             - --websocket-server={{server}}:10000
-            - --alsologtostderr
-            - --v=2
```
fisherxu, line 26: I think these two flag can use to debug, can we also explore how to enable them? Which is exised in other components :) @q77190858

chendave: +1 

> remove these 2 args --alsologtostderr and --v=2, because edgemark now do not support , these args make edgemark exit error

maybe enable them instead.

q77190858:
enable these 2 args --alsologtostderr and --v=2 for edgemark

fisherxu:
> enable these 2 args --alsologtostderr and --v=2 for edgemark

@q77190858  Could you please show the output of cmd `edgemark -h` here? :) Which should include these 2 args now.

edge/cmd/edgemark/hollow_edgecore.go
```diff
@@ -96,6 +97,7 @@ func newHollowEdgeNodeCommand() *cobra.Command {
 
 	fs := cmd.Flags()
 	fs.AddGoFlagSet(flag.CommandLine) // for flags like --docker-only
+	globalflag.AddGolbalFlags(fs, cmd.Name())
```
fisherxu, line 100:
```suggestion
	globalflag.AddGlobalFlags(fs, cmd.Name())
```

PR #5034, opened 2023-09-19
Add admission to cloud helm charts

### LINKED ISSUE #4058: Integrate more cloud components like `admission`, `csidriver`, etc. to `keadm init` container installation
Integrate more cloud components like `admission`, `csidriver`, etc. to `keadm init` container installation
**Why is this needed**:

WillardHu:
**What type of PR is this?**

/kind feature

Supports keadm install the admission component in cloud. The admission component is disabled by default, you can use the `keadm init` option `--set admission.enable=true` to enable it.

Fixes #4058 

```release-note

```

### DIFF
--- a/manifests/charts/cloudcore/templates/deployment_admission.yaml
+++ b/manifests/charts/cloudcore/templates/deployment_admission.yaml
@@ -1,0 +1,72 @@
+{{- if .Values.admission.enable }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  {{- with .Values.admission.annotations }}
+  anntations: {{- toYaml . | nindent 4 }}
+  {{- end }}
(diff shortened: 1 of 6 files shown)


fisherxu (commented):
Not sure whether keadm will install admission after add it to helm charts? @WillardHu

WillardHu:
> Not sure whether keadm will install admission after adding it to helm charts? @WillardHu

Some of the most important checks are in the admission.  Like example NodeUpgradeJob one of the verification rules: "nodeNames and labelSelector cannot be both empty. That means that we must specify at least one valid node. And users cannot set the two fields both either".

WillardHu:
PTAL  @fisherxu  @Shelley-BaoYue

fisherxu (commented):
/lgtm
/approve

PR #5037, opened 2023-09-20
fix: the container of mqtt can't be deleted when exec keadm reset

Onion-of-dreamed:
**What type of PR is this?**
/kind bug

the container of mqtt can't be deleted when exec keadm reset, It turned out to be the way to filter containers using the label "ancestor=eclipse-mosquitto:1.6.15", but i found it can't find the currect container of mqtt, use the container name is a better way to handle this action

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/image.go
+++ b/keadm/cmd/keadm/app/cmd/util/image.go
@@ -154,7 +154,7 @@ func (runtime *DockerRuntime) RemoveMQTT() error {
 		All: true,
 	}
 	options.Filters = filters.NewArgs()
-	options.Filters.Add("ancestor", constants.DefaultMosquittoImage)
+	options.Filters.Add("name", image.EdgeMQTT)
 
 	mqttContainers, err := runtime.Client.ContainerList(runtime.ctx, options)
 	if err != nil {


keadm/cmd/keadm/app/cmd/util/image.go
```diff
@@ -154,7 +154,7 @@ func (runtime *DockerRuntime) RemoveMQTT() error {
 		All: true,
 	}
 	options.Filters = filters.NewArgs()
-	options.Filters.Add("ancestor", constants.DefaultMosquittoImage)
```
fisherxu, line 157: Thanks for the fixing, and could I please ask what's the meaning of `ancestor` here?

fisherxu (commented):
/approve

fisherxu (commented):
cc @WillardHu @ZhengXinwei-F @Shelley-BaoYue

keadm/cmd/keadm/app/cmd/util/image.go
WillardHu: This condition `ancestor` uses image to filter containers.  But we use a constant of image when we reset node.  When we use a custom image repository for join node, it causes changes to the mqtt image. This will result in the mqtt container to not be deleted.

Onion-of-dreamed: yeah, use the special image name only can find special contained,  when we use anthoer image to replace the mosqutto,  the condition `ancestor` can't be work normally,  At the same time, the name of mqtt container is unique, and whatever image you used, it always could find the mqtt container and remove it.

Onion-of-dreamed:
@Shelley-BaoYue please review the PR, thanks

PR #5047, opened 2023-09-22
Fix upgrade time layout and lost time value issue

WillardHu:
**What type of PR is this?**

/kind bug

1. The upgrade time initial value is returned by the `String()` function of the time struct, like `2023-09-22 16:07:13.505954202 +0800 CST m=+19903.570598396`, this time format is difficult to be handle, the k8s's various time format layouts are `2006-01-02T15:04:05Z`, so we need to use that format layout as well;
2. When the upgrade status is updated again after the first update, the upgrade time is lost. The `upgradeTime` of parameter `status` has no value in `UpdateNodeUpgradeJobStatus(..)` function, so we need set the generated upgrade time of old upgrade status to the new value. If generated upgrade time is empty, we initialize it with the current time.

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/nodeupgradejobcontroller/controller/downstream.go
+++ b/cloud/pkg/nodeupgradejobcontroller/controller/downstream.go
@@ -210,7 +210,7 @@ func (dc *DownstreamController) processUpgrade(node string, upgrade *v1alpha1.No
 		State:    v1alpha1.Upgrading,
 		History: v1alpha1.History{
 			HistoryID:   upgradeReq.HistoryID,
-			UpgradeTime: time.Now().String(),
+			UpgradeTime: time.Now().Format(ISO8601UTC),
 		},
 	}
 	err = patchNodeUpgradeJobStatus(dc.crdClient, upgrade, status)
--- a/cloud/pkg/nodeupgradejobcontroller/controller/util.go
+++ b/cloud/pkg/nodeupgradejobcontroller/controller/util.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/distribution/distribution/v3/reference"
 	metav1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


fisherxu (commented):
/assign @ZhengXinwei-F

cloud/pkg/nodeupgradejobcontroller/controller/downstream.go
```diff
@@ -210,7 +210,7 @@ func (dc *DownstreamController) processUpgrade(node string, upgrade *v1alpha1.No
 		State:    v1alpha1.Upgrading,
 		History: v1alpha1.History{
 			HistoryID:   upgradeReq.HistoryID,
-			UpgradeTime: time.Now().String(),
+			UpgradeTime: time.Now().Format(ISO8601UTC),
```
fisherxu, line 213: How about using `RFC3339`, which is a golang const format? @WillardHu

WillardHu:
```golang
package time

const (
    ....
    RFC3339     = "2006-01-02T15:04:05Z07:00"
    ....
)
```

RFC3339 has a numeric time zone offsets(07:00), can we use `metav1.Time.MarshalJSON` to format `updateTime`?
```golang
tt := metav1.Time{Time: time.Now()}
bff, err := tt.MarshalJSON()
if err != nil {
    // handler error
}
string(bff)  // 2023-10-12T03:57:14
```

fisherxu: Seems there's have no difference woth `ISO8601UTC` above, I think it's also fine to use `ISO8601UTC` :)

fisherxu (commented):
/lgtm
/approve

fisherxu:
@WillardHu Please looks at whether we need to cherry-pick to release branchs.

WillardHu:
> @WillardHu Please looks at whether we need to cherry-pick to release branchs.

We are now temporarily formatting the `upgradeTime` in the business layer, and this fix can be released in 1.15 stable.

fisherxu:
Okey, please cherry-pick to release-1.15, release-1.14, release-1.13, release-1.12, thanks :) @WillardHu

PR #5058, opened 2023-10-08
Create a new, empty flag set for keadm

Shelley-BaoYue:
**What type of PR is this?**

/kind cleanup

Create a new, empty flag set for keadm, to prevent keadm reusing the following global flag in kuberntes:

image

### DIFF
--- a/keadm/cmd/keadm/app/keadm.go
+++ b/keadm/cmd/keadm/app/keadm.go
@@ -26,7 +26,8 @@ import (
 
 // Run executes the keadm command
 func Run() error {
-	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+	flagSet := pflag.NewFlagSet("keadm", pflag.ExitOnError)
+	flagSet.AddGoFlagSet(flag.CommandLine)
 
 	cmd := cmd.NewKubeedgeCommand()
 	return cmd.Execute()


fisherxu:
Not sure why keadm will print these flags.. `kubeadm` also init like this `pflag.CommandLine.AddGoFlagSet(flag.CommandLine)`  ref: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/kubeadm.go#L35

cc @WillardHu @ZhengXinwei-F

Shelley-BaoYue:
> Not sure why keadm will print these flags.. `kubeadm` also init like this `pflag.CommandLine.AddGoFlagSet(flag.CommandLine)` ref: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/kubeadm.go#L35
> 
> cc @WillardHu @ZhengXinwei-F

kubeadm uses `klog.InitFlags`, while klog will new a flag set. And kubeadm will reuse global flags of klog, like:
 
image

WillardHu:
There are no global flags yet, right? 

The `pflag.CommandLine` is usually called only in the `cmd` package, and we will not import the `cmd` packages of the dependent projects. But I think this change is okay, we don't use the `pflag.CommandLine` elsewhere yet in the keadm.

fisherxu:
> There are no global flags yet, right?

Yes, now there has no global flags

PR #5068, opened 2023-10-12
Add opengemini api

wbc6080:
**What type of PR is this?**
/kind feature

add opengemini database api in mapper-framework and kubeedge

Fixes #

```release-note

```


build/crds/devices/devices_v1beta1_device.yaml
```diff
@@ -633,6 +633,40 @@ spec:
                                       type: object
                                   type: object
                               type: object
+                            openGemini:
```
WillardHu, line 636: These extensions of `pushMethod` make the device instance CRD very large.

Can we abstract the configuration, for example, into a yaml string?
```yaml
pushMethod:
  type: influxdb2 / openGemini / http / mqtt / ...
  config: |
    hostName: ""
    port: 0
    requestPath: ""
    timeout: 0
```
Or reference a configuration CRD:
```yaml
# device CR
pushMethod:
  crd: http.devicepushmethod.kubeedge.io
  crname: xxx

# http device pushmethod CR
spec:
  hostName: ""
  port: 0
  requestPath: ""
  timeout: 0
```

shilinlee (approved):

build/crds/devices/devices_v1beta1_device.yaml
wbc6080: This is indeed a problem, and I will think about how to make it more lightweight

PR #5070, opened 2023-10-12
mapper framework fix

cl2017:
**What type of PR is this?**
/kind bug

mapper framework fix
- We shloud move `pkg/util/grpcclient` to `pkg/grpcclient`.
- `RegisterMapper` function should use `pkg/grpcclient/config`  instead of function argument

Fixes #

```release-note
NONE
```


wbc6080:
This is a good modification. Indeed, we have stored config in a global variable when initializing grpcclient and do not need to write it in function argument. But I find that you didn't fix the import path in device/devicetwin.go. Please fix it, thank you

cl2017:
> This is a good modification. Indeed, we have stored config in a global variable when initializing grpcclient and do not need to write it in function argument. But I find that you didn't fix the import path in device/devicetwin.go. Please fix it, thank you

Yes, I'll open another PR to rewrite `pkg/config`, in this PR, we fix `pkg/grpcclient` fisrt.
I have already fix `devicetwin.go`, thanks.

fisherxu (commented):
/approve
/assign @WillardHu @wbc6080

PR #5071, opened 2023-10-12
fix ci disk space lack issue

Shelley-BaoYue:
**What type of PR is this?**

/kind failing-test

fix ci disk space lack issue. cleanup dacker builder cache and dangling images(untagged images).

### DIFF
--- a/hack/make-rules/image.sh
+++ b/hack/make-rules/image.sh
@@ -40,7 +40,7 @@ ALL_IMAGES_AND_TARGETS=(
 
 GO_LDFLAGS="$(${KUBEEDGE_ROOT}/hack/make-rules/version.sh)"
 IMAGE_TAG=$(git describe --tags)
-
+DOCKER_BUILD_AND_SYSTEM_PRUNE=${DOCKER_BUILD_AND_SYSTEM_PRUNE:-"false"}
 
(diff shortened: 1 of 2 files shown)


hack/make-rules/image.sh
```diff
@@ -90,6 +90,9 @@ function build_images() {
     set -x
     docker build --build-arg GO_LDFLAGS="${GO_LDFLAGS}" -t kubeedge/${IMAGE_NAME}:${IMAGE_TAG} -f ${DOCKERFILE_PATH} .
     set +x
+
+    docker builder prune -f
+    docker system prune -f
```
fisherxu, line 95:
```suggestion
    if [[ "${CLEAN_BUILD_IMAGE_CACHE}" = "true" ]]; then
    docker builder prune -f
    docker system prune -f
fi
    
```

At L43, CLEAN_BUILD_IMAGE_CACHE=${CLEAN_BUILD_IMAGE_CACHE:-"false"}

How about add a flag to control whether to clean the cache, because in develpers' host, it's no need and should not delete other images. We only need to delete in ci host now   cc @WillardHu @Shelley-BaoYue

fisherxu (commented):
/lgtm
/approve

Issue #5024, opened 2023-09-18
Change astaxie/beego to beego/beego


fisherxu:
Now the go framework `beego` has been migrated to `github.com/beego/beego` from `github.com/astaxie/beego`,  and we still use the  `github.com/astaxie/beego`, let's change to `beego/beego`!

MeenuyD:
Hello, @fisherxu should I need change all files?

fisherxu:
And let's use `v1.12.2` of beego instead :)

yeqiugt:
/assign 

Does it have to be version v1.12.2?
The latest version of v1 is v1.12.12, and the v2 version has also been released.

fisherxu:
> The latest version of v1 is v1.12.12, and the v2 version has also been released.

fisherxu:
> The latest version of v1 is v1.12.12, and the v2 version has also been released.

OK, v1.12.12 shoud be fine, I think I should mean 1.12.12 above, not sure what's the difference of v2 and v1, and the compatibility of them :) @yeqiugt

PR #5089, opened 2023-10-16
 use golang.org/x/text/cases instead of strings.Title(Deprecated)

yeqiugt:
**What type of PR is this?**
/kind cleanup

use golang.org/x/text/cases instead of strings.Title(Deprecated) 

This is my first commit to kubeedge. Please give me your advice.

no

```release-note

```


pkg/metaserver/util/util.go
```diff
@@ -2,6 +2,8 @@ package util
 
 import (
 	"context"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	"strings"
```
fisherxu, line 7:
```
import (
	"context"
	"strings"

	"golang.org/x/text/cases"
	"golang.org/x/text/language"	
        "k8s.io/apimachinery/pkg/api/meta"
	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
	"k8s.io/apimachinery/pkg/fields"
	"k8s.io/apimachinery/pkg/labels"
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apiserver/pkg/storage"
	"k8s.io/client-go/kubernetes/scheme"
	"k8s.io/klog/v2"

	beehiveModel "github.com/kubeedge/beehive/pkg/core/model"
)
```
Sorts the imports as above, the `golang.org` belongs to the thirdpart import.

fisherxu (commented):
Thanks for the fixing, and please run the `make verify` command to update the vendor in your host, thanks @yeqiugt

pkg/metaserver/util/util.go
yeqiugt: I sort the imports. But when I run `make verify` it fails, why and what should I do to fix it?

fisherxu: Then just run `git status` to check and submit the generated files

vendor/github.com/kubeedge/viaduct
```diff
@@ -1 +0,0 @@
-../../../staging/src/github.com/kubeedge/viaduct/
```
fisherxu, line 1: And you should not delete this file and other two files for `beehive` and `mapper-framework`, they are link to staging dirs.

fisherxu (commented):
/approve
/cc @WillardHu @Shelley-BaoYue

fisherxu:
@yeqiugt Would you like to help fix this issue? https://github.com/kubeedge/kubeedge/issues/5024

PR #5100, opened 2023-10-19
Fixed the mqtt would not start when using custom registry

WillardHu:
**What type of PR is this?**

/kind bug

If the image of the `mqtt` uses a custom registry, the container of the `mqtt` will not start properly.

```bash
keadm join --cloudcore-ipport=<"THE-EXPOSED-IP":10000> --token=<token>  --image-repository=<custom registry>/kubeedg

# The edgecore will have an error
start failed in pod mqtt-kubeedge_default(93205b92-a7fe-46a9-b899-c4eaa43ef339): ErrImageNeverPull: Container image "eclipse-mosquitto:1.6.15" is not present with pull policy of Never
10月 17 17:06:14 zdk1 edgecore[23420]: E1017 17:06:14.588945   23420 pod_workers.go:962] "Error syncing pod, skipping" err="failed to \"StartContainer\" for \"mqtt\" with ErrImageNeverPull: \"Container image \\\"eclipse-mosquitto:1.6.15\\\" is not present with pull policy of Never\"" pod="default/mqtt-kubeedge" podUID=93205b92-a7fe-46a9-b899-c4eaa43ef339
```

This is because the `keadm` only pull the custom image `<custom registry>/kubeedg/eclipse-mosquitto:1.6.15`, not the default image `eclipse-mosquitto:1.6.15`, and the `kubelet` image pull policy is used `Never` when the image pull policy is empty.

This pr is fixed:
- Set the default image pull policy to `PullIfNotPresent`;
- Align the image of the `mqtt` with the image pulled by the `keadm`

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -121,12 +121,20 @@ func (e *edged) Enable() bool {
 func (e *edged) Start() {
 	klog.Info("Starting edged...")
 
-	// edged saves the data of mqtt container in sqlite3 and starts it. This is a temporary workaround and will be modified in v1.15.
-	withMqtt, err := strconv.ParseBool(os.Getenv(constants.DeployMqttContainerEnv))
-	if err == nil && withMqtt {
-		err := dao.SaveMQTTMeta(e.nodeName)
-		if err != nil {
-			klog.ErrorS(err, "Start mqtt container failed")
+	// edged saves the data of mqtt container in sqlite3 and starts it.
+	// This is a temporary workaround and will be modified in v1.15.
+	if deployMqtt := os.Getenv(constants.DeployMqttContainerEnv); deployMqtt != "" {
+		if withMqtt, err := strconv.ParseBool(deployMqtt); err == nil && withMqtt {
+			image := constants.DefaultMosquittoImage
+			if customimg := os.Getenv(constants.DeployMqttContainerImageEnv); customimg != "" {
+				image = customimg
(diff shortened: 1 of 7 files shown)


edge/pkg/edged/edged.go
```diff
@@ -121,12 +121,20 @@ func (e *edged) Enable() bool {
 func (e *edged) Start() {
 	klog.Info("Starting edged...")
 
-	// edged saves the data of mqtt container in sqlite3 and starts it. This is a temporary workaround and will be modified in v1.15.
-	withMqtt, err := strconv.ParseBool(os.Getenv(constants.DeployMqttContainerEnv))
-	if err == nil && withMqtt {
-		err := dao.SaveMQTTMeta(e.nodeName)
-		if err != nil {
-			klog.ErrorS(err, "Start mqtt container failed")
+	// edged saves the data of mqtt container in sqlite3 and starts it.
```
ZhengXinwei-F, line 124: The code nesting here is a little deep and simply serves as interim logic. I think it can be extracted into a function for later replacement. It also improves the readability of the code. : )

WillardHu: > The code nesting here is a little deep and simply serves as interim logic. I think it can be extracted into a function for later replacement. It also improves the readability of the code. : ) P.S. The env judgment can be extracted as a function and added to the original if condition.

Thanks, fixed in #5101

PR #5107, opened 2023-10-23
Fix error logs for nodegroup

lishaokai1995:
**What type of PR is this?**
/kind bug

When a node has already been added to the nodegroup and the node is added to another nodegroup again, we can see the error message in CR, but the error message is not correct. The purpose of this PR is to fix it

Fixes #

```release-note

```


fisherxu (commented):
/approve

PR #5120, opened 2023-10-27
modify parameters for ginkgo v2

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

/kind test

- add parameter 'timeout' for ginkgo v2. ginkgo v2 changes timeout default to 1h, while the conformance test takes 1h-1.5h. The ginkgo v1 default timeout value is 24h. 
- add two skip cases, for we don't support them now.
- --noColor has been changed to --no-color in ginkgo v2
- CurrentGinkgoTestDescription() is deprecated in Ginkgo V2.  Use CurrentSpecReport() instead.


fisherxu (commented):
/lgtm
/approve

PR #5140, opened 2023-11-06
Resolve that users do not need to define the status module in device. yaml.

luomengY:
**What type of PR is this?**
/kind bug

The current mapper requires users to define status and twist data in device, but this part of data should be obtained from the device based on the mapper, and should not be defined by the user. Therefore, this PR mainly solves the problem of users no longer defining status in device, and the mapper reconstructs the data structure of twist based on properties to report status.

Fixes #

@wbc6080 @cl2017 
Users no longer need to define the status module in device. yaml.

```release-note

```

### DIFF
--- a/cloud/pkg/devicecontroller/controller/upstream.go
+++ b/cloud/pkg/devicecontroller/controller/upstream.go
@@ -135,8 +135,9 @@ func (uc *UpstreamController) updateDeviceStatus() {
 			}
 			deviceStatus := &DeviceStatus{Status: cacheDevice.Status}
 			for twinName, twin := range msgTwin.Twin {
-				for i, cacheTwin := range deviceStatus.Status.Twins {
-					if twinName == cacheTwin.PropertyName && twin.Actual != nil && twin.Actual.Value != nil {
+				deviceTwin := findTwinByName(twinName, &deviceStatus.Status.Twins)
+				if deviceTwin != nil {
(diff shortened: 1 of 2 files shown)


fisherxu (commented):
Sorry for the delay, thanks for the fixing!
/approve
/assign @cl2017 @WillardHu @wbc6080

cloud/pkg/devicecontroller/controller/upstream.go
```diff
@@ -136,17 +136,32 @@ func (uc *UpstreamController) updateDeviceStatus() {
 			deviceStatus := &DeviceStatus{Status: cacheDevice.Status}
 			for twinName, twin := range msgTwin.Twin {
```
WillardHu, line 137: The code indentation is too deep, you can abstract a function to returns a reference to a device status twin property, and then processing.
```golang
for twinName, twin := range msgTwin.Twin {
    statusTwin := findStatusTwinProperty(twinName, deviceStatus.Status.Twins)
    if twin.Actual != nil && twin.Actual.Value != nil {
        // set statusTwin ...
    }
    if twin.Expected != nil && twin.Expected.Value != nil {
        // set statusTwin ...
    }
}

func findStatusTwinProperty(...) *Twin {
    // todo ...
}
```

luomengY: ok,I have modified.

PR #5161, opened 2023-11-13
 Refactor the code for skip environment check

ZhengXinwei-F:
**What type of PR is this?**

Refactor the code for skip environment check.

Current code that skips environment checks is not elegant. Because environment checking is enabled by default but the default value in golang is false. Changing this parameter to DISABLE_ENVIRONMENT_CHECK is the best answer. However, the change on the user-facing is too large. Only the code is optimized in this case.

Fixes #

Should we consider changing CHECK_EDGECORE_ENVIRONMENT to DISABLE_ENVIRONMENT_CHECK.

```
none
```

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -101,17 +101,11 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 			// To help debugging, immediately log version
 			klog.Infof("Version: %+v", version.Get())
 
-			// Check the running environment by default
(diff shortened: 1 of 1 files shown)


Shelley-BaoYue:
fix the DCO

ZhengXinwei-F:
> fix the DCO

done

fisherxu (commented):
/approve

fisherxu (commented):

fisherxu (commented):
/lgtm

PR #5164, opened 2023-11-14
Optimize ModuleInfo registration and startup process

ZhengXinwei-F:
**What type of PR is this?**
/kind bug

Optimize ModuleInfo registration and startup process
1. ModuleInfo.contextType is an internal variable, it must be determined at initialization. Instead of abnormally exiting during subsequent use.
2. ModuleInfo.ModuleSocket.IsRemote should be false when the contextType is channel and true when the contextType is unixpacket.

Fixes #

I am concerned about the (Shotgun Surgery) issue; please review it for me.

```
none
```

/cc @fisherxu @Shelley-BaoYue

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/core.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/core.go
@@ -19,24 +19,13 @@ func StartModules() {
 	modules := GetModules()
 
 	for name, module := range modules {
-		var m common.ModuleInfo
-		switch module.contextType {
-		case common.MsgCtxTypeChannel:
-			m = common.ModuleInfo{
-				ModuleName: name,
-				ModuleType: module.contextType,
-			}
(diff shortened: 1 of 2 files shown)


staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -19,24 +19,13 @@ func StartModules() {
...
-			m = common.ModuleInfo{
-				ModuleName: name,
-				ModuleType: module.contextType,
-				// the below field ModuleSocket is only required for using socket.
-				ModuleSocket: common.ModuleSocket{
-					IsRemote: module.remote,
-				},
-			}
-		default:
-			klog.Exitf("unsupported context type: %s", module.contextType)
```
fisherxu, line 39: Is it still needed to check for `unsupported context` for future extension? Though the contextType is hard code now..

PR #5169, opened 2023-11-17
fix go version verify

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

fix go version verify, go minor version must >= 20


fisherxu (commented):
/approve
/assign @WillardHu

PR #5171, opened 2023-11-17
Let the podSendboxImage supports custom repository

WillardHu:
**What type of PR is this?**

kind bug

If the edge node can only use the private image registry, the pod sndbox cannot run properly,  because it cannot pull the kubeedge/pause:3.6 image from the docker hub

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/join_others.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join_others.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"time"
@@ -314,6 +315,10 @@ func createV1alpha1EdgeConfigFiles(opt *common.JoinOptions) error {
 
 	if len(opt.Labels) > 0 {
 		edgeCoreConfig.Modules.Edged.Labels = setEdgedNodeLabels(opt)
+	}
+
+	if opt.ImageRepository != "" {
+		edgeCoreConfig.Modules.Edged.PodSandboxImage = path.Join(opt.ImageRepository, constants.DefaultPodSandboxImage)
 	}
 
 	if errs := validationv1alpha1.ValidateEdgeCoreConfiguration(edgeCoreConfig); len(errs) > 0 {


keadm/cmd/keadm/app/cmd/edge/join_others.go
```diff
@@ -316,6 +317,10 @@ func createV1alpha1EdgeConfigFiles(opt *common.JoinOptions) error {
 		edgeCoreConfig.Modules.Edged.Labels = setEdgedNodeLabels(opt)
 	}
 
+	if opt.ImageRepository != "" {
+		edgeCoreConfig.Modules.Edged.PodSandboxImage = path.Join(opt.ImageRepository, constants.DefaultPodSandboxImage)
```
fisherxu, line 321: Looks good, and now this seems to be set in runtime(containerd), the value in edged/kubelet is only used in imageGC, ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/config/flags.go#L53

cc @Shelley-BaoYue

WillardHu: > Looks good, and now this seems to be set in runtime(containerd), the value in edged/kubelet is only used in imageGC, ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/config/flags.go#L53

Yes, I see the sendbox image is configured in containerd  or cri-dockerd. Do we let the user configure it themselves, or do we change the container configuration when executing the join command? @fisherxu @Shelley-BaoYue

Shelley-BaoYue: I think user configure it by themselves is better.  If we change the cri config and restart runtime, will the keadm join command become risk?  Or we can let the users to choose whether it can be configured by keadm.

WillardHu: > I think user configure it by themselves is better. If we change the cri config and restart runtime, will the keadm join command become risk? Or we can let the users to choose whether it can be configured by keadm.

I agree that it is better to let users configure themselves now, and I will close this pr.

PR #5172, opened 2023-11-17
Delete useless return in NegotiateTunnelPort function

ZhengXinwei-F:
**What type of PR is this?**
/kind bug

Delete useless return in NegotiateTunnelPort function

Fixes #

The last return in the NegotiateTunnelPort function can never be reached.

```
none

```


WillardHu:
Please make cicd required checks successful.

cloud/cmd/cloudcore/app/server.go
```diff
@@ -172,86 +173,88 @@ func NegotiateTunnelPort() (*int, error) {
...
+	return &port, nil
+}
+
+func updateTunnelPortConfig(tunnelPort *v1.ConfigMap, localIP string, kubeClient kubernetes.Interface) (*int, error) {
+	var record iptables.TunnelPortRecord
+	recordStr, found := tunnelPort.Annotations[modules.TunnelPortRecordAnnotationKey]
+	recordBytes := []byte(recordStr)
+	if !found {
+		return nil, errors.New("failed to get tunnel port record")
+	}
```
WillardHu, line 230: Before converting a string to bytes, it is better to first determine if an annotation exists.
```golang
if !found {
    return nil, errors.New("failed to get tunnel port record")
}
recordBytes := []byte(recordStr)
```

PR #5177, opened 2023-11-20
passthrough support regexp, like healthz::*

ZhengXinwei-F:
**What type of PR is this?**
/kind feature

passthrough support regexp, like healthz::*
MetaServer pass through rules should support regular matching.

Fixes #

none

```
none

```

### DIFF
--- a/pkg/util/pass-through/pass_through.go
+++ b/pkg/util/pass-through/pass_through.go
@@ -1,5 +1,9 @@
 package passthrough
 
+import (
+	"regexp"
+)
+
 type passRequest string
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/util/pass-through/pass_through.go
```diff
@@ -12,5 +16,14 @@ var passThroughMap = map[passRequest]bool{
 
 // IsPassThroughPath determining whether the uri can be passed through
 func IsPassThroughPath(path, verb string) bool {
```
WillardHu, line 18: You can add some unittests to test this funcion.

pkg/util/pass-through/pass_through.go
```diff
@@ -12,5 +16,14 @@ var passThroughMap = map[passRequest]bool{
...
-	return passThroughMap[passRequest(path+"::"+verb)]
+	for request, b := range passThroughMap {
+		reg := regexp.MustCompile(string(request))
+		if reg == nil {
+			continue
+		}
+		if reg.Match([]byte(path + "::" + verb)) {
```
WillardHu, line 24: Does the arg `path` definitely have a character prefix '/'？This regex `/version::get` does not match `"version" + "::" + "get"`

PR #5190, opened 2023-11-22
fix ginkgo version in schedule test

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

/kind failing-test

fix ginkgo version in schedule test 

image


fisherxu (commented):
/approve

PR #5193, opened 2023-11-22
Cloudhub opens the quic communication protocol by default.

luomengY:
**What type of PR is this?**
/kind feature

When keadm joins, it already supports selecting cloud edge communication protocols through the hub-protocol（PR:https://github.com/kubeedge/kubeedge/pull/5156）. Currently, it supports two protocols: websocket and quic. Therefore, it is necessary to enable the quic communication protocol of the cloud hub by default.

Fixes #

@Shelley-BaoYue  @fisherxu 

```release-note

```


luomengY:
Previously, we discussed whether to simultaneously open the quic protocol in Cloudcore. Keam supports hub protocol. After selecting quic, if it is not opened by default, it may cause inconvenience for users. We need to discuss whether this should be merged into the master branch. @Shelley-BaoYue @fisherxu

Shelley-BaoYue:
> Previously, we discussed whether to simultaneously open the quic protocol in Cloudcore. Keam supports hub protocol. After selecting quic, if it is not opened by default, it may cause inconvenience for users. We need to discuss whether this should be merged into the master branch. @Shelley-BaoYue @fisherxu

After quic is set to true by default, will there be any problems if two edge nodes connect to the cloud through quic and websocket respectively?

luomengY:
> After quic is set to true by default, will there be any problems if two edge nodes connect to the cloud through quic and websocket respectively?

This will not be a problem, as Cloudcore supports enabling both the Quic and WebSocket protocols simultaneously. In other words, in a cluster, some edge nodes use the WebSocket protocol, while others use the Quic protocol. Cloudcore can support both Quic and WebSocket protocol connections simultaneously.

WillardHu:
> If Cloudcore can support both Quic and WebSocket protocol connections simultaneously, I think this PR can be merged. PTAL @fisherxu @WillardHu

I agree.

fisherxu:
/hold 
Need more tests.

luomengY:
add it to v1.17.0 @Shelley-BaoYue

luomengY:
In the cloudhub, both quic and websocket share the same messageHandle, and opening both quic and websocket has no difference for the server side. The code is as follows:
https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/servers/server.go#L18-L28
@fisherxu

PR #5194, opened 2023-11-22
Workflows Updated 

### LINKED ISSUE #5182: Add CI test case for different runtimes

Add tests for more commonly used runtimes like cri-o, docker. 

- [ ] Add a new job and define CONTAINER_RUNTIME in the matrix as either cri-o or docker (also update CONTAINER_RUNTIME to containerd on line 229).
- [ ] Currently, "make e2e" installs kubeedge using the `hack/local-up-kubeedge.sh`, which only supports containerd. It needs to be modified to support installation with different runtime types depend on the CONTAINER_RUNTIME value.
- [ ] When using docker, cri-dockerd also needs to be installed. Therefore, steps for installing cri-dockerd should be included.

**Why is this needed**: 

Now the KubeEdge CI only use containerd as runtime to test, we need to add more test for commonly used runtimes so that we can ensure KubeEdge run properly with different runtimes.

heysourin:
**What type of PR is this?**
Feature: CI-test

This branch has got updated GitHub workflow actions, that fixes first task of the issue: "Add a new job and define CONTAINER_RUNTIME in the matrix as either cri-o or docker (also update CONTAINER_RUNTIME to containerd on line 229)."

Fixes #5182

There may be a change required on line 390

```release-note
NONE
```


Shelley-BaoYue:
If this PR is still in progress, you can add a `wip` first.

PR #5198, opened 2023-11-23
Script updated

### LINKED ISSUE #5182: Add CI test case for different runtimes

heysourin:
**What type of PR is this?**
test: shell-script

It modifies to support installation with different runtime types depend on the CONTAINER_RUNTIME value

Fixes #5182 Task 2

Have a look at command for containerd

```release-note
NONE
```


hack/local-up-kubeedge.sh
```diff
@@ -46,6 +52,22 @@ function check_prerequisites {
...
 
+function verify_crio_installed {
+  # Commands to check if cri-o is installed
+  if ! command -v crio &> /dev/null; then
+    echo "cri-o is not found"
+    exit 1
+  fi
+}
+
+function verify_containerd_installed {
```
Shelley-BaoYue, line 63: `verify_containerd_installed` has achieved in hack/lib/install.sh. 😄

hack/local-up-kubeedge.sh
```diff
@@ -46,6 +52,22 @@ function check_prerequisites {
   fi
 }
 
+function verify_crio_installed {
```
Shelley-BaoYue, line 55: We can also move verify_crio_installed to hack/lib/install.sh.

hack/local-up-kubeedge.sh
heysourin: Resolved

hack/local-up-kubeedge.sh
heysourin: Resolved
image

PR #5133, opened 2023-11-02
delete serviceaccount token in edgedb

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

When pod is deleted, Kubelet can only delete the cache of the serviceaccounttoken, the data stored in edgeDB can not be deleted.  

**Special notes for your reviewer**:

This PR is only for review for now. We need to release a new kubeedge/kubernetes version and then merge it.

PR #5199, opened 2023-11-23
[release-1.15] fix panic when TokenRequest.Spec.BoundObjectRef is null pointer 

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

In some exceptional scenarios, there might be some wrong with the serviceaccount saved in edge DB, resulting in tr.Spec.BoundObjectRef being a null pointer. This leads to a panic occurring at line 56.


edge/pkg/metamanager/client/serviceaccount.go
```diff
@@ -49,7 +49,7 @@ func (c *serviceAccountToken) DeleteServiceAccountToken(podUID types.UID) {
 	for _, sa := range *svcAccounts {
 		var tr authenticationv1.TokenRequest
 		err = json.Unmarshal([]byte(sa.Value), &tr)
-		if err != nil {
+		if err != nil || tr.Spec.BoundObjectRef != nil {
```
fisherxu, line 52:
```suggestion
		if err != nil || tr.Spec.BoundObjectRef == nil {
```
Seems it should be nil?

fisherxu: Is it also need to fix for master?

Shelley-BaoYue: > Seems it should be nil?

fixed. 😢

Shelley-BaoYue: > Is it also need to fix for master?

PR #5133 fixs it for master.

fisherxu (commented):
/approve

edge/pkg/metamanager/client/serviceaccount.go
WillardHu, line 52: Is it not clear here to use one message describing that the error occurred and that 'BoundObjectRef' is nil? The `BoundObjectRef` is nil because it is nil, not because it is lost by unmarshalling.

Shelley-BaoYue: > Is it not clear here to use one message describing that the error occurred and that 'BoundObjectRef' is nil? The `BoundObjectRef` is nil because it is nil, not because it is lost by unmarshalling.

Sorry, I cannot get what you meant. I added this because during my local test, some dirty data appeared in the edge database. It had the serviceaccount record, but `boundObjectRef` was empty, causing a panic when reading the `tr.Spec.BoundObjectRef.UID` here.

WillardHu: > Sorry, I cannot get what you meant. I added this because during my local test, some dirty data appeared in the edge database. It had the serviceaccount record, but `boundObjectRef` was empty, causing a panic when reading the `tr.Spec.BoundObjectRef.UID` here.

If this is a frequent occurrence, it can be processed with two pieces of logic and return the corresponding error message, if it is only due to a small amount of dirty data, it does not matter.
```golang
if err != nil {
    klog.Errorf("unmarshal resource %s token request failed: %v", sa.Key, err)
    continue
}else if tr.Spec.BoundObjectRef == nil {
    klog.Errorf("the specification BoundObjectRef of resource %s is nil", sa.Key)
    continue
}
```

Shelley-BaoYue: > > Sorry, I cannot get what you meant. I added this because during my local test, some dirty data appeared in the edge database. It had the serviceaccount record, but `boundObjectRef` was empty, causing a panic when reading the `tr.Spec.BoundObjectRef.UID` here.
> 
> If this is a frequent occurrence, it can be processed with two pieces of logic and return the corresponding error message, if it is only due to a small amount of dirty data, it does not matter.
> 
> ```go
> if err != nil {
>     klog.Errorf("unmarshal resource %s token request failed: %v", sa.Key, err)
>     continue
> }else if tr.Spec.BoundObjectRef == nil {
>     klog.Errorf("the specification BoundObjectRef of resource %s is nil", sa.Key)
>     continue
> }
> ```

Yes, it occurs only when there are some dirty data. I add this judgement since it will affect other normal data.

PR #5200, opened 2023-11-23
remove deprecated func of create v1alpha1 edgecoreconfig in keadm

Shelley-BaoYue:
**What type of PR is this?**

/kind cleanup

According to the Keadm version compatibility, starting from version 1.16, installation of edgecore prior to version 1.12 is no longer supported. Therefore, there is no need to consider parameter configurations for v1alpha1 edgecore.

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/join_others.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join_others.go
@@ -36,8 +36,6 @@ import (
 	"github.com/kubeedge/kubeedge/common/constants"
 	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
 	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/util"
-	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha1"
-	validationv1alpha1 "github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha1/validation"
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/edge/join_others.go
```diff
@@ -272,84 +267,3 @@ func runEdgeCore(withMqtt bool) error {
 	klog.Infoln(tip)
 	return nil
 }
-
-func createV1alpha1EdgeConfigFiles(opt *common.JoinOptions) error {
```
fisherxu, line 276: Is there another one in join_windows?

Shelley-BaoYue: > Is there another one in join_windows?

No, because edgecore supports windows since v1.15, it does not involve config of v1alpha1.

fisherxu (commented):
/approve
/assign @WillardHu

WillardHu:
Does the version v1.16 not need to be compatible with versions earlier than v1.12, right?
Do we give users a hint when using incompatible versions? Users should have problems installing v1.11 with the latest keadm.

Shelley-BaoYue:
> Does the version v1.16 not need to be compatible with versions earlier than v1.12, right? Do we give users a hint when using incompatible versions? Users should have problems installing v1.11 with the latest keadm.

As discussed on community meeting, Keadm only supports the installation of the current and previous two versions, but it is not described in the document. I will update the document asap.

PR #5174, opened 2023-11-17
update mapper framework config file

wbc6080:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Currently, mapper-framework configures mapper parameters through the config.yaml file and put the config file into the mapper image. Users cannot change the configuration information unless rebuilding the image. This pr mounts the config configuration information into the mapper pod in the form of configmap, which can easily change the mapper configuration method.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5202, opened 2023-11-23
Updateconfig docker

wbc6080:
**What type of PR is this?**
/kind cleanup

After mounting the config configuration information into the mapper pod in the form of configmap in #5174 , we need to remove the config yaml in mapper framework and in Dockerfile. Now we can use configmap to set mapper config.

Fixes #

```release-note

```


WillardHu:
Does the mapper have to be deployed in the Pod? If there are other deployment methods, a default configuration file is necessary.

wbc6080:
Indeed, when debugging locally, you can directly compile and run the mapper code for deployment. I will re-add config.yaml

wbc6080:
Since we found that in scenarios such as e2e testing, we still need to put the config file into the Dockerfile image, we will temporarily close this PR.

PR #5146, opened 2023-11-07
bump go version to 1.20

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

 bump go version to 1.20

#5122

PR #5207, opened 2023-11-27
update go version in mapper Dockerfile

wbc6080:
**What type of PR is this?**
/kind feature

bump go version to 1.20 according to #5146 

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.20-alpine AS builder
+FROM golang:1.20.10-alpine3.18 AS builder
 
 WORKDIR /build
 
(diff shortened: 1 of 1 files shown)


Shelley-BaoYue:
Does the ubuntu version need to be updated as well? Other parts all use 18.04, although I'm not quite sure what the effect of using 16.04 is.  cc @fisherxu 
image

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.17-alpine AS builder
+FROM golang:1.20-alpine AS builder
```
fisherxu, line 1:
```suggestion
FROM golang:1.20.10-alpine3.18
```
Seems we used the above version in all other dockerfiles cc @Shelley-BaoYue

fisherxu:
> Does the ubuntu version need to be updated as well? Other parts all use 18.04, although I'm not quite sure what the effect of using 16.04 is. cc @fisherxu image

+1, let's keep consistent with others

wbc6080:
OK,I will fix them

fisherxu (commented):
/approve
/assign @Shelley-BaoYue

PR #5239, opened 2023-11-30
bump `slsa-framework / slsa-github-generator` to v1.9.0

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

Bump `slsa-framework / slsa-github-generator` from v1.4.0 to v1.9.0.  

We signed images failed because the download from the cosign-releases GCS bucket was deprecated, cosign-installer need to be updated to v3.1.1 or later version, but `slsa-framework / slsa-github-generator` v1.4.0 still uses cosign-installer v2.8.1.  So we need to bump  `slsa-framework / slsa-github-generator`  to the newest version v1.9.0.

The blog about deprecated cosign-releases GCS bucket: https://blog.sigstore.dev/cosign-releases-bucket-deprecation/ 

The discussion about this pr:  https://github.com/slsa-framework/slsa-github-generator/issues/3006

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -92,7 +92,7 @@ jobs:
       actions: read
       id-token: write
       contents: write
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.2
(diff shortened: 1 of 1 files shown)


fisherxu (commented):
/lgtm
/approve

PR #5248, opened 2023-12-01
Automated cherry pick of #5174: update mapper framework config

wbc6080:
Cherry pick of #5174 on release-1.15.

#5174: update mapper framework config

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/resource/configmap.yaml
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/resource/configmap.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cm-mapper
+data:
+  configData: |
+    grpc_server:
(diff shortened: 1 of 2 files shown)


Shelley-BaoYue:
> From the label of #5174, which looks like a feature or enhancement, and we only cherry-pick bug fix to the patch releases, please have a check, thanks. cc @WillardHu @cl2017 @Shelley-BaoYue

+1. It appears to be an enhancement PR that doesn't affect the capability, we don't need to cherry-pick it.

WillardHu:
> From the label of #5174, which looks like a feature or enhancement, and we only cherry-pick bug fix to the patch releases, please have a check, thanks. cc @WillardHu @cl2017 @Shelley-BaoYue

+1

wbc6080:
My original idea was to cherry-pick it to the 1.15 version mapper-framework, so that users can directly find the mapper configuration file when using it. But Users get yaml from the latest version is also OK

PR #5251, opened 2023-12-02
[WIP]add namespace field in dmi-api

### LINKED ISSUE #5220: device model sync and consider the impact between namespaces
**What happened**:
In the pr #5065 , we fixed device model sync problem. When added or deleted device, need to determine whether the device model needs to be sync to edgecore. However, we did not consider that device instance and device model are namespace-level resources. Resources with the same name in different namespaces may interfere with each other and require further repairs. 
Fix device model sync problem with the same name resources in different namespaces

- [x] We need to update dmi-api to fix this problem  #5251 
- [x] There may be similar problems in the **DeviceController** part, which need to be analyzed and solved. #5329
- [x] There may be similar problems in the **DMI** part, which need to be analyzed and solved. #5253 
- [x] There may be similar problems in the **Mapper Framework** part, which need to be analyzed and solved.  #5280 

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

wbc6080:
**What type of PR is this?**
/kind feature

Now the dmi-api does not consider that device instance and device model are namespace-level resources, and the corresponding fields are not configured. This field needs to be added to deal with issue #5220 

Fixes #
This PR will correct the problems in dmi-api mentioned in the issue #5220 

```release-note

```


wbc6080:
PTAL @cl2017 @WillardHu  thanks

pkg/apis/dmi/v1beta1/api.proto
```diff
@@ -111,8 +111,10 @@ message MapperRegisterResponse {
 message DeviceModel {
     // Name of a device model.
     string name = 1;
+    // NameSpace of the device model.
+    string nameSpace = 2;
```
WillardHu, line 115: 1. The namespace is a word, you can replace it all.
2. Do we need to consider downward compatibility? Is it more appropriate to append a field rather than insert it?

wbc6080:
PTAL @fisherxu thanks

luomengY:
Can we reuse the namespace field defined in the metadata of DeviceModel?
image
image

wbc6080:
In the cloud, there is complete device model information, and metadata can be used. However, the device model sent by the cloud to the mapper is the device model in the dmi-api definition. There is no metadata field, so the namespace field needs to be added. @luomengY

luomengY:
> In the cloud, there is complete device model information, and metadata can be used. However, the device model sent by the cloud to the mapper is the device model in the dmi-api definition. There is no metadata field, so the namespace field needs to be added.

What is the reason why the dmi API was not distributed to mapper with metadata? Would it be better to reuse it here?

wbc6080:
You can refer to [page](https://github.com/kubeedge/kubeedge/blob/master/edge/pkg/devicetwin/dtcommon/util.go#L67C20-L67C20), If we reuse metadata, we still need to define the corresponding fields in dmi-api and implement the corresponding parsing function. I think this will not simplify the current code and may get more useless variables.

Shelley-BaoYue (commented):
/lgtm

fisherxu (commented):
/approve

PR #5252, opened 2023-12-02
modification for edged config

Shelley-BaoYue:
**What type of PR is this?**

/kind cleanup
/kind feature

- modification some parameters in edged config to keep consistent with v1.27.7 kubeletconfig and kubeletflag 
- add DEPRECATED label for some parameter which has been removed in k8s v1.27.7, and remind users we will remove them in KubeEdge v1.17 
- Refactoring ValidateModuleEdged.

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -152,6 +152,7 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 		}, nil
 	}
 
+	// initial kubelet config and flag
 	var kubeletConfig kubeletconfig.KubeletConfiguration
 	var kubeletFlags kubeletoptions.KubeletFlags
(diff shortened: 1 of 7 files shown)


ZhengXinwei-F:
lgtm

edge/pkg/edged/edged.go
```diff
@@ -166,11 +167,19 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	if !edgedconfig.Config.RegisterNode {
 		kubeletConfig.RegisterNode = false
 	}
+
+	// set feature gates from initial flags-based config
+	if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(kubeletConfig.FeatureGates); err != nil {
```
fisherxu, line 172: Without the fix above, the FeatureGate will doesn't work, right?

fisherxu: If so, it should be a bug fix? And should be cherry picked to releases

Shelley-BaoYue: > Without the fix above, the FeatureGate will doesn't work, right?

Yes, I will separate it to another bugfix pr.

fisherxu (commented):
/approve
/assign @WillardHu

PR #5254, opened 2023-12-02
fix staticPodPath default value in windows

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

Fix staticPodPath default value in windows


Shelley-BaoYue:
> /approve Any issue relate to this PR? cc @WillardHu @wujunyi792

No, I discovered this problem when I do some modification work for edged parameters.

PR #5260, opened 2023-12-04
update csi-lib-utils version

wbc6080:
**What type of PR is this?**
/kind bug

update csi-lib-utils package version 

Fixes #

```release-note

```

### DIFF
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -574,8 +574,8 @@ github.com/kubeedge/viaduct/pkg/protos/message
 github.com/kubeedge/viaduct/pkg/server
 github.com/kubeedge/viaduct/pkg/smgr
 github.com/kubeedge/viaduct/pkg/translator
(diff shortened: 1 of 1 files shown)


ZhengXinwei-F:
lgtm

go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/gorilla/websocket v1.4.2
 	github.com/kubeedge/beehive v0.0.0
 	github.com/kubeedge/viaduct v0.0.0
-	github.com/kubernetes-csi/csi-lib-utils v0.6.1
```
fisherxu, line 23: Why bump to v0.11.0? Maybe we can update to the latest version.

wbc6080: When I upgrade to the newer 0.14 version, it succeeds directly, but when I upgrade to a version after 0.15, I get an error. It looks like I need to manually update a series of dependencies. Do we still need to upgrade it to the latest version? @fisherxu 
vendor

PR #5261, opened 2023-12-04
update go-sqlite3 version

wbc6080:
**What type of PR is this?**
/kind bug

update go-sqlite3 package version

Fixes #

```release-note

```


ZhengXinwei-F:
lgtm

wbc6080:
Since the go-sqlite3 official document states that the current new version is after v1.14 and not v2, we need to downgrade the version.

wbc6080:
PLAT @Shelley-BaoYue @fisherxu

go.mod
```diff
@@ -56,7 +56,7 @@ require (
 
 require (
 	github.com/agiledragon/gomonkey v2.0.2+incompatible
-	github.com/beego/beego v1.12.12
```
fisherxu, line 59: Will it report error if we don't modify the beego version?

wbc6080: When I downgrade the go-sqlite3 version to 1.14, the go mod automatically downgrades the beego version as well, it looks like go-sqlite3 depends on the 1.10 version of beego

wbc6080: But I found that the latest version of go-sqlite3 is 1.14.18, I will upgrade it to the latest version

Shelley-BaoYue: How about update beego/beego version to v2？

go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/kubeedge/beehive v0.0.0
 	github.com/kubeedge/viaduct v0.0.0
 	github.com/kubernetes-csi/csi-lib-utils v0.6.1
-	github.com/mattn/go-sqlite3 v2.0.3+incompatible
+	github.com/mattn/go-sqlite3 v1.14.18
```
Shelley-BaoYue, line 24: we can update this to newest version 1.14.22

wbc6080: Now beego/beego version has been upgraded to v2 and go-sqlite3 version has been upgraded to 1.14.22

wbc6080:
Since the v2 version of `beego` is quite different from the v1 version, the unit test cases for sqlite data operations need to be adjusted extensively. In order to avoid blocking the subsequent go and k8s version upgrades, the relevant test cases have been temporarily removed from this PR and use #5677  to track subsequent test case refactoring work @fisherxu @Shelley-BaoYue

Shelley-BaoYue (commented):
/lgtm

PR #5269, opened 2023-12-05
Deprecate the edgecore containerized deployment solution

fisherxu:
**What type of PR is this?**
/kind cleanup

Now we only deploy edgecore in container for testing, in 1) performance test in edgemark 2) keink sub-project like kind.
So mark the edgecore containerized deployment solution as deprecated, avoid to mislead users.

Fixes #

```release-note
NONE
```


fisherxu (commented):
cc @WillardHu @Shelley-BaoYue @luomengY

luomengY:
This is a good suggestion. I think we should remove this solution. Deploying edgecore in a container is not reasonable. One of the responsibilities of edgecore is to manage container applications on compilation nodes. If edgecore is also deployed in a container, then the container cannot be managed, which is similar to the kubelet component in the upstream k8s.

fisherxu:
Yes, I've googled this issue before, the biggest problem is the file system, edgecore/kubelet need to manage the file system of the container, and the file system of container is isolated from the host. unlike the network, we can use host mode for container.

> I think we should remove this solution.

For some dockerfile and deployment yaml files, just deprecate and retain them for developers to research related tests. :)

luomengY:
> For some dockerfile and deployment yaml files, just deprecate and retain them for developers to research related tests. :)

ok, it can be marked as deprecate.

WillardHu:
lgtm

PR #5270, opened 2023-12-06
fix featuregates didn't take effect in edged config

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

FeatureGates config need to be set into DefaultMutableFeatureGate Map, but we didn't set it, causing featureGate in edged config didn't take any effect in previous releases.


fisherxu (commented):
/approve
cc @WillardHu @luomengY

luomengY:
lgtm @Shelley-BaoYue need rebase.

PR #5274, opened 2023-12-06
edge node task proposal

ZhengXinwei-F:
**What type of PR is this?**
/kind feature

edge node task proposal

Fixes #

@fisherxu @Shelley-BaoYue @WillardHu 

```release-note

```


docs/proposals/edge-node-tasks-design.md
```diff
@@ -0,0 +1,306 @@
...
+}
+
+// NodeUpgradeJobStatus stores the status of NodeUpgradeJob.
+// contains multiple edge nodes upgrade status.
+// +kubebuilder:validation:Type=object
+type NodeUpgradeJobStatus struct {
+	// State represents for the state phase of the NodeUpgradeJob.
+	// There are three possible state values: "", upgrading and completed.
+	State UpgradeState `json:"state,omitempty"`
+    currentVersion string
```
WillardHu, line 220: On second thought, I found it doesn't make sense to put the CurrentVersion and HistoricVersion in this structure. This is assumed that the base version of the node associated with the node upgrade task is the same. But in reality, several nodes of different older versions will be upgraded to the latest version.

WillardHu:
BackUp, TimeOut.., they're a word, it's better to convert them to lowercase, like: Backup, Timeout ...

WillardHu:
Please update the proposal with the latest code. :-)

WillardHu:
Should we consider how to handle the proposal file edge-node-upgrade.md?

docs/proposals/edge-node-tasks-design.md
```diff
@@ -0,0 +1,306 @@
...
+    // The default Concurrency value is 1.
+    // +optional
+    Concurrency int32 `json:"concurrency,omitempty"`
+    
+    // CheckItems specifies the items need to be checked before the task is executed.
+    // The default CheckItems value is nil.
+    // +optional
+    CheckItems []string `json:"checkItems,omitempty"`
+
+    FailureTolerate float64 `json:"failureTolerate,omitempty"`
```
WillardHu, line 210: Why does the design use the float type here, but the actual code use the string type?

docs/proposals/edge-node-tasks-design.md
ZhengXinwei-F: Yes. Currently, it appears that if the upgrade procedure is only A -> C, you may simply use State+Event to determine whether it was successful. However, the idea of multi-version upgrades can be incorporated later(A -> B -> C). Also, exercise judgment before updating to confirm that the node's historical version is consistent.

ZhengXinwei-F:
> Should we consider how to handle the proposal file edge-node-upgrade.md?

Action should utilize a verb string.

ZhengXinwei-F:
> Please update the proposal with the latest code. :-)

done. Thanks.

docs/proposals/edge-node-tasks-design.md
ZhengXinwei-F: I've updated the proposal.

ZhengXinwei-F:
> Should we consider how to handle the proposal file edge-node-upgrade.md?

I think it could be preserved temporarily, and the version range can be marked (before 1.16).

docs/proposals/edge-node-tasks-design.md
WillardHu: I think a float type would be more appropriate, otherwise you have to determine if the failureTolerate is empty when converting to a floating point number.

ZhengXinwei-F: Yes.  Like this
```
failureTolerate:
        type: number
        format: float64
```

fisherxu (commented):
/approve

PR #5275, opened 2023-12-06
replace panic with error

cl2017:
**What type of PR is this?**
/kind bug

We'd better not use panic in our code, and fix imports order 

Fixes #

```release-note
NONE
```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/redis/client.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/redis/client.go
@@ -3,11 +3,14 @@ package redis
 import (
 	"context"
 	"encoding/json"
-	"github.com/go-redis/redis/v8"
-	"github.com/kubeedge/Template/pkg/common"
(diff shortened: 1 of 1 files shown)


wbc6080:
Just one small question, do we have any plans to implement these features in the future? If these features are not necessary maybe we can just delete them?
@cl2017

cl2017:
> Just one small question, do we have any plans to implement these features in the future? If these features are not necessary maybe we can just delete them? @cl2017

I think we can keep it and replace panic with error first.
As discussed yesterday, if we need to remove db content later, we can do delete this code together at that time.

wbc6080:
OK
lgtm

fisherxu (commented):
/approve
And seems the lint check haven't covered this dir.

cl2017:
> /approve And seems the lint check haven't covered this dir.

It seems so. Let me add it? Expect to find a lot of problems, and we can fix them together

fisherxu:
I've open an issue to track it :) https://github.com/kubeedge/kubeedge/issues/5291 @cl2017

PR #5276, opened 2023-12-06
Replace deprecated `io/ioutil` function

cl2017:
**What type of PR is this?**
/kind feature

As we upgrade to go 1.20, replace `io/ioutil` with `io` or `os`

Fixes #

```release-note
NONE
```


wbc6080:
Very nice modification, looks like we should have changed the above problem as early as version 1.16 👍 
/lgtm

PR #5137, opened 2023-11-03
Init owners for mapper-framework

fisherxu:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:

Add owners for mapper-framework, let's promote the development of mapper-framework :)

@cl2017 @WillardHu and @fisherxu as init approvers, and @wbc6080 as reviewer.

And Thanks for the contributions for all. 

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #5277, opened 2023-12-06
Add DB handler

cl2017:
**What type of PR is this?**
/kind feature

The `dbpush` code is not suitable in the `device.go` file, and it should be writed in the `data/dbmethod` directory, then it can be handled by different db alone.
So in this PR, I add new `handler` in every db directory to push data to db.

Fixes #

```release-note
NONE
```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/influxdb2/handler.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/influxdb2/handler.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2023 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package influxdb2
+
+import (
+	"context"
+	"time"
+
+	"k8s.io/klog/v2"
+
+	"github.com/kubeedge/Template/driver"
+	"github.com/kubeedge/mapper-framework/pkg/common"
+)
+
+func DataHandler(ctx context.Context, twin *common.Twin, client *driver.CustomizedClient, visitorConfig *driver.VisitorConfig, dataModel *common.DataModel) {
+	dbConfig, err := NewDataBaseClient(twin.Property.PushMethod.DBMethod.DBConfig.Influxdb2ClientConfig, twin.Property.PushMethod.DBMethod.DBConfig.Influxdb2DataConfig)
+	if err != nil {
+		klog.Errorf("new database client error: %v", err)
+		return
+	}
+	dbClient := dbConfig.InitDbClient()
+	if err != nil {
+		klog.Errorf("init database client err: %v", err)
+		return
+	}
+	reportCycle := time.Duration(twin.Property.ReportCycle)
+	if reportCycle == 0 {
+		reportCycle = common.DefaultReportCycle
+	}
+	ticker := time.NewTicker(reportCycle)
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				deviceData, err := client.GetDeviceData(visitorConfig)
+				if err != nil {
+					klog.Errorf("publish error: %v", err)
+					continue
+				}
+				sData, err := common.ConvertToString(deviceData)
+				if err != nil {
+					klog.Errorf("Failed to convert publish method data : %v", err)
+					continue
+				}
+				dataModel.SetValue(sData)
+				dataModel.SetTimeStamp()
+
+				err = dbConfig.AddData(dataModel, dbClient)
+				if err != nil {
+					klog.Errorf("influx database add data error: %v", err)
+					return
+				}
+			case <-ctx.Done():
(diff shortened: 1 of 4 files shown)


wbc6080:
nice modification
/lgtm

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/influxdb2/handler.go
```diff
@@ -0,0 +1,57 @@
+package influxdb2
```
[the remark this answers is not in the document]
cl2017: ok, done.

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
```diff
@@ -170,7 +170,7 @@ func pushHandler(ctx context.Context, twin *common.Twin, client *driver.Customiz
 	}
 	reportCycle := time.Duration(twin.Property.ReportCycle)
 	if reportCycle == 0 {
-		reportCycle = common.DefaultReportCycle
+		reportCycle = 1 * time.Second
```
wbc6080, line 173: We change it to common.DefaultReportCycle in #5230  
```suggestion
		reportCycle = common.DefaultReportCycle
```

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
```diff
@@ -202,37 +202,7 @@ func dbHandler(ctx context.Context, twin *common.Twin, client *driver.Customized
...
-						continue
-					}
-					sData, err := common.ConvertToString(deviceData)
-					if err != nil {
-						klog.Errorf("Failed to convert publish method data : %v", err)
-						continue
-					}
-					dataModel.SetValue(sData)
-					dataModel.SetTimeStamp()
+		dbInflux.DataHandler(ctx, twin, client, visitorConfig, dataModel)
```
wbc6080, line 205: Do we need to add dbredis and dbtdengine too?

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
cl2017: yes, I pushed the new code.

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
cl2017: yes, I pushed the new code.

wbc6080 (commented):
/lgtm
cc @WillardHu If you think there is no problem with this PR, you can directly approve it because #5137

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/influxdb2/handler.go
```diff
@@ -0,0 +1,73 @@
...
+					klog.Errorf("Failed to convert publish method data : %v", err)
+					continue
+				}
+				dataModel.SetValue(sData)
+				dataModel.SetTimeStamp()
+
+				err = dbConfig.AddData(dataModel, dbClient)
+				if err != nil {
+					klog.Errorf("influx database add data error: %v", err)
+					return
```
WillardHu, line 65: Is this an attempt to interrupt the loop?

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/influxdb2/handler.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"time"
+
+	"k8s.io/klog/v2"
+
+	"github.com/kubeedge/Template/driver"
+	"github.com/kubeedge/mapper-framework/pkg/common"
+)
+
+func DataHandler(ctx context.Context, twin *common.Twin, client *driver.CustomizedClient, visitorConfig *driver.VisitorConfig, dataModel *common.DataModel) {
```
WillardHu, line 29: Duplicate code looks like it can be abstracted away.

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/influxdb2/handler.go
cl2017: It seems so. Because these code is moving from `device.go`, I think we better not change some logic in this PR.

wbc6080: The original idea here was to exit the dbhandler and not push again if adding data to the database failed. Do you think it would be better to continue trying to push? @WillardHu

WillardHu: If the error is unrecoverable, the processing can be interrupted.

cl2017: If stop push when adding data to the database failed, we can not push again unless restart mapper, right?

wbc6080: When the user updates the device, the initialization of the device will be re-executed. In this case, the database push will also be re-executed. The most common reason for errors here is that the secret key of the user database is incorrect. At this time, the user can update the correct secret key and push again.

WillardHu:
We can optimize the `DataHandler(..)` function in the future. @wbc6080 @cl2017

cl2017:
> We can optimize the `DataHandler(..)` function in the future. @wbc6080 @cl2017

ok, I'll open another PR to do this.

PR #5279, opened 2023-12-06
fix mutex unlock

cl2017:
**What type of PR is this?**
/kind bug

We need to unlock the mutex after the loop ends, instead of defining defer when function ends.

Fixes #

```release-note
NONE
```


wbc6080:
It is indeed better to change this way
lgtm

fisherxu (commented):
Looks good, and just curious any bugs will it cause?
/approve

cl2017:
> Looks good, and just curious any bugs will it cause? /approve

it might be, because `CreateDMIClient` will use another mutex, which can cause blocking in some cases

fisherxu (commented):
> it might be, because CreateDMIClient will use another mutex, which can cause blocking in some cases

Ok, so we should cherry-pick it to the release branchs.

PR #5281, opened 2023-12-06
Update parse/grpc.go in mapper-framework

cl2017:
**What type of PR is this?**
/kind feature

This PR Update `grpc.go`:
- Add constant definition
- Optimize the if logic and reduce one if
- Error content added description
- Refine other code details

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
@@ -16,23 +16,20 @@ type TwinResultResponse struct {
 }
 
 func getProtocolNameFromGrpc(device *dmiapi.Device) (string, error) {
-
 	return device.Spec.Protocol.ProtocolName, nil
 }
 
 func getPushMethodFromGrpc(visitor *dmiapi.DeviceProperty) (string, error) {
-	// TODO add more push method
-	if visitor.PushMethod.Http != nil {
-		return "http", nil
+	if visitor.PushMethod != nil && visitor.PushMethod.Http != nil {
+		return common.PushMethodHTTP, nil
 	}
-	if visitor.PushMethod.Mqtt != nil {
-		return "mqtt", nil
+	if visitor.PushMethod != nil && visitor.PushMethod.Mqtt != nil {
+		return common.PushMethodMQTT, nil
 	}
 	return "", errors.New("can not parse publish method")
 }
 
 func getDBMethodFromGrpc(visitor *dmiapi.DeviceProperty) (string, error) {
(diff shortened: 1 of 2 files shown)


staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
```diff
@@ -132,68 +131,67 @@ func buildPropertiesFromGrpc(device *dmiapi.Device) []common.DeviceProperty {
 		var dbconfig common.DBConfig
 		var pushMethod []byte
 		var pushMethodName string
-		if pptv.PushMethod != nil {
-			if pptv.PushMethod.DBMethod != nil {
-				dbMethodName, err = getDBMethodFromGrpc(pptv)
+		if pptv.PushMethod != nil && pptv.PushMethod.DBMethod != nil {
+			dbMethodName, err = getDBMethodFromGrpc(pptv)
```
wbc6080, line 135: In order to fix the null pointer problem before, we put pptv.PushMethod != nil in a separate if. If you change it like this, if the user does not configure the http or mqtt field of pushMethod, it seems that the null pointer problem will reappear on line 178.

wbc6080: What I mean is that after this if, the pushMethodName field will be parsed on line 175. pptv.PushMethod != nil is not judged on line 175. This part of the original code is in the if of pptv.PushMethod != nil.

wbc6080: pushmethod

Shelley-BaoYue: > What I mean is that after this if, the pushMethodName field will be parsed on line 175. pptv.PushMethod != nil is not judged on line 175. This part of the original code is in the if of pptv.PushMethod != nil.

OK, I get it.

cl2017: Oh, yes. But if that's the case, these `if` is confusing, so let me fix it again

fisherxu (commented):
/assign @WillardHu 
Plz help take a look :)

cl2017:
> This method `buildPropertiesFromGrpc` is not sensitive to errors and we do not know if it will have an impact on the business. If an error occurs in the method, it will only be logged and properties will be lost.
> 
> In addition, I feel that there are many places in this file that can be abstracted and simplified.

ok, I'm going to look at this function more time, and I'll submit it again later.

fisherxu (commented):
/approve

PR #5288, opened 2023-12-08
fix mapper framework generate.sh to build image successfully

### LINKED ISSUE #5287: Build mapper image fail
**What happened**:
When we use mapper-framework to generate a mapper project, an error will occur if we directly build the mapper image.
docker

Successfully built mapper image
Execute the make generate command in mapper-framework to generate the mapper project, and then docker build the image in the mapper project

**Environment**:
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): v1.15.0

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

wbc6080:
**What type of PR is this?**
/kind bug

Fix the mapper generate.sh to successfully build the images refer to #5287 
I tried adding the database dependency package in go.mod of mapper-framework, but when I executed `make verify` at the end, the package I added was automatically removed. It seems that go mod cannot correctly identify the dependency reference under the _template file. So I executed `go mod tidy` in the generate.sh to re-import the database dependencies.
By the way, I also corrected the import order of tdengine.

Fixes #5287 

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/tdengine/client.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/tdengine/client.go
@@ -4,12 +4,14 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
-	"github.com/kubeedge/Template/pkg/common"
-	_ "github.com/taosdata/driver-go/v3/taosRestful"
-	"k8s.io/klog/v2"
 	"os"
 	"strings"
 	"time"
+
+	_ "github.com/taosdata/driver-go/v3/taosRestful"
+	"k8s.io/klog/v2"
+
+	"github.com/kubeedge/Template/pkg/common"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


wbc6080:
PTAL @cl2017  and Do you have any other better way to solve this problem?

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile
```diff
@@ -7,6 +7,8 @@ ENV GO111MODULE=on \
 
 COPY . .
 
+RUN  go mod tidy
```
WillardHu, line 10: It doesn't need to be declared

wbc6080: You can refer to the issue #5287 I raised. Since the _template file seems to be skipped when executing `make verify` to detect dependencies, the dependency on the database cannot be added to go.mod, resulting in an error during docker build, so I executed `go mod tidy` here to refresh go.mod after generating the mapper project. @WillardHu

WillardHu: It really depends on the user generated project, and you can execute go mod tidy in the generated project.

wbc6080: Do you mean that we execute `go mod tidy` directly when generating the user mapper project instead of when building the image?

WillardHu: Yes, this is more suitable for the user's use scenario, and the user local debug will actually update the go.mod

wbc6080: After testing, it can indeed achieve the same effect. I have re-edited

WillardHu:
Please make e2e test successful

fisherxu (commented):
/approve

PR #5290, opened 2023-12-08
Add Device Admission

cl2017:
**What type of PR is this?**
/kind feature

Add admission validation rules for the `device CR` to ensure that the `property.name` is unique.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/admissioncontroller/admit_device.go
+++ b/cloud/pkg/admissioncontroller/admit_device.go
@@ -1,0 +1,62 @@
+package admissioncontroller
+
+import (
+	"net/http"
+	"strings"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+
+	devicesv1beta1 "github.com/kubeedge/kubeedge/pkg/apis/devices/v1beta1"
+)
+
+func admitDevice(review admissionv1.AdmissionReview) *admissionv1.AdmissionResponse {
+	reviewResponse := admissionv1.AdmissionResponse{}
+	reviewResponse.Allowed = true
+	var msg string
+
+	switch review.Request.Operation {
+	case admissionv1.Create, admissionv1.Update:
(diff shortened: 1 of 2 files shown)


WillardHu:
You need to add the webhook configuration to the ValidatingWebhookConfiguration in the method `registerWebhooks(..)`

cl2017:
> You need to add the webhook configuration to the ValidatingWebhookConfiguration in the method `registerWebhooks(..)`

ok, done.

cloud/pkg/admissioncontroller/admit_device.go
```diff
@@ -0,0 +1,60 @@
...
+	//device properties name must be unique.
+	var msg string
+	propertyNameMap := make(map[string]bool)
+	for _, property := range device.Spec.Properties {
+		if _, ok := propertyNameMap[property.Name]; !ok {
+			propertyNameMap[property.Name] = true
+		} else {
+			msg = "property names must be unique."
+			response.Allowed = false
+		}
```
WillardHu, line 53: There is no need to use map to determine whether the name is the same
```golang
size := len(device.Spec.Properties)
for i := range device.Spec.Properties {
    for j := i+1; j < size; j++ {
        if device.Spec.Properties[i].Name == device.Spec.Properties[j].Name {
            msg = "property names must be unique."
            response.Allowed = false
            break
        }
    }
}
```

cl2017: I think the time complexity of using a hash table is O(n), which should be lower than that of a two-level loop.

WillardHu: > I think the time complexity of using a hash table is O(n), which should be lower than that of a two-level loop.

Using map requires memory to be allocated and hash code to be calculated. 

When the amount of data is large, the hash code used as map index performs better than loop data.

The value of map doesn't mean much here either.

cl2017: I have modified

wbc6080:
LGTM

PR #5294, opened 2023-12-09
 Add nodename in edge certificate

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

Add nodename in edge certificate to achieve certificate uniqueness

/sig security

### DIFF
--- a/cloud/pkg/cloudhub/servers/httpserver/server.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/server.go
@@ -36,6 +36,7 @@ import (
 
 	hubconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/config"
 	"github.com/kubeedge/kubeedge/common/constants"
+	"github.com/kubeedge/kubeedge/common/types"
 )
 
 // StartHTTPServer starts the http service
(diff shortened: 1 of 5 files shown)


cloud/pkg/cloudhub/servers/httpserver/server.go
```diff
@@ -86,8 +86,9 @@ func EncodeCertPEM(cert *x509.Certificate) []byte {
 
 // edgeCoreClientCert will verify the certificate of EdgeCore or token then create EdgeCoreCert and return it
 func edgeCoreClientCert(request *restful.Request, response *restful.Response) {
+	nodeName := request.Request.Header.Get("NodeName")
```
WillardHu, line 89: Please use `request.Request.Header.Get(constants.NodeName)` and replase the repeatedly defined code with the nodeName

WillardHu:
Does this check only when the edge node obtains the certs? And does not affect the signed certs in response, right?

cloud/pkg/cloudhub/servers/httpserver/server.go
Shelley-BaoYue: > Please use `request.Request.Header.Get(constants.NodeName)` and replase the repeatedly defined code with the nodeName

fixed.

Shelley-BaoYue:
> Does this check only when the edge node obtains the certs? And does not affect the signed certs in response, right?

The certs subject common name is decided by edgecore. After this PR merged, the signed certs which the new version edgecore request for will contain nodename in certificate subject. So cloudhub will verify if the cert has nodename. Of course, we will continue to maintain compatibility with the certificates obtained by older versions of Edgecore.

cloud/pkg/cloudhub/servers/httpserver/server.go
WillardHu: You can replace codes `request.Request.Header.Get(constants.NodeName)` with `nodeName` at L93 and L106

Shelley-BaoYue: > You can replace codes `request.Request.Header.Get(constants.NodeName)` with `nodeName` at L93 and L106

Done. And I moved the `constants.NodeName` to `types.NodeNameKey` in `common/types/http.go`, since I think it is more appropriate to place the request const in http.go same as the` AuthorizationKey` const.

fisherxu (commented):
/approve

PR #5299, opened 2023-12-11
 remove `--runtimetype` in keadm

### LINKED ISSUE #5285: Remove `--runtimetype` in keadm join in release 1.16

Remove `--runtimetype` in keadm join command 

**Why is this needed**:

Kubernetes v1.27 has removed `ContainerRuntime` flag for kubelet  -->  In KubeEdge next version, v1.16, the `--runtimetype` will no longer have any effect, and therefore it can be removed.

And from Kubernetes v1.24, the only valid value of `ContainerRuntime` is `remote`.  -->  From KubeEdge v1.14, due to the upgrade of Kubernetes dependency to version 1.24, the `--runtimetype` can only be configured as "remote."

I have evaluated the impact of the removal, the following observations have been made:
- Since the v1.14 version, we have already provided a notification that the runtimetype will be removed in v1.16.
- Following community standards, keadm needs to be compatible with the current version and the two previous versions. In other words, the v1.16 keadm needs to be compatible with v1.14 and v1.15. As mentioned earlier, the runtimetype only supports the "remote" from v1.14. Therefore, remove `--runtimetype` will not affect compatibility with older versions.

image

cc @fisherxu @WillardHu

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

remove `--runtimetype` in keadm since the runtimetype is no longer effective, and from v1.14, it only supports `remote` to set. 

Fixes #5285


keadm/cmd/keadm/app/cmd/edge/join_others.go
```diff
@@ -60,9 +59,6 @@ func AddJoinOtherFlags(cmd *cobra.Command, joinOptions *common.JoinOptions) {
 		fmt.Printf("mark flag required failed with error: %v\n", err)
 	}
 
-	cmd.Flags().StringVarP(&joinOptions.RuntimeType, common.RuntimeType, "r", joinOptions.RuntimeType,
```
WillardHu, line 63: Please remove the constant common.RuntimeType and relevant commands: `keadm reset`, `keadm config iamges`

Shelley-BaoYue: > Please remove the constant common.RuntimeType and relevant commands: `keadm reset`, `keadm config iamges`

Sorry, could you please point out the specific location which need to removed? It seems like I have already deleted the constant common.RuntimeType.

WillardHu: Sorry, I didn't see it clearly. I thought you only deleted the flag of the edge join command.

fisherxu (commented):
/approve

PR #5305, opened 2023-12-12
remove unused protocol field

wbc6080:
**What type of PR is this?**
/kind cleanup

The protocols field in the current DevPanel structure is not actually used, and the protocol information is already reflected in the device instance and device model, so I think this field can be removed.

Fixes #

```release-note

```


fisherxu (commented):
/approve

PR #5308, opened 2023-12-13
fix go mod to prepare for mapper upgrade

wbc6080:
**What type of PR is this?**
/kind feature

This PR adds the go.mod file to _template and makes changes to generate.sh. 
In this PR,  `make generate` will only generate files in the _template directory, and the original pkg files will be referenced in the form of packages to facilitate mapper upgrades. 
In addition, this PR fix the module of mapper-framework to the path of mapper-framework repo to correct reference.

Fixes #

This PR is mainly the preparation work for mapper upgrade. After merge and sync to mapper-framework repo, we need to correctly create tags for the mapper-framework repo. On this basis, subsequent PRs will modify the way `pkg` is referenced in _template

```release-note

```


fisherxu (commented):
/approve

Issue #5007, opened 2023-09-12
Support preloading docker image to the edge node


Skilence:
**What would you like to be added/modified**:
Support preloading docker image to the edge node before  deployment.
支持在部署前提前加载镜像到边缘节点上

**Why is this needed**:
When the network between cloud and edge is unstable, we can download the docker image to the node in advance when the network is good. And then perform the upgrade as planned later.
当云边网络不稳定时，可以在网络好的时候提前将镜像下载到边缘节点上，之后再根据计划统一进行升级

PR #5310, opened 2023-12-13
add images prepull proposal

Shelley-BaoYue:
**What type of PR is this?**

/kind design

add images prepull proposal

#5007


docs/proposals/image-prepull.md
```diff
@@ -0,0 +1,293 @@
...
+              operator: "replace"
+              value: "hangzhou.registry.io"
+          secretOverriders: 
+            - component: "SecretName"
+              operator: "replace"
+              value: "hangzhou-secret"              
+
+```
+
+## Plan
```
fisherxu, line 204: Need to add the release plan details.

WillardHu:
We need to update this document with the latest code.

docs/proposals/image-prepull.md
```diff
@@ -0,0 +1,293 @@
...
+          secretOverriders: 
+            - component: "SecretName"
+              operator: "replace"
+              value: "hangzhou-secret"              
+
+```
+
+## Plan 
+
+- Support image and secret overriders
```
WillardHu, line 206: If we define a pre-pull image task for specified of nodes, is it necessary to define a subset of node overrides?

Shelley-BaoYue: > If we define a pre-pull image task for specified of nodes, is it necessary to define a subset of node overrides?

Image and secret overriders are used for private image repository or repository for nodegroups in different regions.  I can't quite get why it's necessary to override nodes. Could you provide an example scenario? I think filtering nodes based on labels or directly specifying a list of nodes should be enough.

WillardHu: image

Yes, we can satisfy multiple scenarios by defining multiple ImagePrePull tasks. I mean probably the imagePrePullOverriders don't need it either. Example of dockerconfig json secret:
```json
"auths": {
  "<registry-1>": {"username": "", "password", ""},
  "<registry-2>": {"username": "", "password", ""}
}
```

docs/proposals/image-prepull.md
```diff
@@ -0,0 +1,293 @@
...
+   // State represents for the state phase of the ImagePrePullJob.
+   // There are four possible state values: "", prepulling, successful and failed.
+   State string `json:"state,omitempty"`
+  
+  // Status contains image prepull status for each edge node.
+   Status []ImagePrePullStatus `json:"status,omitempty"`
+}
+
+// ImagePrePullStatus stores image prepull status for each edge node.
+type ImagePrepullStatus struct {
```
WillardHu, line 134: The latest ImagePrePullStatus structure is:
```golang
type ImagePrePullStatus struct {
	// TaskStatus represents the status for each node
	*TaskStatus `json:"nodeStatus,omitempty"`
	// ImageStatus represents the prepull status for each image
	ImageStatus []ImageStatus `json:"imageStatus,omitempty"`
}
```
The memeber TaskStatus refers to the structure of the node upgrade task. How will its fields(status, event, action, reason, time) be represented in the image pre pull task?

docs/proposals/image-prepull.md
Shelley-BaoYue: The overrider is indeed just a preliminary solution discussed before.  I will compare it with other approaches and choose the best one to implement.

docs/proposals/image-prepull.md
Shelley-BaoYue: TaskStatus is not specifically refer to node upgrade task, it's an abstracted status for all tasks. In Image Prepull task, status means the overall state of the pull task, event represents specific events in images prepull process, such as check, pull, and etc.

WillardHu: > TaskStatus is not specifically refer to node upgrade task, it's an abstracted status for all tasks. In Image Prepull task, status means the overall state of the pull task, event represents specific events in images prepull process, such as check, pull, and etc.

Okey, I think it will improve readability if we define upgrade node task constants and image prepull task constants completely separately, although they will have partially the same value.

PR #5325, opened 2023-12-20
Allow configuration of dmi.sock path

### LINKED ISSUE #5323: local-up-kubeedge script cannot create dmi.sock when e2e testing
**What happened**:
During e2e testing workflow, local-up-kubeedge script is executed to generate a kubeeedge cluster, but the edgecore log shows that dmi.sock cannot be created.
edgecore日志

Create dmi.sock correctly when e2e testing
Execute e2e test
In the code, the listener `/etc/kubeedge/dmi.sock` will be automatically created when dmiserver starts. I tried to output the contents of the current `etc` directory and found that the `/etc/kubeedge` directory was not created, causing the listener created fail.
etc目录

**Environment**:
- Kubernetes version (use `kubectl version`):  v1.26
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): v1.15.1

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

wbc6080:
**What type of PR is this?**
/kind bug

Based on the situation encountered in the issue #5323 , We add the configuration field for the dmi.sock path in edgecore.yaml to allow customization of the dmi.sock path. For example, in the local-up script, the original path /etc/kubeedge.dmi.sock will become /tmp/etc/kubeedge/dmi.sock

Fixes #5323 

```release-note

```

### DIFF
--- a/common/constants/default.go
+++ b/common/constants/default.go
@@ -156,4 +156,7 @@ const (
 	// EdgeHub
 	DefaultWebSocketPort = 10000
 	DefaultQuicPort      = 10001
+
+	// DeviceTwin
+	DefaultDMISockPath = "/etc/kubeedge/dmi.sock"
 )
--- a/edge/pkg/devicetwin/dmiserver/server.go
+++ b/edge/pkg/devicetwin/dmiserver/server.go
@@ -37,6 +37,7 @@ import (
 	"github.com/kubeedge/kubeedge/common/constants"
 	messagepkg "github.com/kubeedge/kubeedge/edge/pkg/common/message"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/modules"
+	deviceconfig "github.com/kubeedge/kubeedge/edge/pkg/devicetwin/config"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dmiclient"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dtcommon"
 	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/dao"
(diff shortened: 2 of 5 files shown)


WillardHu:
The `/etc/kubeedge` folder will be generated after `function start_edgecore` is executed. Should your script be executed before this?

wbc6080:
Because if the `/etc/kubeedge` path does not exist after starting edgecore, dmi.sock cannot be created and will not be retried, so I modify the path created here after building edgecore and before starting the kubeedge cluster. But I didn’t see the command to create` /etc/kubeedge` path in the function `start_edgecore` . Did I miss something? @WillardHu

WillardHu:
> Because if the `/etc/kubeedge` path does not exist after starting edgecore, dmi.sock cannot be created and will not be retried, so I modify the path created here after building edgecore and before starting the kubeedge cluster. But I didn’t see the command to create` /etc/kubeedge` path in the function `start_edgecore` . Did I miss something? @WillardHu

The shell `local-up-kubeedge` generated edgecore to use the configuration path seems to be `/tmp/etc/kubeedge`. Can you see how to get your logic to read the folder `/tmp/etc/kubeedge`?

wbc6080:
Yes, the local-up script will place edgecore's ca certificate and other contents in the `/tmp/etc/kubeedge` path. But we set the sock address of dmi to `/etc/kubeedge/dmi.sock` in the code, and the document also mentioned that the configuration information related to the edge should be under the` /etc/kubeedge` path, so I think the /etc/kubeedge path should be created @WillardHu

WillardHu:
I think it would be better for edgecore to be able to configure path `/etc/kubeedge/` through configuration files `edgecore.yaml`

wbc6080:
OK. I have moved the path creation work to edgecore. @WillardHu

WillardHu:
> OK. I have moved the path creation work to edgecore. @WillardHu

I mean add a EtcPath setting to edgecore.yaml default is `/etc/kubeedge`, and `SockPath = path.Join(config.EtcPath, "dmi.sock")`. This allows you to modify the EtcPath to `/tmp/etc/kubeedge` to be compatible with this localup scenario.

wbc6080:
I have added the dmisock path configuration field to edgecore.yaml as you suggested. PTAL @WillardHu

WillardHu:
> I have added the dmisock path configuration field to edgecore.yaml as you suggested. PTAL @WillardHu

Do you need to change this configuration to make the e2e test passed in local up script?

wbc6080:
> Do you need to change this configuration to make the e2e test passed in local up script?

I have tested that when running the local-up script, the path to dmi.sock will be automatically replaced by /tmp/etc/kubeedge/dmi.sock. It seems that the substitution here has a direct effect. @WillardHu 

tihuan

fisherxu (commented):
/approve

PR #5280, opened 2023-12-06
Modify the same name problem caused by namespace in mapper-framework

wbc6080:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
We did not consider that device instance and device model are namespace-level resources. Resources with the same name in different namespaces may interfere with each other. So we fix it in mapper-framework
**Which issue(s) this PR fixes**:

Fixes #
This PR will correct the problems in mapper-framework mentioned in the issue #5220 
**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5253, opened 2023-12-02
update dmi server to get namespace

wbc6080:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
After adding the namespace field to dmi-api, we need to modify the dmi server part and deliver the namespace of device instance and device model to mapper refer to #5220 
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5329, opened 2023-12-22
fix namespace problem in cloudcore

wbc6080:
**What type of PR is this?**
/kind bug

After adding the namespace field to dmi-api, we need to modify the devicecontroller part and deliver the namespace of device instance and device model to mapper refer to #5220 

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/devicecontroller/manager/common.go
+++ b/cloud/pkg/devicecontroller/manager/common.go
@@ -44,3 +44,8 @@ func (c *CommonResourceEventHandler) OnDelete(obj interface{}) {
 func NewCommonResourceEventHandler(events chan watch.Event) *CommonResourceEventHandler {
 	return &CommonResourceEventHandler{events: events}
 }
+
+// GetResourceID return resource ID
+func GetResourceID(namespace, name string) string {
+	return namespace + "-" + name
+}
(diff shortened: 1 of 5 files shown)


wbc6080:
Could you please help to review this PR and its related #5253 #5280 PRs? The test of #5253 failed because #5251 has not been merged yet. You can check the code first. Thanks @cl2017 @WillardHu

cloud/pkg/devicecontroller/manager/common.go
```diff
@@ -44,3 +44,8 @@ func (c *CommonResourceEventHandler) OnDelete(obj interface{}) {
 func NewCommonResourceEventHandler(events chan watch.Event) *CommonResourceEventHandler {
 	return &CommonResourceEventHandler{events: events}
 }
+
+// GetResourceID return resource ID
+func GetResourceID(namespace, name string) string {
+	return namespace + "-" + name
```
fisherxu, line 50:
```suggestion
	return namespace + "/" + name
```
We should use `/` here, because the namesapce or name maybe include `-`, then you can't split ns/name from the key.

Ref in k8s: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/object-names.go#L44

wbc6080: Done @fisherxu

fisherxu (commented):
/approve

Issue #5264, opened 2023-12-05
keadm reset - Log output suggest optimization


shijinye:
**What happened**:
keadm reset  - Log output suggest optimization

when i execute reset ，log output “failed to check cloudcore is running: failed to create KubeClient, error: get kube config failed with error: stat /root/.kube/config: no such file or directory” ，but in fact，it is succeed

![企业微信截图_00ac12fc-de11-4fb0-8d3b-864dfcf917c1](https://github.com/kubeedge/kubeedge/assets/36811869/9eb5c30b-7726-47b5-a141-399d1bfe46b8)

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`):
- <details><summary>Cloud nodes Environment:</summary>

  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>
- <details><summary>Edge nodes Environment:</summary>

  - edgecore version (e.g. `edgecore --version`):
  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>

Shelley-BaoYue:
Could you please provide the keadm version and cloudcore version :-）

Shelley-BaoYue:
The keadm reset has been optimized in the latest version. You are welcome to use it. If you have any questions, you can always reopen this issue.

Issue #5317, opened 2023-12-15
Enhance the installation tool keadm


WillardHu:
**What would you like to be added/modified**:

1. [x] Refactor the implementation code for the cloud init command, use the gofile `keadm/cmd/keadm/app/cmd/helm/cloudcore.go` func `Install(..)` to complete the install logic.  https://github.com/kubeedge/kubeedge/pull/5461
2. [x] Change the print helm chart resources YAML command name`keadm generate` to `keadm manifest`. https://github.com/kubeedge/kubeedge/pull/5462
3. [x] Refactor the command `reset` to third-level commands, like: `keamd reset cloud` and `keadm reset edge`. If uses `keadm rest`, both the cloud and the edge will be reset. https://github.com/kubeedge/kubeedge/pull/5463
4. [x] Add a parent flag `--log-level` to the keadm and make it hidden for debug. https://github.com/kubeedge/kubeedge/pull/5547
5. [x] Add some custom image repository flags to `keadm init` and `keadm upgrade`. https://github.com/kubeedge/kubeedge/pull/5540
6. [x] Add a flag `--set` to edge command `keadm join` makes the edgecore to modify the configuration file `/etc/kubeedge/config/edgecore.yaml` before startup.  @HT0403
7. [ ] ~~Runtime-endpoint is automatically discrovered when the upgrade and reset commands are executed.~~

**Why is this needed**:

Shelley-BaoYue:
“Add a flag --set to edge command keadm join makes the edgecore to modify the configuration file /etc/kubeedge/config/edgecore.yaml before startup.”

> I completely agree with this point. Could we discuss in community meeting whether to support modifying config using `--set` or allowing users to directly use the local edgecore.yaml?

WillardHu:
> I completely agree with this point. Could we discuss in community meeting whether to support modifying config using `--set` or allowing users to directly use the local edgecore.yaml?

This can be discussed at community meeting.

luomengY:
Add a flag --set to edge command keadm join makes the edgecore to modify the configuration file /etc/kubeedge/config/edgecore.yaml before startup.

WillardHu:
Add a task: Runtime-endpoint is automatically discrovered when the upgrade and reset commands are executed.

528548004:
```
 Runtime-endpoint is automatically discrovered when the upgrade and reset commands are executed.
```

528548004:
[...]

PR #5338, opened 2023-12-29
fix keadm error log

tangming1996:
**What type of PR is this?**
/kind bug

When keadm reset is executed at the edge, the cloud error log is reported, which can mislead users.
#5264

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common_others.go
+++ b/keadm/cmd/keadm/app/cmd/util/common_others.go
@@ -86,7 +86,7 @@ func HasSystemd() bool {
 func RunningModuleV2(opt *types.ResetOptions) types.ModuleRunning {
 	osType := GetOSInterface()
 	cloudCoreRunning, err := IsCloudcoreContainerRunning(constants.SystemNamespace, opt.Kubeconfig)
-	if err != nil {
+	if cloudCoreRunning && err != nil {
 		// just log the error, maybe we do not care
 		klog.Warningf("failed to check cloudcore is running: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/util/common_others.go
```diff
@@ -86,7 +86,7 @@ func HasSystemd() bool {
 func RunningModuleV2(opt *types.ResetOptions) types.ModuleRunning {
 	osType := GetOSInterface()
 	cloudCoreRunning, err := IsCloudcoreContainerRunning(constants.SystemNamespace, opt.Kubeconfig)
-	if err != nil {
+	if cloudCoreRunning && err != nil {
```
WillardHu, line 89: This logical expression will cause the original error message to be lost. You may consider other ways to solve this problem.

keadm/cmd/keadm/app/cmd/util/common_others.go
```diff
@@ -95,7 +95,7 @@ func RunningModuleV2(opt *types.ResetOptions) types.ModuleRunning {
 	}
 
 	edgeCoreRunning, err := osType.IsKubeEdgeProcessRunning(KubeEdgeBinaryName)
-	if err != nil {
+	if edgeCoreRunning && err != nil {
```
WillardHu, line 98: Ditto.

keadm/cmd/keadm/app/cmd/util/common_others.go
tangming1996: Thank you for your review. My purpose here is to shield error logs. Cloud logs should not be output when reset is on the edge. Similarly, cloud logs should not be output on the edge. According to the original logic, when the reset operation is performed on the side, the error of the cloud must be output, but this is meaningless and will mislead the user.

keadm/cmd/keadm/app/cmd/util/common_others.go
tangming1996: Ditto.

keadm/cmd/keadm/app/cmd/util/common_others.go
WillardHu: We can optimize the reset command to solve this issue, and you can look at the third task in the issue #5317

WillardHu:
This solution is not very good, now the cloud and the edge reset commands are separated

Issue #5350, opened 2024-01-09
Request Scope Creater Scheme Implementation


Windrow14:
**What would you like to be added/modified**:

`handlers.RequestScope` needs an implementation of its `Creater`.

**Why is this needed**:

After fixing the panic issue described in https://github.com/kubeedge/kubeedge/pull/5313 by adding `fakeTypeConverter` and `fakeFieldManager`, the panic still goes on. So this issue is also a bug report.

This time, panic happens at `Creater`. which is not defined.
```
		result, err := finisher.FinishRequest(ctx, func() (runtime.Object, error) {
			liveObj, err := scope.Creater.New(scope.Kind)
			if err != nil {
				return nil, fmt.Errorf("failed to create new object (Create for %v): %v", scope.Kind, err)
			}
			obj = scope.FieldManager.UpdateNoErrors(liveObj, obj, managerOrUserAgent(options.FieldManager, req.UserAgent()))
			admit = fieldmanager.NewManagedFieldsValidatingAdmissionController(admit)
			...
```

If we set `Creater` to `runtime.NewScheme()` in `func NewRequestScope() *handlers.RequestScope`, panic no longer happens, but CRD such as `Ciliumendpoint` cannot be created.
Following error appears:

Jan 08 16:42:54 edge121 edgecore[609218]: E0108 16:42:54.998857  609218 status.go:71] apiserver received an error that is not an metav1.Status: &errors.errorString{s: "failed to create new object (Create for cilium.io/v2, Kind=Ciliumendpoint): no kind \"Ciliumendpoint\" is registered for version \"cilium.io/v2\" in scheme \"pkg/metamanager/metaserver/kubernetes/scope/scope.go:49\""}: failed to create new object (Create for cilium.io/v2, Kind=Ciliumendpoint):no kind "Ciliumendpoint" is registered for version "cilium.io/v2" in scheme "pkg/metamanager/metaserver/kubernetes/scope/scope.go:49"

So we need an implementation of this `Creater` which can really do creation.

Shelley-BaoYue:
Does this issue occur with the KubeEdge master branch? Or there is also the same problem in v1.15?

Windrow14:
> Does this issue occur with the KubeEdge master branch? Or there is also the same problem in v1.15?

It does occur on master branch. It is introduced after:
```
commit 7c02a3f52090618df852f3184297a91cd00b2d0a
Author: Shelley-BaoYue <baoyue2@huawei.com>
[...]

PR #5313, opened 2023-12-14
Fix Metaserver Panic when Handling POST 2.

Windrow14:
/kind bug

What this PR does / why we need it:

Fix to the edgecore panic issue:
```
12月 13 17:03:06 windrow-PC edgecore[143516]: E1213 17:03:06.923689  143516 runtime.go:77] Observed a panic: runtime error: invalid memory address or nil pointer dereference
12月 13 17:03:06 windrow-PC edgecore[143516]: goroutine 2535 [running]:
12月 13 17:03:06 windrow-PC edgecore[143516]: k8s.io/apiserver/pkg/endpoints/handlers/finisher.finishRequest.func1.1()
12月 13 17:03:06 windrow-PC edgecore[143516]:         /home/windrow/code/kubeedge/vendor/k8s.io/apiserver/pkg/endpoints/handlers/finisher/finisher.go:105 +0xa5
12月 13 17:03:06 windrow-PC edgecore[143516]: panic({0x2a2eee0, 0x4ab6680})
12月 13 17:03:06 windrow-PC edgecore[143516]:         /usr/local/go/src/runtime/panic.go:884 +0x213
12月 13 17:03:06 windrow-PC edgecore[143516]: k8s.io/apimachinery/pkg/util/managedfields/internal.(*FieldManager).Update(0x0, {0x0, 0x0}, {0x332f300, 0xc0004329e8}, {0xc0019168c0, 0xc})
12月 13 17:03:06 windrow-PC edgecore[143516]:         /home/windrow/code/kubeedge/vendor/k8s.io/apimachinery/pkg/util/managedfields/internal/fieldmanager.go:120 +0x50
12月 13 17:03:06 windrow-PC edgecore[143516]: k8s.io/apimachinery/pkg/util/managedfields/internal.(*FieldManager).UpdateNoErrors(0xc00125c1e0?, {0x0?, 0x0?}, {0x332f300, 0xc0004329e8}, {0xc0019168c0?, 0xc001243ba8?})
12月 13 17:03:06 windrow-PC edgecore[143516]:         /home/windrow/code/kubeedge/vendor/k8s.io/apimachinery/pkg/util/managedfields/internal/fieldmanager.go:143 +0x45
12月 13 17:03:06 windrow-PC edgecore[143516]: k8s.io/apiserver/pkg/endpoints/handlers.UpdateResource.func1.1({0x2e99594?, 0x2?}, {0x332f300, 0xc0004329e8}, {0x0, 0x0})
[...]

Issue #4844, opened 2023-07-12
KubeEdge edgecore supports CNI Cilium?


fujitatomoya:
**What happened and what you expected to happen**:

3rd party CNI implementation support is listed as roadmap here, [Roadmap 2021-H1](https://kubeedge.io/en/docs/roadmap/#2021-h1) but we are having several problems once enabling the cilium on edgecore.

**How to reproduce it (as minimally and precisely as possible)**:

after KubeEdge is enabled, [install cilium](https://docs.cilium.io/en/stable/gettingstarted/k8s-install-default/).

**Anything else we need to know?**:

The followings are related issues aligned with this question.

- https://github.com/kubeedge/kubeedge/issues/4819
- https://github.com/kubeedge/kubeedge/issues/4817

Windrow14:
Let me elaborate our team's investigation on using `cilium` with `KubeEdge` so far.

Here is our process of setting up the cluster (`cloudcore` and `edgecore` in use are including some workarounds for problems encountered, I'll show you the details also.):
1. Create the cluster on cloud node with `kubeadm init --node-name pc-windrow`. Copy `/etc/kubernetes/admin.conf` to `~/.kube/config`.
2. Untaint the master node with `kubectl taint node pc-windrow node-role.kubernetes.io/master:NoSchedule-`.
3. Install `cilium` into the cluster with `cilium --encryption wireguard install`.
4. Add `node-role.kubernetes.io/control-plane: ''` and tolerations, so that `cilium-operator` would always be deployed on cloud node. `kubectl edit deployment -n kube-system cilium-operator`.
5. Duplicate `cilium`'s `daemonsets`, one is still for nodes using `kubelet`, including master node. The other is adjusted for nodes using `edgecore`, `--k8s-api-server=127.0.0.1:10550` is added to `cilium-agent`'s argument. Two `daemonsets` have different names and the argument mentioned above, all the other things remain the same.
6. Initialize `KubeEdge` on master node with `sudo keadm init --advertise-address="xxx.xxx.xxx.xxx" --profile version=v1.12.1 --kube-config=/home/windrow/.kube/config`.
7. Add `cloudcore` in `kubectl edit clusterRolebinding cilium` to give it permission of accessing `cilium`'s resources.
8. Enable `dynamiccontroller` by editing `configmap` with `kubectl edit cm -n kubeedge cloudcore`, and deleting the old pod `kubectl delete pod -n kubeedge cloudcore-5876c76687-52xr5`.
[...]

PR #5352, opened 2024-01-10
Fix edgecore's metaserver panic due to nil initializers in request scope.

Windrow14:
**What type of PR is this?**

/kind bug

Sorry that I didn't succeed in fixing panic problem thoroughly in pull request 5313.
This PR fill schemes for elements not initialized in request scope which lead to new panic issues.

Related tickets:
https://github.com/kubeedge/kubeedge/issues/4844
https://github.com/kubeedge/kubeedge/issues/5350
https://github.com/kubeedge/kubeedge/pull/5313

 No.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -43,6 +43,7 @@ linters-settings:
   misspell:
     ignore-words:
       - mosquitto
+      - creater
 
 linters:
   disable-all: true
--- a/edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
@@ -4,10 +4,12 @@ import (
 	"k8s.io/apiextensions-apiserver/pkg/crdserverscheme"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/unstructuredscheme"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/managedfields"
 	"k8s.io/apiserver/pkg/endpoints/handlers"
+	"k8s.io/apiserver/pkg/registry/rest"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/klog/v2"
 	"k8s.io/kube-openapi/pkg/validation/spec"
@@ -18,16 +20,16 @@ import (
 )
 
 func NewRequestScope() *handlers.RequestScope {
-	fakeTypeConverter, err := managedfields.NewTypeConverter(make(map[string]*spec.Schema), false)
+	typeConverter, err := managedfields.NewTypeConverter(make(map[string]*spec.Schema), false)
 	if err != nil {
 		klog.Errorf("Failed to create TypeConverter: %v\n", err)
 		return nil
 	}
-	fakeFieldManager, err := managedfields.NewDefaultFieldManager(
-		fakeTypeConverter,
-		nil,
+	fieldManager, err := managedfields.NewDefaultFieldManager(
+		typeConverter,
+		fakers.NewFakeObjectConvertor(),
 		fakers.NewFakeObjectDefaulter(),
-		nil,
+		unstructuredscheme.NewUnstructuredCreator(),
 		schema.GroupVersionKind{},
 		schema.GroupVersion{},
 		"",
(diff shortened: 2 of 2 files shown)


edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
```diff
@@ -46,26 +48,32 @@ func NewRequestScope() *handlers.RequestScope {
...
-		//Creater:         nil,
-		Convertor: fakers.NewFakeObjectConvertor(),
-		Defaulter: fakers.NewFakeObjectDefaulter(),
-		Typer:     crdserverscheme.NewUnstructuredObjectTyper(),
-		//UnsafeConvertor: nil,
-		Authorizer: fakers.NewAlwaysAllowAuthorizer(),
+
+		StandardSerializers: make([]runtime.SerializerInfo, 0),
+
+		Creater:         unstructuredscheme.NewUnstructuredCreator(),
```
ZhengXinwei-F, line 54: Perhaps we could try dividing it into three dimensions to generate creator, kubernetes, kubeedge, and other resources.

edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
```diff
@@ -46,26 +48,32 @@ func NewRequestScope() *handlers.RequestScope {
...
 		Serializer:     serializer.NewNegotiatedSerializer(),
 		ParameterCodec: scheme.ParameterCodec,
-		//Creater:         nil,
-		Convertor: fakers.NewFakeObjectConvertor(),
-		Defaulter: fakers.NewFakeObjectDefaulter(),
-		Typer:     crdserverscheme.NewUnstructuredObjectTyper(),
-		//UnsafeConvertor: nil,
-		Authorizer: fakers.NewAlwaysAllowAuthorizer(),
+
+		StandardSerializers: make([]runtime.SerializerInfo, 0),
```
ZhengXinwei-F, line 52: You could add some comments to clarify why newly member variable entities is used.

ZhengXinwei-F (commented):
Could you please provide the results of the self-test in the PR description?

ZhengXinwei-F:
I've added the 1.16 milestone tag to this PR.  cc @fujitatomoya @WillardHu @Shelley-BaoYue

edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
Windrow14: When and how should implemented creaters for CRDs to be injected or registered into it? 

I'm confused about how the creaters will work. Should CRDs be synchronized from `cloudcore` and then `metaserver` update the creaters for them according to the definitions?

edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
```diff
@@ -46,26 +48,32 @@ func NewRequestScope() *handlers.RequestScope {
...
+		Defaulter:       fakers.NewFakeObjectDefaulter(),
+		Typer:           crdserverscheme.NewUnstructuredObjectTyper(),
+		UnsafeConvertor: fakers.NewFakeObjectConvertor(),
+		Authorizer:      fakers.NewAlwaysAllowAuthorizer(),
 
 		EquivalentResourceMapper: runtime.NewEquivalentResourceRegistry(),
 
-		TableConvertor: nil,
-		FieldManager:   fakeFieldManager,
+		TableConvertor: rest.NewDefaultTableConvertor(schema.GroupResource{}),
```
fujitatomoya, line 63: Is this also required to change against the panic? could you share the explanation why this is needed?

.golangci.yml
```diff
@@ -43,6 +43,7 @@ linters-settings:
   misspell:
     ignore-words:
       - mosquitto
+      - creater
```
fujitatomoya, line 46: Do we really want to add this to `kubeedge`? originally this structure member name is defined in `vendor/k8s.io/apiserver`, not `kubeedge`?

edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
Windrow14: `TableConvertor` is used in `kubeedge/vendor/k8s.io/apiserver/pkg/endpoints/handlers/response.go`, if it is called while it is `nil`, another panic will happen. I just want to prevent this in advance.

.golangci.yml
Windrow14: Lint thinks `creater` is a misspelling, which should be `creator`. However, even variables in k8s are named with `creater`, so we need lint to ignore this mistake.

edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
Windrow14: And what about convertor and typer? Do they need to be divided also?

edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
Windrow14: In the definition of `type RequestScope struct` in `vendor/k8s.io/apiserver/pkg/endpoints/handlers/rest.go`, there are a few variables not specified in `edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go`. There is a risk that those `nil` variables lead to panic issues when they are called in somewhere else. So I initialize them here as well.

edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
Windrow14: @ZhengXinwei-F Hi, any further idea? Can we merge the fix to the panic first anyway?

fujitatomoya:
@Windrow14 probably it would be better to add `depends on https://github.com/kubeedge/kubeedge/pull/5313` in the PR header clearly, since this is generated by https://github.com/kubeedge/kubeedge/pull/5313

edge/pkg/metamanager/metaserver/kubernetes/scope/scope.go
Windrow14: @fisherxu Hi, could you please have a look at this PR, can we merge the fix to the panic first anyway?

fujitatomoya:
@Shelley-BaoYue thank you for taking care of this. i guess now it is staged to merge, please let us know if anything else needs to be done on our side.

btw, not related to this PR. assigned reviewers are not responsive at all, this is kinda common problem for open source project. i am not sure how much we could help with our knowledge but we would be happy to help review as well.

fujitatomoya:
@WillardHu thanks for the approval.

Shelley-BaoYue:
@Windrow14  This PR need to be cherry-picked to release-1.16, and other previous releases if needed. 😄

Windrow14:
> @Windrow14 This PR need to be cherry-picked to release-1.16, and other previous releases if needed. 😄

https://github.com/kubeedge/kubeedge/pull/5479 @Shelley-BaoYue

PR #5356, opened 2024-01-12
fix some bug in framework

wbc6080:
**What type of PR is this?**
/kind bug

1. In the previous PR, we removed the protocol field, but it was not fixed in the grpcserver/device.go file, resulting in incorrect function parameters and needs to be corrected.

2. When parsing the Property field in the parse/grpc.go file, if dbMethod or pushMethod is empty, the entire Property field will be directly returned as nil. This is unreasonable and will cause other Property field information to be lost. Here we change that if dbMethod or pushMethod is empty, just change this field to nil, which will not affect the parsing of other fields.

xzzq

Fixes #

@cl2017 @WillardHu 

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
@@ -2,6 +2,7 @@ package parse
 
 import (
 	"encoding/json"
+
 	"k8s.io/klog/v2"
 
 	dmiapi "github.com/kubeedge/kubeedge/pkg/apis/dmi/v1beta1"
@@ -147,7 +148,6 @@ func buildPropertiesFromGrpc(device *dmiapi.Device) []common.DeviceProperty {
 			default:
 				klog.Errorf("get DBMethod err: Unsupported database type")
 			}
-
 		}
 		if pptv.PushMethod != nil {
 			//parse pushmethod filed
--- a/staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
@@ -2,8 +2,6 @@ package parse
 
 import (
 	"encoding/json"
-	"errors"
-
 	"k8s.io/klog/v2"
 
 	dmiapi "github.com/kubeedge/kubeedge/pkg/apis/dmi/v1beta1"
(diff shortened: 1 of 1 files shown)


wbc6080:
PTAL @WillardHu @cl2017 thanks

staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
```diff
@@ -133,64 +133,66 @@ func buildPropertiesFromGrpc(device *dmiapi.Device) []common.DeviceProperty {
 			dbMethodName, err = getDBMethodFromGrpc(pptv)
```
WillardHu, line 133: It doesn't look like the function `getDBMethodFromGrpc(..)` is playing a big role here, we can do this directly with the switch:
```golang
switch pptv
case visitor.PushMethod.DBMethod.Influxdb2 != nil:
    ...
case visitor.PushMethod.DBMethod.Redis != nil:
    ...
case visitor.PushMethod.DBMethod.Tdengine != nil:
    ...
default:
    klog.Error("...")

File: staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go [hunk 2]
```diff
@@ -133,64 +133,66 @@ func buildPropertiesFromGrpc(device *dmiapi.Device) []common.DeviceProperty {
...
+					dbconfig = common.DBConfig{
+						TDEngineClientConfig: clientConfig,
+					}
 				}
 			}
+
 		}
 
 		// get pushMethod filed by grpc device instance
 		pushMethodName, err = getPushMethodFromGrpc(pptv)
```
WillardHu, line 177: Ditto

staging/src/github.com/kubeedge/mapper-framework/pkg/grpcserver/device.go
```diff
@@ -51,7 +51,7 @@ func (s *Server) RegisterDevice(ctx context.Context, request *dmiapi.RegisterDev
 	}
 
 	deviceInstance.PProtocol = protocol
-	s.devPanel.UpdateDev(&model, deviceInstance, &protocol)
+	s.devPanel.UpdateDev(&model, deviceInstance)
```
cl2017, line 54: I have made these changes in https://github.com/kubeedge/kubeedge/pull/5292, do you need to rebase?

cl2017 (commented):
I have made these changes in https://github.com/kubeedge/kubeedge/pull/5292, do you need to rebase?

staging/src/github.com/kubeedge/mapper-framework/pkg/grpcserver/device.go
wbc6080: I have rebased them @cl2017

staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go
wbc6080: Done

PR #5375, opened 2024-01-19
fix the package version in mapper-framework

wbc6080:
**What type of PR is this?**
/kind bug

update the package version in mapper-framework to generate mapper project correctly

Fixes #

```release-note

```


fisherxu (commented):
/kind bug

wbc6080:
In order to match the latest versions of the kubeedge main library and mapper-framework, the version referenced by mapper is upgraded.
PTAL @cl2017 @WillardHu thanks a lot

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/go.mod
```diff
@@ -8,8 +8,8 @@ require (
 	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/influxdata/influxdb-client-go/v2 v2.13.0
-	github.com/kubeedge/kubeedge v1.15.0-beta.0.0.20240110111122-c44efa2389b9
-	github.com/kubeedge/mapper-framework v0.0.0-20240111132845-9ad4739bda66
+	github.com/kubeedge/kubeedge v1.15.0-beta.0.0.20240118080528-4237f6f6805e
```
WillardHu, line 11: Does the kubeedge version need to be upgraded to v1.16?

wbc6080: I import the package according to the method defined in kubernetes like https://github.com/kubernetes/api/blob/master/go.mod#L10 by commit, and v1.16 has not been officially tagged yet, so it seems that it cannot be quoted.

WillardHu: You can try command `go get github.com/kubeedge/kubeedge@f69bb68` to upgrade kubeedge version. The 
f69bb68 is the commit id of the tag `v1.16.0-bate.0`.

wbc6080: Indeed, it can automatically change to v1.16.0-beta.0 after running like this. I have made modifications. Thank you for your advice.

PR #5229, opened 2023-11-29
Construction of cloudcore upgrading capability

WillardHu:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

We can easy upgrade cloud components through the keadm.

**Which issue(s) this PR fixes**:

Fixes #5131

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5383, opened 2024-01-22
fix keadm flag in windows

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

Fix keadm join flag in windows for we have modified these flag in linux #5229 , otherwise some errors will occur like :

image


fisherxu (commented):
/lgtm
/approve

PR #5374, opened 2024-01-19
For backward compatibility of the edgecore trigger the upgrade command

WillardHu:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

If the edgecore version is earlier then v1.16, It still upgrade with the command `keadm upgrade ...`. So we need to think about backward compatibility for that. If the upgrade command no three-level command, keadm upgrades the edge components.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5387, opened 2024-01-24
Fix keadm upgrade command

WillardHu:
**What type of PR is this?**

/kind bug

Sorry, the function reference I used earlier(https://github.com/kubeedge/kubeedge/pull/5374) caused a problem with the scope of the upgrade options. In the RunE, the upgradeOptions pointer variable refers to the upgradeOptions in the NewEdgeUpgrade(), it's just an instance initialized with default values.

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/upgrade.go
+++ b/keadm/cmd/keadm/app/cmd/upgrade.go
@@ -13,6 +13,7 @@ Specify whether to upgrade the cloud or the edge through three-level commands.
 If no three-level command, it upgrades edge components.`
 )
 
+// NewUpgradeCommand creates a upgrade command instance and returns it.
 func NewUpgradeCommand() *cobra.Command {
 	cmds := &cobra.Command{
 		Use:   "upgrade",
@@ -20,15 +21,15 @@ func NewUpgradeCommand() *cobra.Command {
 		Long:  upgradeLongDescription,
 	}
 
-	edgecmd := edge.NewEdgeUpgrade()
-
 	// Used for backward compatibility of the edgecore trigger the upgrade command
 	upgradeOptions := edge.NewUpgradeOptions()
+	cmds.RunE = func(cmd *cobra.Command, args []string) error {
+		return upgradeOptions.Upgrade()
+	}
 	edge.AddUpgradeFlags(cmds, upgradeOptions)
-	cmds.RunE = edgecmd.RunE
 
 	// Register three-level commands
-	cmds.AddCommand(edgecmd)
+	cmds.AddCommand(edge.NewEdgeUpgrade())
 	cmds.AddCommand(cloud.NewCloudUpgrade())
 	return cmds
 }
--- a/keadm/cmd/keadm/app/cmd/edge/upgrade.go
+++ b/keadm/cmd/keadm/app/cmd/edge/upgrade.go
@@ -51,7 +51,7 @@ func NewEdgeUpgrade() *cobra.Command {
 		Long:  "Upgrade edge components. Upgrade the edge node to the desired version.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// upgrade edgecore
-			return upgradeOptions.upgrade()
+			return upgradeOptions.Upgrade()
 		},
 	}
 
@@ -68,7 +68,8 @@ func NewUpgradeOptions() *UpgradeOptions {
 	return opts
 }
 
-func (up *UpgradeOptions) upgrade() error {
+// Upgrade handles upgrade command logic
+func (up *UpgradeOptions) Upgrade() error {
 	// get EdgeCore configuration from edgecore.yaml config file
 	data, err := os.ReadFile(up.Config)
 	if err != nil {


WillardHu:
**Test Data**
```bash
root@ubuntu:~# keadm version
version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.0-beta.0.3+0d13ab32d8df59-dirty", GitCommit:"0d13ab32d8df596b67bafd3d466c414c0988ad87", GitTreeState:"dirty", BuildDate:"2024-01-24T03:41:20Z", GoVersion:"go1.20.10", Compiler:"gc", Platform:"linux/amd64"}
root@ubuntu:~#
root@ubuntu:~# keadm upgrade --upgradeID upgrade-example \
>   --historyID fff83cd0-de7b-48f1-ace8-0063190020d5 \
>   --fromVersion v1.14.3 --toVersion v1.16.0 \
>   --config /etc/kubeedge/config/edgecore.yaml \
>   --image docker.m.daocloud.io/kubeedge/installation-package:v1.16.0
I0124 04:17:02.455487 1619321 upgrade.go:159] Begin to download version v1.16.0 edgecore
Pulling docker.m.daocloud.io/kubeedge/installation-package:v1.16.0 ...
Successfully pulled docker.m.daocloud.io/kubeedge/installation-package:v1.16.0
I0124 04:17:03.522153 1619321 upgrade.go:234] upgrade process start
edgecore is stopped
I0124 04:17:07.900459 1619321 join_others.go:264]
I0124 04:17:07.900637 1619321 join_others.go:265] KubeEdge edgecore is running, For logs visit: journalctl -u edgecore.service -xe
E0124 04:17:07.920390 1619321 common.go:700] report result &{200 OK 200 HTTP/1.1 1 1 map[Content-Length:[2] Content-Type:[text/plain; charset=utf-8] Date:[Wed, 24 Jan 2024 04:17:07 GMT]] 0xc0003782c0 2 [] false false map[] 0xc0000c8400 0xc0006b78c0}
```

The EdgeCore upgrade was successful.

keadm/cmd/keadm/app/cmd/upgrade.go
```diff
@@ -13,22 +13,23 @@ Specify whether to upgrade the cloud or the edge through three-level commands.
...
 
-	edgecmd := edge.NewEdgeUpgrade()
-
 	// Used for backward compatibility of the edgecore trigger the upgrade command
 	upgradeOptions := edge.NewUpgradeOptions()
+	cmds.RunE = func(cmd *cobra.Command, args []string) error {
+		return upgradeOptions.Upgrade()
+	}
 	edge.AddUpgradeFlags(cmds, upgradeOptions)
-	cmds.RunE = edgecmd.RunE
```
wbc6080, line 30: Sorry, I'm a little confused about the modifications here. Why do we need to assign a separate value to cmd.RunE? It seems that the original method can also achieve the same effect.

WillardHu: That's what I thought before, so I assigned `edgecmd.RunE` to `cmd.RunE`. If we look closely, we can see that the upgradeOptions is not passed to the function edgecmd.RunE() in any way, so where do the upgradeOptions come from, that's the scope of the closure. The upgradeOptions is assigned to the closure function RunE() after initialization in the NewEdgeUpgrade(), so the closure RunE use the upgradeOptions in the NewEdgeUpgrade()
```golang
func NewEdgeUpgrade() *cobra.Command {
    opts := NewUpgradeOptions()  // address: 0x01
    RunE: func(...) {
        opts.Upgrade() // opts address: 0x01
    }
}

func NewCloudUpgrade() *cobra.Command {
    edgeCmd := NewEdgeUpgrade()
    opts := newCloudUpgradeOptions() // address: 0x02
    RunE: func(...) {
        opts.Upgrade() // opts address: 0x02
    }
    RunE = edgeCmd.RunE = func(...){ opts.Upgrade() /* opts address: 0x01 */ }
}
```

wbc6080: LGTM.
I understand the reason for your modification here. It is true that RunE in the original code uses the UpgradeOptions defined in NewEdgeUpgrade() instead of in NewUpgradeCommand(). This bug is really hard to find. 
I have another question, what will this bug perform when it is actually run? Did some subsequent module update the UpgradeOptions in NewUpgradeCommand(), but the update did not take effect because we use the Options of NewEdgeUpgrade() before?

WillardHu: Yes, edge nodes that use the default upgrade command `keadm upgrade` will fail to upgrade, because the UpgradeOptions instantce is just a default value created by the function edge.NewUpgradeOptions()

wbc6080: OK, thanks for your detailed answer!
/lgtm
cc @Shelley-BaoYue @fisherxu

wbc6080 (approved):

WillardHu:
PTAL, thanks @Shelley-BaoYue @fisherxu

Shelley-BaoYue (commented):
please rebase the current master branch to fix the container e2e CI test for PR #5443  has been merged

WillardHu:
> please rebase the current master branch to fix the container e2e CI test for PR #5443 has been merged

I have rebased the latest master branch, but there is a few e2e tasks still fail...

wbc6080 (commented):
#5443 mainly adds a retry mechanism when downloading the container runtime to avoid network problems causing e2e test failure. But I feel that the reason here is the failure caused by actually executing the e2e test task after downloading the container runtime.

Shelley-BaoYue (commented):
/lgtm

fisherxu (commented):
/approve

PR #5196, opened 2023-11-23
Supports installing edgecore without installing the CNI plugin.

luomengY:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
When running the container installation-package:v1.15.0 during keadm join, we use the hostNetwork mode to avoid users manually installing CNI plugins before this, which is also not logical.（https://github.com/kubeedge/kubeedge/issues/5195 ）

**Which issue(s) this PR fixes**:

Fixes https://github.com/kubeedge/kubeedge/issues/5195

**Special notes for your reviewer**:
@fisherxu  @Shelley-BaoYue 
**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5397, opened 2024-01-29
fix custom image is cp resource failed

thinkeng:
**What type of PR is this?**
** 当镜像重命名或者不使用docker 镜像仓库的镜像，或者使用私有仓库镜像时，会报镜像无法加载的错误**
`1201 21:35:36.352097 935665 join.go:184] 5. Copy resources from the image to the management directory E1201 21:35:38.461963 935665 remote_runtime.go:176] "RunPodSandbox from runtime service failed" err="rpc error: code = Unknown desc = failed to setup network for sandbox \"f9713b839351af9c989aa02b36aba4152db30f0034041f40a64352744976e001\": plugin type=\"bridge\" failed (add): netplugin failed with no error message: signal: segmentation fault" Error: edge node join failed: copy resources failed: rpc error: code = Unknown desc = failed to setup network for sandbox "f9713b839351af9c989aa02b36aba4152db30f0034041f40a64352744976e001": plugin type="bridge" failed (add): netplugin failed with no error message: signal: segmentation fault execute keadm command failed: edge node join failed: copy resources failed: rpc error: code = Unknown desc = failed to setup network for sandbox "f9713b839351af9c989aa02b36aba4152db30f0034041f40a64352744976e001": plugin type="bridge" failed (add): netplugin failed with no error message: signal: segmentation fault`

Fixes #

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/image.go
+++ b/keadm/cmd/keadm/app/cmd/util/image.go
@@ -115,6 +115,7 @@ func (runtime *CRIRuntime) PullImage(image string, authConfig *runtimeapi.AuthCo
 // CopyResources copies binary and configuration file from the image to the host.
 // The same way as func (runtime *DockerRuntime) CopyResources
 func (runtime *CRIRuntime) CopyResources(edgeImage string, files map[string]string) error {
+	edgeImage = convertCRIImage(edgeImage)
 	psc := &runtimeapi.PodSandboxConfig{
 		Metadata: &runtimeapi.PodSandboxMetadata{
 			Name:      KubeEdgeBinaryName,


WillardHu:
This looks like a problem caused by the CNI plugin. The PR(https://github.com/kubeedge/kubeedge/pull/5196) fixed the the network issue. What version of the keadm are you using?

thinkeng:
> This looks like a problem caused by the CNI plugin. The PR(#5196) fixed the the network issue. What version of the keadm are you using?

Both v1.14.2 and v15.0 have encountered , it doesn't seem to be the same problem

WillardHu:
> Both v1.14.2 and v15.0 have encountered , it doesn't seem to be the same problem

You can try version v1.15.2 or v1.16.0. If the issue is the image path, the message that the image does not exist is returned, but from the error message you provided we can see that there is a issue with the network plugin:
image

fisherxu:
> You can try version v1.15.2 or v1.16.0. If the issue is the image path, the message that the image does not exist is returned, but from the error message you provided we can see that there is a issue with the network plugin:


+1, looks like it's the cni plugin issue

PR #5409, opened 2024-02-02
normalize the hostname in fucntion GetHostname

### LINKED ISSUE #5391: EdgeCore has different behavior in generating node name compared to kubelet.
**What happened**:

EdgeCore and kubelet generates different node names.

EdgeCore and kubelet generates the same node.

Install edgecore on a server with uppercase hostname.

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - edgecore version (e.g. `edgecore --version`): 15.1
  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

Kubelet uses the lower case of hostname for node name.

```golang
// https://github.com/kubernetes/kubernetes/blob/b98e47dd80166935117f890b437c8e30a91da173/staging/src/k8s.io/component-helpers/node/util/hostname.go
func GetHostname(hostnameOverride string) (string, error) {
	hostName := hostnameOverride
	if len(hostName) == 0 {
		nodeName, err := os.Hostname()
		if err != nil {
			return "", fmt.Errorf("couldn't determine hostname: %w", err)
		}
		hostName = nodeName
	}

	// Trim whitespaces first to avoid getting an empty hostname
	// For linux, the hostname is read from file /proc/sys/kernel/hostname directly
	hostName = strings.TrimSpace(hostName)
	if len(hostName) == 0 {
		return "", fmt.Errorf("empty hostname is invalid")
	}

	return strings.ToLower(hostName), nil
}
```

However, edge core generates node name by hostname without a `ToLower()` step.

```golang
// https://github.com/kubeedge/kubeedge/blob/master/pkg/util/util.go
func GetHostname() string {
	hostnameOverride, err := os.Hostname()
	if err != nil {
		return constants.DefaultHostnameOverride
	}
	msgs := validation.ValidateNodeName(hostnameOverride, false)
	if len(msgs) > 0 {
		return constants.DefaultHostnameOverride
	}
	return hostnameOverride
}
```

If there is a node with hostname `Abc`, the node name generated by kubelet is `abc` while edgecore name it `default-edge-node`.

The difference may lead to some compatibility issues when some nodes are migrated from kubelet to kubeedge.

spambot000:
Because hostname is case-insensitive, normalize it before coverting to a DNS subdomain name.

**What type of PR is this?**

/kind bug

Kubeedge generates default node name by hostname if it is a valid DNS subdomain. The default node name will be `default-edge-node` if hostname contains uppercase letters, which may lead to compatibility issues when nodes are migrated from kubelet to kubeedge.

**Which issue(s) this PR fixes**：

Fixes #5391

yes
```release-note
The default name of node which has a uppercase hostname is changed from `default-edge-node` to lowercase hostname.
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	utilnet "k8s.io/apimachinery/pkg/util/net"
+	nodeutil "k8s.io/component-helpers/node/util"
 	"k8s.io/kubernetes/pkg/apis/core/validation"
 
 	"github.com/kubeedge/kubeedge/common/constants"
@@ -131,7 +132,7 @@ func SpliceErrors(errors []error) string {
 
 // GetHostname returns a reasonable hostname
 func GetHostname() string {
-	hostnameOverride, err := os.Hostname()
+	hostnameOverride, err := nodeutil.GetHostname("")
 	if err != nil {
 		return constants.DefaultHostnameOverride
 	}
--- a/go.mod
+++ b/go.mod
@@ -65,6 +65,7 @@ require (
 	golang.org/x/text v0.13.0
 	gopkg.in/square/go-jose.v2 v2.6.0
 	gopkg.in/yaml.v3 v3.0.1
+	k8s.io/component-helpers v0.0.0
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3
 )
 
@@ -259,7 +260,6 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	k8s.io/cloud-provider v0.27.7 // indirect
 	k8s.io/cluster-bootstrap v0.27.7 // indirect
-	k8s.io/component-helpers v0.0.0 // indirect
 	k8s.io/controller-manager v0.0.0 // indirect
 	k8s.io/csi-translation-lib v0.27.7 // indirect
 	k8s.io/dynamic-resource-allocation v0.0.0 // indirect


pkg/util/util.go
```diff
@@ -135,6 +135,9 @@ func GetHostname() string {
 	if err != nil {
 		return constants.DefaultHostnameOverride
 	}
+	// Hostname should be normailzied because it's case-insensitive
```
Shelley-BaoYue, line 138: Thx for your contributions!  Can we call `GetHostName` func of kubernetes mentioned in the issue https://github.com/kubeedge/kubeedge/issues/5391 here?   In this way, if there is any change in Kubernetes, it can be directly synchronized.

spambot000: Yes, it's the better implementation.

pkg/util/util.go
```diff
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	utilnet "k8s.io/apimachinery/pkg/util/net"
+	nodeutil "k8s.io/component-helpers/node/util"
```
Shelley-BaoYue, line 29: Because the import lib has been updated, you need to execute `make verify` and commit the changes to update go.mod and vendor.

spambot000: Fixed. I ran the command under a directory that didn't contains the `.git` dir.

Shelley-BaoYue (commented):
/lgtm 
cc @WillardHu @fisherxu

PR #5416, opened 2024-02-06
Add http-port config in mapper

### LINKED ISSUE #5415: Allow users to customize the port number of mapper pull api
Allow users to customize the port number of mapper pull api
**Why is this needed**:
Now the port number of the device data pull API provided by mapper for users is the immutable port 7777. If the port is already occupied, it will affect the normal provision of data pull service by mapper. Therefore, features need to be added to allow users to customize port numbers

wbc6080:
**What type of PR is this?**
/kind feature

As described in #5415 , the port of pull device data provided by mapper cannot be set currently , so port conflicts may occur. This PR allows users to set the port number themselves in the mapper configuration file to avoid port occupation.

Fixes #5415 

```release-note

```


fisherxu (commented):
/approve

PR #5484, opened 2024-03-25
fix retrytimes in imageprepulljob

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

When retryTimes equals zero(which the default value is zero), we will skip the imagepull step and report successful status, which is not what we want.  We need to pull images one time when it equals zero. 

Also fix the wrong field in crd-sample, timeoutSecondsOnEachNode -> timeoutSeconds

### DIFF
--- a/build/crd-samples/operations/imageprepulljob.yaml
+++ b/build/crd-samples/operations/imageprepulljob.yaml
@@ -10,5 +10,5 @@ spec:
       - busybox:latest
     nodes:
       - edgenode1 # Need to replaced with your own node name
-    timeoutSecondsOnEachNode: 300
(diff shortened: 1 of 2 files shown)


WillardHu:
Please retest CI checks

Shelley-BaoYue:
> Please retest CI checks

I have pushed another PR https://github.com/kubeedge/kubeedge/pull/5485 to fix the CI

WillardHu:
> I have pushed another PR #5485 to fix the CI
Ok, I approved this PR

fisherxu (commented):
/approve

PR #5485, opened 2024-03-25
update controller-runtime setup-envtest version to handle the CI failure 

Shelley-BaoYue:
**What type of PR is this?**

/kind failing-test

integration test always fail with the error as below:

image

So we cannot use the latest version of controller-runtime setup-envtest, because it refers to a Go version that is too new. 

More details and disscussion: https://github.com/kubernetes-sigs/controller-runtime/issues/2720


Shelley-BaoYue:
And I guess this PR may also need to be cherry-picked to previous version

PR #5493, opened 2024-03-26
upgrade slsa-framework version to v1.10.0

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

upgrade slsa-framework version to v1.10.0 to fix the error when we draft a new release, the error is like: 

image

The details about this error: https://github.com/slsa-framework/slsa-github-generator/issues/3031


fisherxu (commented):
/approve
/assign @WillardHu

PR #5499, opened 2024-03-28
[WIP]Add delay to fix docker e2e test

wbc6080:
**What type of PR is this?**
/kind bug

The current failure rate of docker e2e testing is high. The possible reason is that the test is performed immediately after starting the kubeedge cluster, but some components in the cluster have not yet been started. Therefore, this PR adds a delay to avoid this problem.

Fixes #

```release-note

```


Shelley-BaoYue (commented):
/lgtm

fisherxu (commented):
/approve

PR #5503, opened 2024-04-01
fix panic when using controller-runtime: log.SetLogger(...) was never called

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

When enable the requireAuthorization the feature, cloudcore will break with panic:

Due to controller-runtime v0.15.0 bump and changes made in https://github.com/kubernetes-sigs/controller-runtime/pull/2317, you must set a logger for controller-runtime. We have our own, so we will set controller-runtime's logger to Null.


cloud/pkg/policycontroller/policycontroller.go
```diff
@@ -54,6 +56,7 @@ func NewAccessRoleControllerManager(ctx context.Context, kubeCfg *rest.Config) (
 }
 
 func setupControllers(ctx context.Context, mgr manager.Manager) error {
+	ctrlruntimelog.SetLogger(logr.New(ctrlruntimelog.NullLogSink{}))
```
WillardHu, line 59: 1. Would it be better to put it in New() func?
2. We can use the klog instance to set the logger of the runtime-controller.
3. Do other runtime-controllers also need to set logger?
```golang
func NewAccessRoleControllerManager() {
    ...
    ctrlruntimelog.SetLogger(klog.Background())
}
```

PR #5490, opened 2024-03-26
fix the error "failed to create serviceaccount access" when metaserver switches to HTTPS.

luomengY:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
ce99a02c27894d084be12d137bc3fbc
62780d45e990f09a8f3b896bfd5fce6

**Which issue(s) this PR fixes**:

Fixes  https://github.com/kubeedge/kubeedge/issues/5486

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5511, opened 2024-04-06
fix rbac for featureGates requireAuthorization

Links2004:
**What type of PR is this?**
/kind bug

fix the RBACs for the featureGate `requireAuthorization`.
the cloudcore will have a lot of RBAC errors in the log without this the.

```release-note
NONE
```


wbc6080 (commented):
Sorry, it seems that this part of the work has been submitted by other developers in #5490 . Can you help review whether #5490 is the same as what you want to modify?

Links2004:
yes, looks like it tries to fix the same problem only without the seperation by `resources` and with out the `get` verb

PR #5513, opened 2024-04-07
support module restarting feature

### LINKED ISSUE #5134: Support auto-restarting for edge-site modules

Support **auto-restarting** for edge-site modules. 

When a module is panicking or failing, judge whether the whole edgecore should exit immediately. 

On remote edge-site devices released to users, we cannot control what users do and what the running environment is. We don't want core module crash and the whole program stop working such as edged. If it's a recoverable fail, edgecore should not exit and restart the module only.

**Why is this needed**:

We are aiming at Kubeedge in production.

Edge-site nodes run in diverse bad environments:

- Bad network or even no network to cloud
- Not enough power
- Partly enabled linux core functions(for optimization by device producer)
- Unpredicted user operations

On edge-site nodes, we could **never** ensure that they are running i ideal environments. We can't suppose that edge-site nodes always have good connectivity with cloud, or pre-conditions of Edgecore are always prepared correctly.

Here comes three facts:
 
1. Edge-site devices can be **powered down and up frequently**. For example: vehicles, white goods, toys.
2. Edge-site devices are running **unpredictably**. They could be up at any time, and in any condition such as no-network environment, no-systemd or systemd partly enabled operating systems as the producer needs.
3. Edge-site devices are **operated by users** with no knowledge of programming. Once some module works incorrectly, edgecore exits and the operator treats it as the device fails completely. 
    - Rebooting must be users' first choice, but it never works.
    - They may ask for engineers' help while few device engineers have knowledge of container, Kubernetes and Kubeedge.

**Our problem**:

As an example, we have met a problem that:

1. **Everytime** on system boot
2. Edgecore starts.
3. Module `edged` calls `os.Exit(1)` because `Kubelet` fails to connect container runtime.
4. Edgecore crashes.
5. Containerd starts.
6. With a no-restart systemd, edgecore never starts again. All pods(static or not) won't run.

In this case, `edgecore` should be able to recover `edged` by itself. If `edged` was restarted by `edgecore`, the program won't crash.

Here we have some discussions and suggesstions, we were glad if Kubeedge community already has proper solutions to our problem.

1. Support passing `context` when a module is started by BeeHive, which will help BeeHive control modules' lifecycle. Because modules are started by BeeHive, we think that BH is responsble for controlling modules' lifecycle. Currently, module call global context directly and BH does not know if module goroutine exits.

Currently: 

```go
// beehive/pkg/core/module.go
type Module interface {
    Name() string
    Group() string
    Start()
    Enable() bool
}

// beehive/pkg/core/core.go
    // ...
    go moduleKeeper(name, module, m)
```

What we want:

```go
// beehive/pkg/core/module.go
type Module interface {
    Name() string
    Group() string
    Start(context.Context)
    Enable() bool
}

// beehive/pkg/core/core.go or something other
    // ...
    for someCondition {
        mCtx, mCancel := context.WithCancel(beehiveContext.GetContext())
        go module.Start(mCtx)
        <-mCtx.Done()
    }
```

2. Support monitoring modules' running state and expose it so that beehive or other modules would be able to know what to do next. We can use context as modules' running state.

What we want:

```go
// some module for our purpose
// maybe beehive can do it.
for _, module := range allModules {
    if IsAlive(module) {
        // do some things...
    } else {
        // do other things...
    }
}
if !IsAllAlive() {
    // do some other things...
}
```

3. Support `select` from BeeHive channels. Currently, a module cannot shutdown gracefully if channels are empty. It stucks waiting for `beehiveContext.Receive` and never select from `beehiveContext.Done`.

Currently:

```go
// edge/pkg/edged/edged.go and other modules
func (e *edged) syncPod(podCfg *config.PodConfig) {
    time.Sleep(10 * time.Second)

    for {
        select {
        case <-beehiveContext.Done():
            klog.Warning("Sync pod stop")
            return
        default:
        }
        // stuck for empty channel
        result, err := beehiveContext.Receive(e.Name())
        if err != nil {
            klog.Errorf("failed to get pod: %v", err)
            continue
        }
        // ...
    }
}
```

What we want:

```go
// edge/pkg/edged/edged.go and other modules
func (e *edged) syncPod(podCfg *config.PodConfig) {
    for {
        select {
        case <-beehiveContext.Done():
            klog.Warning("Sync pod stop")
            // do some cleaning works if needed
            return
        case result := <-beehiveContext.MustGetChannel(e.Name()):
                // ...
        }
    }
}
```

micplus:
**What type of PR is this?**

/kind feature

To achieve the goal of module restarting on edge site.

1. Modify BeeHive to support restarting for local modules
2. Add feature gate `moduleRestart`
3. Remove additional goroutine of MetaManager's main loop
4. Modify `edged.syncPod` to support early exit
5. Enable module-restarting in edgecore's main function

Fixes #5134

Alpha version

```release-note
NONE
```

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -30,11 +30,11 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha2"
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha2/validation"
 	"github.com/kubeedge/kubeedge/pkg/features"
+	kefeatures "github.com/kubeedge/kubeedge/pkg/features"
 	"github.com/kubeedge/kubeedge/pkg/util"
 	"github.com/kubeedge/kubeedge/pkg/util/flag"
 	utilvalidation "github.com/kubeedge/kubeedge/pkg/util/validation"
 	"github.com/kubeedge/kubeedge/pkg/version"
-	kefeatures "github.com/kubeedge/kubeedge/pkg/features"
 )
 
 // NewEdgeCoreCommand create edgecore cmd
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -34,6 +34,7 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/util/flag"
 	utilvalidation "github.com/kubeedge/kubeedge/pkg/util/validation"
 	"github.com/kubeedge/kubeedge/pkg/version"
+	kefeatures "github.com/kubeedge/kubeedge/pkg/features"
 )
 
 // NewEdgeCoreCommand create edgecore cmd
@@ -123,7 +124,7 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 			registerModules(config)
 
 			// enable module auto-restart feature
-			if config.FeatureGates[options.FeatureModuleAutoStart] {
+			if kefeatures.DefaultFeatureGate.Enabled(kefeatures.ModuleRestart) {
 				core.EnableModuleRestart()
 			}
 
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -30,7 +30,6 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha2"
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha2/validation"
 	"github.com/kubeedge/kubeedge/pkg/features"
-	kefeatures "github.com/kubeedge/kubeedge/pkg/features"
 	"github.com/kubeedge/kubeedge/pkg/util"
 	"github.com/kubeedge/kubeedge/pkg/util/flag"
 	utilvalidation "github.com/kubeedge/kubeedge/pkg/util/validation"
@@ -124,7 +123,7 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 			registerModules(config)
 
 			// enable module auto-restart feature
-			if kefeatures.DefaultFeatureGate.Enabled(kefeatures.ModuleRestart) {
+			if features.DefaultFeatureGate.Enabled(features.ModuleRestart) {
 				core.EnableModuleRestart()
 			}
 
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -121,6 +121,12 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 			}
 
 			registerModules(config)
+
+			// enable module auto-restart feature
+			if config.FeatureGates[options.FeatureModuleAutoStart] {
+				core.EnableModuleRestart()
+			}
+
 			// start all modules
 			core.Run()
 		},
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -60,8 +60,8 @@ import (
 	metaclient "github.com/kubeedge/kubeedge/edge/pkg/metamanager/client"
 	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/dao"
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha2"
-	"github.com/kubeedge/kubeedge/pkg/version"
 	kefeatures "github.com/kubeedge/kubeedge/pkg/features"
+	"github.com/kubeedge/kubeedge/pkg/version"
 )
 
 // GetKubeletDeps returns a Dependencies suitable for lite kubelet being run.
@@ -145,9 +145,6 @@ func (e *edged) Start() {
 				os.Exit(1)
 			}
 			klErrChan <- err
-			// send empty message to wakeup syncPod loop
-			nilMsg := model.NewMessage("").BuildRouter(e.Name(), e.Group(), "", "")
-			beehiveContext.Send(modules.EdgedModuleName, *nilMsg)
 		}
 	}()
 	e.syncPod(e.KubeletDeps.PodConfig, klErrChan)
@@ -254,9 +251,12 @@ func (e *edged) syncPod(podCfg *config.PodConfig, kubeletErrChan <-chan error) {
 		case err := <-kubeletErrChan:
 			klog.Errorf("Start edged failed, err: %v", err)
 			return
-		case <-kubeletWaiter.C:
-			kubeletInited = true
 		default:
+			select {
+			case <-kubeletWaiter.C:
+				kubeletInited = true
+			default:
+			}
 			if !kubeletInited {
 				time.Sleep(time.Second)
 				continue
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -31,7 +31,6 @@ import (
 	"os"
 	"reflect"
 	"strconv"
-	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/protobuf/jsonpb"
@@ -135,14 +134,17 @@ func (e *edged) Start() {
 		}
 	}
 
+	klErrChan := make(chan error, 1)
 	go func() {
 		err := DefaultRunLiteKubelet(e.context, e.KubeletServer, e.KubeletDeps, e.FeatureGate)
 		if err != nil {
-			klog.Errorf("Start edged failed, err: %v", err)
-			os.Exit(1)
+			klErrChan <- err
+			// send empty message to wakeup syncPod loop
+			nilMsg := model.NewMessage("").BuildRouter(e.Name(), e.Group(), e.namespace+"/"+model.ResourceTypePod, "")
+			beehiveContext.Send(modules.EdgedModuleName, *nilMsg)
 		}
 	}()
-	e.syncPod(e.KubeletDeps.PodConfig)
+	e.syncPod(e.KubeletDeps.PodConfig, klErrChan)
 }
 
 // newEdged creates new edged object and initialises it
@@ -220,9 +222,7 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	return ed, nil
 }
 
-func (e *edged) syncPod(podCfg *config.PodConfig) {
-	time.Sleep(10 * time.Second)
-
+func (e *edged) syncPod(podCfg *config.PodConfig, klErrchan <-chan error) {
 	//when starting, send msg to metamanager once to get existing pods
 	info := model.NewMessage("").BuildRouter(e.Name(), e.Group(), e.namespace+"/"+model.ResourceTypePod,
 		model.QueryOperation)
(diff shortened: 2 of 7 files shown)


edge/cmd/edgecore/app/server.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha2"
 	"github.com/kubeedge/kubeedge/pkg/apis/componentconfig/edgecore/v1alpha2/validation"
 	"github.com/kubeedge/kubeedge/pkg/features"
+	kefeatures "github.com/kubeedge/kubeedge/pkg/features"
```
Shelley-BaoYue, line 33: repeated import with Line32

edge/pkg/edged/edged.go
```diff
@@ -220,20 +225,44 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	return ed, nil
 }
 
-func (e *edged) syncPod(podCfg *config.PodConfig) {
-	time.Sleep(10 * time.Second)
+func (e *edged) syncPod(podCfg *config.PodConfig, kubeletErrChan <-chan error) {
+	startWaiter := time.NewTimer(10 * time.Second)
```
Shelley-BaoYue, line 229: Could u plz explain more about the modification of Line 229-249?

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -42,7 +43,12 @@ func StartModules() {
 		beehiveContext.AddModule(&m)
 		beehiveContext.AddModuleGroup(name, module.module.Group())
 
-		go moduleKeeper(name, module, m)
+		if module.remote {
```
Shelley-BaoYue, line 46: Is the module restart determined based on the remote?  Sorry, I may need some time to get the definition of remote.

edge/cmd/edgecore/app/server.go
micplus: Sorry, I've missed it.

edge/pkg/edged/edged.go
micplus: If we use `time.Sleep`, when the goroutine which runs kubelet failed without calling `os.Exit(1)`, `syncPod` cannot returns immediately.

To make `syncPod` return once kubelet fails, use a Timer `startWaiter` to provide a channel instead of just sleep the goroutine.

With the channel `startWaiter.C`, it covers three conditions:
 
1. global context exits, `syncPod` must return
2. kubelet fails, `syncPod` must return
3. `syncPod` has waited for 10s, `syncPod` is ready to sync pods

I've misunderstood godoc for timer and removed the unnecessary draining channel operation here.

staging/src/github.com/kubeedge/beehive/pkg/core/module.go
```diff
@@ -30,6 +32,7 @@ func init() {
 type ModuleInfo struct {
 	contextType string
 	remote      bool
+	active      bool
```
wbc6080, line 35: What is the significance of the newly added **active** field? Does it mean that a module was started normally?

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -84,3 +90,46 @@ func moduleKeeper(name string, moduleInfo *ModuleInfo, m common.ModuleInfo) {
...
+		}
+		m.active = false
+		klog.Errorf("module %s exited, will restart in %ds", m.module.Name(), int(backoffDuration.Seconds()))
+	}
+
+	for {
+		func() {
+			defer afterFunc()
+			m.active = true
+			m.module.Start()
```
wbc6080, line 118: I'm a little confused here. The function here seems to be to start a module. If the startup fails, afterFunc will be executed and an error message will be output. But if the module is successfully started, will afterFunc still be executed? But there seems to be no handling of successful startup in afterFunc. Did I understand something wrong?

staging/src/github.com/kubeedge/beehive/pkg/core/module.go
micplus: Yes, it means a module is started and running normally. In this PR it's `true` if `module.Start()` doesn't exits.

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
micplus: Currently we are focusing on `EdgeCore` and this feature could only be implemented for edge-site modules so we care less about `CloudCore`. 

On edge site, the `Start()` function of `devicetwin`, `edged`, `edgehub`, `edgestream`, `eventbus`, `servicebus` will loop forever. Specially, `metamanager` starts two goroutine and returns. But the function `runMetaManager()` tells that it is Main loop at L425, `metamanager/process.go`. So this PR treats it as the main loop of the module and removes the `go func`.

Ideally, if a goroutine is started by some module, the module does have responsibility to handle the goroutine to prevent producing orphans. But it need plenty of work to modify all modules, especially for `CloudCore`. 

With restarting of cloud-site modules disabled, it works as normal and `afterFunc` will not be called, which post-processes module exiting.

In this PR, cloud-site restarting cannot be enabled, so it will not ruin cloud-site logic for starting modules. To make it further, maybe a `<-beehiveContext.Done()` could be added to `Start()` to keep a main goroutine of modules.

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -84,3 +90,46 @@ func moduleKeeper(name string, moduleInfo *ModuleInfo, m common.ModuleInfo) {
...
+		case <-ctx.Done():
+			klog.Infof("module %s shutdown", m.module.Name())
+			return
+		case <-time.After(backoffDuration):
+		}
+
+		if backoffDuration < 30*time.Second {
+			backoffDuration *= 2
+			if backoffDuration > 30*time.Second {
+				backoffDuration = 30 * time.Second
```
fisherxu, line 131: We need to set the Max retry time, so process can exit after that, and the `backoffDuration` also should be configurable for users.

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
```diff
@@ -84,3 +90,46 @@ func moduleKeeper(name string, moduleInfo *ModuleInfo, m common.ModuleInfo) {
...
+
+	ctx := beehiveContext.GetContext()
+	backoffDuration := time.Second
+
+	// do if module exits
+	afterFunc := func() {
+		if r := recover(); r != nil {
+			klog.Errorf("module %s panicking: %v", m.module.Name(), r)
+		}
+		m.active = false
```
fisherxu, line 110: Where else is this field `active` used? Seems only here and L117

edge/pkg/edged/edged.go
```diff
@@ -135,14 +136,18 @@ func (e *edged) Start() {
...
-			os.Exit(1)
+			if !kefeatures.DefaultFeatureGate.Enabled(kefeatures.ModuleRestart) {
+				klog.Errorf("Start edged failed, err: %v", err)
+				os.Exit(1)
+			}
+			klErrChan <- err
 		}
 	}()
-	e.syncPod(e.KubeletDeps.PodConfig)
+	e.syncPod(e.KubeletDeps.PodConfig, klErrChan)
```
fisherxu, line 150: If run `DefaultRunLiteKubelet` failed, why not return before `e.syncPod` directly？

edge/pkg/edged/edged.go
```diff
@@ -135,14 +136,18 @@ func (e *edged) Start() {
...
 	}
 
+	klErrChan := make(chan error, 1)
 	go func() {
 		err := DefaultRunLiteKubelet(e.context, e.KubeletServer, e.KubeletDeps, e.FeatureGate)
 		if err != nil {
-			klog.Errorf("Start edged failed, err: %v", err)
-			os.Exit(1)
+			if !kefeatures.DefaultFeatureGate.Enabled(kefeatures.ModuleRestart) {
+				klog.Errorf("Start edged failed, err: %v", err)
```
fisherxu, line 144:
```suggestion
				klog.Errorf("Failed to start edged, err: %v", err)
```

edge/pkg/edged/edged.go
```diff
@@ -220,20 +225,37 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
...
-func (e *edged) syncPod(podCfg *config.PodConfig) {
-	time.Sleep(10 * time.Second)
+func (e *edged) syncPod(podCfg *config.PodConfig, kubeletErrChan <-chan error) {
+	startWaiter := time.NewTimer(10 * time.Second)
+	defer startWaiter.Stop() 
+
+	// block until ready to sync pods
+	select {
+	case <-beehiveContext.Done():
+		klog.Warning("Sync pod stop")
```
fisherxu, line 235:
```suggestion
		klog.Warning("Stop sync pod")
```

edge/pkg/edged/edged.go
```diff
@@ -220,20 +225,37 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
...
+		return
+	case err := <-kubeletErrChan:
+		klog.Errorf("Start edged failed, err: %v", err)
```
fisherxu, line 238:
```suggestion
		klog.Errorf("Failed to start edged, err: %v", err)
```

fisherxu (commented):
Overall looks good, thanks for the great work @micplus @ttlv

edge/pkg/metamanager/process.go
```diff
@@ -419,21 +419,19 @@ func (m *metaManager) process(message model.Message) {
...
+		default:
 		}
-	}()
+		msg, err := beehiveContext.Receive(m.Name())
+		if err != nil {
+			klog.Errorf("get a message %+v: %v", msg, err)
+			continue
+		}
+		klog.V(2).Infof("get a message %+v", msg)
+		m.process(msg)
```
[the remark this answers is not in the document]
micplus: Yes, it will rerun.

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
micplus: We can make it configurable in the next phase of development. 

Now it's a feature in development. So we haven't add configurations yet.

staging/src/github.com/kubeedge/beehive/pkg/core/core.go
micplus: It hasn't been used now. I'll remove it.

edge/pkg/edged/edged.go
micplus: Maybe we can move the pre-condition of `edged.syncPod` to `edged.Start` directly

edge/pkg/edged/edged.go
```diff
@@ -221,18 +242,18 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 }
 
 func (e *edged) syncPod(podCfg *config.PodConfig) {
-	time.Sleep(10 * time.Second)
```
fisherxu, line 245: Delete this line to fix the lint error

micplus: Oh sorry, I will fixt it now

Shelley-BaoYue (commented):
/lgtm

PR #5514, opened 2024-04-07
Mapper supports video streaming data reporting

wbc6080:
**What type of PR is this?**
/kind feature

This PR implements mapper’s feature of supporting reporting of streaming data. Mainly includes the following parts:

1. Change the device model api definition to allow the data type of streaming data

2. Add the function of saving frame and saving video to the mapper-framework data processing function.

3. Change the logic of pushing device data in mapper-framework. When the type is streaming data, the device data is no longer pushed to the database and user applications, but processed by saving frames or saving videos.

Fixes ##5401

```release-note

```


wbc6080:
PTAL,thanks @luomengY @WillardHu @fisherxu

luomengY:
Suggest adding a switch control parameter for starting/stopping stream data collection in the next version.

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/handler.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+
+	// parse streamConfig data from device visitorConfig
+	var streamConfig StreamConfig
+	visitorConfigData, err := json.Marshal(visitorConfig.VisitorConfigData)
+	err = json.Unmarshal(visitorConfigData, &streamConfig)
+	if err != nil {
+		klog.Errorf("Unmarshal streamConfigs error: %v", err)
+		return err
```
WillardHu, line 53: Wrap the error message and return it, make troubleshooting easy.
```golang
return fmt.Errorf("unmarshal streamConfigs error: %v", err)
```

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/img.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/sailorvii/goav/swscale"
+	"k8s.io/klog/v2"
+)
+
+// GenFileName generate file name with current time. Formate f<year><month><day><hour><minute><second><millisecond>.<format>
+func GenFileName(dir string, format string) string {
+	return fmt.Sprintf("%s/f%s.%s", dir, time.Now().Format(time.RFC3339Nano), format)
+}
+
+func save(frame *avutil.Frame, width int, height int, dir string, format string) {
```
WillardHu, line 39: Does this function not return an error?

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/img.go
```diff
@@ -0,0 +1,246 @@
...
+	// Find the first video stream
+	streamIndex := -1
+	for i := 0; i < int(pFormatContext.NbStreams()); i++ {
+		if pFormatContext.Streams()[i].CodecParameters().AvCodecGetType() == avformat.AVMEDIA_TYPE_VIDEO {
+			streamIndex = i
+			break
+		}
+	}
+	if streamIndex == -1 {
+		return fmt.Errorf("couldn't find video stream")
```
WillardHu, line 142: You can use the function errors.New() when without formatting.

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/img.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	// Determine required buffer size and allocate buffer
+	numBytes := uintptr(avcodec.AvpictureGetSize(avcodec.AV_PIX_FMT_YUVJ420P, pCodecCtx.Width(),
+		pCodecCtx.Height()))
+	buffer := avutil.AvMalloc(numBytes)
+
+	// Assign appropriate parts of buffer to image planes in pFrameRGB
+	// Note that pFrameRGB is an AVFrame, but AVFrame is a superset
+	// of AVPicture
+	avp := (*avcodec.Picture)(unsafe.Pointer(pFrameRGB))
```
WillardHu, line 178: Should we consider the case where type is invalid?

staging/src/github.com/kubeedge/mapper-framework/README.md
```diff
@@ -42,6 +42,30 @@ mapper
...
+If you want to debug locally first, you can also compile and run the mapper code directly:
+```shell
+  go run -tags Nostream cmd/main.go --v <log level,like 3> --config-file <path to config yaml>
+```
 
+The above operations are mainly for mappers that do not process streaming data. If you need to enable 
+the streaming data processing function, please execute the following command:
+```shell
+  go run -tags stream cmd/main.go --v <log level,like 3> --config-file <path to config yaml>
+```
```
fisherxu, line 65: It is not recommended to write too much here, just write it in the document.

staging/src/github.com/kubeedge/mapper-framework/README.md
```diff
@@ -42,6 +42,30 @@ mapper
...
 # Where does it come from?
-mapper-framework is synced from https://github.com/kubeedge/kubeedge/tree/master/staging/src/github.com/kubeedge/mapper-framework. Code changes are made in that location, merged into kubeedge and later synced here.
+mapper-framework is synced from https://github.com/kubeedge/kubeedge/tree/master/staging/src/github.com/kubeedge/mapper-framework.
+Code changes are made in that location, merged into kubeedge and later synced here.
+# Things you should *NOT* do
+https://github.com/kubeedge/kubeedge/tree/master/staging/src/github.com/kubeedge/mapper-framework is synced to kubeedge/mapper-framework. 
+All changes must be made in the former. The latter is read-only.
```
fisherxu, line 71: It shoud not need this, and keep consistent with other sub-projects :)

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile
```diff
@@ -7,7 +7,7 @@ ENV GO111MODULE=on \
 
 COPY . .
 
-RUN CGO_ENABLED=0 GOOS=linux go build -o main cmd/main.go
+RUN CGO_ENABLED=0 GOOS=linux go build -tags Nostream -o main cmd/main.go
```
fisherxu, line 10: Is there any way to control the feature? Like add a startup parameter to control the mdule to start or not, hdyt @WillardHu @luomengY @wbc6080

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/handler.go
wbc6080: Done

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/img.go
wbc6080: I have add it

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/img.go
wbc6080: I don't quite understand your question. Is there a problem with converting *avutil.Frame to *avcodec.Picture? The official documentation here shows that *avcodec.Picture is a superset of *avutil.Frame. You can refer to [this](https://github.com/giorgisio/goav/blob/master/example/tutorial01.go#L128)

staging/src/github.com/kubeedge/mapper-framework/README.md
wbc6080: I've simplified the README file but still kept some necessary steps to use it

staging/src/github.com/kubeedge/mapper-framework/README.md
wbc6080: Done

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile
wbc6080: I have considered using a switch to set whether to use this feature, but even if we add a switch, it seems that this part of the code will still be compiled, only not run this part, so the third-party package of stream data processing will still be used. In order to prevent this part of the code from being compiled, what I finally found was the method of conditional compilation. As for other methods, I haven't found them yet.

wbc6080:
> Suggest adding a switch control parameter for starting/stopping stream data collection in the next version.

Yes, we can add it with the feature about mapper supporting data writing in v1.18

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile
WillardHu: One idea is that handle the logic(determine which files to copy) when the mapper is generated from the template.

wbc6080: Oh I see, very good idea, I will try to implement it like this @WillardHu

wbc6080:
I have made modifications according to everyone's opinions, mainly:
1. Removed conditional compilation.
2. Modified the mapper generate script to generate different mapper projects depending on whether it is used to process stream data.
3. Provide Dockerfile files for two situations
PTAL,thanks @WillardHu @fisherxu

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/handler_nostream.go
```diff
@@ -0,0 +1,33 @@
+/*
```
fisherxu, line 1: Is this file necessary? The file `handler_nostream.go` under `stream` dir seems a bit unreasonable

wbc6080: This file is required, because stream.Handler needs to be called at mapper data plane capability like picture following. For mappers that do not require stream data processing, if this file is removed, an error will occur during compilation because stream.Handler cannot be found. If for name reasons, I can modify the mapper generation script and name this file to handler.go in the generated mapper project.

device

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/handler.go
```diff
@@ -0,0 +1,66 @@
...
+		err = SaveFrame(streamURI.(string), streamConfig.OutputDir, streamConfig.Format, streamConfig.FrameCount, streamConfig.FrameInterval)
+	case "saveVideo":
+		err = SaveVideo(streamURI.(string), streamConfig.OutputDir, streamConfig.Format, streamConfig.FrameCount, streamConfig.VideoNum)
+	default:
+		err = fmt.Errorf("cannot find the processing method for the corresponding Property %s of the stream data", twin.PropertyName)
+	}
+	if err == nil {
+		klog.V(2).Infof("Successfully processed streaming data by %s", twin.PropertyName)
+	}
+	return err
```
fisherxu, line 65:
```suggestion
	if err != nil {
	        return err
	}
	klog.V(2).Infof("Successfully processed streaming data by %s", twin.PropertyName)
	return nil
```

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/handler.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return fmt.Errorf("Unmarshal streamConfigs error: %v", err)
+	}
+
+	switch twin.PropertyName {
+	// Currently, the function of saving frames and saving videos is built-in according to the configuration.
+	// Other functions can be expanded here.
+	case "saveFrame":
+		err = SaveFrame(streamURI.(string), streamConfig.OutputDir, streamConfig.Format, streamConfig.FrameCount, streamConfig.FrameInterval)
+	case "saveVideo":
```
fisherxu, line 57: The PropertyName name is passed from Device API? If so, It's necessary to add some comments in api and add some const

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile_stream
```diff
@@ -0,0 +1,18 @@
...
+
+RUN apt-get update && \
+    apt-get install -y bzip2 upx-ucl gcc-aarch64-linux-gnu libc6-dev-arm64-cross gcc-arm-linux-gnueabi libc6-dev-armel-cross libva-dev libva-drm2 libx11-dev libvdpau-dev libxext-dev libsdl1.2-dev libxcb1-dev libxau-dev libxdmcp-dev yasm
+
+RUN curl -sLO https://ffmpeg.org/releases/ffmpeg-4.1.6.tar.bz2 && \
+    tar -jx --strip-components=1 -f ffmpeg-4.1.6.tar.bz2 &&  \
+    ./configure &&  make && \
+    make install
+
+RUN GOOS=linux go build -o main cmd/main.go
```
fisherxu, line 18: Is it necessary to separate into two dockerfile? also cc @WillardHu

staging/src/github.com/kubeedge/mapper-framework/hack/make-rules/generate.sh
```diff
@@ -27,6 +33,13 @@ function entry() {
     exit 1
   fi
   cp -r "${ROOT_DIR}/_template/mapper" "${mapperPath}"
+  if [ "${buildMethod}" = "stream" ]; then
+    rm "${mapperPath}/data/stream/handler_nostream.go"
+  fi
+
+  if [ "${buildMethod}" = "nostream" ]; then
+      rm "${mapperPath}/data/stream/handler.go" "${mapperPath}/data/stream/img.go" "${mapperPath}/data/stream/video.go"
+  fi
```
fisherxu, line 42: If add new file for stream, this part will be hard to maintain.

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/handler.go
wbc6080: Done

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/stream/handler.go
wbc6080: Done

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/Dockerfile_stream
wbc6080: Since the stream mapper needs to download additional dependencies, it takes about five minutes to build the image, so I feel it would be better to separate it without affecting the construction of other mapper images.

staging/src/github.com/kubeedge/mapper-framework/hack/make-rules/generate.sh
wbc6080: I modified it to only keep the handler_stream.go file and renamed this file to handler.go after build

fisherxu (commented):
/approve
Looks good for the first version.
/assign @WillardHu

PR #5515, opened 2024-04-07
fix edgeapp validate error

tangming1996:
**What type of PR is this?**
/kind bug

https://github.com/kubeedge/kubeedge/issues/5497

Fixes https://github.com/kubeedge/kubeedge/issues/5497

```release-note

```


WillardHu:
Please add some unit-tests.

WillardHu:
The Admission is not a required component right now, and other processing options need to be considered. It is recommended to complete the proposal first, and then do the code implementation after community discussion.

Issue #5516, opened 2024-04-08
admission deployment run error


tangming1996:
**What happened**:
When running the admission deployment, the error log is as follows:
Error: unknown flag: --alsologtostderr
Usage:
  admission [flags]

Admission flags:

      --ca-cert-file string           File containing the x509 Certificate for HTTPS.
      --kubeconfig string             Path to kubeconfig file with authorization and master location information.
      --master string                 The address of the Kubernetes API server (overrides any value in kubeconfig)
      --port int32                    the port used by admission-controller-server. (default 443)
      --tls-cert-file string          File containing the default x509 Certificate for HTTPS. (CA cert, if any, concatenated after server cert).
      --tls-private-key-file string   File containing the default x509 private key matching --tls-cert-file.
      --webhook-namespace string      The namespace of this webhook (default "kubeedge")
      --webhook-service-name string   The name of this admission service (default "kubeedge-admission-service")

Global flags:

  -h, --help                           help for admission
      --log-flush-frequency duration   Maximum number of seconds between log flushes (default 5s)
  -v, --v Level                        number for the log level verbosity
      --version version[=true]         Print version information and quit
      --vmodule moduleSpec             comma-separated list of pattern=N settings for file-filtered logging (only works for the default text log format)

**What you expected to happen**:
Run successfully

Catherine-monk:
I already fix this bug. https://github.com/kubeedge/kubeedge/pull/5517

tangming1996:
> I already fix this bug. #5517

@WillardHu @Shelley-BaoYue Please review it.

PR #5517, opened 2024-04-08
fix admission deployment run error

Catherine-monk:
**What type of PR is this?**
/kind bug

https://github.com/kubeedge/kubeedge/issues/5516

Fixes https://github.com/kubeedge/kubeedge/issues/5516

### DIFF
--- a/build/admission/deployment.yaml
+++ b/build/admission/deployment.yaml
@@ -30,7 +30,6 @@ spec:
             - --ca-cert-file=/admission.local.config/certificates/ca.crt
             - --webhook-namespace=kubeedge
             - --webhook-service-name=kubeedge-admission-service
-            - --alsologtostderr
             - --port=443
             - -v=4
             - 2>&1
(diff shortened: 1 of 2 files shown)


Shelley-BaoYue (commented):
Would you like to help remove this line in edge-mark https://github.com/kubeedge/kubeedge/blob/master/build/edgemark/hollow_edge_node_template.yaml#L25

Catherine-monk:
> Would you like to help remove this line in edge-mark https://github.com/kubeedge/kubeedge/blob/master/build/edgemark/hollow_edge_node_template.yaml#L25
I already remove this line of the file.

build/admission/deployment.yaml
```diff
@@ -30,7 +30,6 @@ spec:
             - --ca-cert-file=/admission.local.config/certificates/ca.crt
             - --webhook-namespace=kubeedge
             - --webhook-service-name=kubeedge-admission-service
-            - --alsologtostderr
```
fisherxu, line 33: If we use klog, there should be `alsologtostderr` in options, maybe there're have some errors when initiating klog? @tangming1996 @Catherine-monk

tangming1996: image

Yes, I see that the documentation originally supported klog, but it may have been removed later, so here we delete the alsologtostderr parameter from the startup command, otherwise the service will fail to start.

PR #5530, opened 2024-04-11
fix upload codecov failed: update codecov to v4

### LINKED ISSUE #5520: Unit test failed because of the token of CodeCov
**Which jobs are failing**:

Main CI WorkFlow / Unit test, integration test edge

Upload coverage to Codecov

three days ago to present

**Reason for failure**:

It seems that there was an error when uploading the code to CodeCov. There seems to be a problem with the secret key.
codecov失败

There has been a similar problem in the[ issue](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954) of CodeCov. The solution is to **add in the Codecov upload token**. But we seem to have added tokens before, so I think it is a problem like expiration of the secret key. Maybe we can re-add the token according to the steps in the [official documentation](https://docs.codecov.com/docs/codecov-uploader#upload-token), as shown in the picture, and see if this problem can be solved.
上传秘钥

Shelley-BaoYue:
**What type of PR is this?**

/kind failing-test

fix upload codecov failed: update codecov to v4   

image

more details: https://github.com/codecov/codecov-action/issues/1359

Fixes #5520

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -142,7 +142,7 @@ jobs:
         # Prevent running from the forked repository that doesn't need to upload coverage.
         # In addition, running on the forked repository would fail as missing the necessary secret.
         if: ${{ github.repository == 'kubeedge/kubeedge' }}
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
           # Even though token upload token is not required for public repos,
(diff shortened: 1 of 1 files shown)


wbc6080 (commented):
LGTM
Thank you for putting forward this solution. I mistakenly thought it was a problem with the secret token before.  @WillardHu @fisherxu

PR #5560, opened 2024-04-24
Keadm Tool Enhancement

fuchendou:
**The PR is about to add new flags to Keadm for custom configuration support for EdgeCore. Keadm needs to support a new flag to dynamically parse new requirements in representation of string and modify the member variables where exists in the YAML files accordingly.**
**Here is thought of process design:**
image

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/set.go
+++ b/keadm/cmd/keadm/app/cmd/util/set.go
@@ -1,0 +1,338 @@
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


wbc6080 (commented):
Thanks for your submission! But the DCO needs to be fixed, you can modify it as shown in the picture below. In addition, when submitting a PR, the description of the PR needs to be submitted according to the fields defined in the template.
dco

fuchendou:
> Thanks for your submission! But the DCO needs to be fixed, you can modify it as shown in the picture below. In addition, when submitting a PR, the description of the PR needs to be submitted according to the fields defined in the template. dco

luomengY:
> **The PR is about to add new flags to Keadm for custom configuration support for EdgeCore. Keadm needs to support a new flag to dynamically parse new requirements in representation of string and modify the member variables where exists in the YAML files accordingly.**
> **Here is thought of process design**

It is recommended to place this flowchart separately in a proposal.

PR #5582, opened 2024-04-28
Tool that supports automatically generating API documentation for KubeEdge

greensong:
**1. Generate OpenAPI Definitions with openapi-gen:**
   Use the openapi-gen tool to generate Go template code containing OpenAPI definitions based on annotation information. By adding a specific annotation `+k8s:openapigen=true` in the `doc.go` file, openapi-gen scans all types under that package and generates OpenAPI definitions for them, stored in the `zz_generated.openapi.go` file.

**2. Generate OpenAPI Specification:**
   Write `generateswagger.go` to reference the generated OpenAPI definitions (`zz_generated.openapi.go`) and generate the OpenAPI specification (`swagger.json`). The `swagger.json` file contains all the OpenAPI definition information for the API server.
   
***The contents of `generateswagger.go` are as follows:***
* Create Schema objects for type registration and object instantiation.
* Create a REST Mapper Mapper for conversion between GVR (GroupVersionResource) and GVK (GroupVersionKind).
* Add specific mappings for various KubeEdge resources.
* Set information for the OpenAPI specification.

***The specific process is as follows:***
* Run “**go run auto_generate_openapi.go**” to call the openapi-gen tool and generate OpenAPI definitions.
* After generating the OpenAPI definitions, run “**go run generateswagger.go**” to create the OpenAPI specification.


WillardHu:
Thank you for your contribution, you need to check with DCO first.
Refer https://github.com/kubeedge/kubeedge/pull/5582/checks?check_run_id=24349847302

hack/tools/swagger/generateswagger.go
```diff
@@ -0,0 +1,121 @@
...
+	generatedopenapi "github.com/kubeedge/kubeedge/pkg/generated/openapi"
+
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/klog/v2"
+	"k8s.io/kube-openapi/pkg/common"
+	"k8s.io/kube-openapi/pkg/validation/spec"
+)
+
+func main() {
```
WillardHu, line 30: These source codes of a openapi generation tool that does not fit under the hack/ directory, is there a better choice?

hack/tools/swagger/generate-openapi/auto_generate_openapi.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func main() {
```
WillardHu, line 11: I think a shell script would be more appropriate then the gofile for execute openapi-gen command. BTW, we can use the compiled binary openapi-gen.

PR #5600, opened 2024-05-10
support parse PKCS8 private key

### LINKED ISSUE #5599: PCKS8 private key cannot be parsed
support parse PKCS8 private key

**Why is this needed**:
PCKS8 private key cannot be parsed

https://github.com/kubeedge/kubeedge/blob/58bf767f8b763f9e77cf237b9488adfcc8b601be/cloud/pkg/cloudhub/servers/httpserver/server.go#L242-L250

https://github.com/kubeedge/kubeedge/blob/58bf767f8b763f9e77cf237b9488adfcc8b601be/cloud/pkg/cloudhub/servers/httpserver/certutil.go#L85-L88

biningo:
**What type of PR is this?**

CKS8 private key cannot be parsed

Fixes #5599


wbc6080 (commented):
Can you explain in detail under what circumstances you need to parse other types of private keys? If we use ECPrivateKey by default, seems we don't need to parse other types?

biningo:
> Can you explain in detail under what circumstances you need to parse other types of private keys? If we use ECPrivateKey by default, seems we don't need to parse other types?

Hi @wbc6080 <br>
If the CA certificate does not exist in the cloud, cloudcore will automatically generate the CA private key and CA certificate. As you said, the auto-generated private key is EC format.

KubeEdge supports users to generate CA private key and CA certificate by themselves, if users use PKCS1 format or PKCS8 format private key, it will fail to parse.
https://github.com/kubeedge/kubeedge/blob/58bf767f8b763f9e77cf237b9488adfcc8b601be/cloud/pkg/cloudhub/servers/httpserver/server.go#L265-L301

cloud/pkg/cloudhub/servers/httpserver/certutil.go
```diff
@@ -126,3 +125,23 @@ func NewCertFromCa(cfg *certutil.Config, caCert *x509.Certificate, serverKey cry
...
+func ParseX509PrivateKey(der []byte) (crypto.Signer, error) {
+	caKeyEc, ecErr := x509.ParseECPrivateKey(der)
+	if ecErr == nil {
+		return caKeyEc, nil
+	}
+
+	caKeyPKCS1, pkcs1Err := x509.ParsePKCS1PrivateKey(der)
+	if pkcs1Err == nil {
+		return caKeyPKCS1, nil
+	}
```
wbc6080, line 139: Is there a way to distinguish the encryption type of the key that needs to be parsed? Or to put it another way, if the PKCS1 secret key calls the ParseECPrivateKey method, an error will be reported, right?

biningo: yeah
If the private key is not in the correct format, it will return err.
Parsing the private key format is the same as parsing the private key.

forsaken628: We can distinguish the encryption type of the key from caCert.PublicKeyAlgorithm.

forsaken628: [tls.X509KeyPair](https://github.com/golang/go/blob/go1.22.3/src/crypto/tls/tls.go#L245) is better. but look at [here](https://github.com/kubeedge/kubeedge/blob/v1.17.0/cloud/pkg/cloudhub/config/config.go#L44)

PR #5603, opened 2024-05-12
Bump artifact action version to v4

ErikJiang:
**What type of PR is this?**
/kind feature

The artifact-related action version has been updated to v4, and [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates) has been added to perform a weekly check for updates to the action version in workflows. If the bot detects that an action version is outdated, it will automatically submit a pull request for updating the version.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: github-actions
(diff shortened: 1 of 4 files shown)


wbc6080 (commented):
Is the reason for upgrading the version because the original version will be deprecated, or because there are some bugs?

ErikJiang:
@wbc6080  I noticed that the artifact component has relevant deprecation warnings.

image
image

ErikJiang:
@chendave @fisherxu PTAL 😊

Shelley-BaoYue (commented):
There are still some warning like: 
image

Shouldn't they be upgraded together?

ErikJiang:
@Shelley-BaoYue 😀 I believe we can entrust the task of updating the next version of the action to Dependabot. However, if you think it's necessary, I can also take care of it.

PR #5609, opened 2024-05-13
auto

Catherine-monk:
### Introduction
A tool that automatically generates API documentation for kubeedge.

### Implementation Method
1. Generate OpenAPI definitions using openapi-gen: The `openapi-gen` tool generates Go template code containing OpenAPI definitions from comment information. By adding a specific annotation `+k8s:openapigen=true` in the `doc.go` file, `openapi-gen` will scan all types in the package to generate OpenAPI definitions, which are stored in the `zz_generated.openapi.go` file.

2. Generate OpenAPI Specification:
Write `generateswagger.go` referencing the generated OpenAPI definitions (`zz_generated.openapi.go`) to create the OpenAPI specification (`swagger.json`). The `swagger.json` file contains all the OpenAPI definition information of the apiserver.

Contents of `generateswagger.go` include:
- Create a Schema object for type registration and object instantiation.
- Create a REST Mapper for converting between GVR (Group Version Resource) and GVK (Group Version Kind).
- Add specific mappings for various Kubeedge resources.
- Set the information for the OpenAPI specification.


wbc6080 (commented):
It seems that there is a problem with the CI test of make verify. You can run hack/update-codegen.sh and push again.

PR #5610, opened 2024-05-14
Clean duplicate code in redis

wbc6080:
**What type of PR is this?**
/kind cleanup

When adding device data to the redis database of mapper-framework, it will be divided into two situations: whether the order set exists or not. But in fact, the code is exactly the same in both cases, so the duplicate code is cleaned up in this PR

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/redis/client.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/redis/client.go
@@ -69,32 +69,15 @@ func (d *DataBaseConfig) AddData(data *common.DataModel) error {
 	// The key to construct the ordered set, here DeviceName is used as the key
 	klog.V(1).Infof("deviceName:%s", data.DeviceName)
 	// Check if the current ordered set exists
-	exists, err := RedisCli.Exists(ctx, data.DeviceName).Result()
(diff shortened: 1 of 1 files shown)


staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/redis/client.go
```diff
@@ -69,33 +69,16 @@ func (d *DataBaseConfig) AddData(data *common.DataModel) error {
 	// The key to construct the ordered set, here DeviceName is used as the key
 	klog.V(1).Infof("deviceName:%s", data.DeviceName)
 	// Check if the current ordered set exists
-	exists, err := RedisCli.Exists(ctx, data.DeviceName).Result()
+	deviceData := "TimeStamp: " + strconv.FormatInt(data.TimeStamp, 10) + " PropertyName: " + data.PropertyName + " data: " + data.Value
+	// Add data to ordered set. If the ordered set does not exist, it will be created.
+	_, err = RedisCli.ZAdd(ctx, data.DeviceName, &redis.Z{
```
WillardHu, line 74: Does the key need a namespace? Do we need to add another recognizable prefix?

wbc6080: Yes, the prefix here needs to be replaced by deviceID. I will submit another PR to modify the influx、redis、mysql 、tdengine uniformly.

Shelley-BaoYue (commented):
/approve

PR #5622, opened 2024-05-21
fix: 修复 token过期

imneov:
**What type of PR is this?**

/kind bug

In the original getTokenLocally logic of version 1.13:

Retrieve the token locally.
If the token is expired, delete the token from the local cache.
If an error is detected, fetch the token from the cloud, completing the token refresh process.
In this logic, if there is a network outage and the token is expired, the token will be deleted from the local cache, but it cannot be fetched from the cloud, causing the Pod to fail to start.

Therefore, I made the following changes:

Moved the expiration logic out of getTokenLocally.
The expired token will only be deleted if getTokenRemotely is successful.
In case getTokenRemotely fails, return the expired token.

Fixes #

```release-note

```


fisherxu (commented):
@imneov Thanks for the fixing, and is it possible to fix in master branch first?

imneov:
OK, do I need to submit a new PR?

imneov:
> @imneov Thanks for the fixing, and is it possible to fix in master branch first?

@fisherxu   The PR #5625 is fix in  master branch

WillardHu:
This can be a temporary solution, but it's not a good one. If metaserver starts using serviceaccount authentication, getting an expired token won't make apiserver dependent Pods work

PR #5623, opened 2024-05-21
 Image PrePull Feature Enhancement Code

hyp4293:
**What type of PR is this?**
/kind feature

Implement the override function for image and secret in multiple mirror repositories.

Fixes #

```release-note

```


Shelley-BaoYue (commented):
/hold
API definition needs to continue to be discussed

pkg/apis/operations/v1alpha1/imageprepull_types.go
```diff
@@ -61,6 +61,7 @@ type ImagePrePullJobList struct {
 type ImagePrePullJobSpec struct {
 	// ImagePrepullTemplate represents original templates of imagePrePull
 	ImagePrePullTemplate ImagePrePullTemplate `json:"imagePrePullTemplate,omitempty"`
+	ImagePrePullOverrides []ImagePrePullTemplate `json:"imagePrePullOverrides,omitempty"`
```
Shelley-BaoYue, line 64: I don't think it's elegant to use template directly in override. If the API is written like this, why don't users just write multiple imagePrepull jobs?

Issue #4441, opened 2022-11-24
Exited container can not be up when edge node in offline condition


shao77622:
**What happened**:

When in off-line condition, edgecore does not recreate pod which was exited, even when reboot node many times.
The only way is to make node on-line, then pod will be up successfully by edgecore.

[root@kubeedge1 ~]# docker ps -a
CONTAINER ID        IMAGE                COMMAND             CREATED             STATUS                       PORTS               NAMES
b226c485a2b1        kubeedge/pause:3.1   "/pause"            23 hours ago        Exited (255) 9 minutes ago                       k8s_POD_redis-3.2.9-n8zlv_edge_d1bd7b12-91b1-4b61-b61b-c79dc7ce8713_1

**What you expected to happen**:

In offline condition, edgecore could manage pod all the time,  making node truely edge autonomy.

**How to reproduce it (as minimally and precisely as possible)**:

offline  and then shutdown for a day ,and  reboot. 

**Anything else we need to know?**:

edgecore logs

`11月 24 22:09:23 kubeedge1 edgecore[2267]: E1124 22:09:23.784636    2267 ws.go:78] dial websocket error(dial tcp 47.98.143.107:10000: connect: connection refused), response message:
11月 24 22:09:23 kubeedge1 edgecore[2267]: E1124 22:09:23.784737    2267 websocket.go:90] Init websocket connection failed dial tcp 47.98.143.107:10000: connect: connection refused
11月 24 22:09:27 kubeedge1 edgecore[2267]: E1124 22:09:27.725619    2267 controller.go:144] failed to ensure lease exists, will retry in 7s, error: unmarshal message to lease failed, err: json: cannot unmarshal array into Go value of type client.LeaseResp
11月 24 22:09:29 kubeedge1 edgecore[2267]: E1124 22:09:29.788215    2267 ws.go:78] dial websocket error(dial tcp 47.98.143.107:10000: connect: connection refused), response message:
11月 24 22:09:29 kubeedge1 edgecore[2267]: E1124 22:09:29.788241    2267 websocket.go:90] Init websocket connection failed dial tcp 47.98.143.107:10000: connect: connection refused
11月 24 22:09:32 kubeedge1 edgecore[2267]: E1124 22:09:32.799665    2267 metaclient.go:117] send sync message edge/podpatch/redis-3.2.9-n8zlv failed, error:timeout to get response for message 4a72ee87-5f52-4cb5-adda-c5e0c1b2d053, retries: 2
[...]

PR #5625, opened 2024-05-22
fix: token expiration

imneov:
**What type of PR is this?**

/kind bug

In the original getTokenLocally logic of version 1.13:

1. Retrieve the token locally.
1. If the token is expired, delete the token from the local cache.
1. If an error is detected, fetch the token from the cloud, completing the token refresh process.

In this logic, if there is a network outage and the token is expired, the token will be deleted from the local cache, but it cannot be fetched from the cloud, causing the Pod to fail to start.

Therefore, I made the following changes:

1. Moved the expiration logic out of getTokenLocally.
1. The expired token will only be deleted if getTokenRemotely is successful.
1. In case getTokenRemotely fails, return the expired token.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/client/serviceaccount.go
+++ b/edge/pkg/metamanager/client/serviceaccount.go
@@ -118,15 +118,6 @@ func getTokenLocally(name, namespace string, tr *authenticationv1.TokenRequest) 
 		klog.Errorf("unmarshal resource %s token request failed: %v", resKey, err)
 		return nil, err
 	}
-	if requiresRefresh(&tokenRequest) {
-		err := dao.DeleteMetaByKey(resKey)
-		if err != nil {
-			klog.Errorf("delete meta %s failed: %v", resKey, err)
-			return nil, err
-		}
-		klog.Errorf("resource %s token expired", resKey)
-		return nil, fmt.Errorf("resource %s token expired", resKey)
-	}
 	return &tokenRequest, nil
 }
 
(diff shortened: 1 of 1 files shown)


imneov:
cc @fisherxu

WillardHu:
For offline, there some scenarios for getting a serviceaccount are problematic:
- Restart a pod, its id hasn't changed;
- Restart a pod, its id has changed;
- The token has expired;
- Clean up invalid serviceaccount data;

In theory, we can use the logic of the cloud to generate a token and resave it to the db in the edge, this token same as the cloud generated, Instead of using a expired token. Maybe this will solve these problems.

imneov:
> For offline, there some scenarios for getting a serviceaccount are problematic:
> 
> * Restart a pod, its id hasn't changed;
> * Restart a pod, its id has changed;
> * The token has expired;
> * Clean up invalid serviceaccount data;
> 
> In theory, we can use the logic of the cloud to generate a token and resave it to the db in the edge, this token same as the cloud generated, Instead of using a expired token. Maybe this will solve these problems.

Offline means that the cloud apiserver cannot be accessed. At this time, the service account will not actually be used. What is the meaning of generating a token at this time?

WillardHu:
> Offline means that the cloud apiserver cannot be accessed. At this time, the service account will not actually be used.
> 
> Is it necessary for us to regenerate a un-expiring token according to the cloud logic?

Security for metaserver. We can provide some local cache resource queries and token authentication when offline. In addition, the offline may only be temporary, but you've already mounted an expired token to the pod, the token is not remounted after the hub is reconnected.

imneov:
> > Offline means that the cloud apiserver cannot be accessed. At this time, the service account will not actually be used.
> > Is it necessary for us to regenerate a un-expiring token according to the cloud logic?
> 
> Security for metaserver. We can provide some local cache resource queries and token authentication when offline. In addition, the offline may only be temporary, but you've already mounted an expired token to the pod, the token is not remounted after the hub is reconnected.

image

If we strictly follow the behavior of kubelet, the token needs to be refreshed automatically.   [CHANGELOG-1.21.md#api-change-3](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md#api-change-3)



However, I understand that is a big project. Before that, how to solve the problem of pod restart failure when offline?

imneov:
> @imneov Could you please submit a new issue to track it? And we can condider merge this temporary solution into some release

OK，I will submit to 1.13

imneov:
> @imneov Could you please submit a new issue to track it? And we can condider merge this temporary solution into some release

I have submitted another PR #5622

Do you mean I should create another issue? I see there are issues discussing this issue #4441

Issue #5601, opened 2024-05-11
keadm init failed with `--profile version=v1.17.0`


Shelley-BaoYue:
**What happened**:

When using `keadm init --profile version=v1.17.0` to deploy cloudcore,  it will failed with error like: 

![image](https://github.com/kubeedge/kubeedge/assets/48788150/fe97ce69-cc82-46df-beb2-33a5c4c12d82)

**What you expected to happen**: 

deploy cloudcore successfully

**How to reproduce it (as minimally and precisely as possible)**:

execute `keadm init --advertise-address=xxx --profile version=v1.17.0`

**Anything else we need to know?**:

Using keadm --kubeedge-version=v1.17.0 can deploy cloudcore successfully

**Environment**:
- Kubernetes version (use `kubectl version`): 1.27
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): cloudcore v1.17.0   keadm version: v1.17.0
- <details><summary>Cloud nodes Environment:</summary>

  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>
- <details><summary>Edge nodes Environment:</summary>

  - edgecore version (e.g. `edgecore --version`):
  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>

Shelley-BaoYue:
cc @luomengY

luomengY:
> cc @luomengY

Keam init has restructured its architecture on v1.17, and now only supports internal profiles/version.yaml for profiles. If you want to specify a version, it is recommended to use `--kubeedge-version` directly. The `--profile` parameter only supports version.

luomengY:
> > cc @luomengY
> 
> Keam init has restructured its architecture on v1.17, and now only supports internal `profiles/version.yaml` for profiles. If you want to specify a version, it is recommended to use `--kubeedge-version` directly. The `--profile` parameter only supports version.

Shelley-BaoYue:
`The --profile parameter only supports version.`
--> but now I cannot using `--profile version` to specify cloudcore version. 

[...]

PR #5461, opened 2024-03-11
Refactor the implementation code for the cloud init command

luomengY:
…gofile keadm/cmd/keadm/app/cmd/helm/cloudcore.go func Install(..) to complete the install logic.

**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
image

**Which issue(s) this PR fixes**:

Fixes #https://github.com/kubeedge/kubeedge/issues/5317

**Special notes for your reviewer**:
@WillardHu 
**Does this PR introduce a user-facing change?**:

```release-note

```

PR #5632, opened 2024-05-24
fix keadm init profile in v1.17.0 for keadm_e2e test.

luomengY:
**What type of PR is this?**
/kind bug

Fixes #

```release-note

```


luomengY:
PTAL @Shelley-BaoYue @WillardHu

Shelley-BaoYue (commented):
/lgtm 
And as disscussed in issue [#5601](https://github.com/kubeedge/kubeedge/issues/5601), I think we should still support `--profile version` in v1.17 (maybe 1.18 also), and claim it will be deprecated at 1.19. During these two release, users have enough time to get this change and modify it.

luomengY:
> And as disscussed in issue [#5601](https://github.com/kubeedge/kubeedge/issues/5601), I think we should still support `--profile version` in v1.17 (maybe 1.18 also), and claim it will be deprecated at 1.19. During these two release, users have enough time to get this change and modify it.

ok， I will propose another PR to handle compatibility.

luomengY:
@fisherxu

Shelley-BaoYue:
fix CI issue cause by https://github.com/kubeedge/kubeedge/pull/5461

PR #5635, opened 2024-05-27
Proposal:Add router message high availability.

### LINKED ISSUE #5561: Router Manager Support HA
Router Manager Support HA

**Why is this needed**:
Users need to communicate between the cloud and the edge. For example, the cloud calls the rest interface of the edge service. In this case, the routing management function of kubeedge can be used. Currently, Kubeedge's routing management function has some problems in the case of multiple cloudcore copies. The main problem is that when there are multiple copies of cloudcore, whether the cloud sends messages to the edge or reports the message to the cloud, it is not known which cloudcore is sent to it for processing, and there is confusion in message management in the cloud.

luomengY:
**What type of PR is this?**
/kind feature
/kind design

Fixes #5561 

```release-note

```


Shelley-BaoYue (commented):
Please link related issue https://github.com/kubeedge/kubeedge/issues/5561 in PR description.

fisherxu (commented):
/approve

Shelley-BaoYue (commented):
/lgtm

Issue #5542, opened 2024-04-17
 Add device status reporting information


JiaweiGithub:
**What would you like to be added/modified**:
We hope to be able to count the status of the device, such as offline, running, and abnormal. However, the current device dmi design does not report this information, so we hope to add it.
**Why is this needed**:

JiaweiGithub:
1.Add state data in mapper's PushToEdgeCore
![image](https://github.com/kubeedge/kubeedge/assets/8078735/db59e23f-3007-4a57-a549-f92231868d13)
2.edgecore modification
![image](https://github.com/kubeedge/kubeedge/assets/8078735/e0aa821e-7e72-44f1-883c-824c4acf4a4f)
3.cloudcore modification
![image](https://github.com/kubeedge/kubeedge/assets/8078735/0f37e4bc-8024-4aa7-9f0a-d100ba12d5c5)

PR #5651, opened 2024-06-03
Proposal: Support device state message

JiaweiGithub:

**What type of PR is this?**
/kind design

Fixes #
https://github.com/kubeedge/kubeedge/issues/5542

```release-note

```


docs/proposals/device-state.md
```diff
@@ -0,0 +1,215 @@
...
+		return nil
+	}
+
+	// state refers to definition in mappers-go/pkg/common/const.go
+	state := strings.ToLower(updatedDevice.State)
+	switch state {
+	case "online", "offline", "ok", "unknown", "disconnected":
+	default:
+		return nil
+	}
```
WillardHu, line 89: Do these scenarios that return nil need to consider printing some logs?  Otherwise, the device status updatation may not be executed due to some unexpected reasons.

docs/proposals/device-state.md
```diff
@@ -0,0 +1,215 @@
...
+	return nil
+}
+```
+
+### Cloudcore design    
+- Add device states-related definitions to the crd definition of device.
+```yaml
+  last_online:
+    description: 'Optional: The last time the device was online.'
+    type: string
```
WillardHu, line 135: Whether to set an enumeration range for type?

docs/proposals/device-state.md
```diff
@@ -0,0 +1,215 @@
...
+  state:
+    description: 'Optional: The state of the device.'
+    type: string
```
WillardHu, line 138: Whether to set an enumeration range for type?

WillardHu:
It is not recommended to have more than one piece of detailed code logic in the proposal, you can add some pseudo-code.

WillardHu:
PTAL @wbc6080

docs/proposals/device-state.md
```diff
@@ -0,0 +1,215 @@
...
+- Mapper only provides a general framework and does not provide device states collection code.
+
+### Device state design
+<img src="../images/device-crd/device-state.png">
+
+### Mapper design
+- In the dmi module, add a grpc interface for ReportDeviceState.
+- Call getDeviceStates based on goroutine in Mapper's DevStart.
+- GetDeviceStates regularly obtains the device states, and then sends it to edgecore through the ReportDeviceState 
+  interface.
```
wbc6080, line 31: Here you can put the modifications to the grpc api

docs/proposals/device-state.md
```diff
@@ -0,0 +1,215 @@
...
+	statesRequest := &dmiapi.ReportDeviceStatesRequest{
+		DeviceName:      gs.DeviceName,
+		State:           states,
+		DeviceNamespace: gs.DeviceNamespace,
+	}
+
+	log.Printf("send statesRequest", statesRequest.DeviceName, statesRequest.State)
+	if err := grpcclient.ReportDeviceStates(statesRequest); err != nil {
+		klog.Errorf("fail to report device states of %s with err: %+v", gs.DeviceName, err)
+	}
```
wbc6080, line 50: I don’t seem to see how to report regularly from the code. Maybe you can use pseudocode to illustrate the overall functionality

docs/proposals/device-state.md
```diff
@@ -0,0 +1,215 @@
...
+	for i := 1; i <= dtcommon.RetryTimes; i++ {
+		err = dtclient.UpdateDeviceFields(
+			device.ID,
+			map[string]interface{}{
+				"last_online": lastOnline,
+				"state":       updatedDevice.State,
+			})
+		if err == nil {
+			break
+		}
```
wbc6080, line 103: We can use the retry-go package

docs/proposals/device-state.md
wbc6080, line 135: There should be a `status` field in the crd of device-instance before. Can we add some content directly to this field?

docs/proposals/device-state.md
```diff
@@ -0,0 +1,215 @@
...
+```
+- Add a deviceStatesChan in UpstreamController, specifically used to process device states reporting messages.
+```Golang 
+type UpstreamController struct {
+    crdClient    crdClientset.Interface
+    messageLayer messagelayer.MessageLayer
+    // devicestates message channel
+    devicestatesChan chan model.Message
+    // deviceStates message channel
+    deviceStatesChan chan model.Message
```
wbc6080, line 148: The names of `deviceStatesChan` and `deviceStatesChan` are too similar. Is the original channel used to report twin messages? Maybe we can change it to `twinstatesChan`?

docs/proposals/device-state.md
```diff
@@ -0,0 +1,77 @@
...
+- In the dmiserver module, implement ReportDeviceStates and send the received device states information to TwinGroup through beehive.
+- In TwinGroup, match the "state/update" topic, send it to the device moudle, and execute the action DeviceStateUpdate.
+- In DeviceStateUpdate, update the device states information of the sqlite database and transmit it to the cloud.
+
+
+
+### Cloudcore design    
+- Add device states-related definitions to the crd definition of device.
+```yaml
+  lastOnline:
```
fisherxu, line 59: The api struct modification is also needed :)

PR #5123, opened 2023-10-30
update go version to 1.20.10 for docker build tool

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

update go version to 1.20.10 for docker build tool

**Which issue(s) this PR fixes**:

Fixes #5122

PR #5652, opened 2024-06-05
update go 1.21 for kubeedge build image

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

update go 1.21 for kubeedge build tool image, keep consistent with Kubernetes

### DIFF
--- a/build/docker/build-tools/build-tools.dockerfile
+++ b/build/docker/build-tools/build-tools.dockerfile
@@ -10,9 +10,9 @@ RUN apt-get autoremove -y &&\
     apt-get clean &&\
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


WillardHu:
Does the version of golang also need to be upgraded in the go.mod?

Shelley-BaoYue:
> Does the version of golang also need to be upgraded in the go.mod?

Yes, it will be put at a seperate PR. We need to upgrate the go version in KubeEdge basic image, and then upgrate it in other parts.  Ref to: https://github.com/kubeedge/kubeedge/pull/5123 and https://github.com/kubeedge/kubeedge/pull/5146

Shelley-BaoYue:
PTAL @fisherxu @wbc6080

wbc6080 (commented):
/lgtm

fisherxu (commented):
/approve

PR #5653, opened 2024-06-05
upgrade golang to v1.21 

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

upgrade golang to v1.21, keep consistent with Kubernetes v1.29


Shelley-BaoYue:
PTAL @fisherxu  @WillardHu @wbc6080

Issue #4403, opened 2022-11-11
use testify package in unit tests


falser101:
**What would you like to be added/modified**:
Use github.com/stretchr/testify/require and github.com/stretchr/testify/assert packages in all unit tests.

Demo PR: #4402 

**Why is this needed**:
This removes much boilerplate from the tests, and improves speed and quality of understanding and modifying them.

PR #2793, opened 2021-04-20
add unit test for edge/pkg/devicetwin/dtmodule

charles-woshicai:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**:
Imporve unit test coverage of one go module.

**Which issue(s) this PR fixes**:
Ref #2240

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
None

```release-note

```

PR #5675, opened 2024-06-16
Test coverage for `application.go`

### LINKED ISSUE #4765: increase test coverage for `application.go`
I would like to add unit test for `pkg/metaserver/application.go`
**Why is this needed**:

1Shubham7:
**What type of PR is this?**

/kind test

I have added unit tests for `pkg/metaserver/application.go`

Fixes a part of #4765 

```release-note
NONE
```


pkg/metaserver/application_test.go
```diff
@@ -0,0 +1,409 @@
+package metaserver
```
Shelley-BaoYue, line 1: We need at copyright for new file

pkg/metaserver/application_test.go
```diff
@@ -0,0 +1,409 @@
...
+
+import (
+	"context"
+	"crypto/sha256"
+	"fmt"
+	"testing"
+
+	"github.com/kubeedge/beehive/pkg/core/model"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/runtime/schema"
```
Shelley-BaoYue, line 11: The import should be sorted by format: Standard lib; third-party lib; local package

pkg/metaserver/application_test.go
```diff
@@ -0,0 +1,409 @@
+package metaserver
+
+import (
+	"context"
+	"crypto/sha256"
+	"fmt"
+	"testing"
+
+	"github.com/kubeedge/beehive/pkg/core/model"
+	"github.com/stretchr/testify/assert"
```
Shelley-BaoYue, line 10: Is this a new import lib? Is there an existing lib that can be used as its substitute? If not, we ned to add this lib add go.mod file and the run `make verify` to update the vendor.

1Shubham7: this is a new import lib (we have `github.com/stretchr/testify` in go.mod ) . this is an assertion lib that makes the code shorter and much more readable. yes, it can be removed and I can simply just use the testing package. But I think using this lib will be better for reasons I mentioned ([there also somebody who created an issue about it](https://github.com/kubeedge/kubeedge/issues/4403), as he said "This removes much boilerplate from the tests, and improves speed and quality of understanding and modifying them." ) . my suggestion would be using testify, you can tell me which one do you prefer and I will use that in coming unit tests as well :)

1Shubham7: I just saw this lib being used in these [tests](https://github.com/kubeedge/kubeedge/pull/2793/files)

pkg/metaserver/application_test.go
Shelley-BaoYue: yes

pkg/metaserver/application_test.go
Shelley-BaoYue: I think `github.com/stretchr/testify` is fine. Go ahead! Or we can use it to do a full replacement in all unit tests in another PR.  
I have find `github.com/stretchr/testify` in go.mod. If we need to use it, we may need to remove `//indirect` and run `make verify` to update the related vendor.

1Shubham7: Shelley, I did as you said, I removed `//indirect` and then run `make verify`. But it gives error at :
```
        /go/pkg/mod/sigs.k8s.io/controller-tools@v0.6.2/cmd/controller-gen/main.go:200 +0x2f6
make: *** [Makefile:96: verify-crds] Error 2
```

Shelley-BaoYue: `make verify` has succeed in CI. I guess the error in your environment is caused by golang version?

1Shubham7: my go version is `1.22.3`

1Shubham7: one test case still failing, I don't understand why, I didn't even change the code for this to fail at 

```
 [FAILED] in [It] - /home/runner/work/kubeedge/kubeedge/edge/test/integration/utils/helpers/helpers.go:265
```

I have simply added tests.

Shelley-BaoYue: This error should be unrelated with you. I will retry the test : )

1Shubham7: edit: now all tests are passing.

1Shubham7:
@Shelley-BaoYue, its asking me to remove `//indirect` from go.mod for resolving conflict, I shouldn't do that right?

pkg/metaserver/application_test.go
```diff
@@ -0,0 +1,426 @@
+/*
+Copyright 2021 The KubeEdge Authors.
```
Shelley-BaoYue, line 2:
```suggestion
Copyright 2024 The KubeEdge Authors.
```

Shelley-BaoYue:
> @Shelley-BaoYue, its asking me to remove `//indirect` from go.mod for resolving conflict, I shouldn't do that right?

I tried it in my environment. After `git rebase master`, `vi go.mod` like:
image

I solve this and save this file 
image



> @Shelley-BaoYue, its asking me to remove `//indirect` from go.mod for resolving conflict, I shouldn't do that right?

I tried it in my environment. When I move this this lib like: 
image

then rebase --continue, it seems the conflict has been solved. You can try it.

Shelley-BaoYue (commented):
/lgtm

PR #5690, opened 2024-06-25
Added test coverage for `pkg/stream` package

1Shubham7:
**What type of PR is this?**

/kind test

I have added unit tests for `pkg/stream` package

```release-note
NONE
```

### DIFF
--- a/pkg/stream/edgedmetricsconnection.go
+++ b/pkg/stream/edgedmetricsconnection.go
@@ -65,7 +65,7 @@ func (ms *EdgedMetricsConnection) CreateConnectMessage() (*Message, error) {
 }
 
 func (ms *EdgedMetricsConnection) String() string {
-	return fmt.Sprintf("EDGE_Metrics_CONNECTOR Message MessageID %v", ms.MessID)
+	return fmt.Sprintf("EDGE_METRICS_CONNECTOR Message MessageID %v", ms.MessID)
 }
 
 func (ms *EdgedMetricsConnection) receiveFromCloudStream(stop chan struct{}) {
--- a/pkg/stream/edgedattachconnection_test.go
+++ b/pkg/stream/edgedattachconnection_test.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2024 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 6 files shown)


pkg/stream/edgedmetricsconnection.go
```diff
@@ -34,10 +34,22 @@ type EdgedMetricsConnection struct {
...
+	}
+	return NewMessage(ms.MessID, MessageTypeMetricConnect, data), nil
+}
+
 func (ms *EdgedMetricsConnection) GetMessageID() uint64 {
 	return ms.MessID
 }
 
+func (ms *EdgedMetricsConnection) String() string {
+	return fmt.Sprintf("EDGE_Metrics_CONNECTOR Message MessageID %v", ms.MessID)
```
1Shubham7, line 50: Should we make it "EDGE_METRICS_CONNECTOR" like other connections?

pkg/stream/edgedlogconnection.go
```diff
@@ -36,10 +36,22 @@ type EdgedLogsConnection struct {
 	Header   http.Header   `json:"header"`
 }
 
+func (l *EdgedLogsConnection) CreateConnectMessage() (*Message, error) {
+	data, err := json.Marshal(l)
+	if err != nil {
+		return nil, err
```
1Shubham7, line 42: Codecov is asking me to cover case where `json.Marshal` gives an error. I think tests are for testing logic of the functions and not standard libraries like `json.Marshal` (how can I make `json.Marshal` give error?). what do you guys have to say about it ? (also I haven't created this func I have simply moved its place so that all funcs are consistently placed in all files)

pkg/stream/edgedmetricsconnection.go
WillardHu: It seems that this function is only used for log and error messages, I think it can be modified.

pkg/stream/edgedlogconnection.go
WillardHu: I think we needn't to cover this case.

pkg/stream/edgedexecconnection.go
```diff
@@ -1,3 +1,19 @@
+/*
```
Shelley-BaoYue, line 1: Please remove this copyright : )

Shelley-BaoYue (commented):
/lgtm

fisherxu (commented):
/approve

PR #5696, opened 2024-07-01
Remove useless updatenodestatus and updatepodstatus func

### LINKED ISSUE #5647: Bug: Unnecessary permissions
Hi community!

I just found that the [Deployment cloudcore](https://github.com/kubeedge/kubeedge/blob/master/manifests/charts/cloudcore/templates/deployment_cloudcore.yaml) in the charts has `update` verb of the `pods` resource and `create/update` verb of the services resource([aggregation_rbac_cloudcore.yaml](https://github.com/kubeedge/kubeedge/blob/77891f2afa51bb20e43f861d4d2b17006f712da9/manifests/charts/cloudcore/templates/aggregation_rbac_cloudcore.yaml#L3)). However, after reading the source code of , I didn't find any Kubernetes API usages that require these permissions. Therefore, for security reasons, I suggest checking these permissions to determine if they are truly unnecessary. If they are, the issue should be fixed by removing the unnecessary permission or other feasible methods. 

**To Reproduce**

Use helm chart with default values.

Shelley-BaoYue:
**What type of PR is this?**

/kind cleanup

Remove useless updatenodestatus and updatepodstatus func. from v1.12, we have used patchnode and patchpod to replace them.

Fixes #5647


Shelley-BaoYue:
@copilot rebase this PR with master branch

PR #5698, opened 2024-07-01
[wip]Add unit test after upgrade beego version

### LINKED ISSUE #5677: Update Unit test case about sqlite after upgrade beego/beego to v2
**Which jobs are failing**:
All job about operation sqlite database in edgecore
Unit test, integration test edge
**Reason for failure**:
In #5261 , we need to upgrade the `go-sqlite` package version. Since packages reference each other, we also need to upgrade the `beego` package to v2. However, the usage method of `beego` in v2 version has been extensively adjusted compared with v1, so the test case related to the sqlite database operation part needs to be reconstructed.

In order to avoid blocking the subsequent upgrade of the Go language and k8s version, the test cases related to the sqlite database operation are temporarily removed in #5261, so this issue is raised to track the subsequent test case reconstruction work.

wbc6080:
**What type of PR is this?**
/kind test

As mentioned in #5677, some unit test cases were temporarily removed after upgrading the `beego` version. Now this part of the test cases has been refactored and adapted to the v2 version of `beego`, so they are added back to the unit tests.

Fixes #5677 

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtclient/dbtest.go
+++ b/edge/pkg/devicetwin/dtclient/dbtest.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 The KubeEdge Authors.
+Copyright 2024 The KubeEdge Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -33,12 +33,8 @@ var errFailedDBOperation = errors.New("Failed DB Operation")
 type CasesSaveStr []struct {
 	// name is name of the testcase
 	name string
-	// returnInt is first return of mock interface ormerMock
-	returnInt int64
-	// returnErr is second return of mock interface ormerMock which is also expected error
-	returnErr error
-	// beginReturn is the return of mock interface ormerMock's Begin function
-	beginReturn orm.TxOrmer
+	// returnErr is return of mock interface ormerMock's DoTX function
+	doTXReturnErr error
 }
 
 // CasesDeleteStr is a struct for cases of delete
@@ -53,8 +49,8 @@ type CasesDeleteStr []struct {
 	deleteReturnErr error
 	// queryTableReturn is the return of mock interface ormerMock's QueryTable function
 	queryTableReturn orm.QuerySeter
-	// beginReturn is the return of mock interface ormerMock's Begin function
-	beginReturn orm.TxOrmer
+	// returnErr is return of mock interface ormerMock's DoTX function
+	doTXReturnErr error
 }
 
 // CasesUpdateStr is a struct for cases of update
@@ -69,8 +65,6 @@ type CasesUpdateStr []struct {
 	updateReturnErr error
 	// queryTableReturn is the return of mock interface ormerMock's QueryTable function
 	queryTableReturn orm.QuerySeter
-	// beginReturn is the return of mock interface ormerMock's Begin function
-	beginReturn orm.TxOrmer
 }
 
 // CasesQueryStr is a struct for cases of query
@@ -127,8 +121,6 @@ type CasesTransStr []struct {
 	queryTableTimes int
 	// wantErr is expected error
 	wantErr error
-	// beginReturn is the return of mock interface ormerMock's Begin function
-	beginReturn orm.TxOrmer
 }
 
 // GetCasesSave get cases for save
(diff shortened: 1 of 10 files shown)


edge/pkg/devicetwin/dtclient/dbtest.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 The KubeEdge Authors.
+Copyright 2024 The KubeEdge Authors.
```
Shelley-BaoYue, line 2: Don't need to change the copyright time

edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -17,20 +19,41 @@ type Device struct {
...
+		// or it will be committed
+		return e
+	})
+
 	if err != nil {
-		klog.Errorf("Something wrong when deleting data: %v", err)
+		klog.Errorf("Something wrong when deleting Device data: %v", err)
 		return err
 	}
-	klog.V(4).Infof("Delete affected Num: %d", num)
```
Shelley-BaoYue, line 33: Wouldn't it be better to keep the affected num?

edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -17,20 +19,41 @@ type Device struct {
...
 }
 
 // DeleteDeviceByID delete device by id
-func DeleteDeviceByID(to orm.TxOrmer, id string) error {
-	num, err := to.QueryTable(DeviceTableName).Filter("id", id).Delete()
+func DeleteDeviceByID(o orm.Ormer, id string) error {
+	//num, err := o.QueryTable(DeviceTableName).Filter("id", id).Delete()
+
+	err := o.DoTx(func(ctx context.Context, txOrm orm.TxOrmer) error {
+		// insert data
```
Shelley-BaoYue, line 44:
```suggestion
		// delete data
```

edge/pkg/devicetwin/dtclient/device_db.go
```diff
@@ -17,20 +19,41 @@ type Device struct {
...
+	}
+	klog.V(4).Info("insert Device data successfully")
+	return nil
 }
 
 // DeleteDeviceByID delete device by id
-func DeleteDeviceByID(to orm.TxOrmer, id string) error {
-	num, err := to.QueryTable(DeviceTableName).Filter("id", id).Delete()
+func DeleteDeviceByID(o orm.Ormer, id string) error {
+	//num, err := o.QueryTable(DeviceTableName).Filter("id", id).Delete()
```
Shelley-BaoYue, line 41: remove useless line

edge/pkg/devicetwin/dtclient/deviceattr_db_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The KubeEdge Authors.
+Copyright 2024 The KubeEdge Authors.
```
Shelley-BaoYue, line 2: ditto

edge/pkg/devicetwin/dtclient/dbtest.go
wbc6080: Done

edge/pkg/devicetwin/dtclient/device_db.go
wbc6080: Done

edge/pkg/devicetwin/dtclient/device_db.go
wbc6080: Done

edge/pkg/devicetwin/dtclient/deviceattr_db_test.go
wbc6080: Done

edge/pkg/devicetwin/dtclient/device_db.go
wbc6080: In the v2 version of beego, it is officially recommended to use `o.DoTx` to handle transactions. `o.DoTx` only returns errors and no longer returns numbers as before. https://github.com/beego/beedoc/blob/master/en-US/mvc/model/transaction.md#transaction

Shelley-BaoYue: I get it. Does txOrm.Insert (user) return num and error? It looks like beego's example doesn't use this return parameter either.

wbc6080: Yes, `txOrm.Insert` will indeed return `num` and `error`. However, `num` parameter cannot be returned in `o.DoTx` (the input parameter function defined by `o.DoTx` can only return err)

Shelley-BaoYue (commented):
/lgtm

fisherxu (commented):
/approve

PR #5721, opened 2024-07-12
fix: func NewErrorMessage

luomengY:
**What type of PR is this?**

/kind bug

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
+++ b/staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
@@ -254,7 +254,7 @@ func (msg *Message) NewRespByMessage(message *Message, content interface{}) *Mes
 
 // NewErrorMessage returns a new error message by a message received
 func NewErrorMessage(message *Message, errContent string) *Message {
-	return NewMessage(message.Header.ParentID).
+	return NewMessage(message.GetID()).
 		SetResourceOperation(message.Router.Resource, ResponseErrorOperation).
 		FillBody(errContent)
 }


staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
```diff
@@ -254,7 +254,7 @@ func (msg *Message) NewRespByMessage(message *Message, content interface{}) *Mes
 
 // NewErrorMessage returns a new error message by a message received
 func NewErrorMessage(message *Message, errContent string) *Message {
-	return NewMessage(message.Header.ParentID).
```
fisherxu, line 257: In NewMessage, the parameter is `ParentID` :)

luomengY: > In NewMessage, the parameter is `ParentID` :)

Here, the `feedbackError` function should return the error to the request from `Sendsyn`, but now the `NewError Message` function uses the ParentID of the requested message as the ParentID of the ErrorMessage. Because the ParentID of the sent message is empty, the ID of the Error Message will also be empty, which will cause the message to be discarded and there will be no receiving channel.

fisherxu: Sounds good.

fisherxu (commented):
/approve

staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
fisherxu: Sounds good.

PR #5747, opened 2024-07-23
fix edge device state sync error

JiaweiGithub:

**What type of PR is this?**
/kind bug

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dttype/types.go
+++ b/edge/pkg/devicetwin/dttype/types.go
@@ -23,6 +23,17 @@ type Device struct {
 	Twin        map[string]*MsgTwin `json:"twin,omitempty"`
 }
 
+// DeviceCloudMsg used to synchronize device data to the cloud
+type DeviceCloudMsg struct {
+	ID             string              `json:"id,omitempty"`
+	Name           string              `json:"name,omitempty"`
+	Description    string              `json:"description,omitempty"`
+	State          string              `json:"state,omitempty"`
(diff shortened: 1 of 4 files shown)


edge/pkg/devicetwin/dttype/types.go
```diff
@@ -23,6 +23,17 @@ type Device struct {
...
+// DeviceCloudMsg used to synchronize device data to the cloud
+type DeviceCloudMsg struct {
+	ID             string              `json:"id,omitempty"`
+	Name           string              `json:"name,omitempty"`
+	Description    string              `json:"description,omitempty"`
+	State          string              `json:"state,omitempty"`
+	LastOnlineTime string              `json:"lastOnlineTime,omitempty"`
+	Attributes     map[string]*MsgAttr `json:"attributes,omitempty"`
+	Twin           map[string]*MsgTwin `json:"twin,omitempty"`
+}
```
fisherxu, line 35: What's the difference with `Device`?

fisherxu (commented):
/approve

edge/pkg/devicetwin/dttype/types.go
JiaweiGithub: The original device has a field last_online, which is related to the database, but this name is not suitable for use on the cloud. The database fields are not suitable for adjustment, so I think this is the smallest change plan.

Shelley-BaoYue:
cc @WillardHu

Shelley-BaoYue (commented):
Please cherry-pick to release-1.18

PR #5763, opened 2024-07-29
Tests for `storage/v1` module

1Shubham7:
**What type of PR is this?**

/kind test

I have added tests for the `edge/pkg/edged/kubeclientbridge/typed/storage/v1` module.

```release-note
NONE
```


1Shubham7:
I have used this approach of creating mocks for `VolumeAttachmentsInterface` and `metaClient` because `metaClient` can't be accessed and only way to use it is New() func where we can't set a mock `send`. If we want to reduce mocks, we will have to do :

```
metaClient := &client.MetaClient{
        Send: mockSend,
    }
```

and directly use it in `VolumeAttachmentsBridge`, otherwise this is the best approach.

fisherxu (commented):
/lgtm
/approve

Issue #5679, opened 2024-06-20
Add a CI job to check openapi and swagger 


Shelley-BaoYue:
**What would you like to be added/modified**: 

Add a CI job to check openapi and swagger 

**Why is this needed**:

As PR [#](https://github.com/kubeedge/kubeedge/pull/5668) has achieved auto-generate API docs. In this PR, we have add generate-openapi.sh and generate-swagger.sh to generate openapi and swagger. We need add a CI job to check this item.

EraKin575:
@Shelley-BaoYue can I work on this issue? Please assign me this issue

Shelley-BaoYue:
@EraKin575  Of course, please go ahead. Feel free to communicate here if you have any question. 
/assign @EraKin575

EraKin575:
Thanks. @Shelley-BaoYue !I have to add a CI-Job to test if the openapi definitions and specifications are generated or not, right?

Shelley-BaoYue:
In PR https://github.com/kubeedge/kubeedge/pull/5668, we can use the scripts to generate openapi and swagger file. But if we modified the API, these files won't be auto modified. We need to add a CI job to check whether these files need to be modified. If yes, the system prompts the developer to submit the corresponding updates. (I think you can ref to https://github.com/kubeedge/kubeedge/blob/master/hack/verify-crds.sh)

EraKin575:
Thanks!

EraKin575:
I am really sorry for such kind of inconvinience for opening multiple PRs

PR #5768, opened 2024-07-30
added ci job for open api gen

EraKin575:
**What type of PR is this?**

**Which issue(s) this PR fixes**: #5679 

Fixes #

```release-note

```


Shelley-BaoYue (commented):
Hi, @EraKin575  Thanks for your contributions!  

If you still work in progress, you can add a [wip] at title. 
I think put this job in https://github.com/kubeedge/kubeedge/blob/master/.github/workflows/main.yaml is more suitable. : )

.github/workflows/main.yaml
```diff
@@ -392,6 +392,21 @@ jobs:
           path: |
             /tmp/cloudcore.log
             /tmp/edgecore.log
+  check-api-docs:
```
Shelley-BaoYue, line 395: I have a new idea:  we don't need add a new job here. We can add a verify under `make verify`  in https://github.com/kubeedge/kubeedge/blob/master/Makefile#L67.  This job https://github.com/kubeedge/kubeedge/blob/master/.github/workflows/main.yaml#L25 will run `make verify`.

tests/scripts/check-api-docs.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
Shelley-BaoYue, line 1: Would it be better name this script `verify-apidoc.sh` to consistent with other verify scripts? And as said above, just add a `verify-apidoc` under make verify https://github.com/kubeedge/kubeedge/blob/master/Makefile#L67 is enough.

EraKin575: thanks for the review! I will make the changes

EraKin575: Can I just implement the changes to the generated file if it needs to be generated after verifying or just leave it at verification?

Makefile
```diff
@@ -91,9 +91,12 @@ verify-codegen:
 .PHONY: verify-vendor-licenses
 verify-vendor-licenses:
 	hack/verify-vendor-licenses.sh
-.PHONY: verify-crds
-verify-crds:
-	hack/verify-crds.sh
+# .PHONY: verify-crds
```
Shelley-BaoYue, line 94: ditto

hack/verify-apidoc.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 The KubeEdge Authors.
```
Shelley-BaoYue, line 3:
```suggestion
# Copyright 2024 The KubeEdge Authors.
```

Shelley-BaoYue (commented):
why apidoc/tools/update-swagger-docs.sh updated? 
image

Makefile
EraKin575: apologies, verify-crds is not working properly, I had to remove it to run make-verify which was after it. I forgot to put it back

EraKin575:
closing this pr

EraKin575:
@Shelley-BaoYue Closing this PR, and opening another one with clean commit history

PR #5774, opened 2024-07-31
Go workspace for kubeedge

WillardHu:
**What type of PR is this?**

/kind feature

Use go workspace to maintains kubeedge and staging/* that syncs dependencies between projects.  This also resolves the compilation error caused by a part of IDEA tools maintaining the staging/* projects in the KubeEdge.

Remove some unreasonable replace in staging/*/go.mod.

Fixes #

```release-note

```


WillardHu:
I've created a https://github.com/cncf/cncf-fuzzing/pull/495 to fix the CIFuzz issue.

fisherxu (commented):
Thanks, looks good! And is it necessary to add a scipt to update and check the go workspace as in K8s? @WillardHu

WillardHu:
> Thanks, looks good! And is it necessary to add a scipt to update and check the go workspace as in K8s? @WillardHu
> 
> Ref: https://github.com/kubernetes/kubernetes/blob/master/hack/update-go-workspace.sh
> 
> https://github.com/kubernetes/kubernetes/blob/master/hack/verify-go-workspace.sh

Kubernetes has changed the way it verifies go workspace, refer to: https://github.com/kubernetes/kubernetes/pull/123627

Do we also put it in the verify-vendor? In this case, the adjustment will be slightly larger. @fisherxu

fisherxu:
As the comments in K8s, when user update the vendor, they will also need to update the go workspace, I think we can also merge the vendor update to go workspace check. But we can add the go workspace check first at present. How do you think? @WillardHu @Shelley-BaoYue @luomengY

WillardHu:
> As the comments in K8s, when user update the vendor, they will also need to update the go workspace, I think we can also merge the vendor update to go workspace check. But we can add the go workspace check first at present. How do you think? @WillardHu @Shelley-BaoYue @luomengY

Done.

WillardHu:
Resolving checklist errors...

Shelley-BaoYue (commented):
/lgtm

fisherxu (commented):
/approve

PR #5775, opened 2024-08-01
Fixed the golangci config and some code validation issues

WillardHu:
**What type of PR is this?**

/kind bug

Some golangci configuration are out of date, and fixed some code validation issues that occur with higher versions.

Fixes #

```release-note

```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,21 +12,51 @@ run:
   # exit code when at least one issue was found, default is 1
   issues-exit-code: 1
 
-  # which dirs to skip: issues from them won't be reported;
-  # can use regexp here: generated.*, regexp is applied on full path;
-  # default value is empty list, but default dirs are skipped independently
-  # from this option's value (see skip-dirs-use-default).
-  # "/" will be replaced by current OS file path separator to properly work
-  # on Windows.
-  skip-dirs:
+issues:
+  # Which dirs to exclude: issues from them won't be reported.
+  # Can use regexp here: `generated.*`, regexp is applied on full path,
+  # including the path prefix if one is set.
+  # Default dirs are skipped independently of this option's value (see exclude-dirs-use-default).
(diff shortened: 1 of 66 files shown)


.golangci.yml
```diff
@@ -59,10 +102,24 @@ linters:
...
-    max-same-issues: 0
-    # Excluding configuration per-path, per-linter, per-text and per-source
-    exclude-rules:
-      - path: edge
-        linters:
-          - errcheck
+  # Maximum count of issues with the same text. Set to 0 to disable. Default is 3.
+  max-same-issues: 0
+  # Excluding configuration per-path, per-linter, per-text and per-source
+  exclude-rules:
```
Shelley-BaoYue, line 108: I'm not very clearly know why we need exclude edge-errcheck rules before?  But I guess with new edged (from v1.12), we may no longer need to set this rules? 
Would you mind try to remove this setting and re-run golint test in this PR?

WillardHu: Lint checked failure, so many errors are not checked. Shoud we handle these errs or recover this rule?

Shelley-BaoYue: I think we should handle these errors. If there are a large num of errs, we can first open an issue and split this task by directory, and then solve it together.

WillardHu: I'll try to handle it first.

WillardHu: Done. For unhandled errors, I print the error first to minimize its logical impact, for example:
```golang
if err != nil {
    // TODO: handle error
    klog.Error(err)
}
```
In addition, I disabled the checks for unused-parameter and unused-receiver, which did not seem to have much effect.

WillardHu:
PTAL @Shelley-BaoYue @fisherxu

Shelley-BaoYue (commented):
/lgtm 
/assign @fisherxu

fisherxu (commented):
/approve

PR #5785, opened 2024-08-07
kubeedge/build/helm -> kubeedge/helm

rohitsinghnaruka:
this PR moves helm to root directory from build directory


fisherxu (commented):
/close
As I've replied in the issue

PR #5799, opened 2024-08-12
Get the ConfigMap and Secret from the edge node

luomengY:
**What type of PR is this?**
/kind feature

When the cloud edge network is not good, starting a pod to obtain the `configmap` from the cloud will result in a longer delay in pod startup, so it is necessary to directly obtain the configmap from the edge nodes.

image

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/client/configmap.go
+++ b/edge/pkg/metamanager/client/configmap.go
@@ -9,6 +9,7 @@ import (
 	"github.com/kubeedge/beehive/pkg/core/model"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/message"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/modules"
+	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/dao"
 )
 
 // ConfigMapsGetter has a method to return a ConfigMapInterface.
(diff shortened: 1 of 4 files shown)


edge/pkg/metamanager/client/configmap.go
```diff
@@ -51,24 +52,33 @@ func (c *configMaps) Delete(string) error {
...
-	}
-	errContent, ok := msg.GetContent().(error)
-	if ok {
-		return nil, errContent
-	}
-	content, err := msg.GetContentData()
-	if err != nil {
-		return nil, fmt.Errorf("parse message to configmap failed, err: %v", err)
+	metas, err := dao.QueryMeta("key", resource)
+	if err != nil || len(*metas) == 0 {
```
wbc6080, line 56: According to the logic here, if the `configmap` information can be queried in the database, it will not be obtained from the cloud, right? But if there is a problem with the information in the database and an error occurs in the subsequent `handleConfigMapFromMetaDB` parsing,  according to the above logic, we also cannot execute the part of obtaining data from the cloud because database have configmap information(Even the data in the database is wrong)?

luomengY: > According to the logic here, if the `configmap` information can be queried in the database, it will not be obtained from the cloud, right? But if there is a problem with the information in the database and an error occurs in the subsequent `handleConfigMapFromMetaDB` parsing, according to the above logic, we also cannot execute the part of obtaining data from the cloud because database have configmap information(Even the data in the database is wrong)?

When the configmap changes in the cloud, it will be updated to the edge sqlite. Prioritizing obtaining it from the edge sqliet can solve the problem of poor cloud edge network.

fujitatomoya: @luomengY just to be sure about this fallback behavior.

> When the configmap changes in the cloud, it will be updated to the edge sqlite.

is there any racy situation that this might get `not-up-to-date` configmap from edge sqlite then? in other word, edge and cloud configmap is always synced, so that edge cannot get `not-up-to-date` configmap from edge sqlite at all?

i was thinking what we would want to do as fallback behavior, maybe we can try cloud and if response is none (e.g network connectivity or anything), then we can fall back to call edge sqlite? this guarantees that edge can get the latest configmap and usually this case just works unless network connectivity problems?

luomengY:
@Shelley-BaoYue @wbc6080 @WillardHu

edge/pkg/metamanager/client/configmap.go
```diff
@@ -51,24 +52,37 @@ func (c *configMaps) Delete(string) error {
...
+		errContent, ok := msg.GetContent().(error)
+		if ok {
+			return nil, errContent
+		}
+		content, err := msg.GetContentData()
+		if err != nil {
+			return nil, fmt.Errorf("parse message to configmap failed, err: %v", err)
+		}
+
+		if msg.GetOperation() == model.ResponseOperation && msg.GetSource() == modules.MetaManagerModuleName {
```
Shelley-BaoYue, line 70: Is this situation(Line 70-72) still necessary?

Shelley-BaoYue (commented):
Overall looks good. Please rebase the new master branch.

luomengY:
> Overall looks good. Please rebase the new master branch.

done

Shelley-BaoYue (commented):
/lgtm 
/assign @WillardHu @fisherxu

PR #5800, opened 2024-08-12
Change apt source to fix container runtime CI failures

### LINKED ISSUE #5798: Container runtime e2e test often fails to run
**Which jobs are failing**:
https://github.com/kubeedge/kubeedge/actions/workflows/main.yaml
Container runtime e2e test
about one week
**Reason for failure**:
Even though we had added a retry mechanism before, the download still failed after 5 retries.
image

wbc6080:
**What type of PR is this?**
/kind failing-test

As mentioned in #5798 , container runtime e2e test CI often fails now because the apt source is inaccessible and dependent components cannot be downloaded. Therefore, this PR adds an additional mirror source replacement mechanism with the base of the previously failure retry mechanism. When failures reaches a certain number, the apt source will be replaced to solve the problem of CI failure.

Fixes #5798

```release-note

```

### DIFF
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -36,6 +36,11 @@ function install_cr() {
 
   while [ $attempt_num -lt $max_attempts ]
   do
+    if  [ $attempt_num -eq 3 ]; then
+      echo "Download failed multiple times, try to change apt source ..."
+      sudo sed -i 's@//.*archive.ubuntu.com@//mirrors.ustc.edu.cn@g' /etc/apt/sources.list
+      sudo apt-get update
+    fi
     if [[ "${CONTAINER_RUNTIME}" = "docker" ]]; then
       install_docker
       verify_docker_installed


wbc6080 (commented):
After this PR implements the mechanism of changing the apt source, we can see from the CI log shows that after three times downloads failed, the apt source will be automatically changed and downloaded again. Finally, the CI test was successfully passed.
x888

fisherxu (commented):
Is this an occasional error, or will it always fail if we don't replace it?

luomengY:
> Is this an occasional error, or will it always fail if we don't replace it?

This should be an occasional error, but this modification can indeed make CI pass more stably.

luomengY:
Let `conformance e2e test` pass. @wbc6080

wbc6080:
> Is this an occasional error, or will it always fail if we don't replace it?


A similar problem occurred three months ago, and in the past week, all PR container runtime CI test failed. The purpose of this PR is to replace the apt source when the original source is inaccessible, making CI testing more stable.

wbc6080:
> Let `conformance e2e test` pass. @wbc6080

The conformance e2e test has also failed in recent PRs. We use #5798  to track this issue. This PR will not fix the conformance e2e test yet.

fisherxu:
@luomengY You need to add yourself to the owner file under hack first ：）
/approve

fisherxu (commented):

PR #5832, opened 2024-09-01
add skip cr install option for local-up-kubeedge.sh

magichan:
**What type of PR is this?**

/kind feature 

`hack/local-up-kubeedge.sh ` try to install non-containerd container runtime in every execution. the pr provide option (`SKIP_CR_INSTALL`, default is false)  to skip the install cr step 

Fixes #

Self Test: 

test case one: default option

image

test case two: docker runtime

image

test case three: skip docker runtime 

image

test case four:  wrong cr 

image
 

```release-note

```

### DIFF
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -22,6 +22,7 @@ TIMEOUT=${TIMEOUT:-60}s
 PROTOCOL=${PROTOCOL:-"WebSocket"}
 CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"containerd"}
 KIND_IMAGE=${1:-"kindest/node:v1.29.0"}
+SKIP_CR_INSTALL=${SKIP_CR_INSTALL:-"false"}
 echo -e "The installation of the cni plugin will overwrite the cni config file. Use export CNI_CONF_OVERWRITE=false to disable it."
 
 if [[ "${CLUSTER_NAME}x" == "x" ]]; then
@@ -31,6 +32,13 @@ fi
 export CLUSTER_CONTEXT="--name ${CLUSTER_NAME}"
 
 function install_cr() {
+
+  if [[ "${CONTAINER_RUNTIME}" = "containerd" ]]; then
+      echo "No need to download container runtime for containerd"
+      return
+  fi
+
+  echo -e "The installation of the container runtime will try to install ${CONTAINER_RUNTIME}. Use export SKIP_CR_INSTALL=true to skip it."
   attempt_num=0
   max_attempts=5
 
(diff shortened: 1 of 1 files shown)


hack/local-up-kubeedge.sh
```diff
@@ -31,6 +31,19 @@ fi
...
+  if [[ "${CONTAINER_RUNTIME}" = "containerd" ]]; then
+      echo "No need to download container runtime for containerd"
+      return
+  fi
+
+  SKIP_CR_INSTALL=${SKIP_CR_INSTALL:-"false"}
+  if [[ "${SKIP_CR_INSTALL}" = "true" ]]; then
+    echo "Skip container runtime installation"
+    return
+  fi
```
fisherxu, line 44: How about making the SKIP_CR_INSTALL as global variable?
  
  if [[ "${SKIP_CR_INSTALL}" = "false" ]]; then
    install_cr
  else
    echo "Skip container runtime installation"
  fi

wbc6080 (commented):
I have an idea, can we directly verify whether container runtime has been installed before downloading? For example, firstly execute the `verify_docker_installed` function in the `install_docker` function. If docker has been downloaded, we can exit directly. In this way, there is no need to add the SKIP_CR_INSTALL parameter, avoiding too many parameters in script.

hack/local-up-kubeedge.sh
magichan: follow the comment, 

move the `SKIP_CR_INSTALL` env out of the `install_cr` function.

fisherxu:
`SKIP_CR_INSTALL`  should be also fine :)  If the container runtime is not installed, but the user also doesn't want to install it with our script, this parameter can be used

magichan:
> I have an idea, can we directly verify whether container runtime has been installed before downloading? For example, firstly execute the `verify_docker_installed` function in the `install_docker` function. If docker has been downloaded, we can exit directly. In this way, there is no need to add the SKIP_CR_INSTALL parameter, avoiding too many parameters in script.

@wbc6080  From a more user-friendly perspective, this is a good idea. However, the workload is somewhat significant. This is because the relevant contributors need to build various container runtime environments to verify the functionality of the script. Would you mind creating a separate issue to track this feature, and first merge this MR as the first contribution of my to KubeEdge?

WillardHu:
Please retry the pipeline checks

magichan:
> Please retry the pipeline checks

@WillardHu  I have rebase master,  but i don't know the next step. 

image

wbc6080:
> > Please retry the pipeline checks
> 
> @WillardHu I have rebase master, but i don't know the next step.

Now all CI check task have passed. If some CI reports error occasionally in subsequent PR，you can retry them by close and re-open the pr

fisherxu (commented):
/approve
Congratulations for your first feature/PR in kubeedeg repo! @magichan  :)

PR #5834, opened 2024-09-03
Proposal plan for edgeapplication overrides

### LINKED ISSUE #5755: Decouple the node cooperation ability and batch management ability of the edgeapplication
**What would you like to be added/modified:**

1. EdgeApplication can be overrides deployment spec via the node group or node label selector.
1. Fix the issue of closed-loop flow control. 

**Why is this needed:**

EdgeApplication can be overrides deployment spec(i.e. replicas, image, commands and environments) via the node group, and pod traffics are closed-loop in a node group(Deployments managed by EdgeApplication share a Service). But in the real scenario, the scope of nodes that need batch operations is different from that of nodes that need to collaborate with each other.

For example, in the scenario of a smart park, there are many smart parks in every cities, we need application traffics are closed-loop in ones smart park, but the scope of batch management may be city level or province level.

Therefore, we need to have a solution to decouple these.

Relevant issue https://github.com/kubeedge/kubeedge/issues/5669

**Expected Outcomes:**

1. Create a proposol of this iessue's solution.
1. Finish the coding. EdgeApplication can be overrides deployment spec via the node label selector.
1. The closed-loop flow control feature is bug-free.

**Recommended Skills:**

1. Understand K8s and KubeEdge.
1. Go language foundation.

**Useful links:**

- [Node Group Management Proposal](https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/node-group-management.md)

EraKin575:
**What type of PR is this?**

/kind design

A part of #5755

```release-note

```

### DIFF
--- a/docs/proposals/edgeapplication-proposal.md
+++ b/docs/proposals/edgeapplication-proposal.md
@@ -1,0 +1,111 @@
+---
+title: Deployment Spec Override via NodeLabels
+status: implementable
+authors:
+    - "@EraKin575"
+approvers:
+  - TBD
+  - TBD
+creation-date: 2024-09-09
+last-updated: 2024-09-17
+---
+
+# Solution Proposal: Enhancing EdgeApplication for Deployment Specification Overrides and Closed-Loop Flow Control
+
+## 1. Introduction
+
+**Deployment Specification Overrides:**  
+Currently, EdgeApplication allows overrides based on node groups, but this functionality lacks the flexibility to deploy applications to specific nodes based on their labels. This proposal aims to extend this feature by allowing deployment specification overrides based on node labels/selectors.
+
+## 2. Problem Statement
+
+1. **Deployment Specification Overrides:**  
+   EdgeApplication currently supports deployment overrides based on specs like replicas, images, etc., at the node group level. However, this method restricts deployment management's granularity and flexibility. There is a need for deployment overrides that can be applied to specific nodes based on node labels, allowing more fine-tuned control over deployments.
+
+## 3. Proposed Solution
+
+### Deployment Specification Overrides via NodeLabels
+
+To enhance the deployment management capabilities of EdgeApplication, the solution proposes to:
+
+#### 1. API Changes:
+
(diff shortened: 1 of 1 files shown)


wbc6080 (commented):
DCO need to be fixed refer to https://kubeedge.io/docs/developer/Supplemental_developer_guide#dco

On the other hand, the `fix `field in the PR description is inaccurate. This PR is just a proposal and cannot solve the problems in the issue. Merging this PR will cause the issue to be closed.

docs/proposals/edgeapplication-proposal.md
```diff
@@ -0,0 +1,49 @@
...
+    ```yaml
+    spec:
+      deploymentOverrides:
+        nodeLabelSelectors:
+          - key: "example.com/label"
+            operator: "In"
+            values:
+              - "value1"
+              - "value2"
+    ```
```
WillardHu, line 31: Stay consistent with the code.
```yaml
spec:
  ...
  workloadScope:
    targetNode:
       labelselector:
          - <to be added>
          - <to be added>
      overriders:
        ..
```

docs/proposals/edgeapplication-proposal.md
```diff
@@ -0,0 +1,49 @@
...
+
+**2. Controller Logic:**
+
+- Update the EdgeApplicationController to handle the new `nodeLabelSelectors` field.
+- Implement logic to match nodes based on labels and apply deployment overrides accordingly.
```
wbc6080, line 36: It would be better if it could be explained in the form of a flow chart or framework diagram like https://kubeedge.io/docs/architecture/cloud/edge_controller#sync-addupdatedelete-event-to-edge

WillardHu:
Please rebase your commits to one commit. @EraKin575

fujitatomoya (commented):
thanks for sharing the info. this looks really reasonable to have for more fine-granularity development configuration for edge application.

fisherxu (commented):
/approve

PR #5844, opened 2024-09-05
Remove some unused constants in api and commons

wbc6080:
**What type of PR is this?**
/kind cleanup

Some constants in the original API will be placed in the common directory. But as the API is moved to the staging directory, the constants used in the API part and the main part of kubeedge need to be adjusted, and some unreferenced and repeated constants need to be removed.

Fixes #

```release-note

```

### DIFF
--- a/common/constants/default.go
+++ b/common/constants/default.go
@@ -10,14 +10,11 @@ import (
 const (
 	// SyncController
 	DefaultContextSendModuleName = "cloudhub"
-
-	ProjectName = "KubeEdge"
-
-	SystemName      = "kubeedge"
-	SystemNamespace = SystemName
-
-	CloudConfigMapName  = "cloudcore"
-	EdgeMappingCloudKey = "cloudcore"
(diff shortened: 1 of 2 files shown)


common/constants/default.go
```diff
@@ -71,58 +49,6 @@ const (
...
-	DefaultDeviceModelEventBuffer    = 1
-	DefaultUpdateDeviceStatusWorkers = 1
-
-	// TaskManager
-	DefaultNodeUpgradeJobStatusBuffer = 1024
-	DefaultNodeUpgradeJobEventBuffer  = 1
-	DefaultNodeUpgradeJobWorkers      = 1
-
 	// ImagePrePullController
 	DefaultImagePrePullJobStatusBuffer = 1024
```
Shelley-BaoYue, line 53: ImagePrePull related constants need also be removed

wbc6080: Done

Shelley-BaoYue (commented):
/lgtm

PR #5845, opened 2024-09-05
adding deployment spec overrides via node label selector

EraKin575:
**What type of PR is this?**
Adding deployment spec overrides via node label selector to have granular node characteristics.

/kind feature

A part of #5755

```release-note

```


WillardHu:
You can add some description in the "What this PR does / why we need it" part.

EraKin575:
@WillardHu can you please review if I am going to the right track?

EraKin575:
question: 1. How can I update the edgeapplication crd yaml to test these changes?
2.
If both TargetNodeGroups and TargetNodeLabels are defined for the same workload, and each specifies different overrides for the same property (e.g., different NodeSelector values), there could be conflicts. What should I prioritize here?

WillardHu:
> question: 1. How can I update the edgeapplication crd yaml to test these changes? 2. If both TargetNodeGroups and TargetNodeLabels are defined for the same workload, and each specifies different overrides for the same property (e.g., different NodeSelector values), there could be conflicts. What should I prioritize here?

Sorry, I've been a little busy lately, so I'm replying late.

1. You can execute `make generate` to update crds.
2. Each item configuration of the `nodeSelector` and `nodeGroup` generates a new Deployment. If a user configuration error causes multiple Pods of the same application to be deployed to one node, resulting in resource conflict, We can reflect the operation result of deployment in the edgeapplication status first. Then we can judge whether verification is necessary based on user feedback. I think validation becomes expensive when there are a large number of nodes and overriders configurations.  What do you think? @EraKin575  @Shelley-BaoYue  @wbc6080 

However, for deployment generated by `nodeSelector`, we need to define their names in other ways. For example, `<edgeapplication>-<random number>`.

wbc6080 (commented):
You can try to run hack/update-codegen.sh to pass the make verify CI test and run hack/generate-crds.sh to generate new crds

wbc6080:
> What do you think? @EraKin575  @Shelley-BaoYue  @wbc6080

Maybe we can discuss more details in community meeting

EraKin575:
> You can try to run hack/update-codegen.sh to pass the make verify CI test and run hack/generate-crds.sh to generate new crds

Its still generating the same error. I tried to upgrade the controller gen version to latest version and it apparently worked. Should I go ahead with it?

wbc6080 (commented):
@EraKin575  Do you mean that after upgrading the `controller gen` version,` hack/generate-crds.sh` was executed successfully? Based on this success, you can run `hack/update-codegen.sh `again, and then execute `git status`. You should be able to see that some files have been automatically modified. Push these modifications up again, and you should be able to solve the CI error.

EraKin575:
@wbc6080 I have a question, edgeapplicationcontroller uses GVK to identify which resources to override, right?

EraKin575:
Hi! I have a question. Can you please tell me about the matching logic for targetnodegroup overriders? How does it find the node group to override the deployment spec?

wbc6080 (commented):
> Hi! I have a question. Can you please tell me about the matching logic for targetnodegroup overriders? How does it find the node group to override the deployment spec?

Sorry, I didn't understand your question clearly. Are you asking how to build nodes into node groups? Or how to fully populate edge application deployment in different node groups?
If it is the former, we define the `NodeGroup` CRD to describe a node group. Nodes with specific labels can be add to a node group management through the matchlable field.

If it is the latter, in `edge application` CRD , the `Workload Templates` field can define the `Deployment Template` of the edge application, and `WorkloadScopes` can be configured differently according to the needs of different node groups, such as configuring an image repository address. `Workload Templates` and `WorkloadScopes` can be added together to form a complete deployment for deployment.

You can also read nodegroup's proposal for more information https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/node-group-management.md

WillardHu:
> Hi! I have a question. Can you please tell me about the matching logic for targetnodegroup overriders? How does it find the node group to override the deployment spec?

1. Assign a instance `overridemanager.OverrideManager` to `Overrider` failed in controller management.https://github.com/kubeedge/kubeedge/blob/841368a8b412169dff3cd7bb34b12f3fe8cfa9a7/cloud/pkg/controllermanager/controllermanager.go#L63
2. Loops overrider-infos to overrides the Deployment yaml https://github.com/kubeedge/kubeedge/blob/841368a8b412169dff3cd7bb34b12f3fe8cfa9a7/cloud/pkg/controllermanager/edgeapplication/edgeapplicationcontroller.go#L130
3. Loops overriders to executes func `ApplyOverrides(..)` of each Overrider implementation(argsoverrider.go, envoverrider.go, imageoverrider.go, etc...).
4. Updates, creates or deletes k8s resources.

EraKin575:
> > Hi! I have a question. Can you please tell me about the matching logic for targetnodegroup overriders? How does it find the node group to override the deployment spec?
> 
>     1. Assign a instance `overridemanager.OverrideManager` to `Overrider` field in controller management.https://github.com/kubeedge/kubeedge/blob/841368a8b412169dff3cd7bb34b12f3fe8cfa9a7/cloud/pkg/controllermanager/controllermanager.go#L63
> 
>     2. Loops overrider-infos to overrides the Deployment yaml https://github.com/kubeedge/kubeedge/blob/841368a8b412169dff3cd7bb34b12f3fe8cfa9a7/cloud/pkg/controllermanager/edgeapplication/edgeapplicationcontroller.go#L130
> 
>     3. Loops overriders to executes func `ApplyOverrides(..)` of each Overrider implementation(argsoverrider.go, envoverrider.go, imageoverrider.go, etc...).
>        https://github.com/kubeedge/kubeedge/blob/841368a8b412169dff3cd7bb34b12f3fe8cfa9a7/cloud/pkg/controllermanager/edgeapplication/overridemanager/overridemanager.go#L35
> 
>     4. Updates, creates or deletes k8s resources.

thanks for the help @WillardHu , but I knew this part. I am struggling to  find the logic for finding the correct targetnodegroup that  implements the deployments on. Example; 

```
targetNodeGroups:
      - name: beijing
        overriders:
          resourcesOverriders:
            - containerName: container-1
              value: {}
      - name: shanghai
        overriders:
          resourcesOverriders:
            - containerName: container-1
              value: {}
```
where is the code to match the name of targetnodegroup with the actual NodeGroup?

EraKin575:
Hi, @WillardHu can you review this PR ? I have made the required changes and it seems to be working. For now, I have added tests for command overrider and it passed. If you approve the changes, I will add test for every overrider
cc @wbc6080

WillardHu:
You need to restore the changes related to the device @EraKin575

cloud/pkg/controllermanager/edgeapplication/edgeapplicationcontroller.go
```diff
@@ -122,20 +125,52 @@ func (c *Controller) syncEdgeApplication(ctx context.Context, edgeApp *appsv1alp
...
+						errs = append(errs, err)
+						continue
+					}
+
+					if err := utils.ApplyNodeAffinity(tmplCopy, info.TargetNodeLabelSelector); err != nil {
+						klog.Errorf("failed to apply node affinity to obj %s/%s of gvk %s, %v",
+							tmplCopy.GetNamespace(), tmplCopy.GetName(), tmplCopy.GroupVersionKind(), err)
+						errs = append(errs, err)
+						continue
+					}
```
WillardHu, line 166: This code can be put in the nodeselectoroverrider.go

cloud/pkg/controllermanager/edgeapplication/edgeapplicationcontroller.go
```diff
@@ -122,20 +125,52 @@ func (c *Controller) syncEdgeApplication(ctx context.Context, edgeApp *appsv1alp
...
+				} else {
+					klog.Warningf("No nodes found matching labels for obj %s/%s of gvk %s",
+						tmplCopy.GetNamespace(), tmplCopy.GetName(), tmplCopy.GroupVersionKind())
+					continue
```
WillardHu, line 170: It would be best to write the message to the EdgeApplication Status. And the historical error handling doesn't look  good, the Reconcile() function returns an error only make it constant retried.

cloud/pkg/controllermanager/edgeapplication/overridemanager/commandoverrider.go
```diff
@@ -33,9 +34,44 @@ func (o *CommandOverrider) ApplyOverrides(rawObj *unstructured.Unstructured, ove
...
+
+	for key, value := range selector.MatchLabels {
+		affinity["nodeAffinity"].(map[string]interface{})["requiredDuringSchedulingIgnoredDuringExecution"].(map[string]interface{})["nodeSelectorTerms"].([]interface{})[0].(map[string]interface{})["matchExpressions"] = append(
+			affinity["nodeAffinity"].(map[string]interface{})["requiredDuringSchedulingIgnoredDuringExecution"].(map[string]interface{})["nodeSelectorTerms"].([]interface{})[0].(map[string]interface{})["matchExpressions"].([]interface{}),
+			map[string]interface{}{
+				"key":      key,
+				"operator": "In",
+				"values":   []interface{}{value},
+			},
+		)
```
WillardHu, line 69: Here, the processing performance is better with map, but the readability is worse. Is it better to use jsonpath?

WillardHu:
The nameoverrider.go file is also to be modified

WillardHu:
> where is the code to match the name of targetnodegroup with the actual NodeGroup where it does the deployment?

Refer to nodeselectoroverrider.go, when each node is bound by a node group, the node automatically adds a label `apps.kubeedge.io/belonging-to`

cloud/pkg/controllermanager/edgeapplication/overridemanager/imageoverrider.go
```diff
@@ -37,6 +37,12 @@ func (o *ImageOverrider) ApplyOverrides(rawObj *unstructured.Unstructured, overr
 			return err
 		}
 	}
+	// Apply node affinity based on TargetNodeLabelSelector
+	if len(overriders.TargetNodeLabelSelector.MatchLabels) > 0 {
+		if err := ApplyNodeAffinity(rawObj, overriders.TargetNodeLabelSelector); err != nil {
+			return fmt.Errorf("failed to apply node affinity: %v", err)
```
WillardHu, line 43: There is no need to set up the node affinity again

cloud/pkg/controllermanager/edgeapplication/edgeapplicationcontroller.go
```diff
@@ -122,20 +125,28 @@ func (c *Controller) syncEdgeApplication(ctx context.Context, edgeApp *appsv1alp
...
+				}
+			}
+			if info.TargetNodeLabelSelector.MatchLabels != nil {
+				if err := utils.ApplyNodeAffinity(tmplCopy, info.TargetNodeLabelSelector); err != nil {
+					klog.Errorf("failed to apply node affinity to obj %s/%s of gvk %s, %v",
+						tmplCopy.GetNamespace(), tmplCopy.GetName(), tmplCopy.GroupVersionKind(), err)
+				}
 				continue
 			}
```
WillardHu, line 147: This piece of logic moved to the nodeselectoroverrider.go will be more

EraKin575:
@WillardHu hi! I have passed the integration test but there is error in verify test and I think its due to device CRD.

EraKin575:
@WillardHu please review the PR
CC @wbc6080

Issue #5060, opened 2023-10-09
边缘节点在断网情况下重启，使用了localpath PV卷的pod无法自动启动


youthantennalemon:
问题：边缘节点在断网情况下重启，使用了localpath PV卷的pod无法自动启动
希望的结果：边缘节点在断网情况下重启，使用了localpath PV卷的pod可以自动被edgecore拉起

复现过程
1. 安装local-path-provisioner作为localPV卷的提供者
2. 在边缘节点启动一个使用deployment控制器控制并且使用localPV卷的应用
3. 断开边缘节点的网络连接（拔掉网线，模拟边缘节点离线情况）
4. 重启边缘节点服务器后发现使用localPV卷的应用pod无法自动启动，其他不带PV卷的pod可以自动启动

查看边缘节点上edgecore的日志发现edgecore需要与云端的API server通信获取PV卷的什么信息，导致无法拉起pod
edgecore[1680]: E1009 11:26:12.084523    1680 pod_workers.go:915] "Error syncing pod, skipping" err="unmounted volumes=[database], unattached volumes=[database kube-api-access-j7vqd protocol-conf]: error processing PVC gw-test-01/elinker-database: failed to fetch PV pvc-3c4b66d1-86a1-4084-aef3-ebabd683e125 from API server: persistentvolume length from meta db is 0" pod="gw-test-01/edge-gw-5f48d56fd6-g8lpn" podUID=9a9d3349-48c4-4183-bb18-5dfb4cfad395

有没有办法让边缘节点在离线情况下重启后能自动拉起使用了localpath PV卷的pod？

版本信息
- Kubernetes version  v1.22.12
- KubeEdge version v1.12.1

Shelley-BaoYue:
It seems the edge DB doesn't save the persistentvolume data, so edged cannot get persistentvolume date when offline.  I checked the code of edged that metemanager does save the persistentvolume data in edge DB.  
Could you please repeat your step 1&2? Raise the log level and after deploying the pod successfully, using `sqlite3 /var/lib/kubeedge/edgecore.db` to see if persistentvolume data has been stored in edge DB.

youthantennalemon:
在启动了带有PV卷的pod后，在/var/lib/kubeedge/edgecore.db的meta表里面可以看到一条关于pvc的信息

youthantennalemon:
[...]

PR #5855, opened 2024-09-11
Fix PersistentVolume data deleted at edge abnormally

Shelley-BaoYue:
**What type of PR is this?**

/kind bug

SyncController  send delete persistentvolume msg to edge, causing persistentvolume data at edge deleted. When edge node restart, pod cannot query pv data successfully. 

**Reason**:

1. When edged query pv from cloud, mataclient sets a default ns in msg. 
2. In CloudHub https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/session/node_session.go#L381, it will be operated as a `NamespaceResource`,  CloudHub will create a corresponding objectSync for pv (ref: https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/cloudhub/session/node_session.go#L441).  But actually **it should create a clusterObjectSync for pv** because pv is a cluster resource. 
3. SyncController lists pv as a namespace resource but it cannot get pv by namespace, so it mistakenly thinking that the pv was deleted, and send a deletion msg to edge.   

**in this PR**:

set `null` namespace for pv msg

Fixes https://github.com/kubeedge/kubeedge/issues/5060


fisherxu (commented):
/approve
/assign @WillardHu

fisherxu (commented):
/lgtm

PR #5856, opened 2024-09-11
Upgrade github actions to v4

Shelley-BaoYue:
**What type of PR is this?**

/kind failing-test

Upgrade github actions to v4

Some CI failed because we still use v1/v2 github action, which are deprecated. Ref to: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/


fisherxu (commented):
/lgtm
/approve

Issue #5851, opened 2024-09-09
the meaning of lastversion shown in https://kubeedge.io/latestversion


magichan:
**What happened and what you expected to happen**:

While reading the code, I found that the result of https://kubeedge.io/latestversion is v1.15.1, but the latest version of KubeEdge is v1.18.0.

Could v1.15.1 be a different manifest version, such as the Helm package for KubeEdge?

**How to reproduce it (as minimally and precisely as possible)**:
```sh
wget https://kubeedge.io/latestversion
```
**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`):

magichan:
@Shelley-BaoYue  
If you agree that this is a bug,
could you please help me change the label from kind/question to kind/bug?

Shelley-BaoYue:
You can help update version here https://github.com/kubeedge/website/blob/master/functions/latestversion.js#L3

PR #5859, opened 2024-09-12
update default kubeedge version in keadm

magichan:
**What type of PR is this?**

/kind bug

When keadm runs without specifying a KubeEdge version, it defaults to installing version v1.15.1, which is outdated.

I suggest updating the default version to v1.18.0. Although the code will be released in v1.19.0, I believe v1.18.0 is more appropriate as it is more stable.

Test: all help information use v1.18.0
image

Test:  In scenarios where both the  client version and the remote version provided by the KubeEdge website are unavailable, use the v1.18.0

image

Test: mock remote verison fail suition , use client verison 
image

Fixes https://github.com/kubeedge/kubeedge/issues/5851

When the remote version provided by the KubeEdge website is functioning, the KubeEdge version defaults to v1.15.1 if no specific version is specified.  but the condition will change after the https://github.com/kubeedge/website/pull/638 merge

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/helm/cloudcore.go
+++ b/keadm/cmd/keadm/app/cmd/helm/cloudcore.go
@@ -124,7 +124,7 @@ func (c *CloudCoreHelmTool) Install(opts *types.InitOptions) error {
 		return fmt.Errorf("failed to verify k8s component installed, err: %v", err)
 	}
 
-	fmt.Println("Kubernetes version verification passed, KubeEdge installation will start...")
+	fmt.Printf("Kubernetes version verification passed, KubeEdge %s installation will start...\n", opts.KubeEdgeVersion)
 
 	appendDefaultSets(opts.KubeEdgeVersion, opts.AdvertiseAddress, &opts.CloudInitUpdateBase)
 	// Load profile values, and merges the sets flag
--- a/keadm/cmd/keadm/app/cmd/common/constant.go
+++ b/keadm/cmd/keadm/app/cmd/common/constant.go
@@ -120,7 +120,7 @@ const (
 	DefaultK8SMinimumVersion = 11
 
 	// DefaultKubeEdgeVersion is the default KubeEdge version, it must have no prefix 'v'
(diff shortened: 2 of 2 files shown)


magichan:
@fisherxu  @WillardHu @Shelley-BaoYue  The MR is ready for merge, please review.

keadm/cmd/keadm/app/cmd/helm/cloudcore.go
```diff
@@ -124,7 +124,7 @@ func (c *CloudCoreHelmTool) Install(opts *types.InitOptions) error {
 		return fmt.Errorf("failed to verify k8s component installed, err: %v", err)
 	}
 
-	fmt.Println("Kubernetes version verification passed, KubeEdge installation will start...")
+	fmt.Printf("Kubernetes version verification passed, KubeEdge(version: %s) installation will start...\n", opts.KubeEdgeVersion)
```
wbc6080, line 127: Will it be better if you don’t add brackets? For example, `KubeEdge v1.18.0 installation will start...`

magichan: Thanks, 
Update by review.

image

fisherxu: Is it possible that version here is empty or other wrong msg? Should be not.

fisherxu: Or users can't understand what's msg meaning after KubeEdge

magichan: @fisherxu  Combining code and experiments, version information will always emerge.

In the most extreme cases:

Incorrect version input in the command
Incorrect version information of the Keadm client
Incorrect remote version information
The version information will appear as the default version information hardcoded in the code.

fisherxu: Sounds good, thanks :)

fisherxu (commented):
/lgtm

fisherxu (commented):
/approve

PR #5860, opened 2024-09-13
Change deviceName to deviceID in mapper data panel

wbc6080:
**What type of PR is this?**
/kind bug

Previously, #5329  changed the key for device storage in mapper, edgecore, and cloudcore from deviceName to deviceID, which can avoid interference between the same devicenames in different namespaces. Based on this, we also changed the key of the database in the mapper data plane from devicenames to deviceID to avoid possible interference caused by devices with the same name.

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/tdengine/client.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/tdengine/client.go
@@ -60,17 +60,17 @@ func (d *DataBaseConfig) CloseSessio() {
 func (d *DataBaseConfig) AddData(data *common.DataModel) error {
 
 	tableName := data.Namespace + "/" + data.DeviceName
-	legal_table := strings.Replace(tableName, "-", "_", -1)
-	legal_tag := strings.Replace(data.PropertyName, "-", "_", -1)
+	legalTable := strings.Replace(tableName, "-", "_", -1)
+	legalTag := strings.Replace(data.PropertyName, "-", "_", -1)
 
-	stable_name := fmt.Sprintf("SHOW STABLES LIKE '%s'", legal_table)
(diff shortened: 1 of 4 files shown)


magichan:
@wbc6080  I noticed different table structures and write methods in the MR. From my personal experience, we should be cautious about potential breaking changes in such cases.

However, when reviewing the code, I only found the `AddData` method for various databases, but I couldn't locate any read methods.

If there are no read methods, there won't be any breaking changes. But if we're not reading the data, why are we saving it?

Could you please help me clarify my confusion?"

wbc6080:
> Could you please help me clarify my confusion?"

The data plane of the mapper is used to push device data. When designing, we provide the ability to push data to the user's application and user database, which is the `AddData` method here. As for the user reading of data, it is not done in the mapper. It is handled by the user, so there is no function to read data in mapper.

magichan:
>  As for the user reading of data, it is not done in the mapper. It is handled by the user, so there is no function to read data in mapper.

Thx, 

If user know the change, it is ok.

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/tdengine/client.go
```diff
@@ -59,15 +59,16 @@ func (d *DataBaseConfig) CloseSessio() {
 
 func (d *DataBaseConfig) AddData(data *common.DataModel) error {
 
-	legal_table := strings.Replace(data.DeviceName, "-", "_", -1)
+	tableName := data.Namespace + "/" + data.DeviceName
+	legal_table := strings.Replace(tableName, "-", "_", -1)
```
WillardHu, line 63: Would it be more appropriate to name all variables as the hump specification?

wbc6080: Done

fisherxu (commented):
/approve

PR #5885, opened 2024-09-24
Fix too many device status messages

JiaweiGithub:

**What type of PR is this?**
/kind bug

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
@@ -150,10 +150,11 @@ func dataHandler(ctx context.Context, dev *driver.CustomizedDev) {
 		}
 		go twinData.Run(ctx)
 
-		//handle status
-		getStates := &DeviceStates{Client: dev.CustomizedClient, DeviceName: dev.Instance.Name,
-			DeviceNamespace: dev.Instance.Namespace}
-		go getStates.Run(ctx)
+		// todo added report to cloud flag and frequency control
+		// handle status
+		// getStates := &DeviceStates{Client: dev.CustomizedClient, DeviceName: dev.Instance.Name,
+		//	DeviceNamespace: dev.Instance.Namespace}
+		// go getStates.Run(ctx)
 
 		// handle push method
 		if twin.Property.PushMethod.MethodConfig != nil && twin.Property.PushMethod.MethodName != "" {


wbc6080 (commented):
As I mentioned in the issue #5892 , reporting device status to the cloud is currently enabled by default, and the frequency of status reporting cannot be set in the device CRD. This will have an impact on our use and debugging of the mapper. So we may be able to firstly set this feature to off by default, and we are waiting for subsequent PRs to implement the feature that mapper can set the status reporting frequency in v1.19.
/lgtm
cc @WillardHu @fisherxu @luomengY

WillardHu:
> As I mentioned in the issue #5892 , reporting device status to the cloud is currently enabled by default, and the frequency of status reporting cannot be set in the device CRD. This will have an impact on our use and debugging of the mapper. So we may be able to firstly set this feature to off by default, and we are waiting for subsequent PRs to implement the feature that mapper can set the status reporting frequency in v1.19. /lgtm cc @WillardHu @fisherxu @luomengY

If this is a temporary solution, it would be more appropriate to comment the calling code. The ReportToCloud field is also inputted as a parameter in here.
```golang
// _template/mapper/device/device.go#154

// TODO .....
// getStates := &DeviceStates{Client: dev.CustomizedClient, DeviceName: dev.Instance.Name,
//     DeviceNamespace: dev.Instance.Namespace}
// go getStates.Run(ctx)
```

fisherxu (commented):
IIRC, we have already `reported to cloud` field in device API?

wbc6080:
> IIRC, we have already `reported to cloud` field in device API?

Yes, but in device status report process, we didn't use this field in initial version. Therefore, the current mapper will always report the device status and cannot be controlled. Subsequent PRs need to include reporting to cloud controls, as well as reporting frequency controls.

wbc6080:
> If this is a temporary solution, it would be more appropriate to comment the calling code. The ReportToCloud field is also inputted as a parameter in here.


That's okay too. All in all, we really need to add control of reporting to the cloud and control of reporting frequency in our follow-up work.

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/devicestatus.go
```diff
@@ -56,6 +56,12 @@ func (deviceStates *DeviceStates) PushStatesToEdgeCore() {
 }
 
 func (deviceStates *DeviceStates) Run(ctx context.Context) {
+	// TODO Too many message reports will affect the debug of other feature.
+	// Default close, report to cloud flag and frequency control will be added later.
+	reportToCloud := false
+	if !reportToCloud {
+		return
+	}
```
WillardHu, line 64: Remove these codes

JiaweiGithub: done

wbc6080 (commented):
/lgtm

fisherxu (commented):
It's worth adding a todo above the commented code and open a tracking issue :)

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
```diff
@@ -151,9 +151,9 @@ func dataHandler(ctx context.Context, dev *driver.CustomizedDev) {
 		go twinData.Run(ctx)
 
 		//handle status
-		getStates := &DeviceStates{Client: dev.CustomizedClient, DeviceName: dev.Instance.Name,
-			DeviceNamespace: dev.Instance.Namespace}
-		go getStates.Run(ctx)
+		//getStates := &DeviceStates{Client: dev.CustomizedClient, DeviceName: dev.Instance.Name,
+		//	DeviceNamespace: dev.Instance.Namespace}
+		//go getStates.Run(ctx)
```
fujitatomoya, line 156: I would suggest that instead of commented code, we probably want to `TODO` description what needs to be done here, so that we can always see the previous code in git history. after all, i do not think commented code is useful here.

JiaweiGithub:
> It's worth adding a todo above the commented code and open a tracking issue :)值得在注释代码上方添加待办事项并打开跟踪问题:)

done

Issue #4471, opened 2022-12-06
Failed to list *v1.Namespace using metaServer


bluven:
**What happened**:
I was trying to run coredns using metaServer on edge node, but encountered two problems:
1. When I used coredns with higher version than 1.8.0, the coredns always complained "kubernetes is not ready", so I use coredns 1.8.0 now
```
[INFO] plugin/ready: Still waiting on: "kubernetes"
[INFO] plugin/ready: Still waiting on: "kubernetes"
[INFO] plugin/ready: Still waiting on: "kubernetes"
[INFO] plugin/ready: Still waiting on: "kubernetes"
[INFO] plugin/ready: Still waiting on: "kubernetes"
```

2. Coredns kept complaining something like this:
```
[INFO] plugin/ready: Still waiting on: "kubernetes"
[INFO] plugin/ready: Still waiting on: "kubernetes"
[INFO] plugin/ready: Still waiting on: "kubernetes"
.:53 on 169.254.25.10
[INFO] plugin/reload: Running configuration MD5 = d0ee041a71b7c0461847e95e0216c16b
CoreDNS-1.8.0
linux/amd64, go1.15.3, 054c9ae
I1206 05:13:23.885230       1 trace.go:205] Trace[2019727887]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.19.2/tools/cache/reflector.go:156 (06-Dec-2022 05:13:13.877) (total time: 10008ms):
Trace[2019727887]: ---"Objects listed" 10008ms (05:13:00.885)
Trace[2019727887]: [10.008080997s] [10.008080997s] END
I1206 05:13:23.886757       1 trace.go:205] Trace[939984059]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.19.2/tools/cache/reflector.go:156 (06-Dec-2022 05:13:13.876) (total time: 10009ms):
Trace[939984059]: ---"Objects listed" 10009ms (05:13:00.886)
Trace[939984059]: [10.00997647s] [10.00997647s] END
I1206 05:13:23.894182       1 trace.go:205] Trace[1474941318]: "Reflector ListAndWatch" name:pkg/mod/k8s.io/client-go@v0.19.2/tools/cache/reflector.go:156 (06-Dec-2022 05:13:13.875) (total time: 10018ms):
Trace[1474941318]: [10.018210657s] [10.018210657s] END
[...]

PR #5894, opened 2024-09-26
Fix errors due to singular and plural conversion in metaserver

### LINKED ISSUE #5890: Get a DeviceList through the API provided by MetaServer on the edge. The Kind field in the return information is incorrect.
**What happened**:

When I get `*v1beta1.DeviceList` using the `clientSet` provided by kubeedge:
```go
	deviceList, err := versionedClient.DevicesV1beta1().Devices(deviceRequest.Namespace).List(ctx, metaV1.ListOptions{
		LabelSelector: deviceRequest.LabelSelector,
	})
	if err != nil {
		return nil, fmt.Errorf("failed to list devices: %v", err)
	}
```
an error occurs:
error

Then I call MetaServer API directly: 
```shell
curl 'http://127.0.0.1:10550/apis/devices.kubeedge.io/v1beta1/namespaces/default/devices'
```
The Kind field in the return information is `DevicList` :
image

The Kind field in the return information is `DeviceList`.

This bug might be related to suffix handling.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.28.14
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): v1.18.0
(I also found this bug in kubeedge1.17)

wbc6080:
**What type of PR is this?**
/kind bug

As mentioned in the #5890 and #4471 , metaserver does not accurately convert singular and plural type when parsing resource types. It converts `devices` into `devic` and `namesapces` into `namespac`, so the resource type obtained is incorrect. So we need to modify the method of singular and plural conversion.

Related to #5890 

```release-note

```


wbc6080 (commented):
Since the rules for converting English words between singular and plural are complicated, we use the `github.com/gertd/go-pluralize `package for conversion. I tested the performance of this package, and the results are as follows. We can see that this package can more accurately convert various types of words between singular and plural.
image

pkg/metaserver/util/util.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"strings"
 
+	pluralize "github.com/gertd/go-pluralize"
```
fisherxu, line 7: Looks good, but it's better to keep same with Kubernetes. Does K8s also use this pkg?

fisherxu: It's better to keep same with K8s.

fujitatomoya: @wbc6080 thanks for taking care of this. this is tricky but annoying problem.

as @fisherxu mentions, probably we would want to go with k8s approach for now. but your concern also stands, we would want keep the note for future possible problems though.

wbc6080: I have made modifications based on the processing method in k8s, and also retained the original issue as a reference.

pkg/metaserver/util/util.go
```diff
@@ -66,7 +66,7 @@ func UnsafeResourceToKind(r string) string {
 	switch {
 	case strings.HasSuffix(k, "ies"):
 		return strings.TrimSuffix(k, "ies") + "y"
-	case strings.HasSuffix(k, "es"):
+	case strings.HasSuffix(k, "ses"):
```
fisherxu, line 69: Is it need for case `es`?

wbc6080: In the implementation of k8s, `es` is added only when the word ends with `s`. https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/restmapper.go#L142

wbc6080:
PTAL, thanks
cc @WillardHu @Shelley-BaoYue @fisherxu

fisherxu (commented):
/approve

PR #5897, opened 2024-09-27
Skip kube-proxy during restarting edgecore.

### LINKED ISSUE #5865: `systemctl restart edgecore` fails since `kube-proxy` is always running.
**What happened**:

Every single time after we change the `/etc/kubeedge/config/edgecore.yaml` configuration such as `nodeIP` or anything else, `systemctl restart edgecore` always fails because `kube-proxy` is running.

`systemctl restart edgecore` should succeed.

1. start `edgecore` with `keadm` with https://kubeedge.io/docs/setup/install-with-keadm
2. `systemctl restart edgecore`

I believe that basic expectation here is that `edgecore` is statically configured with `/etc/kubeedge/config/edgecore.yaml` before starting the system service, and the start the service. in production phase this sound reasonable, but this is actually the problem because this behavior blocks the system service auto-restarting for sure. that is said, `edgecore` will never be running because of `kube-proxy` process.
besides, during development, it is likely to restart the edgecore with updating the configuration. eventually user needs to kill the kube-proxy every time, and i think that is not really good user-experience.

related code is,

https://github.com/kubeedge/kubeedge/blob/486b978826c20958948193430519c97a536ebd80/edge/cmd/edgecore/app/server.go#L186-L190

- checking `kubelet` sounds correct error here to stop starting `edgecore` since it should never be running on edge node.
- `kube-proxy` is expected to be running on edge node once it joins to the cluster, so it should kill it during restart if it has to? or skip checking it, so that k8s can recognize this pod?

fujitatomoya:
**What type of PR is this?**

/kind bug

This PR is minor fix to skip checking `kube-proxy` when it restarts the edgecore.
Since `edgecore` and `kube-proxy` does not conflict each other, and lifecycle is not dependent, proceed to restart the edgecore.

Fixes #5865

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -185,10 +185,7 @@ func environmentCheck(skipCheck bool) error {
 		}
 		switch processName {
 		case "kubelet": // if kubelet is running, return error
-			return errors.New("kubelet should not running on edge node when running edgecore")
-		case "kube-proxy":
-			// kube-proxy pod does not conflict with edgecore, proceed to restart edgecore
-			klog.Infof("kube-proxy is running when restarting edgecore")
+			return errors.New("kubelet should not be running on edge node when starting edgecore")
 		}
 	}
 
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -186,8 +186,9 @@ func environmentCheck(skipCheck bool) error {
 		switch processName {
 		case "kubelet": // if kubelet is running, return error
 			return errors.New("kubelet should not running on edge node when running edgecore")
-		case "kube-proxy": // if kube-proxy is running, return error
-			return errors.New("kube-proxy should not running on edge node when running edgecore")
+		case "kube-proxy":
+			// kube-proxy pod does not conflict with edgecore, proceed to restart edgecore
+			klog.Infof("kube-proxy is running when restarting edgecore")
 		}
 	}
 


fujitatomoya:
@wbc6080 @WillardHu could either of you take a look at this?

edge/cmd/edgecore/app/server.go
```diff
@@ -186,8 +186,9 @@ func environmentCheck(skipCheck bool) error {
 		switch processName {
 		case "kubelet": // if kubelet is running, return error
 			return errors.New("kubelet should not running on edge node when running edgecore")
-		case "kube-proxy": // if kube-proxy is running, return error
-			return errors.New("kube-proxy should not running on edge node when running edgecore")
+		case "kube-proxy":
+			// kube-proxy pod does not conflict with edgecore, proceed to restart edgecore
+			klog.Infof("kube-proxy is running when restarting edgecore")
```
fisherxu, line 191: Now `kube-proxy` can indeed run on edge nodes! And this log will also be printed when edgecore is started for the first time, could users understand what it means? :)

fujitatomoya: Ah, that is true. that does not really explain what to do for user...
So we should do keep it silence? not checking nor printing anything? i guess this sounds fine too.
any thoughts?

fisherxu: I think we can skip the kube-proxy check here. Because it can run on edgenode now, plz take a look. cc @WillardHu @Shelley-BaoYue

fujitatomoya: agree on this, fixed in https://github.com/kubeedge/kubeedge/pull/5897/commits/2e8b274056e5724dc47d8e4dee7673af6b04022c

edge/cmd/edgecore/app/server.go
```diff
@@ -185,9 +185,7 @@ func environmentCheck(skipCheck bool) error {
 		}
 		switch processName {
 		case "kubelet": // if kubelet is running, return error
-			return errors.New("kubelet should not running on edge node when running edgecore")
-		case "kube-proxy": // if kube-proxy is running, return error
-			return errors.New("kube-proxy should not running on edge node when running edgecore")
+			return errors.New("kubelet should not be running on edge node when starting edgecore")
```
fujitatomoya, line 188: this is just a cosmetic message fix.

fisherxu (commented):
Looks good, thanks!
/approve

PR #5898, opened 2024-09-27
Add an action in the conformance CI to increase available disk space 

bzsuni:
**What type of PR is this?**
/kind test

Add an action in the conformance CI to increase available disk space to prevent test execution failures due to insufficient disk capacity.

ref #5774
https://github.com/kubeedge/kubeedge/actions/runs/11026517530/job/30634159887?pr=5774

As follows, there may be issues with test failures due to disk pressure：
```yaml
status:
            conditions:
            - lastProbeTime: null
              lastTransitionTime: "2024-09-25T05:44:10Z"
              message: '0/2 nodes are available: 1 node(s) had untolerated taint {node.kubeedge.io/conformance:
                remove-when-completed}, 1 node(s) had untolerated taint {node.kubernetes.io/disk-pressure:
                }. preemption: 0/2 nodes are available: 2 Preemption is not helpful for scheduling.'
              reason: Unschedulable
```

Fixes #

Before cleanup:
```bash
$ dh -h /

Filesystem      Size  Used Avail Use% Mounted on
/dev/root        73G   54G   20G  74% /
```

After cleanup:
```bash
+ df -Th
Filesystem     Type   Size  Used Avail Use% Mounted on
/dev/root      ext4    73G   28G   45G  39% /
```

```release-note
none
```

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -420,7 +420,6 @@ jobs:
 
         - name: Install dependencies
           run: |
-            command -v ginkgo || go install github.com/onsi/ginkgo/v2/ginkgo@${{ env.GINKGO_VERSION }}
             go install sigs.k8s.io/kind@v0.22.0
             curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.29.5/bin/linux/amd64/kubectl && sudo install kubectl /usr/local/bin/kubectl
 
@@ -428,9 +427,6 @@ jobs:
           uses: actions/checkout@v4
           with:
             fetch-depth: 0
-
-        - name: Cleanup images
-          run: docker system prune -a -f
 
         - name: Retrieve saved kubeedge/build-tools image
           uses: actions/download-artifact@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -441,11 +437,22 @@ jobs:
...
 
+        - name: Increase disk space
+          uses: fcbg-platforms/increase-disk-space@latest
+          with:
+            tool-cache: true
+            android: true
+            dotnet: true
+            haskell: true
+            large-packages: true
+            old-dotnet: true
```
fisherxu, line 454: Any docs for what's meaning of these fields?

bzsuni: On the official page https://github.com/jlumbroso/free-disk-space, there isn't a detailed explanation, and it only states "Most of the options are self-explanatory."
image


However, it seems that:
- android, dotnet, haskell refer to whether you want to clean up libraries related to Android, .NET, and Haskell, respectively.
- docker-images indicates whether you want to clean up Docker images.
- swap-storage refers to whether to disable the swap partition.
- The large-packages option is not very clear regarding which contents it cleans up, but it doesn't seem to negatively impact the conformance workflow.

fisherxu: Thanks, we can add `docker-images` in future if the disk space is still not enough.

fisherxu (commented):
/lgtm
/assign @Shelley-BaoYue @WillardHu

fisherxu (commented):
/lgtm

Issue #5857, opened 2024-09-12
Follow-up Maintenance Strategy for `Viaduct` Repository


Shelley-BaoYue:
**What would you like to be added/modified**: 

Migrate `viaduct` directly to `https://github.com/kubeedge/kubeedge/tree/master/pkg`, and no longer maintain https://github.com/kubeedge/viaduct repository. 

**Why is this needed**:

After research and dicussions on community meetings and TSC meetings, we have found that there are currently no scenarios for using [viaduct](https://github.com/kubeedge/viaduct)  repository independently. To better maintain this part, we have decided to migrate `viaduct` directly to `https://github.com/kubeedge/kubeedge/tree/master/pkg`. 

The original [viaduct](https://github.com/kubeedge/viaduct)  repository will be temporarily retained. If you need this repository to be maintained independently, please reply in this issue, and we will adjust our maintenance stragegy. 

If  there are no objections within a week, I will begin the migration process on Sept. 20th.

PR #5907, opened 2024-10-11
Migrate `viaduct` to kubeedge/pkg

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

Migrate viaduct directly to https://github.com/kubeedge/kubeedge/tree/master/pkg as #5857 described. 

Tempararily remove the unit-test in viaduct because this ut do not match the quic-go version and cannot run successfully. I also open an issue #5910  to track the guic-go version update and unit-test supplement.  

Fixes https://github.com/kubeedge/kubeedge/issues/5857


Shelley-BaoYue:
The fuzzing test failed for the viaduct path has been changed, need https://github.com/cncf/cncf-fuzzing/pull/527 merged first.

Shelley-BaoYue (commented):
/assign @fisherxu @WillardHu @luomengY @wbc6080 @tangming1996

PR #5913, opened 2024-10-17
Mapper support set device status reportcycle

### LINKED ISSUE #5892: Mapper supports setting devicestatus reporting frequency

Users can set whether to report device status to the cloud in device CRD, and set the frequency of reporting.

**Why is this needed**:

In 1.18, we implemented the feature of device status collection and reporting. But currently the device status is reported to the cloud by default, and the frequency of device status reporting cannot be controlled through device CRD. This will result in frequent output of logs reporting device status when debugging and using mapper, and the default reporting cycle may cause too many requests error.  So we need to add the `reportToCloud` flag to control whether the device status is reported to the cloud, and users can also control the status reporting cycle through device CRD.

自动上报status

status太多

wbc6080:
**What type of PR is this?**
/kind feature

In #5892, we find the device status is reported to the cloud by default currently, and the frequency of device status reporting cannot be controlled through device CRD. In this PR, we add the `reportToCloud` and `reportCycle` field in `device.status` of device-instance crd, allowing users to set whether to push the device status to the cloud and the push frequency in the device-instance yaml.

Fixes #5892

```release-note

```


wbc6080:
In the original `device-instance CRD`, user can set whether to report to the cloud and the reporting frequency in each device property. 
image

Since the reporting frequency of each device property may be different, device status cannot reuse the value set in the device property.In this PR, we set these two fields separately for the device status in the `device-instance CRD` to control whether the device status is reported to the cloud and the frequency of reporting. An example configuration file of device-instance after adding them is as follows:
image

wbc6080:
Ready for review. 
/assign @luomengY @JiaweiGithub @WillardHu

luomengY:
`ReportToCloud` seems to already have this feature.

luomengY:
image

wbc6080:
> `ReportToCloud` seems to already have this feature.

This is true, but this field is used to control device property and cannot be used directly to control the reporting of device status.

staging/src/github.com/kubeedge/api/apis/dmi/v1beta1/api.proto
```diff
@@ -367,6 +367,11 @@ message ReportDeviceStatesRequest {
 message DeviceStatus {
     // the device twins of the device.
     repeated Twin twins = 1;
+    // whether be reported to the cloud
+    bool ReportToCloud = 2;
```
fisherxu, line 371: Start with the lowercase as below?

fisherxu (commented):
/approve
cc @WillardHu @JiaweiGithub

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/devicestatus.go
```diff
@@ -49,15 +51,22 @@ func (deviceStates *DeviceStates) PushStatesToEdgeCore() {
...
-	ticker := time.NewTicker(2 * time.Second)
+	// No need to report device status to the cloud
+	if !deviceStates.ReportToCloud {
+		return
+	}
+	// Set device status report cycle
+	if deviceStates.ReportCycle == 0 {
+		deviceStates.ReportCycle = common.DefaultReportCycle
+	}
+	ticker := time.NewTicker(deviceStates.ReportCycle)
```
WillardHu, line 69: Define a minimum unit of time. `deviceStates.ReportCycle*time.Second` or `deviceStates.ReportCycle*time.Millisecond`

WillardHu:
After this PR is merged, will the other one https://github.com/kubeedge/kubeedge/pull/5885 not be needed?

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/devicestatus.go
wbc6080: We have defined the time unit during initialization
image

PR #5914, opened 2024-10-17
Add a note comment for default value modification of `edged.rootDirectory`

Shelley-BaoYue:
**What type of PR is this?**

/kind documentation

Add a note comment for change default value of `edged.rootDirectory`, from `/var/lib/edged` to `/var/lib/kubelet`, to keep consistent with kubelet.  

File directories consistent with kubelet can reduce maintain costs. In addition, kubeedge can seamlessly connect to native Kubernetes resources.

Fixes #

```release-note
In v1.20, the default value of edged.rootDirectory will be changed to "/var/lib/kubelet". If you want to continue using "/var/lib/edged", please modify the edgecore configuration. 
```

### DIFF
--- a/staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/types.go
+++ b/staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/types.go
@@ -766,6 +766,7 @@ type TailoredKubeletFlag struct {
 	// rootDirectory is the directory path to place kubelet files (volume
 	// mounts,etc).
 	// default "/var/lib/edged"
+	// The default will change to "/var/lib/kubelet" in KubeEdge v1.20.
 	RootDirectory string `json:"rootDirectory,omitempty"`
 	// WindowsService should be set to true if kubelet is running as a service on Windows.
 	// Its corresponding flag only gets registered in Windows builds.


Shelley-BaoYue:
/hold
wait for discussion.

fujitatomoya (commented):
@Shelley-BaoYue i believe that this is a right way to go, but only questions is there would be conflict possibility with `kubelet` in any ways? for example, i do not think this is likely but if user is trying both `kubelet` and `edgecore` for testing or something (production phase, this is not gonna happen), they do not want to see the conflicts and unexpected problems.

Shelley-BaoYue:
> @Shelley-BaoYue i believe that this is a right way to go, but only questions is there would be conflict possibility with `kubelet` in any ways? for example, i do not think this is likely but if user is trying both `kubelet` and `edgecore` for testing or something (production phase, this is not gonna happen), they do not want to see the conflicts and unexpected problems.

@fujitatomoya Thank you for your advice. It's true that the question you raised has not been considered before. Although kubelet and edgecore cannot run on edge nodes at the same time https://github.com/kubeedge/kubeedge/blob/master/edge/cmd/edgecore/app/server.go#L187, I will seriously consider whether there is a possibility of running at the same time in the future.

fujitatomoya:
> Although kubelet and edgecore cannot run on edge nodes at the same time

yeah, i am aware of this. i was thinking more like, deleting the files under `/var/lib/kubelet` that kind of unexpected problem for each other. sorry i do not have actual conflict use case to point that out, maybe this is false alarm... just wanted to know if there is a possibility. thanks for checking!

Shelley-BaoYue:
> > Although kubelet and edgecore cannot run on edge nodes at the same time
> 
> yeah, i am aware of this. i was thinking more like, deleting the files under `/var/lib/kubelet` that kind of unexpected problem for each other. sorry i do not have actual conflict use case to point that out, maybe this is false alarm... just wanted to know if there is a possibility. thanks for checking!

I have thought this problem, but as edged and kubelet are same in functionalities, there may be relatively few scenarios deploy both on same node. If such a situation does raise in the future, we'll need to consider not only directory conflict, but also port conflicts and other problem. 😄  

To ensure compatibility during version upgrades, I plan to first update the node comment in the upcoming 1.19 and note it in changelog. And make this modification in v1.20.

fisherxu (commented):
/approve
/assign @WillardHu

PR #5935, opened 2024-10-28
Add windows crossbuild in CI

fisherxu:
**What type of PR is this?**
/kind test

Now the windows release is mainly focus on the edgecore/keadm, so adding edgecore/keadm/cloudcore crossbuild check in CI, to ensure that code changes would not cause problems with compilation on windows.

In previous versions, we only compiled the windows version when we published the release.

Fixes #

```release-note
NONE
```

### DIFF
--- a/hack/lib/golang.sh
+++ b/hack/lib/golang.sh
@@ -308,6 +308,7 @@ kubeedge::golang::cross_build_place_binaries() {
         set +x
       fi
     elif [ "${goos}" == "windows" ]; then
+      sudo apt-get install -y mingw-w64
       echo "cross building $bin ${goos} ${goarch}"
       set -x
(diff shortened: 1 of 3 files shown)


hack/lib/golang.sh
```diff
@@ -308,6 +308,7 @@ kubeedge::golang::cross_build_place_binaries() {
         set +x
       fi
     elif [ "${goos}" == "windows" ]; then
+      sudo apt-get install -y mingw-w64
```
fisherxu, line 311: Copied from https://github.com/kubeedge/kubeedge/blob/master/Makefile#L446

fisherxu:
Ready for review :) cc @WillardHu @Shelley-BaoYue @luomengY @wbc6080

Shelley-BaoYue (commented):
/lgtm 
It's really helpful for checking the windows code.

luomengY:
lgtm

Issue #5940, opened 2024-11-01
[EPIC] Enhance the keadm edge commands


WillardHu:
## What would you like to be added/modified:

- [x] Same node name check when join a node; https://github.com/kubeedge/kubeedge/pull/5965
- [x] Add prerun and postrun shell hooks for keadm edge commands;  https://github.com/kubeedge/kubeedge/pull/5946
- [ ] The batch-join command supprots sign in to edge nodes with the SSH private key; https://github.com/kubeedge/kubeedge/pull/5968
- [ ] The batch-join command supprots scp some files to edge nodes; ditto

### Same node name check when join a node

When we join two nodes with the same name, only one node will be registered in the cloud, and then the status of the nodes will be affected by each other. This is due to the logic of EdgeCore, EdgeCore registers nodes to the cloud when it starts, when EdgeCore restarts, it cannot become a new registered node. Therefore, the logic in the cloud is that if the node exists, no processing will be performed.

We need to call the https server of cloudcore to verify whether the node name already exists when executing the `keadm join` command. Calling the api-server will increase the complexity of executing `keadm join`, because we need to obtain kubeconf or token to call the api-server.

Kubeadm will also check for duplicate node names when joining a node. Refer to: https://github.com/kubernetes/kubernetes/blob/c6669ea7d61af98da3a2aa8c1d2cdc9c2c57080a/cmd/kubeadm/app/cmd/phases/join/kubelet.go#L214

### Add prerun and postrun shell hooks for keadm edge commands

Previously, we discussed in the issue #5745 about automatically installing the container runtime before executing the `keadm join` command. The conclusion is that there are many implementations and versions of container runtimes, and it will be a huge effort for the KubeEdge community to maintain them.

Therefore, we thought of using prerun and postrun shell hooks to handle container runtime installation and uninstallation, as well as more flexible applications. E.g., `keadm join --preun install-constainerd.sh`, `keadm reset --postrun uninstall-containerd.sh`.

### Enhance the batch-join command 

We have just implemented a Bate version of the `keadm batchjoin` command to batch join edge nodes.
- issue: https://github.com/kubeedge/kubeedge/issues/5590
- proposal: https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/batch-node-access.md
- pr: https://github.com/kubeedge/kubeedge/pull/5889

#### Supprots login to edge nodes with the SSH private key

[...]

Issue #5745, opened 2024-07-22
keadm tool function enhancement


tangming1996:
surport container runtime auto install

**What would you like to be added/modified**:

**Why is this needed**:

EraKin575:
Hi, can I work on this? You can assign me this issue after I have submitted a PR

EraKin575:
@tangming1996 can you please assist me about the details?

EraKin575:
Hi, @tangming1996 ! I want to work on this issue. Can you please assign it to me?

EraKin575:
Hi @Shelley-BaoYue ! I apologize for messaging you here. Can you please provide me some details for this issue.

magichan:
@EraKin575   noticed that you've been focusing mainly on #5755 . Do you still have time to handle this issue?

If you don't have time at the moment, would you mind unassigning yourself from the issue for now so that I can see if I can take care of it?

EraKin575:
@magichan please refer to this:https://cloud-native.slack.com/archives/C066UJZJKQE/p1724256854980599?thread_ts=1724256854.980599&cid=C066UJZJKQE

Shelley-BaoYue:
@tangming1996  Please update the conclusion of discussion in this issue 
![image](https://github.com/user-attachments/assets/2fd6f7e3-fbd4-4b7a-a802-086cd0b888df)

PR #5946, opened 2024-11-06
add prerun and postrun shell hooks for keadm edge commands

tangming1996:
**What type of PR is this?**
/kind feature

Previously, we discussed in the issue https://github.com/kubeedge/kubeedge/issues/5745 about automatically installing the container runtime before executing the keadm join command. The conclusion is that there are many implementations and versions of container runtimes, and it will be a huge effort for the KubeEdge community to maintain them.

Therefore, we thought of using prerun and postrun shell hooks to handle container runtime installation and uninstallation, as well as more flexible applications. E.g., keadm join --preun install-constainerd.sh, keadm reset --postrun uninstall-containerd.sh.

Fixes a part of https://github.com/kubeedge/kubeedge/issues/5940

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/common/constant.go
+++ b/keadm/cmd/keadm/app/cmd/common/constant.go
@@ -23,6 +23,12 @@ const (
 
 	// FlagNameKubeEdgeVersion sets the version of KubeEdge to be used
 	FlagNameKubeEdgeVersion = "kubeedge-version"
+
+	// FlagNamePreRun ...
+	FlagNamePreRun = "pre-run"
+
+	// FlagNamePostRun ...
+	FlagNamePostRun = "post-run"
 )
 
 // Cloud init and upgrade common flag names
--- a/keadm/cmd/keadm/app/cmd/edge/join.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join.go
@@ -59,6 +59,12 @@ func NewEdgeJoin() *cobra.Command {
 		Example:      edgeJoinExample,
 		SilenceUsage: true,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if joinOptions.PreRun != "" {
+				step.Printf("Executing pre-run script: %s\n", joinOptions.PreRun)
+				if err := util.RunScript(joinOptions.PreRun); err != nil {
+					return err
+				}
+			}
 			step.Printf("Check KubeEdge edgecore process status")
 			running, err := util.IsKubeEdgeProcessRunning(util.KubeEdgeBinaryName)
 			if err != nil {
@@ -95,6 +101,15 @@ func NewEdgeJoin() *cobra.Command {
 				return fmt.Errorf("edge node join failed: %v", err)
 			}
 
+			return nil
+		},
+		PostRunE: func(cmd *cobra.Command, args []string) error {
+			if joinOptions.PostRun != "" {
+				fmt.Printf("Executing post-run script: %s\n", joinOptions.PostRun)
+				if err := util.RunScript(joinOptions.PostRun); err != nil {
+					fmt.Printf("Execute post-run script: %s failed: %v\n", joinOptions.PostRun, err)
+				}
+			}
 			return nil
 		},
 	}
--- a/keadm/cmd/keadm/app/cmd/edge/join_others.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join_others.go
@@ -95,6 +95,12 @@ func AddJoinOtherFlags(cmd *cobra.Command, joinOptions *common.JoinOptions) {
 
 	cmd.Flags().StringVar(&joinOptions.Sets, common.FlagNameSet, joinOptions.Sets,
 		`Set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)`)
+
+	cmd.Flags().StringVar(&joinOptions.PreRun, common.FlagNamePreRun, joinOptions.PreRun,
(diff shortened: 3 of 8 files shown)


luomengY:
Let `e2e` pass the test first

tangming1996:
> Let `e2e` pass the test first让 `e2e` 先通过测试

OK

keadm/cmd/keadm/app/cmd/util/scriptrun.go
```diff
@@ -0,0 +1,39 @@
...
+
+package util
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func RunScript(scriptPath string) error {
```
WillardHu, line 25: This function is very simple now, I don't think it needs to be abstracted yet. In keadm, the `exec.Command` is used in many places, We can cansider how to abstract them later.

keadm/cmd/keadm/app/cmd/edge/join_others.go
```diff
@@ -181,6 +184,12 @@ func createEdgeConfigFiles(opt *common.JoinOptions) error {
 }
 
 func join(opt *common.JoinOptions, step *common.Step) error {
+	if opt.PreRun != "" {
+		step.Printf("Execute the prescript before joining the node")
+		if err := util.RunScript(opt.PreRun); err != nil {
+			return err
+		}
+	}
```
WillardHu, line 192: Move to PreRunE，and add `--post-run` flag. Reset need add `--pre-rune` flag too.
```goalng
PreRunE: func(cmd *cobra.Command, args []string) error {
    ...
    if opt.PreRun != "" {
        // run script
    }
}
PostRun: func(cmd *cobra.Command, args []string) error {
    ....
    if opt.PostRun != "" {
        // run script without err handle
    }
}
```

tangming1996: Thank you for your suggestion. I have modified it.

keadm/cmd/keadm/app/cmd/common/constant.go
```diff
@@ -50,6 +50,9 @@ const (
 
 	// FlagNamePrintFinalValues ...
 	FlagNamePrintFinalValues = "print-final-values"
+
+	// FlagNamePostRun ...
+	FlagNamePostRun = "post-run"
```
fujitatomoya, line 55: why is this assigned to `Cloud init` flag constant? according to https://github.com/kubeedge/kubeedge/issues/5940, it is supposed to be under `Edge join flag names`? Or if those are common flags, they can be under `Common flag names`?

keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -97,6 +103,15 @@ func NewEdgeJoin() *cobra.Command {
 
 			return nil
 		},
+		PostRunE: func(cmd *cobra.Command, args []string) error {
+			if joinOptions.PostRun != "" {
+				step.Printf("Execute the prescript before joining the node")
```
fujitatomoya, line 108:
```suggestion
				step.Printf("Execute the postscript after joining the node")
```

keadm/cmd/keadm/app/cmd/edge/reset_others.go
```diff
@@ -51,6 +51,12 @@ func NewOtherEdgeReset() *cobra.Command {
 		Long:    resetLongDescription,
 		Example: resetExample,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if reset.PreRun != "" {
+				fmt.Printf("Execute the prescript before joining the node")
```
fujitatomoya, line 55:
```suggestion
				fmt.Printf("Execute the prescript before resetting the node")
```

fujitatomoya (commented):
IMO pre/post script hook is generally useful to do user specific things.

a few minor comments need to be addressed.

btw,

> https://github.com/kubeedge/kubeedge/issues/5745
> surport container runtime auto install

if that is the major case, we would want to provide the example script to install container runtime? this would be useful against the original issue description?

keadm/cmd/keadm/app/cmd/common/constant.go
tangming1996: Thank you for your suggestion. I have modified it.

keadm/cmd/keadm/app/cmd/edge/join.go
tangming1996: Thank you for your suggestion. I have modified it.

keadm/cmd/keadm/app/cmd/edge/reset_others.go
tangming1996: Thank you for your suggestion. I have modified it.

tangming1996:
> IMO pre/post script hook is generally useful to do user specific things.IMO 前/后脚本钩子通常对于执行用户特定的事情很有用。
> 
> a few minor comments need to be addressed.一些小的评论需要解决。
> 
> btw, 增值税
> 
> > #5745
> > surport container runtime auto installsurport 容器运行时自动安装
> 
> if that is the major case, we would want to provide the example script to install container runtime? this would be useful against the original issue description?如果这是主要情况，我们是否要提供示例脚本来安装 Container Runtime？这对原始问题描述有用吗？

Supporting the container runtime is just a scenario we assume. There may be other requirements from users, such as environment checks and so on. The specific scripts needed should be customized according to the user's own scenarios. Of course, you are quite right. It is very necessary to provide a case. Later, we will submit a separate pull request (pr) to provide a case of installing the container runtime.

PR #5947, opened 2024-11-07
Sync OWNER COC and LICENSE file from sub repo

wbc6080:
**What type of PR is this?**
/kind documentation

When updating the sub-repo versions of `api` and `mapper-framework`, it was found that some files in the independent sub-repo were not synchronized with the staging directory. For example, there are OWNER, LISENCE and COC files in the `api` repo, but not in the staging directory. So this PR synchronizes the differences between them. By the way, the sub-repo in the k8s staging directory also contains the above files like https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/api, so adding these files can be more standardized.

Fixes #

```release-note

```


fisherxu (commented):
/approve

PR #5948, opened 2024-11-08
edgepp resource overrider enhancement

tangming1996:
**What type of PR is this?**
/kind feature

The previous design of Resource override was to fully overwrite the Resource field in the template. This led to the situation where when users wanted to replace a single independent value, they also needed to repeatedly fill in the values that they did not want to overwrite, which did not conform to the user's intuitive perception. The main modification in this optimization is that it can support overwriting only the independent fields with actual values filled in by the users. For the fields without filled - in values, the template values are still used.

for example:
```yaml
template:
   Resource:
       limits:
         cpu: 800m
         memory: 800Mi
       requests:
         cpu: 600m
         memory: 600Mi

overrider:
   Resource:
       limits:
         cpu: 400m
       requests:
         memory: 400Mi

final result:
   Resource:
       limits:
         cpu: 400m
         memory: 800Mi
       requests:
         cpu: 600m
         memory: 400Mi
```

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/controllermanager/edgeapplication/overridemanager/resourcesoverrider.go
+++ b/cloud/pkg/controllermanager/edgeapplication/overridemanager/resourcesoverrider.go
@@ -17,6 +17,7 @@ limitations under the License.
 package overridemanager
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
lgtm with green CI.

Note: i am not even sure why `acquireOverride` appends the list no matter what... this PR makes more sense to me.

tangming1996:
> lgtm with green CI.
> 
> Note: i am not even sure why `acquireOverride` appends the list no matter what... this PR makes more sense to me.


The acquireOverride function appends patches to the list based on non-zero values in resourcesValue.Requests and resourcesValue.Limits. This means that as long as there are any non-zero resource requests or limits, patches will be generated. If resourcesValue has only zero values for both Requests and Limits, then no patches will be appended, and an empty list will be returned.

cloud/pkg/controllermanager/edgeapplication/overridemanager/resourcesoverrider.go
```diff
@@ -79,25 +79,53 @@ func buildResourcesPatchesWithPath(specContainersPath string, rawObj *unstructur
...
 	return patches, nil
 }
 
-// acquireOverrideOption for adding resources override.
-func acquireOverride(resourcesPath string, resourcesValue corev1.ResourceRequirements) (overrideOption, error) {
+// acquireOverrideOptions for adding selective resources override.
+func acquireOverride(resourcesPath string, resourcesValue corev1.ResourceRequirements) ([]overrideOption, error) {
 	if !strings.HasPrefix(resourcesPath, pathSplit) {
-		return overrideOption{}, fmt.Errorf("internal error: [acquireOverride] resourcesPath should start with / character")
+		return nil, fmt.Errorf("internal error: [acquireOverride] resourcesPath should start with / character")
```
wbc6080, line 99: Use `errors.New()` maybe better than `fmt.Errorf`

wbc6080 (commented):
Please rebase the latest change to pass the fuzzing CI.

cloud/pkg/controllermanager/edgeapplication/overridemanager/resourcesoverrider.go
tangming1996: Thanks for the correction, I have made the modification.

PR #5965, opened 2024-11-15
Check for duplicate names when joining the node

tangming1996:
**What type of PR is this?**
/kind feature

When we join two nodes with the same name, only one node will be registered in the cloud, and then the status of the nodes will be affected by each other. This is due to the logic of EdgeCore, EdgeCore registers nodes to the cloud when it starts, when EdgeCore restarts, it cannot become a new registered node. Therefore, the logic in the cloud is that if the node exists, no processing will be performed.

Fixes a part of https://github.com/kubeedge/kubeedge/issues/5940

```release-note

```

### DIFF
--- a/common/constants/default.go
+++ b/common/constants/default.go
@@ -25,6 +25,7 @@ const (
 const (
 	DefaultCAURL              = "/ca.crt"
 	DefaultCertURL            = "/edge.crt"
+	DefaultCheckNodeURL       = "/node/{nodename}"
 	DefaultNodeUpgradeURL     = "/nodeupgrade"
 	DefaultTaskStateReportURL = "/task/{taskType}/name/{taskID}/node/{nodeID}/status"
 
--- a/keadm/cmd/keadm/app/cmd/edge/join.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join.go
@@ -17,13 +17,18 @@ limitations under the License.
 package edge
 
 import (
+	"crypto/tls"
 	"fmt"
+	"net"
+	"net/http"
 	"os"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2"
+	apiutil "github.com/kubeedge/api/apis/util"
 	"github.com/kubeedge/kubeedge/common/constants"
 	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/common"
 	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/util"
@@ -76,19 +81,21 @@ func NewEdgeJoin() *cobra.Command {
 
 			step.Printf("Check if the management directory is clean")
 			if _, err := os.Stat(util.KubeEdgePath); err != nil {
-				if os.IsNotExist(err) {
-					return nil
+				if !os.IsNotExist(err) {
+					return fmt.Errorf("Stat management directory %s failed: %v", util.KubeEdgePath, err)
 				}
-				return fmt.Errorf("Stat management directory %s failed: %v", util.KubeEdgePath, err)
+			} else {
+				entries, err := os.ReadDir(util.KubeEdgePath)
+				if err != nil {
+					return fmt.Errorf("read management directory %s failed: %v", util.KubeEdgePath, err)
+				}
+				if len(entries) > 0 {
+					return fmt.Errorf("the management directory %s is not clean, please remove it first", util.KubeEdgePath)
+				}
 			}
-			entries, err := os.ReadDir(util.KubeEdgePath)
-			if err != nil {
-				return fmt.Errorf("read management directory %s failed: %v", util.KubeEdgePath, err)
-			}
-			if len(entries) > 0 {
-				return fmt.Errorf("the management directory %s is not clean, please remove it first", util.KubeEdgePath)
-			}
-			return nil
+
+			step.Printf("Check if the node name is valid")
+			return isNodeExist(joinOptions)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ver, err := util.GetCurrentVersion(joinOptions.KubeEdgeVersion)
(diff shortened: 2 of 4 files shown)


cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
```diff
@@ -0,0 +1,43 @@
+package nodetask
```
WillardHu, line 1: The package name can be changed to `node` or create a new package.

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
```diff
@@ -0,0 +1,43 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
+)
+
+func CheckNodeName(request *restful.Request, response *restful.Response) {
+	// get nodeName
+	nodeName := request.QueryParameter("nodeName")
```
WillardHu, line 17: You could consider using path parameters `request.PathParameter(..)`.

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
```diff
@@ -0,0 +1,43 @@
...
+	if nodeName == "" {
+		err := response.WriteErrorString(http.StatusBadRequest, "nodeName parameter is required")
+		if err != nil {
+			klog.Errorf("failed to send the task resp to edge , err: %v", err)
+		}
+		return
+	}
+
+	_, err := client.GetKubeClient().CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+	if err != nil {
```
WillardHu, line 27: This API is used to check whether the node name conflicts. If the node is not found, 200 should be returned, and 409 should be returned if the node is found.

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
tangming1996: Thank you for your suggestion, but my API is designed as `/nodenamecheck?nodeName=%s`, so `request.PathParameter(..)`is not appropriate here.

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
tangming1996: Thank you for your suggestion, but I think the original meaning should be followed on the server side, and if the returned status code is used, it should be controlled by the client logic, because the interface is not necessarily used for node name checking later, it may be directly used to query nodes, so the return of the original status code will be easier to understand.

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
tangming1996: Thanks for your suggestion, I will create a new package.

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
WillardHu: In that case you need to change the service name. The meaning of checking the node name here is to check whether the node name conflicts...

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
WillardHu: Yes, would it be more RESTful to use path parameters?

tangming1996:
Demonstration effect
image

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
tangming1996: Ok, thank you for your suggestion. I have made corresponding modifications

cloud/pkg/cloudhub/servers/httpserver/nodetask/checknodename.go
tangming1996: Thank you for your suggestion. I have made corresponding modifications.

cloud/pkg/cloudhub/servers/httpserver/node/nodename.go
```diff
@@ -0,0 +1,43 @@
+package node
```
Shelley-BaoYue, line 1: Add a copyright

cloud/pkg/cloudhub/servers/httpserver/node/nodename.go
```diff
@@ -0,0 +1,43 @@
...
+	if nodeName == "" {
+		err := response.WriteErrorString(http.StatusBadRequest, "nodename parameter is required")
+		if err != nil {
+			klog.Errorf("failed to send the task resp to edge , err: %v", err)
+		}
+		return
+	}
+
+	_, err := client.GetKubeClient().CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
+	if err != nil {
```
Shelley-BaoYue, line 27: if err != nil and !errors.IsNotFound(err) , what should we do here? It seems now we response `Node founded` also for this err?

common/constants/default.go
```diff
@@ -25,6 +25,7 @@ const (
 const (
 	DefaultCAURL              = "/ca.crt"
 	DefaultCertURL            = "/edge.crt"
+	DefaultHasNodeNameURL     = "/node/{nodename}"
```
Shelley-BaoYue, line 28: Would it be better using `CheckNodeName` ?

cloud/pkg/cloudhub/servers/httpserver/node/nodename.go
tangming1996: copyrighted

common/constants/default.go
tangming1996: Yes, it was indeed called CheckNodeName before. However, later on we considered that this interface might not necessarily be used for node name checking in the future. Perhaps it can simply be used to determine whether a node exists or not.

cloud/pkg/cloudhub/servers/httpserver/node/nodename.go
tangming1996: copyrighted

cloud/pkg/cloudhub/servers/httpserver/node/nodename.go
tangming1996: Thank you for pointing out the problem. I have modified it.

common/constants/default.go
Shelley-BaoYue: Or we can call it `CheckNode`? It'll be better expanded, not just for node name. But this name won't cause any impact, it's just my suggestion.

tangming1996: Thank you for your suggestion. I have made corresponding modifications.

keadm/cmd/keadm/app/cmd/edge/join.go
```diff
@@ -176,3 +183,43 @@ func createBootstrapFile(opt *common.JoinOptions) error {
...
+	if err != nil {
+		return errors.Errorf("error making request: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		return errors.New("node already exists, cannot register again")
+	}
+
+	return nil
```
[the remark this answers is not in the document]
tangming1996: Thank you for your suggestion. I have modified it.

fisherxu (commented):
/lgtm
Thanks for the great feature :)

PR #5980, opened 2024-11-21
Debug the location of the tools cache folder

martin-g:
**What type of PR is this?**

Fixes #

```release-note

```


wbc6080 (commented):
Looks good as a temporary way. But I am a little curious why this step is no longer needed after CNCF hosted/ephemeral runners are used. Is it because after CNCF hosted, will create a new environment every time to run a PR?

martin-g:
Yes!
At the moment this is a self-hosted runner that is running all the time. 
With the CNCF setup it will use runners which are started for each job run and then discarded.

wbc6080 (commented):
/lgtm
cc @Shelley-BaoYue @WillardHu @fisherxu

.github/workflows/main-arm64.yaml
```diff
@@ -45,6 +45,10 @@ jobs:
     name: Multiple build
     needs: image-prepare
     steps:
+        # this step should be deleted once CNCF hosted/ephemeral runners are in use
```
fisherxu, line 48:
```suggestion
        # TODO: this step should be deleted once CNCF hosted/ephemeral runners are in use
```
Thsnks for the clarification, and it's better add a todo here. BTW when will the CNCF hosted/ephemeral runners be used?  :)

PR #4184, opened 2022-09-15
[Feature-New-Edged] integrate Kubelet in new edged

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

#4157 

**What this PR does / why we need it**:

- integrate Kubelet in edged, include init Kubelet config and execute  Kubelet `Run` func.
- in gomod，replace kubernetes with kubernetes in kubeedge project [https://github.com/kubeedge/kubernetes].  And update vendor & licenses

PR #6000, opened 2024-11-27
add version view and fixed a bug where non-existent commands could be…

tangming1996:
… run for edgecore

**What type of PR is this?**
/kind bug

1、edgecore does not have a command to view the version.
2、edgecore executes a non-existent command with no error and runs directly.
image

Fixes #

```release-note

```


edge/cmd/edgecore/app/server.go
```diff
@@ -54,6 +55,11 @@ is the message processor between edged and edgehub. It is also responsible for s
 to/from a lightweight database (SQLite).ServiceBus is a HTTP client to interact with HTTP servers (REST),
 offering HTTP client capabilities to components of cloud to reach HTTP servers running at edge. `,
 		Run: func(cmd *cobra.Command, args []string) {
+			// print version
+			if opts.ShowVersion {
+				fmt.Printf("version: %#v\n", version.Get())
+				return
+			}
```
wbc6080, line 62: If we develop based on a certain release version, what does the returned version look like? For example, `v1.19.0-dirty` or still `v1.19.0`?

tangming1996: This may depend on the implementation of the `version.Get()` function, which I call directly.

Shelley-BaoYue:
> We have version flag in old release, I think we can also add it like before, ref: https://github.com/kubeedge/kubeedge/blob/release-1.11/edge/cmd/edgecore/app/server.go#L126 PTAL @tangming1996
> 
> Not sure why it's deleted in this PR: #4184 cc @Shelley-BaoYue

It conflicted with the flag of kubelet, so it was deleted directly without careful consideration. Now we can find the solution for this.

tangming1996:
> > We have version flag in old release, I think we can also add it like before, ref: https://github.com/kubeedge/kubeedge/blob/release-1.11/edge/cmd/edgecore/app/server.go#L126 PTAL 我们在旧版本中有版本标志，我想我们也可以像以前一样添加它，参考：https://github.com/kubeedge/kubeedge/blob/release-1.11/edge/cmd/edgecore/app/server.go#L126 PTAL@tangming1996
> > Not sure why it's deleted in this PR: #4184 cc 不知道为什么在此 PR 中删除它： #4184 cc@Shelley-BaoYue
> 
> It conflicted with the flag of kubelet, so it was deleted directly without careful consideration. Now we can find the solution for this.它与 kubelet 的标志冲突，因此没有仔细考虑就直接删除了。现在我们可以找到解决方案。

Yes, there will be conflict. The errors are as follows:
(base) tangming@MacBook-Pro-6 edgecore % ./edgecore -h                                
./edgecore flag redefined: version
panic: ./edgecore flag redefined: version

PR #6022, opened 2024-12-05
fix device desired value type error

JiaweiGithub:

**What type of PR is this?**
/kind bug

Desired.Value is a fixed string type in the latest version and does not require type matching. When the value is not set, it will match the int type by default and report an error.

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
@@ -266,12 +266,18 @@ func setVisitor(visitorConfig *driver.VisitorConfig, twin *common.Twin, dev *dri
 		return nil
 	}
 	klog.V(2).Infof("Convert type: %s, value: %s ", twin.Property.PProperty.DataType, twin.ObservedDesired.Value)
-	value, err := common.Convert(twin.Property.PProperty.DataType, twin.ObservedDesired.Value)
-	if err != nil {
-		klog.Errorf("Failed to convert value as %s : %v", twin.Property.PProperty.DataType, err)
-		return err
+	var value interface{}
+	if twin.ObservedDesired.Value != "" {
+		convertedValue, err := common.Convert(twin.Property.PProperty.DataType, twin.ObservedDesired.Value)
+		if err != nil {
+			klog.Errorf("Failed to convert value as %s : %v", twin.Property.PProperty.DataType, err)
+			return err
+		}
+		value = convertedValue
+	} else {
+		value = twin.ObservedDesired.Value
 	}
-	err = dev.CustomizedClient.SetDeviceData(value, visitorConfig)
+	err := dev.CustomizedClient.SetDeviceData(value, visitorConfig)
 	if err != nil {
 		return fmt.Errorf("%s set device data error: %v", twin.PropertyName, err)
 	}


staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
```diff
@@ -266,12 +266,7 @@ func setVisitor(visitorConfig *driver.VisitorConfig, twin *common.Twin, dev *dri
 		return nil
 	}
 	klog.V(2).Infof("Convert type: %s, value: %s ", twin.Property.PProperty.DataType, twin.ObservedDesired.Value)
-	value, err := common.Convert(twin.Property.PProperty.DataType, twin.ObservedDesired.Value)
-	if err != nil {
-		klog.Errorf("Failed to convert value as %s : %v", twin.Property.PProperty.DataType, err)
-		return err
-	}
-	err = dev.CustomizedClient.SetDeviceData(value, visitorConfig)
+	err := dev.CustomizedClient.SetDeviceData(twin.ObservedDesired.Value, visitorConfig)
```
wbc6080, line 269: The purpose of convert here is to convert string type `value` into other types, such as int, etc. This type is consistent with the user-defined device property type.

wbc6080: The data type of device property is universal, so we define it in `devicemodel` https://github.com/kubeedge/kubeedge/blob/master/staging/src/github.com/kubeedge/api/apis/devices/v1beta1/device_model_types.go#L41 . `deviceinstance` needs to obtain the data type of the device property from` devicemodel` when building https://github.com/kubeedge/kubeedge/blob/master/staging/src/github.com/kubeedge/mapper-framework/pkg/util/parse/grpc.go#L280-L303

JiaweiGithub: My field on the model is int, but if the expected value is not set by default, it will be "", which will cause the mapper to fail to start by default. This is equivalent to all my int type fields having to be set to a number by default.

fujitatomoya: is this expected behavior because user specifies that empty string value into the int type, that sounds like syntax error? this PR returns to the application with empty string without any error which appears to be the success of conversion (from int in your particular case), but internally it does not succeed any conversion at all?

WillardHu: I also agree that we need to think about whether an empty string is an illegal input or a zero value according to the scenario.

wbc6080: +1, and I suggest that you can add null value verification to the place where this convert function is called, that is, `setvisitor`. If the expected value is null, you can output the log and returned directly. There is no need to enter the `convert` function.

JiaweiGithub: > +1, and I suggest that you can add null value verification to the place where this convert function is called, that is, `setvisitor`. If the expected value is null, you can output the log and returned directly. There is no need to enter the `convert` function.+1，我建议你可以在调用这个convert函数的地方，即`setvisitor`添加空值验证。如果期望值为null，则可以输出日志并直接返回。无需输入`convert`函数。

I think this suggestion is good.

fujitatomoya (approved):
IMO this is better, lgtm with green CI

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
```diff
@@ -266,12 +266,18 @@ func setVisitor(visitorConfig *driver.VisitorConfig, twin *common.Twin, dev *dri
...
-		klog.Errorf("Failed to convert value as %s : %v", twin.Property.PProperty.DataType, err)
-		return err
+	var value interface{}
+	if twin.ObservedDesired.Value != "" {
+		convertedValue, err := common.Convert(twin.Property.PProperty.DataType, twin.ObservedDesired.Value)
+		if err != nil {
+			klog.Errorf("Failed to convert value as %s : %v", twin.Property.PProperty.DataType, err)
+			return err
+		}
+		value = convertedValue
```
wbc6080, line 276:
```suggestion
		value, err := common.Convert(twin.Property.PProperty.DataType, twin.ObservedDesired.Value)
		if err != nil {
			klog.Errorf("Failed to convert value as %s : %v", twin.Property.PProperty.DataType, err)
			return err
		}
```

wbc6080 (commented):
It seems no need to define a new variable `convertedValue`

staging/src/github.com/kubeedge/mapper-framework/_template/mapper/device/device.go
JiaweiGithub: done

JiaweiGithub: The scopes are different.

image

wbc6080: OK, I see

Issue #6008, opened 2024-11-30
Support update edgecore config from cloud


Shelley-BaoYue:
**What would you like to be added/modified**:  

Support update edgecore config from cloud 

**Why is this needed**: 

Support updating the configuration of EdgeCore from the cloud, especially in scenarios involving unified operations for batch nodes, which can improve operational efficiency.

fujitatomoya:
@Shelley-BaoYue i believe this is really great and useful feature for user experience.

Shelley-BaoYue:
> @Shelley-BaoYue i believe this is really great and useful feature for user experience.

I have submitted proposal for this feature. Welcome to review it : )

PR #6024, opened 2024-12-06
Add proposal for supporting update edge configuration from cloud side

Shelley-BaoYue:
**What type of PR is this?**

/kind design

Add proposal for supporting update edge configuration from cloud side

#6008 

Fixes #

```release-note

```


docs/proposals/edge-config-update.md
```diff
@@ -0,0 +1,212 @@
...
+- Check the received config-update task.  
+
+### Keadm
+
+Add n new subcommand for keadm, `keadm update-config`, to process this task. keadm update-config will do: 
+
+- use `/etc/kubeedge/idempotency_record` file to make sure only one task(config update or node upgrade) can be performed at a time  
+- Back up `edgecore.db, edgecore.yaml` to backup path `/etc/kubeedge/backup/update-config` 
+- Update the `edgecore.yaml` and restart edgecore.service 
+- rollback: if update failed, keadm will do rollback operation to start edgecore with origin configuration.
```
fujitatomoya, line 54: if one of the thousand edgenodes happened to fail eventually, what would happen? it looks like this job failed gracefully and rollback all requested node after all?

fujitatomoya (commented):
this is pretty exciting feature to manage thousands devices at edge environment :rocket: :rocket: :rocket:

docs/proposals/edge-config-update.md
Shelley-BaoYue: Only the nodes that fail to be updated will be rolled back. The proposal may not be described clearly, I will optimize it again.

docs/proposals/edge-config-update.md
```diff
@@ -0,0 +1,212 @@
+---
+title: Support update EdgeCore config
+authors:
+- "@Shelley-BaoYue"
+  approvers:
+  creation-date: 2024-11-30
+  last-updated: 2024-11-30
+  status: implementable
+---
```
tangming1996, line 9: There seems to be a problem with the format here. Maybe something like this?
```yaml
---
title: Support update EdgeCore config
authors:
  - "@Shelley-BaoYue"
approvers:
  - "@Approver1"
  - "@Approver2"
creation-date: 2024-11-30
last-updated: 2024-11-30
status: implementable
---
```

docs/proposals/edge-config-update.md
```diff
@@ -0,0 +1,212 @@
...
+# Support update EdgeCore config
+
+## Motivation 
+
+In many scenarios, being able to update the EdgeCore configuration files for edge nodes directly from the cloud provides significant convenience compared to logging into each edge node for manual updates. This is especially beneficial for batch operations, where managing multiple nodes at once can greatly enhance efficiency and streamline the operational process.
+
+### Goals
+
+- Provide APIs for creating config update tasks from cloud.
+- Support update edge config on multi edge nodes or nodegroup.
```
tangming1996, line 20: It is not appropriate to use `nodegroup` directly here, because there is no direct reuse of `nodegroup` apis in the edge-config-update` api design, which may be misleading.

docs/proposals/edge-config-update.md
```diff
@@ -0,0 +1,212 @@
...
+### EdgeHub
+
+- Check the received config-update task.  
+
+### Keadm
+
+Add n new subcommand for keadm, `keadm update-config`, to process this task. For each edgenode, keadm update-config will do: 
+
+- use `/etc/kubeedge/idempotency_record` file to make sure only one task(config update or node upgrade) can be performed at a time  
+- Back up `edgecore.db, edgecore.yaml` to backup path `/etc/kubeedge/backup/update-config`
```
WillardHu, line 52: Modifying the configuration file may not change the edgecore.db, so should we still back it up? However, this raises to a question. If the configuration modifies the path-related properties, should we process the files in the these paths, or assume that the file is ready by users?

PR #6029, opened 2024-12-09
fix generate crds error

JiaweiGithub:

**What type of PR is this?**
/kind bug

Fixes #

```release-note

```


Makefile
```diff
@@ -199,9 +199,14 @@ endif
 
 CRD_VERSIONS=v1
 CRD_OUTPUTS=build/crds
-DEVICES_VERSION=v1alpha2
+DEVICES_VERSION=v1beta1
+OPERATIONS_VERSION=v1alpha1
```
Shelley-BaoYue, line 203: This pr mainly fix the error caused by updating device version to v1beta1?  Please provide detail description in this PR.

fujitatomoya: it would be also nice to update the doc section below accordingly.

JiaweiGithub: I'm having some issues with the test, so let's ignore it for now。

Issue #5997, opened 2024-11-27
Dependency update - Kubernetes 1.30


Shelley-BaoYue:
**What would you like to be added/modified**:

Upgrade Kubernetes version to v1.30

**Why is this needed**:

As KubeEdge is built upon Kubernetes, we need to update Kubernetes to compatible with new release of Kubernetes and support more new feature of Kubernetes.

PR #6033, opened 2024-12-13
Dependency update - Kubernetes 1.30.7 

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

Bump Kubernetes version to v1.30.7 

#5997


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/kubeedge/kubeedge
 
-go 1.22
+go 1.22.0
```
WillardHu, line 3: Could the go version be upgrede to 1.22.7?

Shelley-BaoYue: > Could the go version be upgrede to 1.22.7?

I have modified it to v1.22.9

staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/types.go
```diff
@@ -528,6 +534,19 @@ type TailoredKubeletConfiguration struct {
...
+	// Default: 1
+	// +optional
+	ContainerLogMaxWorkers *int32 `json:"containerLogMaxWorkers,omitempty"`
+	// ContainerLogMonitorInterval specifies the duration at which the container logs are monitored
+	// for performing the log rotate operation. This defaults to 10 * time.Seconds. But can be
+	// customized to a smaller value based on the log generation rate and the size required to be
+	// rotated against
+	// Default: 10s
+	// +optional
+	ContainerLogMonitorInterval *metav1.Duration `json:"containerLogMonitorInterval,omitempty"`
```
fisherxu, line 549: Why add these three fields, because it need to set the default values？

Shelley-BaoYue: New features in Kubelet v1.30 add these three new fields and need set the default values.

fisherxu (commented):
/lgtm
/assign @WillardHu

PR #6046, opened 2024-12-24
add bandwidth limit

JiaweiGithub:

**What type of PR is this?**
/kind feature

Fixes #

```release-note

```


edge/pkg/edged/bandwidth/tclimit/container_cmd.go
```diff
@@ -0,0 +1,112 @@
...
+# 获取 PID
+CONTAINER_ID="%s"
+PID=$(ctr -n k8s.io tasks ls | grep $CONTAINER_ID | awk '{print $2}' | tr -d '\n')
+
+# 进入网络命名空间并提取 eth0@if
+IFACE_NAME=$(nsenter -t $PID -n ip a | grep 'eth0@if' | awk '{print $2}' | tr -d '\n')
+
+# 输出接口名称
+echo "$IFACE_NAME"
+`, containerID)
```
tangming1996, line 72: It’s unclear whether the use of embedded scripts here is due to the inability to utilize standard Go libraries. If possible, using standard Go code might be more appropriate. At the very least, ctr can call containerd’s standard APIs to achieve this, and there’s no guarantee that the ctr tool will be available in the user’s environment.

JiaweiGithub: The Go standard library is not as flexible; it does not support many operations.

PR #6063, opened 2025-01-07
fix: multiple --set parameters in keadm join command

### LINKED ISSUE #6050: `keadm join --set` command only applies the last `--set` parameter, which is different from the behavior of `keadm init --set`.
**What happened**:

When using the --set parameter multiple times, only the last --set will take effect. For example:

```
keadm join --kubeedge-version=$KEADM_VERSION \
  --cgroupdriver=systemd \
  --set=modules.deviceTwin.enable=false \
  --set=modules.eventBus.enable=false \
  --set=modules.edgeStream.enable=true
```

The final generated configuration file will only have modules.edgeStream.enable=true take effect.

However, the `keadm join --help` states that it supports multiple `--set` parameters, just like `keadm init`.

```
> keadm join --help
...

Flags:
	  ...
      --set string                       Set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
	 ...

> keadm init --help
...

Flags:
      ...
      --set stringArray             Set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
      ...
```

Behave the same as the keadm init command.

[`CloudInitUpdateBase.Sets`](https://github.com/kubeedge/kubeedge/blob/master/keadm/cmd/keadm/app/cmd/common/types.go#L33)

[`JoinOptions.Sets`](https://github.com/kubeedge/kubeedge/blob/master/keadm/cmd/keadm/app/cmd/common/types.go#L96)

**Environment**:
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): 1.19.1

  </details>

XmchxUp:

Fixes #6050

### DIFF
--- a/keadm/cmd/keadm/app/cmd/common/types.go
+++ b/keadm/cmd/keadm/app/cmd/common/types.go
@@ -93,7 +93,7 @@ type JoinOptions struct {
 	CertPort              string
 	CGroupDriver          string
 	Labels                []string
-	Sets                  string
+	Sets                  []string
 	PreRun                string
 	PostRun               string
 
--- a/keadm/cmd/keadm/app/cmd/edge/join_others.go
+++ b/keadm/cmd/keadm/app/cmd/edge/join_others.go
@@ -93,7 +93,7 @@ func AddJoinOtherFlags(cmd *cobra.Command, joinOptions *common.JoinOptions) {
 	cmd.Flags().StringVar(&joinOptions.HubProtocol, common.HubProtocol, joinOptions.HubProtocol,
 		`Use this key to decide which communication protocol the edge node adopts.`)
 
-	cmd.Flags().StringVar(&joinOptions.Sets, common.FlagNameSet, joinOptions.Sets,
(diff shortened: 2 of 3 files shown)


fujitatomoya (commented):
This change allows multiple Sets values to be specified, which aligns with the functionality described in documentation. lgtm 👍

Shelley-BaoYue (commented):
/approve 
👍

wbc6080 (commented):
It seems that this bug-fix can be cherry-pick to the 1.18 and 1.19 release. Could you help cherry-pick it? @XmchxUp  You can refer to our cherry-pick docs https://kubeedge.io/docs/developer/cherry_pick

XmchxUp:
> It seems that this bug-fix can be cherry-pick to the 1.18 and 1.19 release. Could you help cherry-pick it? @XmchxUp You can refer to our cherry-pick docs [kubeedge.io/docs/developer/cherry_pick](https://kubeedge.io/docs/developer/cherry_pick)

Ok.

Issue #3860, opened 2022-05-21
run systemd edgecore.service in pure go


gy95:
**What would you like to be added/modified**:
when we run `keadm beta join` command to start an edge node, we run systemd related commands in our source codes to start edgeocre systemd service, such as 
https://github.com/kubeedge/kubeedge/blob/bef4c51ee551d315e3eb5e75db4878dd023b867f/keadm/cmd/keadm/app/cmd/edge/joinbeta.go#L250-L254

I find a go package https://github.com/coreos/go-systemd , maybe we could import this package to run edgecore systemd service in pure go

/help
/good-first-issue
**Why is this needed**:

robpickerill:
I think to switch this over we could use the dbus package from go-systemd, and it would look something like this:

```
func runEdgeCore() error {
	systemdExist := util.HasSystemd()

	var binExec, tip string
	if systemdExist {
		tip = fmt.Sprintf("KubeEdge edgecore is running, For logs visit: journalctl -u %s.service -xe", common.EdgeCore)

		ctx := context.TODO()
		d, err := dbus.NewSystemConnectionContext(ctx)
		if err != nil {
			return err
		}
		defer d.Close()

		enabled, _, err := d.EnableUnitFilesContext(ctx, []string{string(common.EdgeCore)}, true, true)
		if err != nil {
			return err
		}
		if !enabled {
			return fmt.Errorf("enable %s.service failed", common.EdgeCore)
		}

		doneChan := make(chan string)
		d.StartUnitContext(ctx, string(common.EdgeCore), "replace", doneChan)
		result := <-doneChan
		if result != "done" {
			return fmt.Errorf("start edgecore failed: %s", result)
		}

	} else {
```

What do you think? I'll tidy it up, and publish the PR if this is the kind of approach you were seeking here.

Ris-code:
Hello I am Rishav Aich, 
I am new to the community and looking to contribute to it. I just wanted to start with some good first issues. So please could anyone explain this issue to me so that I could understand the flow and the codes.

nox1134:
Hi @fisherxu @gy95! Is this issue still open? If no one is working on this issue, I'd like to work on it.
Could you please assign the issue to me?

ABirHAsan123-beep:
#3860

inosmeet:
Hey @forsaken628!
are you working on this? or is it okay for me to take?

forsaken628:
[...]

PR #6075, opened 2025-01-11
use go-systemd to run edgecore systemd service in pure go 

inosmeet:
**What type of PR is this?**
/kind cleanup

run edgecore systemd service in pure go

Fixes: #3860 

```release-note
NONE
```


inosmeet:
Thank you for your feedback!

Since I was aiming to fix issue #3860, I didn't initially focus on the feasibility of replacing the current approach. @gy95 might have more to say in this.

inosmeet:
On a later note, after a bit of digging, I came across these key points:

1. **Unprivileged Operation**

    With go-systemd, you can interact with services via DBus. This allows non-root users to manage their own services (start/stop/enable/disable) without requiring sudo access, aligning with the principle of least privilege.
    Avoids risks of escalating privileges unnecessarily, ensuring a safer execution environment.

2. **Robust Error Handling**

    go-systemd communicates directly with the DBus API, providing structured and granular error messages.
    In contrast, systemctl relies on parsing command-line output, which is prone to inconsistencies across systemd versions or custom configurations.
    Example: You can distinguish between a missing service, a permissions issue, or a transient failure—critical for debugging and reliability.

3. **Eliminates Shell Dependency**

    Direct DBus interactions remove the need for spawning subprocesses (exec.Command) and parsing shell output.
    This improves:
        Performance: No overhead from creating subprocesses for systemctl commands.
        Security: Reduces risk of shell injection vulnerabilities.
        Cross-environment Consistency: Works uniformly across distributions that use systemd, regardless of systemctl CLI behavior.

4. **Better Integration with Go Applications**

    Being a Go library, go-systemd integrates naturally with your application's workflow:
        Makes your service management functionality idiomatic and maintainable within Go code.
        Avoids the need for brittle and verbose shell commands or external dependency wrappers.

WillardHu:
Great, you can refer to https://github.com/kubeedge/kubeedge/blob/04ea218964a20f2b2018ce53436e6347b6b245e1/keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_unix.go#L94 to add an implementation that dbus management systemd, and then continue to let the caller uses the ExtSystem interface, which can better maintain the code and switch the implementation.

inosmeet:
Do I also update methods like enable/disable here or in separate PR?

inosmeet:
Hey @Shelley-BaoYue! can you approve this for tests?
Thanks!

PR #6085, opened 2025-01-13
change `edged.rootDirectory` default to `/var/lib/kubelet`

Shelley-BaoYue:
**What type of PR is this?**

/kind feature

As #https://github.com/kubeedge/kubeedge/pull/5914 disscussed, we plan to change the default value of `edged.rootDirectory` to `/var/lib/kubelet`, which is consistent with kubelet. It will help to reduce maintain costs. In addition, kubeedge can seamlessly connect to native Kubernetes resources.


fisherxu (commented):
/lgtm
/assign @WillardHu

fujitatomoya (approved):
lgtm with green CI, thank you for follow-up.

PR #6090, opened 2025-01-16
Fix mapper lint error

wbc6080:
**What type of PR is this?**
/kind bug

Fix lint error in mapper-framework.

Fixes #

```release-note

```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/influxdb2/handler.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/influxdb2/handler.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"k8s.io/klog/v2"
+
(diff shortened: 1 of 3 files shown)


wbc6080:
cc @luomengY @WillardHu @fisherxu

Shelley-BaoYue (commented):
CI cannot check lint error in mapper? Can we add lint check for mapper in CI?

wbc6080:
> CI cannot check lint error in mapper? Can we add lint check for mapper in CI?

We have previously added mapper checks to lint. https://github.com/kubeedge/kubeedge/blob/master/hack/lib/lint.sh#L81 The currently problem code is located in the `_template` directory of mapper-framework. This part is the mapper template and cannot be compiled and run normally, which should be the reason why lint cannot check it. Only after executing `make generate` to generate the mapper project can this part of the code run normally. I will see if there is a suitable solution later.

Shelley-BaoYue (commented):
/approve

Issue #6101, opened 2025-01-21
[lfx-mentorship] Enhance KubeEdge testing coverage


wbc6080:
**What would you like to be added/modified**:

KubeEdge would like to improve the UT coverage of the code to better maintain the quality of the code and reduce the introduction of defects.

Increase the UT coverage rate to 60% to 70% (currently, the UT coverage rate is 38.69% ).

It is important to note that in addition to requiring the overall UT coverage of KubeEdge to meet the requirements, the UT coverage of **each core code directory** also needs to exceed 60%. These directories include:

```
cloud/
edge/
keadm/
pkg/
```
You can get the code coverage of different files under the current master branch from [here](https://app.codecov.io/gh/kubeedge/kubeedge).

I hope that the students who apply for this project can give the UT coverage rate of each directory plan to achieve, we will move towards these specific goals.

**Why is this needed**:

ravjot07:
Hey @wbc6080,
I’m Ravjot Singh, a developer with experience in Go, backend development, and testing frameworks. I'm excited to contribute to KubeEdge by improving UT coverage, ensuring better code quality and defect prevention. Could you guide me on where to start and how to proceed with the coverage improvements? Looking forward to collaborating!

wbc6080:
Hi @ravjot07 , We are preparing for this project and will release it on the LFX mentorship official website later. You can follow the LFX website and submit applications in time. Welcome to participate in the community!

yashpandey06:
@wbc6080  will be contributing to it ! thanks !

upsaurav12:
@wbc6080 i would like to work in this project for the upcoming lfx term!!. Thank you.

vaidikcode:
Hi @wbc6080 ,  

I’m interested in contributing to improving the unit test (UT) coverage for KubeEdge. I understand that the goal is to increase overall UT coverage from **38.69% to 60-70%**, while ensuring that each of the core directories (`cloud/`, `edge/`, `keadm/`, and `pkg/`) also meets at least **60% coverage**.  

I have a couple of clarifications:  
1. Should we aim for **full repo** coverage improvement, or is the focus strictly on the mentioned core directories?  
2. When referring to **each core directory exceeding 60%**, does this mean every subpackage within these directories must individually reach 60%+, or does it apply to the entire directory as an aggregate?  

Thanks for your guidance! Looking forward to contributing.

wbc6080:
[...]

PR #6119, opened 2025-02-05
[Tests]: improve code coverage context.go 21% to 100%

Dhiren-Mhatre:
**What type of PR is this?**
/kind test
/kind coverage

This PR significantly improves test coverage for the cloud/pkg/common/messagelayer/context.go package from 21% to 100%. It adds comprehensive unit tests for:
- Message sending/receiving workflows
- Router message validation
- All controller implementations (EdgeController, DeviceController, etc.)
- Message comparison logic

Part of [lfx-mentorship] Enhance KubeEdge testing coverage initiative   #6101

- File path: `cloud/pkg/common/messagelayer/context_test.go`
- Added 4 new test cases covering normal/edge scenarios
- Implemented message comparison logic ignoring generated IDs
- Verified with `go test -coverprofile=coverage.out ./cloud/pkg/common/messagelayer/...`
- Followed KubeEdge's test pattern conventions

no
**Snapshots**
image


cloud/pkg/common/messagelayer/context_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The KubeEdge Authors.
+Copyright 2025 The KubeEdge Authors.
```
Shelley-BaoYue, line 2: We don't need to modify the copyright time for any old files. : )

Dhiren-Mhatre: My bad ,I will fix it ASAP

Dhiren-Mhatre:
@wbc6080 PTAL

cloud/pkg/common/messagelayer/context_test.go
```diff
@@ -18,78 +18,205 @@ package messagelayer
...
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
 	"github.com/kubeedge/beehive/pkg/core/model"
+	"github.com/kubeedge/kubeedge/cloud/pkg/common/modules"
 )
 
 const (
 	sendModuleName     = "testcore"
 	receiveModuleName  = "testcore"
 	responseModuleName = "testcore"
+	routerModuleName   = "router"
```
fujitatomoya, line 34: any specific intention here? instead, can we just align with the other names here for test purpose?

```suggestion
	routerModuleName   = "testrouter"
```

cloud/pkg/common/messagelayer/context_test.go
```diff
@@ -18,78 +18,205 @@ package messagelayer
...
+				ModuleName: module,
+				ModuleType: common.MsgCtxTypeChannel,
+			}
+			beehiveContext.AddModule(add)
+			beehiveContext.AddModuleGroup(module, module)
+		}
+	})
+}
+
+// compareMessages compares two messages ignoring the ID field
```
fujitatomoya, line 54:
```suggestion
// compareMessages compares two messages ignoring the UUID field
```

fujitatomoya (approved):
a couple of minor comments, but lgtm with green CI

cloud/pkg/common/messagelayer/context_test.go
```diff
@@ -18,78 +18,205 @@ package messagelayer
...
+			expected: ContextMessageLayer{
+				SendModuleName:     modules.CloudHubModuleName,
+				ReceiveModuleName:  modules.PolicyControllerModuleName,
+				ResponseModuleName: modules.CloudHubModuleName,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt // capture range variable
```
wbc6080, line 209: I'm thinking about whether it's necessary to limit the variable scope like this. I checked the test code of the upstream kubernetes and found that when executing test cases in parallel, it is indeed necessary to limit the variable scope like https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/testing/store_tests.go#L201-L211. However, if we don't have parallel execution content like https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/storage/testing/store_tests.go#L60-L67, Is this sentence unnecessary? Please correct me if my understanding is wrong

Dhiren-Mhatre: Thanks for the detailed explanation! Shall I keep the tt := tt variable capture as a defensive programming practice, even though we're not currently using parallel execution. This aligns with Kubernetes' conventions and guards against future test parallelization changes

PR #6094, opened 2025-01-18
E2E Tests for Device Plugins (deploy and register)

1Shubham7:
**What type of PR is this?**

/kind test

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note

```

PR #6136, opened 2025-02-09
[Integration Test] - Fixing "Add a device with Twin attributes" Integration Test with query optimization and correct tests

1Shubham7:
**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/edge/test/integration/device/device_test.go
+++ b/edge/test/integration/device/device_test.go
@@ -310,11 +310,13 @@ var _ = Describe("Event Bus Testing", func() {
 			common.PrintTestcaseNameandStatus()
 		})
 		It("TC_TEST_EBUS_9: Add a sample device with device attributes to kubeedge node", func() {
-			//Generating Device ID
+			// Generate Device ID
 			DeviceIDWithAttr = GenerateDeviceID("kubeedge-device-WithDeviceAttributes")
-			//Generate a Device
+			// Create a Device
 			DeviceATT = CreateDevice(DeviceIDWithAttr, "DeviceATT", "unknown")
-			//Add Attribute to device
+			common.Infof("Created a device: %v, of DeviceId: %s", DeviceATT.Name, DeviceIDWithAttr)
+
+			// Add Attribute to device
 			AddDeviceAttribute(DeviceATT, "Temperature", "25.25", "float")
 
 			IsDeviceAdded := HandleAddAndDeleteDevice(http.MethodPut, ctx.Cfg.TestManager+Devicehandler, DeviceATT)
@@ -329,11 +331,13 @@ var _ = Describe("Event Bus Testing", func() {
 		})
 
 		It("TC_TEST_EBUS_10: Add a sample device with Twin attributes to kubeedge node", func() {
-			//Generating Device ID
+			// Generate Device ID
 			DeviceIDWithTwin = GenerateDeviceID("kubeedge-device-WithTwinAttributes")
-			//Generate a Device
+			// Create a Device
 			DeviceTW = CreateDevice(DeviceIDWithTwin, "DeviceTW", "unknown")
-			//Add twin attribute
+			common.Infof("Created a device: %v, of DeviceId: %s", DeviceTW.Name, DeviceIDWithTwin)
+
+			// Add twin attribute
 			AddTwinAttribute(DeviceTW, "Temperature", "25.25", "float")
 
 			IsDeviceAdded := HandleAddAndDeleteDevice(http.MethodPut, ctx.Cfg.TestManager+Devicehandler, DeviceTW)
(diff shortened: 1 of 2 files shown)


1Shubham7:
@fisherxu  @Shelley-BaoYue @WillardHu @wbc6080 , the testcase 11, 12 were not correct.

- Secondly, previously we were querying all rows, I think because of that database goes in a locked state from previous operations. hence, I think this should solve test case failing at #6094. Even if not this change is anyways important.

PTAL

Shelley-BaoYue (commented):
Is this PR used to fix failed integration test case or is it just an optimization?

1Shubham7:
Hey @wbc6080 @WillardHu @Shelley-BaoYue, reminder to review this  : )

edge/test/integration/device/device_test.go
```diff
@@ -344,17 +348,15 @@ var _ = Describe("Event Bus Testing", func() {
...
 
-			//Generate a Device
-			device := CreateDevice(DeviceIDWithAttr, "DeviceATT", "unknown")
-			//Add Attribute to device
-			AddDeviceAttribute(device, "Temperature", "50.50", "float")
+			// Update Attribute
+			AddDeviceAttribute(DeviceATT, "Temperature", "50.50", "float")
 
-			IsDeviceAdded := HandleAddAndDeleteDevice(http.MethodPut, ctx.Cfg.TestManager+Devicehandler, device)
+			IsDeviceAdded := HandleAddAndDeleteDevice(http.MethodPut, ctx.Cfg.TestManager+Devicehandler, DeviceATT)
```
wbc6080, line 359: I'm a little confused about the modification here. Is it because the device named `DeviceATT` has been created in the previous `testcase9`, so there is no need to recreate it here? Will there be an error if I create it repeatedly?

edge/test/integration/utils/helpers/helpers.go
```diff
@@ -157,31 +157,26 @@ func GetTwinAttributesFromDB(deviceID string, Name string) TwinAttribute {
...
-			&twinAttribute.Type,
-			&twinAttribute.MetaData)
-
-		if err != nil {
-			common.Fatalf("Failed to scan DB rows: %v", err)
-		}
-		if string(twinAttribute.DeviceID) == deviceID && twinAttribute.Name == Name {
-			break
-		}
+	row := db.QueryRow("SELECT * FROM device_twin WHERE deviceid = ? AND name = ? ORDER BY id DESC LIMIT 1", deviceID, Name)
```
wbc6080, line 160: That is to say, by modifying the query conditions, we do not need to traverse the entire table, which improves efficiency and prevents the database from being locked, right?

edge/test/integration/device/device_test.go
```diff
@@ -310,11 +310,13 @@ var _ = Describe("Event Bus Testing", func() {
 			common.PrintTestcaseNameandStatus()
 		})
 		It("TC_TEST_EBUS_9: Add a sample device with device attributes to kubeedge node", func() {
-			//Generating Device ID
+			// Generate Device ID
```
wbc6080, line 313: plz squash some commits https://kubeedge.io/docs/developer/Supplemental_developer_guide#ensuring-to-squash-similiar-commits

edge/test/integration/utils/helpers/helpers.go
1Shubham7: yes, that's exactly what's happening.

edge/test/integration/device/device_test.go
1Shubham7: >> so there is no need to recreate it here?

yes, DeviceATT and DeviceTW are created once and used in multiple testscases. this is done to have lesser devices otherwise creating and deleting them cause DB to go to lock state.
BTW this is how it was previously, I didn't change that logic.

>> Will there be an error if I create it repeatedly?

It may fail the ending testcases.

**detailed explanation:** here the testcase is "Update existing device with new attributes". But previously we were just creating a device and adding attibutes, rather than updating an existing device.

Previously:
```
It("TC_TEST_EBUS_11: Update existing device with new attributes", func() {

			//Generate a Device
			device := CreateDevice(DeviceIDWithAttr, "DeviceATT", "unknown")
			//Add Attribute to device
			AddDeviceAttribute(device, "Temperature", "50.50", "float")

			IsDeviceAdded := HandleAddAndDeleteDevice(http.MethodPut, ctx.Cfg.TestManager+Devicehandler, device)
			Expect(IsDeviceAdded).Should(BeTrue())

			Eventually(func() string {
				attributeDB := GetDeviceAttributesFromDB(DeviceIDWithAttr, "Temperature")
				common.Infof("DeviceID= %s, Value= %s", attributeDB.DeviceID, attributeDB.Value)
				return attributeDB.Value
			}, "60s", "2s").Should(Equal("50.50"), "Device Attributes are not updated within specified time")

		})
```

If you notice we just created a device and added attibute to it, this does not actually test what the testcase aims for ("Update existing device with new attributes"), that's another thing I fixed.

edge/test/integration/device/device_test.go
1Shubham7: sure

1Shubham7:
done, PTAL @wbc6080 : )

1Shubham7:
PTAL @wbc6080  
Now some tests are failing, these are some other issue, not a problem with this PR : )

wbc6080:
I tried to rerun the Unit test several times and found that similar errors still existed. According to your point, is there a problem with the test cases 13 and 14, causing CI to still have a probability of failure? @1Shubham7

1Shubham7:
> I tried to rerun the Unit test several times and found that similar errors still existed. According to your point, is there a problem with the test cases 13 and 14 or other reasons, causing CI to still have a probability of failure? @1Shubham7

hmm I see, still some issue with the aftersuite. the issue is not with any test case now. let me figure out what's breaking.

PR #6139, opened 2025-02-11
[Test]:Improve of ./pkg/stream/edgedlogconnection.go

vivekbisen04:
/kind test

This PR significantly improves test coverage for the `./pkg/stream/` package by adding comprehensive test cases for file   ./pkg/stream/edgedlogconnection . 
 

Part of **[lfx-mentorship] Enhance KubeEdge testing coverage initiative #6101**  

### **Before**  
 existing test coverage (**11.54%**)  

### **After**  
Achieved  **81.5%** test coverage

##  Screenshots  
Screenshot from 2025-02-11 16-00-50

## Commands to Test  
- **File path:** `./pkg/stream/  
- **Test execution commands:**  

  ```sh
  go test -coverprofile=coverage.out ./pkg/stream/...
  go tool cover -html=coverage.out -o coverage.html

### DIFF
--- a/pkg/stream/edgedlogconnection_test.go
+++ b/pkg/stream/edgedlogconnection_test.go
@@ -17,12 +17,68 @@ limitations under the License.
 package stream
 
 import (
+	"bufio"
+	"bytes"
 	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 )
+
+const (
+	defaultTestTimeout = 100 * time.Millisecond
+	defaultWaitTime    = 1 * time.Second
+)
+
+type mockSafeWriteTunneler struct {
+	written  []*Message
+	writeErr error
+	closeErr error
+	nextErr  error
+	nextType int
+	nextData []byte
+}
+
+func (m *mockSafeWriteTunneler) WriteMessage(msg *Message) error {
+	if m.writeErr != nil {
+		return m.writeErr
+	}
+	m.written = append(m.written, msg)
+	return nil
+}
+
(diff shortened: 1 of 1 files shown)


pkg/stream/edgedlogconnection_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The KubeEdge Authors.
+Copyright 2025 The KubeEdge Authors.
```
wbc6080, line 2: Don't change copyright time. We only need to add copyright when add new files

pkg/stream/edgedlogconnection_test.go
```diff
@@ -88,3 +139,183 @@ func TestLogConnection_CloseReadChannel(t *testing.T) {
...
+	assert := assert.New(t)
+
+	stop := make(chan struct{}, 1)
+	readChan := make(chan *Message, 2)
+
+	logs := &EdgedLogsConnection{
+		ReadChan: readChan,
+		MessID:   1,
+	}
+
```
wbc6080, line 153: Will it be better to use `defer close(logs.ReadChan)` here? It can effectively avoid forgetting to close the channel

pkg/stream/edgedlogconnection_test.go
```diff
@@ -88,3 +139,183 @@ func TestLogConnection_CloseReadChannel(t *testing.T) {
...
+			mockTunnel := &mockSafeWriteTunneler{}
+
+			errChan := make(chan error)
+			go func() {
+				errChan <- logs.Serve(mockTunnel)
+			}()
+
+			go func() {
+				time.Sleep(100 * time.Millisecond)
```
wbc6080, line 309: Would it be better to use `constant definitions`?

PR #6143, opened 2025-02-12
Upgrade ginkgo version in CI check

wbc6080:
**What type of PR is this?**
/kind feature

Currently, the `ginkgo` version in each CI check is set to **v2.9.5** https://github.com/kubeedge/kubeedge/blob/master/.github/workflows/main.yaml#L22, but actually in e2e and other tests, the dependent `ginkgo` version is **v2.17.1**. https://github.com/kubeedge/kubeedge/blob/master/go.mod#L25. Because they do not match, an error message will be output in the CI check log. Therefore, this PR completes the upgrade of the ginkgo version in the CI check.

image

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/main-arm64.yaml
+++ b/.github/workflows/main-arm64.yaml
@@ -19,7 +19,7 @@ on:
 
 env:
   CONTAINER_RUN_OPTIONS: " "
-  GINKGO_VERSION: "v2.9.5"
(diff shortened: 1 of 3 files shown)


fisherxu (commented):
/lgtm

Shelley-BaoYue (commented):
/approve

PR #6147, opened 2025-02-12
remove protocol from device models

### LINKED ISSUE #5969: Remove duplicate protocol definitions on devices and models
Remove the protocol definition on the model

**Why is this needed**:
Currently there are protocol definition fields on devices and device models, which is very confusing.Device supports multiple protocols and should not be constrained by the device model.

vaidikcode:
**What type of PR is this?**
/kind feature

This PR removes protocol definitions from DeviceModel CRDs while keeping them in Device CRDs. This change improves the separation of concerns between device models and device instances.

### Justification for removing protocol from DeviceModel:
Its better if
   - DeviceModel should define *what* a device can do (capabilities/properties)
   - Device CRD should define *how* to connect to the device (protocol/connectivity)

&
   - Same device model can be used with different protocols
   - Better supports multi-protocol devices

**Reduced Duplication too**
   - Eliminates redundant protocol definitions
   - Single source of truth for protocol configuration
   - Simplifies maintenance and updates

## Changes Made:
1. Removed protocol field from DeviceModel CRD in:
   ```yaml
   # filepath: /build/crds/devices/devices_v1beta1_devicemodel.yaml
   # filepath: /manifests/charts/cloudcore/crds/devices_v1beta1_devicemodel.yaml
   ```

Fixes #5969 

```release-note
NONE
```

### DIFF
--- a/build/crds/devices/devices_v1beta1_devicemodel.yaml
+++ b/build/crds/devices/devices_v1beta1_devicemodel.yaml
@@ -154,9 +154,6 @@ spec:
                       type: object
                   type: object
                 type: array
-              protocol:
-                description: 'Required for DMI: Protocol name used by the device.'
-                type: string
             type: object
         type: object
     served: true
@@ -229,9 +226,6 @@ spec:
                       type: string
                   type: object
                 type: array
-              protocol:
-                description: 'Required: Protocol name used by the device.'
-                type: string
             type: object
         type: object
     served: true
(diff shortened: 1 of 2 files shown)


WillardHu:
Thanks for your contribution, but there will be another solution for this issue. In addition, just deleting the protocol field will affect the work of the devicetwin module. I will close this PR.

PR #6148, opened 2025-02-12
feat: Add support for local image repository in keadm init

### LINKED ISSUE #6011: Can you provide a way to pull images from a local repository when "keadm init"?
I think that you should add a feature that the user can specify a local container image repository when "keadm init".
**Why is this needed**:
Because in some scenario there is no Internet connection.

vaidikcode:
Signed-off-by: vaidikcode <vaidikbhardwaj00@gmail.com>

**What type of PR is this?**
/kind feature

Adds support for pulling images from a local container registry during `keadm init`. This is particularly useful in air-gapped environments where internet access is restricted.

### Feature Implementation
- Added `ImageRepository` field to `KubeCloudHelmInstTool` struct
- Modified `handleProfile` to support custom image repositories
- Added unit tests to verify functionality

### Usage Example
```bash
# Initialize KubeEdge with local registry
keadm init --image-repository=localhost:5000
```

Fixes #6011 

```release-note
NONE
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/helm/installer.go
+++ b/keadm/cmd/keadm/app/cmd/helm/installer.go
@@ -61,6 +61,7 @@ type KubeCloudHelmInstTool struct {
 	DryRun           bool
 	Action           string
 	existsProfile    bool
+	ImageRepository  string
 }
 
 // InstallTools downloads KubeEdge for the specified version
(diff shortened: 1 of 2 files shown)


WillardHu:
https://github.com/kubeedge/kubeedge/blob/d27bd0f12380ff103ec3f207a1ffdd1fea2e87e1/keadm/cmd/keadm/app/cmd/helm/cloudcore.go#L322-L329

The private repository processing is supported.

Issue #6088, opened 2025-01-15
Service supports traffic closed-loop


JiaweiGithub:
**What would you like to be added/modified**:
![service-closed-loop](https://github.com/user-attachments/assets/5963738c-ba87-49a8-b2f6-10c3883ba953)**Why is this needed**:
Edge computing nodes are increasingly deployed in multi-region environments. Native Kubernetes topology-aware routing provides a mechanism to help keep traffic within the region where it originated.When calculating service endpoints, the EndpointSlice controller considers the physical topology of each endpoint (region and zone) and populates the hint fields to assign them to the appropriate region.

Native Kubernetes relies on kube-proxy for traffic management. However, in the KubeEdge architecture, there is no kube-proxy, and service communication relies on the edgemesh component.Therefore, this solution is based on edgemesh to implement a closed-loop for service communication traffic.

https://kubernetes.io/zh-cn/docs/concepts/services-networking/topology-aware-routing/

JiaweiGithub:
Currently, node group management supports functions similar to topology-aware.https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/node-group-management.md

JiaweiGithub:
Based on the debugging results of the actual code, there are some issues with this functionality:

1.The cloud does not watch the endpointSlice resource.
2.The NeedFilter method has a type conversion error; the content is actually of a type similar to endpointSlice resources and cannot be cast to UnstructuredList and Unstructured.

<img width="963" alt="Image" src="https://github.com/user-attachments/assets/d8c2afdb-be6f-435e-8b53-cf9aea024ab4" />

JiaweiGithub:
According to the latest research, endpointslice is not required in the traffic closed loop of kubeedge. There are two problems as follows.
1.unstructured type conversion error.
2.Filter method does not work when endpoints are synchronized.

PR #6149, opened 2025-02-13
Solve endpoints synchronization filtering not work

JiaweiGithub:
**What type of PR is this?**

/kind bug

There is no cache in the informer when synchronization starts, causing the filtering method to return false and not work. There is a cache later, and the edge  resource will not be updated if it has not changed.

Fixes #
https://github.com/kubeedge/kubeedge/issues/6088

```release-note

```

### DIFF
--- a/cloud/pkg/dynamiccontroller/filter/endpointresource/endpointresource.go
+++ b/cloud/pkg/dynamiccontroller/filter/endpointresource/endpointresource.go
@@ -1,13 +1,17 @@
 package endpointresource
 
 import (
+	"context"
+
 	v1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 
+	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/cloud/pkg/controllermanager/nodegroup"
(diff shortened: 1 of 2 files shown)


cloud/pkg/dynamiccontroller/filter/endpointresource/endpointresource.go
```diff
@@ -123,19 +127,32 @@ func filterEndpoints(targetNode string, obj runtime.Object) {
...
 	}
 	svcName := ep.GetName()
-	svcRaw, err := filter.GetDynamicResourceInformer(v1.SchemeGroupVersion.WithResource("services")).Lister().ByNamespace(ep.Namespace).Get(svcName)
-	if err != nil {
-		klog.Errorf("filter endpoint slice for svc %s error: %v", svcName, err)
-		return
+	var svcRaw interface{}
+
+	if !filter.GetDynamicResourceInformer(v1.SchemeGroupVersion.WithResource("services")).Informer().HasSynced() {
+		klog.Infof("services informer has not synced yet")
```
wbc6080, line 133:
```suggestion
		klog.Info("services informer has not synced yet")
```

cloud/pkg/dynamiccontroller/filter/util.go
```diff
@@ -1,36 +1,57 @@
...
+	if targetNodeName == epNodeName {
 		return true
 	}
-	targetNode, err := GetDynamicResourceInformer(v1.SchemeGroupVersion.WithResource("nodes")).Lister().Get(targetNodeName)
+
+	var getNode func(string) (interface{}, error)
+
+	// Define a function to get the node based on whether the informer is synced
+	if !GetDynamicResourceInformer(v1.SchemeGroupVersion.WithResource("nodes")).Informer().HasSynced() {
+		klog.Infof("nodes informer has not synced yet")
```
wbc6080, line 28:
```suggestion
		klog.Info("nodes informer has not synced yet")
```

PR #6150, opened 2025-02-13
Dynamiccontroller filterEvent object type convert error

JiaweiGithub:
**What type of PR is this?**

/kind bug

filterEvent.Object is of runtime.Object type and cannot be forced to *unstructured.Unstructured, causing the MessageFilter method to not work.

Fixes #
https://github.com/kubeedge/kubeedge/issues/6088

```release-note

```


Shelley-BaoYue (commented):
/approve 
/assign @WillardHu @wbc6080

PR #6188, opened 2025-03-06
[test]:improve code cov of server.go in cloud/cloudcore/app

Dhiren-Mhatre:
**What type of PR is this?**
/kind test
/kind bug

This PR improves code coverage for the cloud/cmd/cloudcore/app package by adding tests for previously uncovered function.

The changes increase test coverage for this package from approximately 28% to 52%.

Additionally, this PR fixes existing tests in server_test.go that were failing when run locally. I've made server.go more testable without changing its functionality, ensuring all tests now pass consistently.

NONE


cloud/cmd/cloudcore/app/server.go
```diff
@@ -268,34 +283,53 @@ func NegotiateTunnelPort() (*int, error) {
...
+	return nil
+}
+
+var updateCloudCoreConfigMapWarningf = klog.Warningf
+
+func updateCloudCoreConfigMap(c *v1alpha1.CloudCoreConfig) {
+	err := UpdateCloudCoreConfigMapWithClient(c, kubeClientGetter())
+	if err != nil {
+		updateCloudCoreConfigMapWarningf("Failed to update cloudcore config: %v", err)
+	}
```
wbc6080, line 323: Why do we need to redefine `klog.Warningf` here? Can't we just use `klog.Warningf` directly?

cloud/cmd/cloudcore/app/server_test.go
```diff
@@ -1,88 +1,709 @@
...
 	}
 }
+
+// TestNegotiateTunnelPortWithClient_ErrorCases tests error handling
+func TestNegotiateTunnelPortWithClient_ErrorCases(t *testing.T) {
+	// Test case where createNamespaceIfNeeded fails
+	t.Run("CreateNamespace fails", func(t *testing.T) {
+		// Save global functions and restore after test
+		restore := saveGlobals()
+		defer restore()
```
wbc6080, line 155: Can this function be combined with the subsequent `TestNegotiateTunnelPortWithClient`? I don't quite understand the point of having a separate function for the failed test case

wbc6080:
I found that the modification to the original code was just to improve the test, not that the original code had a bug, right? So I think the `kind-bug` tag can be removed.
/remove-kind bug

Dhiren-Mhatre:
> I found that the modification to the original code was just to improve the test, not that the original code had a bug, right? So I think the `kind-bug` tag can be removed. /remove-kind bug

The original code was not functioning correctly when I ran `go test -coverprofile=coverage.out ./cloud/cloudcore/app`. That's why I added the kind /bug label. Please correct me if I am mistaken.

cloud/cmd/cloudcore/app/server.go
Dhiren-Mhatre: It allows mocking warning logs in unit tests. Otherwise, they mostly remain uncovered.

cloud/cmd/cloudcore/app/server_test.go
Dhiren-Mhatre: The `TestNegotiateTunnelPortWithClient_ErrorCases` focuses specifically on error cases with separate test scenarios, while the `TestNegotiateTunnelPortWithClient` tests the happy path with multiple successful scenarios. If you suggest, I can combine them into one function.

Issue #6186, opened 2025-03-04
[LFX-Mentorship-Term1-2025-Tracking-Issue] Enhance KubeEdge testing coverage


Dhiren-Mhatre:
## Upstream Issue

#6101

## Legend

- ⬜: No PR issued yet
- ⏳: In Review
- ✅: Merged

## Pre-Existing Pull requests:

| Status                       | Period         | Directory                                              | Old Coverage                     | New Coverage                    | PR                              | Code Quality                  | Docs Quality                  | Bug Discovery                 |
| ---------------------------- | -------------- | ------------------------------------------------------ | -------------------------------- | ------------------------------- | ------------------------------- | ----------------------------- | ----------------------------- | ----------------------------- |
| <div align="center">⏳</div> | Before March 4 | edge/pkg/common/util/resource                          | <div align="center">0%</div>     | <div align="center">N/A</div>   | <div align="center">#6108</div> | <div align="center">N/A</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">✅</div> | Before March 4 | edge/pkg/common/message                                | <div align="center">0%</div>     | <div align="center">100%</div>  | <div align="center">#6112</div> | <div align="center">N/A</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">✅</div> | Before March 4 | pkg/stream/flushwriter/writer                          | <div align="center">0%</div>     | <div align="center">100%</div>  | <div align="center">#6114</div> | <div align="center">N/A</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">⏳</div> | Before March 4 | pkg/stream/tunnel                                      | <div align="center">0%</div>     | <div align="center">100%</div>  | <div align="center">#6116</div> | <div align="center">N/A</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">✅</div> | Before March 4 | pkg/stream/message                                     | <div align="center">N/A</div>    | <div align="center">100%</div>  | <div align="center">#6118</div> | <div align="center">N/A</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
| <div align="center">⏳</div> | Before March 4 | cloud/pkg/common/messagelayer/context                  | <div align="center">21%</div>    | <div align="center">100%</div>  | <div align="center">#6119</div> | <div align="center">N/A</div> | <div align="center">N/A</div> | <div align="center">N/A</div> |
[...]

PR #6196, opened 2025-03-12
[test]:improve coverage of containerexec_connection_test.go in cloud/…

Dhiren-Mhatre:
/kind test

This PR adds additional unit tests for the ContainerExecConnection in the cloudstream package to increase test coverage. Test coverage has been improved from minimal coverage to 82%, covering various edge cases and logging paths.

Key improvements include:
- Added tests for error handling in the Serve method
- Coverage for EOF and read error paths
- Tests for context cancellation behavior
- Coverage for retry logic in sendCloseMessage

Part of #6186 (Enhance KubeEdge testing coverage)

The PR focuses solely on improving test coverage without modifying the implementation code. Custom mocks were created to test various edge cases.

```release-note
NONE

### DIFF
--- a/cloud/pkg/cloudstream/containerexec_connection_test.go
+++ b/cloud/pkg/cloudstream/containerexec_connection_test.go
@@ -17,9 +17,12 @@ limitations under the License.
 package cloudstream
 
 import (
+	"context"
+	"errors"
 	"net/http"
 	"net/url"
 	"testing"
+	"time"
 
 	"github.com/emicklei/go-restful"
 	"github.com/stretchr/testify/assert"
@@ -81,16 +84,19 @@ func TestSetEdgePeerDone_Exec(t *testing.T) {
 		closeChan:    make(chan bool),
 	}
 
+	received := make(chan bool, 1)
 	go func() {
-		execConn.SetEdgePeerDone()
+		select {
+		case <-execConn.edgePeerStop:
+			received <- true
+		case <-time.After(100 * time.Millisecond):
+			received <- false
+		}
 	}()
 
-	select {
-	case <-execConn.edgePeerStop:
-		assert.True(true)
-	case <-execConn.closeChan:
-		assert.Fail("Expected edgePeerStop to receive but got closeChan")
-	}
+	execConn.SetEdgePeerDone()
+
+	assert.True(<-received, "edgePeerStop channel should receive a signal")
 }
 
 func TestEdgePeerDone_Exec(t *testing.T) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cloud/pkg/cloudstream/containerexec_connection_test.go [hunk 1]
```diff
@@ -164,3 +167,175 @@ func TestSendConnection_Exec(t *testing.T) {
...
+		MessageID:    1,
+		edgePeerStop: make(chan struct{}),
+		closeChan:    make(chan bool),
+	}
+
+	close(execConn.closeChan)
+
+	execConn.SetEdgePeerDone()
+
+	assert.True(true)
```
wbc6080, line 309: It simply closes the `closeChan` channel and then calls the `SetEdgePeerDone` method without checking whether the `edgePeerStop` channel receives the signal, so the correctness of the `SetEdgePeerDone` method cannot be verified.

I think you can start a goroutine to try to receive a signal from the `edgePeerStop` channel to verify that the method is correct like:
```
close(execConn.closeChan)

received := make(chan bool, 1)
    go func() {
        select {
        case <-execConn.edgePeerStop:
            received <- true
        default:
            received <- false
        }
    }()

execConn.SetEdgePeerDone()

```

Then use assert to determine the content of received chan

Dhiren-Mhatre: Thanks for the review and suggestion. I've applied your recommended approach using a goroutine to monitor the edgePeerStop channel and confirm it receives the signal.

```
func TestSetEdgePeerDone_Exec(t *testing.T) {
	assert := assert.New(t)

	execConn := &ContainerExecConnection{
		MessageID:    1,
		edgePeerStop: make(chan struct{}),
		closeChan:    make(chan bool),
	}

	received := make(chan bool, 1)
	go func() {
		select {
		case <-execConn.edgePeerStop:
			received <- true
		case <-time.After(100 * time.Millisecond):
			received <- false
		}
	}()

	execConn.SetEdgePeerDone()
	
	assert.True(<-received, "edgePeerStop channel should receive a signal")
}

```

wbc6080: Good job! And I think L299-L312 also need to be changed like this :)

Dhiren-Mhatre: Thank you for the follow-up!

wbc6080:
lgtm
cc @fisherxu @Shelley-BaoYue @WillardHu

PR #6225, opened 2025-03-26
Enhance keadm upgrade

### LINKED ISSUE #6211: Enhance keadm upgrade command
## What would you like to be added/modified

1. Add a `keadm backup edge` command;
2. If the upgrade fails in the upgrade command, it will no longer roll back automatically;
3. Trun the rollback command into a third-level command `keadm rollback edge` (Compatibility);
4. Make `keadm upgrade` no longer be an executable command, `keadm upgrade cloud/edge` was modified in v1.16;
5. Prevent backup, upgrade and rollback from being executed at the same time;
6. Remove the interference flags of the upgrade command and modify the way to repot the results to the node upgrade job;
## Why is this needed

Image

The node upgrade job executor will first back up the edge files in EdgeCore and then execute the `keadm upgrade` command, but the command does not have a disable-backup flag, this makes the backup also execute in the upgrade command.

If the upgrade fails, the upgrade command will automatically roll back. When the node upgrade job receives a report of upgrade failure, `keadm rollback` will also be triggered in EdgeCore. This means that upgrade command executed by the node upgrade job executor may trigger 2 backups and 2 rollbacks. Making upgrade-related commands(backup, upgrade and rollback) do just one thing will give users more flexibility.

The idempotency_record file prevents upgrade command from being executed at the same time. However, rollback and backup are also mutually exclusive with upgrade, so they need to be handled the same way.

There are too many flags(upgradeID, historyID, type, etc) that are not related to the upgrade itself. They are only used for node upgrade job, but for users who only use upgrade command, these flags are noise. We can use other way, such as recording the results of the command throgh a structed file, so that the node upgrade job knows the execution results of the command.

WillardHu:
**What type of PR is this?**

/kind feature

Mainly solved the problems mentioned in #6211. This PR is divided into 3 commits. The first is used to organize constants. The second is to move the meta client codes into the keadm/cmd/keadm/app/cmd/util/metaclient package. The third is to enhance the keadm upgrade.

It is easier to maintain common constants in one plase, moving them in the kubeedge/api allows users to use them without depending on the main project kubeedge. Dependency on kubeedge will lead to indirect dependency on some K8s related projects, which will greatly increase the management cost of go.mod.

I think it is a good idea to use edgecore to get the currently deployed version. The first solution that comes to mind is to add a `--version` flag to edgecore, bug it is not compatible with older versions because the edgecore will be started again after running `edgecore --version`. Therefore, for new versions, edgecore writes the version as a file to the /etc/kubeedge directory. For old versions, get  node resource `node.status.nodeInfo.kubeletVersion` through the metaserver to get the currently version. At this time, both the keadm ctl command and the keadm upgrade related commands will use the meta client, so it is moved to a more appropriate package.

Fixes #6211

```release-note

```


keadm/cmd/keadm/app/cmd/edge/upgrade.go
```diff
@@ -1,11 +1,11 @@
 /*
-Copyright 2022 The KubeEdge Authors.
+Copyright 2025 The KubeEdge Authors.
```
Shelley-BaoYue, line 2:
```suggestion
Copyright 2022 The KubeEdge Authors.
```

keadm/cmd/keadm/app/cmd/edge/upgrade_backup.go
```diff
@@ -0,0 +1,111 @@
...
+		Use:   "edge",
+		Short: "Back up important files for rollback edgecore.",
+		RunE: func(_cmd *cobra.Command, _args []string) error {
+			var err error
+			defer func() {
+				// Report the result of the backup process.
+				reporter := upgrdeedge.NewJSONFileReporter(upgrdeedge.EventTypeBackup,
+					"", executor.currentVersion)
+				if reperr := reporter.Report(err); reperr != nil {
+					klog.Errorf("failed to report upgrade result: %v", reperr)
```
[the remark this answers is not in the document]
WillardHu: Done

keadm/cmd/keadm/app/cmd/edge/upgrade_rollback.go
```diff
@@ -0,0 +1,123 @@
+package edge
```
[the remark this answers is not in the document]
WillardHu: Done

keadm/cmd/keadm/app/cmd/util/metaclient/metaclient.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The KubeEdge Authors.
+Copyright 2025 The KubeEdge Authors.
```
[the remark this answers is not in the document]
WillardHu: Done

keadm/cmd/keadm/app/cmd/edge/upgrade_backup.go
```diff
@@ -0,0 +1,111 @@
+/*
```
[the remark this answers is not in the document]
WillardHu: Done

keadm/cmd/keadm/app/cmd/edge/upgrade.go
WillardHu: Done

edge/cmd/edgecore/app/server.go
```diff
@@ -101,6 +101,9 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 
 			// To help debugging, immediately log version
 			klog.Infof("Version: %+v", version.Get())
+			if err := version.WriteEdgeCoreVersion(opts.ConfigFile); err != nil {
```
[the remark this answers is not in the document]
WillardHu: I have removed the code for obtaining the edgecore version through the version file and changed it to directly obtain the edgecore version through node info queried by metaserver. The node upgrade task will be highly dependent on metaserver.

I used a new commit to do this that can be used for revert the code.

edge/test/integration/utils/edge/tmpfile.go
```diff
@@ -1,11 +1,11 @@
 /*
-Copyright 2019 The KubeEdge Authors.
+Copyright 2025 The KubeEdge Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
wbc6080, line 8: It looks like extra spaces are added

keadm/cmd/keadm/app/cmd/edge/backup.go
```diff
@@ -0,0 +1,111 @@
...
+		return errors.New("cannot get the required current version")
+	}
+	if err := executor.baseUpgradeExecutor.postPreRun(opts.PreRun); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (executor *backupExecutor) backup(opts BaseOptions) error {
+	klog.Infof("backup process start ...")
```
wbc6080, line 93:
```suggestion
	klog.Info("backup process start ...")
```

keadm/cmd/keadm/app/cmd/edge/backup.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := os.MkdirAll(backupPath, os.ModePerm); err != nil {
+		return fmt.Errorf("failed to create backup path %s, err: %v", backupPath, err)
+	}
+	for _, file := range backupFiles {
+		dest := filepath.Join(backupPath, filepath.Base(file))
+		if err := files.FileCopy(file, dest); err != nil {
+			return fmt.Errorf("failed to backup file %s, err: %v", file, err)
+		}
+	}
+	klog.Infof("backup process successful")
```
wbc6080, line 109:
```suggestion
	klog.Info("backup process successful")
```

keadm/cmd/keadm/app/cmd/edge/join_others.go
```diff
@@ -239,6 +241,31 @@ func join(opt *common.JoinOptions, step *common.Step) error {
...
+	ctx := context.Background()
+	imageSet := util.EdgeSet(opt)
+	images := imageSet.List()
+
+	runtime, err := util.NewContainerRuntime(opt.RemoteRuntimeEndpoint, opt.CGroupDriver)
+	if err != nil {
+		return err
+	}
+
+	step.Printf("Pull Images")
```
wbc6080, line 254: Do we have `step.Print` ? Use it may be better

keadm/cmd/keadm/app/cmd/edge/backup.go
WillardHu: Done

keadm/cmd/keadm/app/cmd/edge/backup.go
WillardHu: Done

keadm/cmd/keadm/app/cmd/edge/join_others.go
WillardHu: I will create a new PR to replace the output of frequently used commands (join/upgrade/rollback/backup/reset) on the edge with the new output framework.

edge/test/integration/utils/edge/tmpfile.go
WillardHu: I used vscode to automatically generate license information, which automatically converted  4 spaces to Tab (because golang's indentation uses Tab by default). I have replaced them with a new commit.

WillardHu:
Isulad e2e test falis because access to gitee.com is unstable now.

Shelley-BaoYue (commented):
/approve 
/assign @wbc6080 @tangming1996 @luomengY

PR #6227, opened 2025-03-26
Edge node certificates should not be shared

WillardHu:
**What type of PR is this?**

/kind bug

Currently, if we copy /etc/kubeedge/certs of an already joined node to the new node, we can successfully connect to the  cloud without  a token. What we expect is that we cannot connect to the cloud successfully.

Fixes #

```release-note

```


WillardHu:
PTAL @Shelley-BaoYue @wbc6080 @fisherxu

wbc6080 (commented):
/lgtm

PR #6259, opened 2025-04-09
feat: add DisableCSINode feature gate to skip CSI initialization

### LINKED ISSUE #6232: In offline, edgecore error: `csi_plugin.go:300] Failed to initialize CSINode: error updating CSINode annotation: timed out waiting for the condition; caused by: node length from meta db is 0`

When starting EdgeCore offline, it fails with Failed to initialize CSINode after retrying: timed out waiting for the condition, causing EdgeCore to restart continuously.

`edgecore[3460]: F0328 18:11:07.681015    3460 csi_plugin.go:300] Failed to initialize CSINode: error updating CSINode annotation: timed out waiting for the condition; caused by: node length from meta db is 0`

Is this behavior expected? Is there any way to bypass it?

1. Disable API Server access on the edge node(Simulate offline).
2. Start EdgeCore using `systemd`(token is entered arbitrarily).
3. Observe the logs; EdgeCore repeatedly crashes and restarts.

**Environment**:

XmchxUp:
Introduce a new edgecore feature gate, DisableCSINode, to optionally disable CSI node initialization. This prevents Kubelet crashes in offline environments where CSINode creation fails due to API server unavailability, enhancing edgecore resilience.

**What type of PR is this?**

Fixes #6232

```release-note

```

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -37,6 +37,8 @@ const (
 	// DisableNodeTaskV1alpha2 disables the node task v1alpha2 feature, uses v1alpha1.
 	// TODO: After v1.23, this switch will be removed and only v1alpha2+ will be supported.
 	DisableNodeTaskV1alpha2 featuregate.Feature = "disableNodeTaskV1alpha2"
+	// DisableCSIVolumePlugin disables the in-tree CSI volume plugin support.
+	DisableCSIVolumePlugin featuregate.Feature = "DisableCSIVolumePlugin"
 )
 
 // defaultFeatureGates consists of all known Kubeedge-specific feature keys.
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -47,6 +47,7 @@ import (
 	"k8s.io/kubernetes/pkg/kubelet/config"
 	"k8s.io/kubernetes/pkg/kubelet/nodestatus"
 	kubelettypes "k8s.io/kubernetes/pkg/kubelet/types"
+	"k8s.io/kubernetes/pkg/volume"
 	csiplugin "k8s.io/kubernetes/pkg/volume/csi"
 
 	"github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2"
@@ -218,6 +219,14 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	}
 	edgedconfig.ConvertConfigEdgedFlagToConfigKubeletFlag(&edgedconfig.Config.TailoredKubeletFlag, &kubeletFlags)
 
+	// Register the DisableCSIVolumePlugin Custom feature gate.
+	featurekey := featuregate.Feature(kefeatures.DisableCSIVolumePlugin)
+	if err := utilfeature.DefaultMutableFeatureGate.Add(map[featuregate.Feature]featuregate.FeatureSpec{
+		featurekey: {Default: false, PreRelease: featuregate.Alpha},
+	}); err != nil {
+		return nil, fmt.Errorf("failed to set default DisableCSIVolumePlugin feature gate : %w", err)
+	}
+
 	// set feature gates from initial flags-based config
 	if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(kubeletConfig.FeatureGates); err != nil {
 		return nil, fmt.Errorf("failed to set feature gates from initial flags-based config: %w", err)
@@ -249,6 +258,8 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	}
 	MakeKubeClientBridge(kubeletDeps)
 
+	kubeletDeps.VolumePlugins = filterVolumePluginsByFeatureGate(kubeletDeps.VolumePlugins)
+
 	// source of all configuration
 	kubeletDeps.PodConfig = config.NewPodConfig(config.PodConfigNotificationIncremental, kubeletDeps.Recorder, kubeletDeps.PodStartupLatencyTracker)
 
@@ -263,6 +274,22 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	}
 
 	return ed, nil
+}
+
+// filterVolumePluginsByFeatureGate filters the list of volume plugins based on feature gate settings.
+func filterVolumePluginsByFeatureGate(plugins []volume.VolumePlugin) []volume.VolumePlugin {
+	// filter the current edged-supported CSI plugin enable/disable control through feature gates
+	res := []volume.VolumePlugin{}
+	for _, plugin := range plugins {
+		if plugin.GetPluginName() == csiplugin.CSIPluginName {
+			if utilfeature.DefaultFeatureGate.Enabled(kefeatures.DisableCSIVolumePlugin) {
+				klog.Infof("CSI plugin disabled by configuration, skipping %q", plugin.GetPluginName())
+				continue
+			}
+		}
+		res = append(res, plugin)
+	}
+	return res
 }
 
 func (e *edged) syncPod(podCfg *config.PodConfig) {


XmchxUp:
CC: @fujitatomoya @FengGaoCSC

fujitatomoya:
@XmchxUp this looks good to me. with this PR, we can disable the CSINode via edgecore.yaml, right? anyway, i will take a look at this PR tomorrow.

XmchxUp:
> we can disable the CSINode via edgecore.yaml, right?

yes, like this

```sh
~ # cat /etc/kubeedge/config/edgecore.yaml
apiVersion: edgecore.config.kubeedge.io/v1alpha2
database:
  aliasName: default
  dataSource: /var/lib/kubeedge/edgecore.db
  driverName: sqlite3
kind: EdgeCore
featureGates:
  disableCSI: true
```

pkg/features/features.go
```diff
@@ -41,4 +47,5 @@ const (
 var defaultFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	RequireAuthorization: {Default: false, PreRelease: featuregate.Alpha},
 	ModuleRestart:        {Default: false, PreRelease: featuregate.Alpha},
+	DisableCSI:           {Default: false, PreRelease: featuregate.Alpha},
```
fujitatomoya, line 50: the variable name should be the meaning of the flag without boolean? having disable in the name could be complicated for user to set.

```suggestion
	SupportCSINode:           {Default: true, PreRelease: featuregate.Alpha},
```

fujitatomoya (approved):
IMO, excluding `csiplugin.CSIPluginName` from volume plugin list looks good at edge. in default, edged still expects that CSINode resource to instantiate during start up. I am not sure if this is reasonable for edgecore use case, probably we can switch the default to false to disable CSINode at edge. but this is behavioral change for user, this needs to be discussed otherwise keeping current behavior is reasonable.

fujitatomoya:
@fisherxu @shenkonghui @wbc6080 what do you think about this feature gate to add for edged? we have particular use case that edgecore starts completely offline scenario, this has been one of the constraints that edgecore restarts always. we would love to share more information if there are some places that we can chat online.

pkg/features/features.go
```diff
@@ -33,6 +33,12 @@ const (
 	// alpha: v1.17
 	// owner: @micplus
 	ModuleRestart featuregate.Feature = "moduleRestart"
+	// SupportCSINode enables CSI (Container Storage Interface) support in edgecore.
+	// When set to true (default), CSI drivers are used for storage operations.
+	// Disabling this (set to false) will turn off all CSI-related functionality.
+	// This is useful in environments that do not require CSI or for troubleshooting.
+	// alpha: v1.21
+	SupportCSINode featuregate.Feature = "supportCSINode"
```
WillardHu, line 41: Feature gates are used to enable alpha or beta features. Is it more appropriate to use the edged configuration property to disable CIS?

WillardHu: And the default value can be enabled.

edge/pkg/edged/features/features.go
```diff
@@ -0,0 +1,33 @@
+package features
```
WillardHu, line 1: You can add a bool field(zero value is false) to Edged,  like `DisableCSI`? 
https://github.com/kubeedge/kubeedge/blob/833ba71817960f5c4453b70b29fd87a2c23993f2/staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/types.go#L99
```golang
type Edged struct {
    ...
    // DisableCSI disables CSI (Container Storage Interface) support in edgecore.
    // The default value is false, CSI drivers are used for storage operations.
    // Disabling this (set to true) will turn off all CSI-related functionality.
    // This is useful in environments that do not require CSI or for troubleshooting.
    DisableCSI bool `json:"disableCSI"`
}
```

pkg/features/features.go
XmchxUp: ok

WillardHu:
/lgtm

cc @Shelley-BaoYue

staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/types.go
```diff
@@ -117,6 +117,12 @@ type Edged struct {
 	// RegisterNodeNamespace indicates register node namespace
 	// default "default"
 	RegisterNodeNamespace string `json:"registerNodeNamespace,omitempty"`
+	// DisableCSI disables CSI (Container Storage Interface) support in edgecore.
+	// The default value is false, CSI drivers are used for storage operations.
+	// Disabling this (set to true) will turn off all CSI-related functionality.
+	// This is useful in environments that do not require CSI or for troubleshooting.
+	// default false
```
fujitatomoya, line 124: should we add this to `v1alpha2/default.go`?

fujitatomoya (commented):
@XmchxUp a minor comment just came up, can you check?

staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/types.go
XmchxUp: I think it's unnecessary, because the FeatureGates in the edgecore config will be initialized here.

https://github.com/kubeedge/kubeedge/blob/00f824650b1c424443e8dfe5458b0914d337a4c5/edge/cmd/edgecore/app/server.go#L98

If a feature doesn't exist in the map, it's considered false by default.

fujitatomoya:
@WillardHu that sounds okay to me, and probably better for the KubeEdge maintainability. but kinda my concern is if that takes some time to have the consensus and reintroduce the related code from kubelet? there could be obvious downside for edge footprint and resource consumption.

my suggestion here is that,

- This PR is aligned with current feature gate implementation of KubeEdge, so we could take this as it is to fix the blocking issue that is complete offline scenario at the boot time.
- In the mean time, we could reintroduce feature gate implementation to kubeedge/kubernetes? and then replace the feature gate implementation build on top of that?

i may be mistaken something, so please correct me if you are saying something different.

WillardHu:
Hi @XmchxUp 

Sorry, for future compatibility, you still need to modify the configuration definition, the last modification!
- If we want to disable CSI at edge node, modify the configuration edgecore.yaml
  ```yaml
  ..
  modules:
    edged:
      tailoredKubeletConfig:
        featureGates:
           DisableCSIVolumePlugin: true
  ```
- Add a constant in pkg/features/features.go. In the next version (v1.22) we will move it to kubeedge/kubernetes.(The first letters of K8s features are all capitalized https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go)
  ```golang
  DisableCSIVolumePlugin featuregate.Feature = "DisableCSIVolumePlugin"
  ```
- Modify the if condition
  ```golang
  import (
      utilfeature "k8s.io/apiserver/pkg/util/feature"

      kefeatures "github.com/kubeedge/kubeedge/pkg/features"
  )
  
  if utilfeature.DefaultMutableFeatureGate.Enable(kefeatures.DisableCSIVolumePlugin) {
      ...
  }
  ```
Please verify that this solution is feasible. If there are no problems, this PR will be merged.

cc @fujitatomoya  @Shelley-BaoYue

WillardHu:
From the unit test error, it seems that you need to add default values for k8s features before setting the feature gates from the configruation(The configuration file does not set the function gate).
https://github.com/kubeedge/kubeedge/blob/3ecf1217c7b782ad964764d40d49b58283c1e8fe/edge/pkg/edged/edged.go#L222

```diff
+ featurekey := featuregate.Feature(kefeatures.ModuleRestart)
+ utilfeature.DefaultMutableFeatureGate.Add(map[featuregate.Feature]featuregate.FeatureSpec{
+     featurekey: {Default: false, PreRelease: featuregate.Alpha},
+ })
// set feature gates from initial flags-based config
if err := utilfeature.DefaultMutableFeatureGate.SetFromMap(kubeletConfig.FeatureGates); err != nil {
    return nil, fmt.Errorf("failed to set feature gates from initial flags-based config: %w", err)
}
```

fujitatomoya:
see https://github.com/kubeedge/kubeedge/actions/runs/15438666147/job/43451012301?pr=6259

```console
edgecore log:
 W0604 09:34:45.911501   63454 validation.go:71] NodeIP is empty , use default ip which can connect to cloud.
I0604 09:34:45.911561   63454 server.go:104] Version: v1.20.0-beta.0.218+194d6310ae977f
I0604 09:34:45.928394   63454 sql.go:21] Begin to register twin db model
I0604 09:34:45.928488   63454 module.go:54] Module twin registered successfully
panic: feature "DisableCSIVolumePlugin" is not registered in FeatureGate "k8s.io/apiserver/pkg/util/feature/feature_gate.go:28"
```

XmchxUp:
The recently merged test code in the master branch has issues. I think it needs to be reverted.
@WillardHu 

#6344

WillardHu:
The https://github.com/kubeedge/kubeedge/pull/6345 PR is merged, please rebase the latest master.

WillardHu:
lgtm  cc @Shelley-BaoYue

fujitatomoya:
@WillardHu @Shelley-BaoYue @XmchxUp thanks for taking care of this, appreciate that 👍

PR #6281, opened 2025-04-17
[test]:improve coverage for nssm.go in keadm/app/cmd/util

Dhiren-Mhatre:
**What type of PR is this?**
/kind test

These changes are part of an effort to improve code quality and increase test coverage for the NSSM utility functions from 0% to 74%.

Part of #6186

NONE

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/nssm_test.go
+++ b/keadm/cmd/keadm/app/cmd/util/nssm_test.go
@@ -1,24 +1,210 @@
 //go:build windows
+/*
+Copyright 2025 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
 
 package util
 
 import (
-	"fmt"
+	"errors"
+	"reflect"
 	"testing"
+
+	"github.com/agiledragon/gomonkey/v2"
+	"github.com/stretchr/testify/assert"
 )
 
-func TestInstallNSSM(t *testing.T) {
-	err := InstallNSSM()
-	if err != nil {
-		t.Error(err)
+const testServiceName = "testService"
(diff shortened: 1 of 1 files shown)


keadm/cmd/keadm/app/cmd/util/nssm_test.go
```diff
@@ -1,24 +1,293 @@
-//go:build windows
```
wbc6080, line 1: It seems that nssm is not only for Windows platform, right?

keadm/cmd/keadm/app/cmd/util/nssm_test.go
```diff
@@ -1,24 +1,293 @@
...
+	defer p2.Reset()
+
+	result := IsNSSMInstalled()
+	assert.True(t, result)
+
+	p3 := gomonkey.ApplyMethod(reflect.TypeOf(cmd), "Exec",
+		func(_ *Command) error {
+			return errors.New("command failed")
+		})
+	defer p3.Reset()
```
wbc6080, line 51: the initialization of P1, P2, and P3 seems to appear repeatedly in multiple test functions. Can some of them be abstracted into a function for calling?

keadm/cmd/keadm/app/cmd/util/nssm_test.go
```diff
@@ -1,24 +1,293 @@
...
+
+	err = InstallNSSM()
+	assert.Error(t, err)
 }
 
-func TestIsServiceExist(t *testing.T) {
-	t.Log(IsServiceExist("Power"))
+func TestInstallNSSMService(t *testing.T) {
+	path := "C:\\path\\to\\service.exe"
```
wbc6080, line 86: The same question, is this only used in Windows environment? Should we consider Linux environment?

Dhiren-Mhatre: Regarding NSSM being Windows-only  yes, you're absolutely right. NSSM is specifically designed for Windows services. The challenge I'm facing is that when I include this build tag, my coverage reports show 0% since I'm developing on Linux. Without the tag, I get 74% coverage. I'm trying to balance accurate testing with meaningful coverage metrics. Since NSSM is Windows-specific, there's no need to consider Linux environments for these tests,they should only run on Windows platforms where NSSM can actually function. What do you think about the build tag issue?I'm leaning toward including the Windows build tag for correctness but wanted to get your thoughts.

wbc6080: I agree with you that showing the right build tag is more important than coverage reports :)

PR #6297, opened 2025-04-24
Reconstruct the metaserver db and replace beego with gorm

### LINKED ISSUE #6296: Refactor the db module in the metaserver on the edge.
The current edge SQLite database uses the Beego framework, and only the ORM module is utilized in the project. In this scenario, using Beego appears relatively heavyweight. Replacing Beego with GORM would be more lightweight. GORM is a complete ORM (Object-Relational Mapping) library that supports mapping between models and database tables, and can automatically handle common CRUD operations. GORM also provides many advanced features such as preloading, transactions, hooks, auto-migration, etc., which can meet complex database operation requirements. Additionally, it is more convenient for testing.
Currently, database operations in the project are scattered. In some places, operation functions are encapsulated, while in others, the underlying functions of Beego are called directly. Therefore, we can uniformly provide a database operation entry in the metaserver, and all parts requiring database operations should call the operation interfaces provided by this module.
After this refactoring, database-related operations will become clearer and more maintainable.
**Why is this needed**:

tangming1996:
**What type of PR is this?**
/kind feature

#6296 

Fixes #6296 

```release-note

```


edge/cmd/edgecore/app/server.go
```diff
@@ -194,6 +194,14 @@ func environmentCheck(skipCheck bool) error {
 
 // registerModules register all the modules started in edgecore
 func registerModules(c *v1alpha2.EdgeCoreConfig) {
+	dao.Init(
```
Shelley-BaoYue, line 205: Should we put this `Init` after all models registed as the note said at line 205?

edge/pkg/devicetwin/devicetwin.go
```diff
@@ -1,13 +1,12 @@
 package devicetwin
 
 import (
-	"k8s.io/klog/v2"
-
 	"github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2"
 	"github.com/kubeedge/beehive/pkg/core"
+	"k8s.io/klog/v2"
```
Shelley-BaoYue: Please fix the import sort

PR #6311, opened 2025-05-13
feat: Clean up non-Running Pods on edged shutdown 

### LINKED ISSUE #6303: Cleanup non-Running Pods on edged shutdown to avoid containerd name conflicts after restart
Enhance edged to clean up Pods that are not in the Running state upon shutdown, especially those whose containers are still transitioning from Created to Running. This would prevent errors like containerd: failed to reserve container name when edgecore restarts and attempts to recreate containers.

**Why is this needed**:
When Pods are initialized, there can be a delay between the container entering Created and Running states. If edgecore restarts during this window, the previous container may still exist but not yet be fully running, causing the new container creation request to fail due to a reserved container name in containerd. Cleaning up non-Running Pods on shutdown would avoid these conflicts, ensuring smoother recovery and Pod recreation after restarts.

Image

A similar logic, summarized in containerd.

> https://github.com/containerd/containerd/issues/4604#issuecomment-1027293621

XmchxUp:
This change ensures that edgecore cleans up Pods not in the Running state—especially those stuck in the Created-to-Running transition—during shutdown. It prevents container name reservation errors in containerd when edgecore restarts and tries to recreate containers.

**What type of PR is this?**

/kind feature

Fixes #6303 

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -38,6 +38,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/component-base/featuregate"
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
 	"k8s.io/klog/v2"
 	kubeletserver "k8s.io/kubernetes/cmd/kubelet/app"
 	kubeletoptions "k8s.io/kubernetes/cmd/kubelet/app/options"
@@ -121,16 +122,26 @@ func (e *edged) Enable() bool {
 func (e *edged) Start() {
 	klog.Info("Starting edged...")
 	kubeletErrChan := make(chan error, 1)
+
 	go func() {
+		origOsExit := klog.OsExit
+		klog.OsExit = func(code int) {
+			klog.Flush()
+			e.cleanupContainers()
+			origOsExit(code)
+		}
+		defer func() { klog.OsExit = origOsExit }()
+
 		err := DefaultRunLiteKubelet(e.context, e.KubeletServer, e.KubeletDeps, e.FeatureGate)
 		if err != nil {
 			if !kefeatures.DefaultFeatureGate.Enabled(kefeatures.ModuleRestart) {
-				klog.Errorf("Start edged failed, err: %v", err)
-				os.Exit(1)
+				klog.Exitf("Start edged failed, err: %v", err)
 			}
 			kubeletErrChan <- err
 		}
 	}()
+
+	defer e.cleanupContainers()
 
 	kubeletReadyChan := make(chan struct{}, 1)
 	go kubeletHealthCheck(e.KubeletServer.ReadOnlyPort, kubeletReadyChan)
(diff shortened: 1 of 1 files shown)


XmchxUp:
CC: @fujitatomoya @FengGaoCSC

edge/pkg/edged/edged.go
```diff
@@ -149,6 +162,43 @@ func (e *edged) Start() {
...
+		ctx, &runtimeapi.ContainerFilter{},
+	)
+	if err != nil {
+		klog.Errorf("cleanupContainers: CRI ListContainers failed: %v", err)
+		return
+	}
+
+	var toRemove []string
+	for _, c := range containers {
+		klog.Warningf("listResp: %v", c)
```
WillardHu, line 178: Why use warning levels here?

XmchxUp: debug reason

WillardHu:
I'm not sure if there are any hidden dangers in this. I think after edged(kubelet) calls CRI to start the container, the container app should be managed by the container runtime. Kubelet also has the same problem, how is it solved?

edge/pkg/edged/edged.go
WillardHu: This seems to be used to output the reason why the container is not running. Using the warning level does not seem appropriate, and `KubeletDeps.RemoteRuntimeService.ListContainers` returns `[]*Container` instead of `[]*ContainerStatusResponse`. You can use the info level to output the necessary fields.

XmchxUp:
> Kubelet also has the same problem, how is it solved?

Kubelet hasn't solved it.

This thinking was initially triggered by [this issue](https://github.com/kubeedge/kubeedge/issues/6232#issuecomment-2785470994), I think you can take a look at that issue first.


Maybe there's another way, perhaps this PR is unnecessary.

edge/pkg/edged/edged.go
XmchxUp: sorry, this is a log I added for my own testing; it's already been removed.

WillardHu:
This can be used as a temporary solution. After kubelet solves this problem, we will clean it up. What do you think? @fujitatomoya @Shelley-BaoYue

fujitatomoya:
i am good to go with that.

one thing that i want to mention about kubelet is, maybe kubelet leaves the pods as it is intentionally even if kubelet itself needs to restart. basically in cloud infrastructure, the priority is service availability as much as possible. that says, even kubelet needs to restart, leave the all containers as its current state and hopefully these containers can start running before kubelet restarts the service.
i mean, maybe this is the designed behavior for kubelet, in other word they would not fix this behavior at all.

but at edge, situation can be different. constrained resource platform takes more time to complete the pod initialization up to running state, that is why we meet this racy condition, i guess.

either way, i think that makes sense. for maintainability, maybe we are inclined to be compatible behavior with kubelet? if that is so, i think totally understandable.

WillardHu:
HI @XmchxUp, you can rebase the latest master branch to fix compilation issues.
```
gti fetch <remote>
git rebase <remote>/master
```
And rebase the pointless commit "update".
```
git rebase -i HEAD~2
```

edge/pkg/edged/edged.go
```diff
@@ -121,17 +122,27 @@ func (e *edged) Enable() bool {
...
 			if !kefeatures.DefaultFeatureGate.Enabled(kefeatures.ModuleRestart) {
-				klog.Errorf("Start edged failed, err: %v", err)
-				os.Exit(1)
+				klog.Exitf("Start edged failed, err: %v", err)
 			}
 			kubeletErrChan <- err
 		}
 	}()
 
+	go e.cleanupContainers()
```
WillardHu, line 144: defer e.cleanupContainers()

WillardHu:
> In my understanding, the reviewer can able to do this(Github provides this), right?

It is possible to do it via the page button, but may not be possible with comment <label>. It is better for contributors to rebase commits.

WillardHu:
/lgtm 

cc @Shelley-BaoYue

XmchxUp:
> /lgtm
> 
> cc @Shelley-BaoYue

Thanks.

Shelley-BaoYue (commented):
/approve 
keadm often encounters this issue when installing EdgeCore. Althouth I haven't encounter this issue when start applications at edge, this problem does exist. 
I agree with @WillardHu  I'm not sure whether this modification will cause other problems, I will merge thie PR. If there is any update in kubelet, we will update it here. Thanks for your contributions!

PR #6321, opened 2025-05-15
fix keadm batch genconfig

tangming1996:
**What type of PR is this?**
/kind bug

When join nodes in batch, the architecture of the nodes will be automatically obtained via commands, eliminating the need for manual configuration by users. 

Code link: https://github.com/kubeedge/kubeedge/blob/master/keadm/cmd/keadm/app/cmd/edge/batch.go#L278

Fixes #

```release-note

```


Shelley-BaoYue (commented):
/lgtm

Issue #6339, opened 2025-05-28
Resource Update Control at Edge


fujitatomoya:
**What happened and what you expected to happen**:

For the use cases such as drones, robotics/robots and autonomous cars, the edge system application should be able to hold/release the lock for CRD (Pods, ConfigMap, Deployment and so on) update at edge.
So that user application (system application running in the host system where edgecore runs) can confirm and control the application pods update timing to release the lock.
For example, when the drone is on the ground and only in the updating state, edged redeploy the pods and deployment with requesting to containerd. The same requirement for robotics/robots, only when the system application determines it safe to update the application pods or any other custom resources, it can do so.
Otherwise, edged sends the request to containerd to recreate the application pods once queued to MetaManager via `CloudCore(EdgeController) ---(WebSocket/QUIC)---> EdgeCore(EdgeHub) ---> MetaManager ---> edged ---> container`.

Usually for the use cases such as sensing data collection or completely stateless application, this is not gonna be any problems because it can update the application pods anytime done by CRD updates via Kubernetes. Probably just temporarily sensing data can be lost, but it does not really matter for those use cases. The sensing data will be coming once the new application pods are deployed and started.

This sounds pretty much for common use case at edge application lifecycle management.
But i could not find the similar features or functionalities to support this requirement, so i would like to bring this up as question to get more feedback from the community.

i would love to know how the developers are approaching or doing work-around for this requirement.

**How to reproduce it (as minimally and precisely as possible)**:

This is not issue but new use case and possible requirement discussion.

**Anything else we need to know?**:

i found https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/edge-node-upgrade.md and https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/OTA(over-the-air)-upgrades-for-edge-node.md, but this seems different use case to support the requirement and feature to upgrade `edgecore`, not the application pods or CRDs controlled by `edgecore`.

**Environment**:
- Kubernetes version (use `kubectl version`): Any
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): Any

fujitatomoya:
@Shelley-BaoYue @WillardHu any thoughts about this?

we are open for any solutions or this could be possible new feature for KubeEdge? what do you think?

CC: @XmchxUp @FengGaoCSC

PR #6348, opened 2025-06-12
Proposal: Resource Update Control at Edge

fujitatomoya:
> [!NOTE]
> This PR is still under draft

**What type of PR is this?**

/kind feature

Some edge application use cases require the full control when to update and apply the resource at edge.
This is the proposal for new feature to add the control mechanism at edge.

Part of https://github.com/kubeedge/kubeedge/issues/6339

```release-note
NONE
```


fujitatomoya:
CC: @FengGaoCSC @XmchxUp

fujitatomoya:
need to rebase on https://github.com/kubeedge/kubeedge/pull/6362

fujitatomoya:
@WillardHu @Shelley-BaoYue this is ready to review. is there any specific meeting that we should bring this topic for further discussion? if there is, i would love to join and share this in that meeting.

fujitatomoya:
@XmchxUp thanks for pushing the change though 👍

docs/proposals/sig-node/edge-resource-upgrade-control.md
```diff
@@ -0,0 +1,157 @@
...
+
+  | Resource           | Control Required | Description / Reason                                                               |
+  | ------------------ | ---------------- | ---------------------------------------------------------------------------------- |
+  | **Pods**           | ✅ Yes            | Primary unit of runtime workload. Restart affects running application.             |
+  | **Deployments**    | ✅ Yes            | Controls rollout strategy. Might recreate Pods if spec changes.                    |
+  | **StatefulSets**   | ✅ Yes            | Stateful services (like databases); restart or scale could be dangerous.           |
+  | **DaemonSets**     | ✅ Yes            | Often used in edge use-cases for agents, telemetry, etc.                           |
+  | **ConfigMaps**     | ⚠️ *Maybe*        | Apps *mount* or *env-inject* values; updates have no effect unless pod restarts.   |
+  | **Secrets**        | ⚠️ *Maybe*        | Same behavior as ConfigMaps — no automatic propagation into running Pods.          |
+  | **CRDs / Volumes** | ❓ Maybe          | Depending on use case. Might not require gating, but may be referenced indirectly. |
```
Shelley-BaoYue, line 237: Thanks @fujitatomoya @XmchxUp  I think it's a great feature and I appreciate this proposal. 
CRDs could also be considered for design, such as controlling whether to prepull images at the edge, which is useful in smart driving scenarios. 
The feature about `node upgrade OTA` that is similiar to this feature. I'm not sure if there are any parts that can common design or modules in implementation. You can ref to https://github.com/kubeedge/kubeedge/blob/4f0664f94e0401d669b420870736b12eb2293f59/docs/proposals/sig-node/5589-ota-upgrades-for-edge-node/README.md 
If this proposal is in a completed state, please remove the draft label : )

fujitatomoya: Definitely controlling the CRDs like pre-pull image via this feature would be nice to have. we can have it in the scope, but 1st thing that we need to have is `Pods`, `Deployment` and `DaemonSets`, to switch over the application based on the confirmation from edge node itself. as you pointed out, there could be possibility that some common areas to be shared with `node upgrade OTA`.

@FengGaoCSC @XmchxUp any objections?

XmchxUp: While creating a new Pod upgrade workflow like OTA is certainly feasible—it enforces an upgrade only after explicit confirmation from the edge—it effectively bypasses the native behavior of Kubernetes resources like Deployments and DaemonSets. So this method cannot be directly applied within those native controllers.

If we want to support upgrade control without changing how users interact with standard resources like Deployments or DaemonSets, we might need to hook into their reconciliation logic, intercept updates, and translate them into corresponding custom CRDs to be sent to the edge. This approach could enable confirmation-based upgrades within the standard resource framework but would likely require significant development effort.

I'm not entirely sure if this line of thinking is correct,

FengGaoCSC: @Shelley-BaoYue :

I also consider that OTA like method targets on the no existed CRDs but currently for switch over the application that target on the kubernete native resource like `pod`, `deployment`, `daemonset` etc, create a new CRDs that we should consider to hook into kubernete   reconciliation logic and then update and translate them into edge with new CRDs, this approach may invoke significant modification base on the current architecture and significant effort.

How about you consideration?
 Do you think @WillardHu 's suggestion on https://github.com/kubeedge/kubeedge/issues/6339#issuecomment-2918512744 is also a choice?

regards

FengGaoCSC: @Shelley-BaoYue 

friendly ping

Shelley-BaoYue: Sorry for late reply. As for native resources like pod, deployment, I agree with this proposal and https://github.com/kubeedge/kubeedge/issues/6339#issuecomment-2918512744, using annotations in yaml and `keadm ctl` at edge.  
I would like to cinfirm again, for native resources, we will choose this approach introduced in this proposal. And for CRD, we can handle them in the way similiar to `node upgrade OTA`?  I'm not sure if I understood correctly, welcome any further discussion.

FengGaoCSC: >  I would like to cinfirm again, for native resources, we will choose this approach introduced in this proposal. And for CRD, we can handle them in the way similiar to `node upgrade OTA`? I'm not sure if I understood correctly, welcome any further discussion.

- Yes, I think so. your understand is  correct!

FengGaoCSC: @Shelley-BaoYue :

Friendly could you take this proposal  for discussion, so we can implement and do some testing work.

Shelley-BaoYue: > Friendly could you take this proposal for discussion, so we can implement and do some testing work.

@FengGaoCSC  I have took this proposal for discussion in community meeting this afternoon. There are no other objections with this proposal. I believe we can do the implement work now : )  

/approve

WillardHu:
The overall process is look great to me. I think we could add some details about how resources are holded and recovered.

XmchxUp:
> The overall process is look great to me. I think we could add some details about how resources are holded and recovered.

Hi @WillardHu,

I’d like to confirm the recovery behavior you're referring to.

Do you mean that after the edge node restarts, it should recover held pod upgrade messages and possibly allow them to be unheld?

Currently, the implementation behaves as follows:
    - When edged starts, it retrieves pod information from metamanager.
    - If a pod has the annotation edge.kubeedge.io/hold-upgrade=true and is in Pending phase, we treat it as an upgrade pod and add it into heldPodUpdates for holding.

```go
func (e *edged) handlePodListFromMetaManager(content []byte, updatesChan chan<- interface{}) (err error) {
            ....
		if filterPodByNodeName(&pod, e.nodeName) {
			if pod.Annotations["edge.kubeedge.io/hold-upgrade"] == "true" && pod.Status.Phase == v1.PodPending {
				key := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
				e.heldPodUpdates[key] = append(e.heldPodUpdates[key], kubelettypes.PodUpdate{Op: kubelettypes.SET, Pods: []*v1.Pod{&pod}, Source: kubelettypes.ApiserverSource})
				continue
			}
```

WillardHu:
> Hi @WillardHu,
> 
> I’d like to confirm the recovery behavior you're referring to.

Currently, the PodUpdate channel is a function parameter. When the user confirms to update the Pod, we need to consider how to put the Pod to be updated back into the PodUpdate channel.

XmchxUp:
> Currently, the PodUpdate channel is a function parameter. When the user confirms to update the Pod, we need to consider how to put the Pod to be updated back into the PodUpdate channel.

Updated.

PR #6356, opened 2025-06-13
Fix some bugs in NodeUpgradeJob

WillardHu:
**What type of PR is this?**

/kind bug

1. After confirming the upgrade, the node upgrade action should start from the backup.
2. After the node task recovers from timeout, the reason should be set to empty.

Fixes #

```release-note

```


Shelley-BaoYue (commented):
/lgtm 
/approve

PR #6361, opened 2025-06-15
feat: add EC key support, improve script flexibility, fix unsafe variable usage

### LINKED ISSUE #6360: ENHC: Update certgen.sh: Add EC Key Support and Password Flexibility
**What would you like to be added or changed?**
Support EC key generation in the `certgen.sh` script. The script should also support generating keys with or without passwords and fix unquoted variable usage.

**Why is this needed?**
Currently, KubeEdge supports EC keys generated automatically, but it doesn't support PKCS#8 keys as mentioned in the issue https://github.com/kubeedge/kubeedge/issues/5599. The existing `certgen.sh` script is outdated, only supports RSA (with password), and doesn't align with the current key requirements, which leads to confusion.

To ensure consistency and flexibility, the script should:

* Add EC key generation support
* Allow optional password protection
* Retain backward compatibility with RSA
* Fix unsafe use of unquoted variables

Spiral-Memory:
**What type of PR is this?**

/kind feature
/kind cleanup

This PR enhances the `certgen.sh` script by:

* Adding support for EC key generation alongside existing RSA
* Allowing keys to be generated with or without passwords
* Preserving backward compatibility
* Improving script safety by quoting variables properly

The current script only supports RSA with a password, which is inconsistent with the modern usage of EC keys in KubeEdge. This update brings flexibility and better alignment with existing KubeEdge practices.

Fixes #6360 

The changes are backward compatible. The default behavior remains RSA with password unless explicitly configured. Extensive care was taken to avoid regressions.

```release-note
certgen.sh now supports EC key generation and the option to skip password protection. Variables are safely quoted to avoid script execution issues.
```

### DIFF
--- a/build/tools/certgen.sh
+++ b/build/tools/certgen.sh
@@ -234,4 +234,111 @@ $(pr -T -o 4 "${certPath}/${name}.key")
 EOF
 }
 
-$@
+cmd=$1
+shift || true
+
+if [ -z "$cmd" ]; then
+  echo ""
+  echo "Usage: ./certgen.sh <command> [args]"
+  echo ""
+  echo "Available commands:"
+  echo "  genCertAndKey     Generate CSR and certificate"
+  echo "  genCert           Sign CSR into certificate"
+  echo "  genCsr            Generate private key and CSR"
+  echo "  stream            Generate stream certs using K8S CA"
+  echo ""
+  echo "Run './certgen.sh help' to see environment variables and examples"
+  echo ""
+  exit 0
+fi
+
+case "$cmd" in
+  "genCertAndKey")
+    genCertAndKey "$@"
+    ;;
+  "genCert")
+    genCert "$@"
+    ;;
+  "genCsr")
+    genCsr "$@"
+    ;;
+  "stream")
+    stream "$@"
+    ;;
+  "help"|"-h"|"--help")
+    subcmd=$(echo "$1" | tr '[:upper:]' '[:lower:]')
+    case "$subcmd" in
+      "gencertandkey")
+        echo ""
+        echo "genCertAndKey <name> [ip1 ip2 ...]"
+        echo "  - Generates key, CSR, and cert signed by local CA"
+        echo "  - Uses ENV variables:"
+        echo "      SUBJECT, KEY_TYPE (rsa/ec), USE_PASSWORD, KEY_PASSWORD"
+        echo "      CA_PATH, CERT_PATH"
+        echo "  - Example:"
+        echo "      KEY_TYPE=rsa USE_PASSWORD=true KEY_PASSWORD=secret123 \\"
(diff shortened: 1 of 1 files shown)


build/tools/certgen.sh
```diff
@@ -39,47 +81,62 @@ ensureCommand() {
...
+    )
+
+    if [ "$usePassword" = "true" ]; then
+        cmd+=( -passin pass:$keyPassword )
     fi
+
+    "${cmd[@]}"
 }
 
 genCertAndKey() {
```
WillardHu, line 122: Thank you for your contribution!

For now we probably only need `genCertAndKey`, `opts` and dependent functions. 
Could you limit the scope of the command? For example:
```shell
function help() {
}

function generate() {
}

cmd=$1
shift
case "$cmd" in
  "help")
    help $@
    ;;
  "generate")
    generate $@
    ;;
  *)
    # error
    ;;
)
esac
```

Spiral-Memory: Here's a quick example that (if I understood you right) scopes the command's limit, like you showed:

```
# This should NOT be publicly exposed
stream() {
}

# This should NOT be publicly exposed
edgesiteServer(){
}

#Same for other few private functions

cmd=$1
shift || true

case "$cmd" in
  "genCertAndKey")
    genCertAndKey "$@"
    ;;
  "opts")
    opts "$@"
    ;;
  *)
    echo "Unknown command: $cmd"
    echo "Allowed: genCertAndKey, opts"
    exit 1
    ;;
esac

```
If that’s correct, just one thing to flag - I noticed the docs mention generating stream certs using this same certgen.sh script. So scoping it like this might conflict with that:
https://kubeedge.io/docs/advanced/debug

image

WillardHu: Currently, CloudCore installed with helm chart(keadm init) will automatically create a certificate for the stream module. For compatibility, we could also keep stream() function.

In addition, the stream module is used to hijack the traffic of api-sever accessing the kubelet debugging capabilities (exec/logs/metric), so it requires the mtls between the api-server(client) and the kubelet(server), this is a issue that needs to be fixed.

Spiral-Memory: Sure @WillardHu 
I've limited the scope as per our discussion and kept stream() for compatibility. The other functions are now internal.

Along with that, I felt it would be helpful to add help commands - as a new user, it makes the utility much easier to use.

Regarding the other issue you mentioned, I’m still quite new to it. I’ve got a basic understanding, but I’ll do some deeper research first. Once I have a better grasp, we can create an issue for it and work on it.

Thanks!

Spiral-Memory:
> New changes are detected. LGTM label has been removed.

Rebased with master branch to run docker e2e test again

Spiral-Memory:
@WillardHu Please re-review this

Shelley-BaoYue (commented):
/approve

PR #6365, opened 2025-06-21
Decouple e2e

### LINKED ISSUE #6212: CI enhancement on e2e test case.
In current e2e test case, we relay on build result and the invoke path like
https://github.com/SamYuan1990/kubeedge/blob/master/tests/scripts/execute.sh#L46
https://github.com/SamYuan1990/kubeedge/blob/master/hack/local-up-kubeedge.sh#L162

which there are two items discuss:
1. the local-up-kubeedge.sh invokes https://github.com/SamYuan1990/kubeedge/blob/master/hack/lib/golang.sh#L106
which means the e2e test case is not only start a kind cluster for test, as 
https://github.com/SamYuan1990/kubeedge/blob/master/tests/scripts/execute.sh#L46
but also including compile e2e test case and build containers.

ref to our current tooling image impls a pass build result by artifactory, so, for e2e test case, we can decouple compile and build from create cluster, as log msg as https://github.com/SamYuan1990/kubeedge/blob/master/tests/scripts/execute.sh#L47

2. as it's not decoupled today, which if we test CI pipeline on own fork. 
> the case is for CI developer, we should test CI on personal fork before submit to upstream as PR.

The fork creation usually missing tag, just default branch.
where https://github.com/SamYuan1990/kubeedge/blob/master/hack/lib/golang.sh#L103-L106 the e2e test case in fact breaks here as tag missing.
and... the log lost as https://github.com/SamYuan1990/kubeedge/blob/master/tests/scripts/execute.sh#L47

3. optional, if possible, can we considering remove or enhance tag checking in non default CI ? as leave it with a default value v0.0.0(which always error with a version checking?), instead of exit 1.

**Why is this needed**:

From previous, we can see current running path has:
- log is not correct, as build error exits with log as cluster start fail as hint.
- compile once and run any where as pass build result by artifactory.

Hackmey:
**What type of PR is this?**

Adds a check for existing cloudcore and edgecore binaries before triggering a build.

Prevents unnecessary rebuilding during E2E runs if the binaries are already present.

Helps speed up local/dev and CI workflows by reusing previously built binaries when available.

Also introduces a fallback mechanism for missing git tags, defaulting to v0.0.0 instead of failing.

Fallbacks to a default version
Uses the existing binaries while tests.

Fixes #6212

After this change is applied:
Screenshot from 2025-06-21 19-28-44

Exits with build error as the git tag is set to defualt.

### DIFF
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -24,6 +24,8 @@ CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"containerd"}
 KIND_IMAGE=${1:-"kindest/node:v1.30.0"}
 SKIP_CR_INSTALL=${SKIP_CR_INSTALL:-"false"}
 echo -e "The installation of the cni plugin will overwrite the cni config file. Use export CNI_CONF_OVERWRITE=false to disable it."
+CLOUDCORE_BIN="${KUBEEDGE_ROOT}/_output/local/bin/cloudcore"
+EDGECORE_BIN="${KUBEEDGE_ROOT}/_output/local/bin/edgecore"
 
 if [[ "${CLUSTER_NAME}x" == "x" ]]; then
   CLUSTER_NAME="test"
@@ -345,6 +347,7 @@ cleanup
 source "${KUBEEDGE_ROOT}/hack/lib/init.sh"
 source "${KUBEEDGE_ROOT}/hack/lib/install.sh"
 
+echo "[INFO] Setting up environment"
 kubeedge::golang::setup_env
 
 if [[ "${SKIP_CR_INSTALL}" = "false" ]]; then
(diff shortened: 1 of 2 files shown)


hack/local-up-kubeedge.sh
```diff
@@ -345,6 +347,7 @@ cleanup
 source "${KUBEEDGE_ROOT}/hack/lib/init.sh"
 source "${KUBEEDGE_ROOT}/hack/lib/install.sh"
 
+echo "[INFO] Setting up environment"
 kubeedge::golang::setup_env
```
fujitatomoya, line 351: Why do we not put this `INFO` message in the function `kubeedge::golang::setup_env`, so that we can make sure the `INFO` message will be printed whenever this function called, less maintenance to avoid putting the same `INFO` message elsewhere. (the same comment goes to the following `INFO` printing below.)

Hackmey: Yes that would be great, will work on it and add more such wherever needed to properly debug if something went wrong during this cycle is invoked.

hack/local-up-kubeedge.sh
```diff
@@ -353,14 +356,21 @@ else
...
 
-build_cloudcore
-build_edgecore
+if [ ! -f "$CLOUDCORE_BIN" ] || [ ! -f "$EDGECORE_BIN" ]; then
+  echo "[INFO] Building cloudcore and edgecore..."
+  build_cloudcore
+  build_edgecore
+else
+  echo "[INFO] cloudcore and edgecore already exist. Skipping build."
+fi
```
wbc6080, line 371: local-up script is mainly used by developers for debugging. In this case, cloudcore and edgecore need to be rebuilt based on code modifications. Therefore, even if there are compiled files in the corresponding paths, we still need to rebuild them.

Hackmey: Okay, got it. sorry for the hassle. Is there any other way to solve this ? otherwise i will just keep the fallback to tag.

wbc6080: I suggest that you do not need to change this part. Rebuilding cloudcore and edgecore is necessary.

wbc6080:
close it because #6406 
/close

PR #6406, opened 2025-07-10
fix: fallback to default tag if none found

Hackmey:
What type of PR is this?

Introduces a fallback mechanism for missing git tags, defaulting to v0.0.0 instead of failing.

NOTE : Creating this PR because #6365  became messy.

Which issue(s) this PR fixes:
Fallbacks to a default version

Fixes https://github.com/kubeedge/kubeedge/issues/6212

Exits with build error as the git tag is set to defualt.


wbc6080 (commented):
/lgtm

Shelley-BaoYue (commented):
/approve

PR #6412, opened 2025-07-14
Improving read timeout handling

PulkitDadwal:
/kind bug

Corrected a bug in which the WebSocket read timeout was not being enforced correctly. This would cause connections to hang forever if the server doesn't reply, which is a functional flaw. By ensuring the deadline for reading is set and enforced, the incorrect behavior is fixed.

```
Now, if the cloud server does not reply within the read timeout configured, the edge node will notice the timeout and recover from it (e.g., reconnect), rather than waiting forever. This enhances user reliability and error recovery with edge nodes.

```

### DIFF
--- a/pkg/viaduct/pkg/conn/ws.go
+++ b/pkg/viaduct/pkg/conn/ws.go
@@ -151,6 +151,10 @@ func (conn *WSConnection) handleMessage() {
 
 func (conn *WSConnection) SetReadDeadline(t time.Time) error {
 	conn.ReadDeadline = t
+	// Actually set the deadline on the underlying websocket connection
+	if conn.wsConn != nil {
+		return conn.wsConn.SetReadDeadline(t)
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


WillardHu:
Thanks for your contribution, DCO is missing, you can follow this [docs](https://kubeedge.io/docs/developer/Supplemental_developer_guide#dco) to fix it.

pkg/viaduct/pkg/conn/ws.go
```diff
@@ -151,6 +151,11 @@ func (conn *WSConnection) handleMessage() {
 
 func (conn *WSConnection) SetReadDeadline(t time.Time) error {
 	conn.ReadDeadline = t
+	conn.locker.Lock()
```
WillardHu: It seems that no locks need to be added here, the wsConn field is created at initialization, and these methods also seem to have no concurrency issues.

PulkitDadwal: DCO added and locks removed

PR #6418, opened 2025-07-17
feat(edge): add hold-and-release mechanism for resource updates via a…

### LINKED ISSUE #6339: Resource Update Control at Edge

XmchxUp:
…nnotation and keadm

**What type of PR is this?**

/kind feature

Fixes #6339

```release-note

```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -319,7 +319,7 @@ func (e *edged) syncPod(podCfg *config.PodConfig) {
 		}
 
 		op := result.GetOperation()
-		_, resType, resID, err := commonmsg.ParseResourceEdge(result.GetResource(), op)
+		ns, resType, resID, err := commonmsg.ParseResourceEdge(result.GetResource(), op)
 		if err != nil {
 			klog.Errorf("failed to parse the Resource: %v", err)
 			continue
@@ -332,17 +332,6 @@ func (e *edged) syncPod(podCfg *config.PodConfig) {
 		}
 
 		switch resType {
-		case model.UnholdUpgradeOperation:
-			key := string(content)
-			if updates, exists := e.heldPodUpdates[key]; exists {
-				klog.V(4).Infof("Unholding pod upgrade: %s", key)
-				for _, update := range updates {
-					rawUpdateChan <- update
-				}
-				delete(e.heldPodUpdates, key)
-			} else {
-				klog.V(4).Infof("No held updates found for pod %s", key)
-			}
 		case model.ResourceTypePod:
 			if op == model.ResponseOperation && resID == "" && result.GetSource() == modules.MetaManagerModuleName {
 				err := e.handlePodListFromMetaManager(content, rawUpdateChan)
@@ -358,6 +347,18 @@ func (e *edged) syncPod(podCfg *config.PodConfig) {
 					continue
 				}
 				podCfg.SetInitPodReady(true)
+			} else if op == model.UnholdUpgradeOperation {
+				key := fmt.Sprintf("%s/%s", ns, resID)
+				if updates, exists := e.heldPodUpdates[key]; exists {
+					klog.V(4).Infof("Unholding pod upgrade: %s", key)
+					for _, update := range updates {
+						rawUpdateChan <- update
+					}
+					delete(e.heldPodUpdates, key)
+				} else {
+					klog.V(4).Infof("No held updates found for pod %s", key)
+				}
+				continue
 			} else {
 				err = e.handlePod(op, content, rawUpdateChan)
 				if err != nil {
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -36,6 +36,7 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/protobuf/jsonpb"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/component-base/featuregate"
 	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
@@ -85,16 +86,19 @@ var DefaultRunLiteKubelet RunLiteKubelet = kubeletserver.Run
 
 // edged is the main edged implementation.
 type edged struct {
-	enable        bool
-	KubeletServer *kubeletoptions.KubeletServer
-	KubeletDeps   *kubelet.Dependencies
-	FeatureGate   featuregate.FeatureGate
-	context       context.Context
-	nodeName      string
-	namespace     string
+	enable         bool
+	KubeletServer  *kubeletoptions.KubeletServer
+	KubeletDeps    *kubelet.Dependencies
+	FeatureGate    featuregate.FeatureGate
+	context        context.Context
+	nodeName       string
+	namespace      string
+	heldPodUpdates map[string][]kubelettypes.PodUpdate
 }
 
 var _ core.Module = (*edged)(nil)
+
+const holdUpgradeLabel = "edge.kubeedge.io/hold-upgrade"
 
 // Register register edged
 func Register(e *v1alpha2.Edged) {
@@ -264,13 +268,14 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	kubeletDeps.PodConfig = config.NewPodConfig(config.PodConfigNotificationIncremental, kubeletDeps.Recorder, kubeletDeps.PodStartupLatencyTracker)
 
 	ed = &edged{
-		enable:        true,
-		context:       context.Background(),
-		KubeletServer: &kubeletServer,
-		KubeletDeps:   kubeletDeps,
-		FeatureGate:   utilfeature.DefaultFeatureGate,
-		nodeName:      nodeName,
-		namespace:     namespace,
+		enable:         true,
+		context:        context.Background(),
+		KubeletServer:  &kubeletServer,
+		KubeletDeps:    kubeletDeps,
+		FeatureGate:    utilfeature.DefaultFeatureGate,
+		nodeName:       nodeName,
+		namespace:      namespace,
+		heldPodUpdates: make(map[string][]kubelettypes.PodUpdate),
 	}
 
 	return ed, nil
@@ -313,12 +318,12 @@ func (e *edged) syncPod(podCfg *config.PodConfig) {
 			continue
 		}
 
-		_, resType, resID, err := commonmsg.ParseResourceEdge(result.GetResource(), result.GetOperation())
+		op := result.GetOperation()
+		_, resType, resID, err := commonmsg.ParseResourceEdge(result.GetResource(), op)
 		if err != nil {
 			klog.Errorf("failed to parse the Resource: %v", err)
 			continue
 		}
-		op := result.GetOperation()
 
 		content, err := result.GetContentData()
 		if err != nil {
@@ -327,6 +332,17 @@ func (e *edged) syncPod(podCfg *config.PodConfig) {
 		}
 
 		switch resType {
+		case model.UnholdUpgradeOperation:
+			key := string(content)
+			if updates, exists := e.heldPodUpdates[key]; exists {
+				klog.V(4).Infof("Unholding pod upgrade: %s", key)
+				for _, update := range updates {
+					rawUpdateChan <- update
+				}
+				delete(e.heldPodUpdates, key)
+			} else {
+				klog.V(4).Infof("No held updates found for pod %s", key)
+			}
 		case model.ResourceTypePod:
 			if op == model.ResponseOperation && resID == "" && result.GetSource() == modules.MetaManagerModuleName {
 				err := e.handlePodListFromMetaManager(content, rawUpdateChan)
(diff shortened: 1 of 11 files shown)


XmchxUp:
@WillardHu @Shelley-BaoYue @fujitatomoya @FengGaoCSC


## Problem 

### 1. 
To address the issue where users are unable to see which pods are currently held (related to the unhold pod name operation), we plan to introduce a list command that displays all pods currently held for upgrade.

At present, keadm has no direct way to fetch data from edged via beehive.

Is there an alternative approach—such as querying directly from SQLite?



### 2.
Currently, I believe there is a potential conflict between the default **rolling update strategy** in Kubernetes and our custom **hold mechanism** for Pod upgrades. However, whether this conflict actually occurs depends on specific scenarios.

#### Case 1: Two nodes — one cloud, one edge

Consider a `Deployment` with the following configuration:
```yaml
replicas: 2
strategy:
  type: RollingUpdate
  rollingUpdate:
    maxSurge: 25%
    maxUnavailable: 25%

```


During an upgrade, due to the `maxUnavailable: 25%` setting, Kubernetes will only delete **one old Pod at a time** and create one new Pod. If the Pod on the **edge node** is selected for upgrade first and is held (e.g., not marked Ready due to a hold condition), then the Pod on the **cloud node** will also be blocked from upgrading — because Kubernetes won't proceed to delete another Pod until the first one is Ready.


**Workaround**

By adjusting the strategy to allow more concurrency:

```yaml
replicas: 2[0001-feat-edge-add-hold-and-release-mechanism-for-resourc.patch](/uploads/97fb53bcf8c4331c7b3c1b2dc095ea40/0001-feat-edge-add-hold-and-release-mechanism-for-resourc.patch)
strategy:
  type: RollingUpdate
  rollingUpdate:
    maxSurge: 2
    maxUnavailable: 2

```

Kubernetes can delete and create **both Pods simultaneously**, so that the hold on the edge node won't block the upgrade on the cloud node.


#### Caveat:

It's important to note that the order in which Kubernetes selects Pods (and thus Nodes) for upgrade is **not predictable**. In hybrid environments (e.g., cloud + edge), the default rolling update strategy might lead to unexpected behavior if certain nodes hold upgrades. Therefore, it's essential for users to tune the `rollingUpdate` parameters appropriately to avoid upgrade bottlenecks and better align with custom upgrade control mechanisms like **hold**.

keadm/cmd/keadm/app/cmd/ctl/ctl.go
```diff
@@ -41,6 +42,7 @@ func NewCtl() *cobra.Command {
 	cmd.AddCommand(get.NewEdgeGet())
 	cmd.AddCommand(restart.NewEdgeRestart())
 	cmd.AddCommand(confirm.NewEdgeConfirm())
+	cmd.AddCommand(unhold.NewEdgeUnholdUpgrade())
```
WillardHu, line 45: Commands need to be designed with reusability and similarity in mind, otherwise the number of commands will continue to increase.
```bash
unhold-upgrade(confirm-upgrade) node <node-name>
unhold-upgrade(confirm-upgrade) pod <pod-name> --namespace(-n) <namespace>
```

edge/pkg/metamanager/metaserver/handlerfactory/extend_unhold_upgrade.go
```diff
@@ -0,0 +1,52 @@
...
+
+	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
+	"github.com/kubeedge/beehive/pkg/core/model"
+	"github.com/kubeedge/kubeedge/edge/pkg/common/modules"
+
+	"k8s.io/klog/v2"
+)
+
+func (f *Factory) UnholdUpgrade() http.Handler {
+	h := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
```
WillardHu, line 37: We can get the Pod from the metadata and then verify it, so that we can give users feedback faster, otherwise users can only look for logs.

staging/src/github.com/kubeedge/beehive/pkg/core/model/message.go
```diff
@@ -39,6 +39,8 @@ const (
 	ResourceTypeSaAccess            = "serviceaccountaccess"
 	ResourceTypeCSR                 = "certificatesigningrequest"
 
+	UnholdPodUpgrade = "unhold-pod-upgrade"
```
WillardHu: I think it can be defined as an operation, and "unhold-upgrade" is also an operation, not a resource type. Although we can treat it as a resource type to perform simple logical judgment, the readability is not very good.

edge/pkg/edged/edged.go
```diff
@@ -264,13 +268,14 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
...
-		nodeName:      nodeName,
-		namespace:     namespace,
+		enable:         true,
+		context:        context.Background(),
+		KubeletServer:  &kubeletServer,
+		KubeletDeps:    kubeletDeps,
+		FeatureGate:    utilfeature.DefaultFeatureGate,
+		nodeName:       nodeName,
+		namespace:      namespace,
+		heldPodUpdates: make(map[string][]kubelettypes.PodUpdate),
```
WillardHu, line 278: Why define the map value as a slice?

XmchxUp: Use the slice retains the order of multiple updates, ensuring that they are applied in sequence when unholding.

keadm/cmd/keadm/app/cmd/ctl/ctl.go
XmchxUp: The current implementation does not account for executing unhold-upgrade pod on the cloud (it only works on the edge) because it uses metaclient.KubeClient(). Should we support this? (Add a flag for edge/cloud.) If executed on the cloud, should it send a message to the metaserver?

WillardHu:
> ## Problem  问题
> ### 1.
> To address the issue where users are unable to see which pods are currently held (related to the unhold pod name operation), we plan to introduce a list command that displays all pods currently held for upgrade.为了解决用户无法看到当前持有的 Pod 的问题（与取消保留 Pod 名称作相关），我们计划引入一个 list 命令，显示当前等待升级的所有 Pod。
> ...
> ### 2.
> Currently, I believe there is a potential conflict between the default **rolling update strategy** in Kubernetes and our custom **hold mechanism** for Pod upgrades. However, whether this conflict actually occurs depends on specific scenarios.目前，我认为 Kubernetes 中的默认**滚动更新策略**与我们用于 Pod 升级的自定义**保留机制**之间存在潜在冲突。然而，这种冲突是否真的发生取决于具体场景。
> ...

Hi @XmchxUp 

Sorry for the late response! I've been tied up with work recently.

Regarding your question:
1. As we discussed before, we can try to add a custom message to the Pod status to indicate that it is waiting for unhold update. The information will be stored in the metadata and can also be synchronized to the cloud. The keadm also has pod read commands.
2. Usually, edge applications are singletons. Even if high availability is required across multiple edge nodes, edge applications should not be scheduled to cloud nodes, right? Users should use node affinity to ensure that edge applications are scheduled to edge nodes.

keadm/cmd/keadm/app/cmd/ctl/ctl.go
WillardHu: These are the commands executed by `keadm ctl` on the edge. We can refer to the command format of kubectl: `kubectl <operation(get/list/create/update/..)> <resource type> [name] [flags...]`. For OTA operations on multiple resources, nodes and pods can be placed under one subcommand.

WillardHu:
You could add some unit tests for your new code

XmchxUp:
@WillardHu 
For testing, I am currently thinking of adding new test logic using Ginkgo under /tests/e2e/apps/deployment.go, but from the GitHub Actions, it seems that the environment for e2e tests does not depend on keadm. Is it reasonable to add a new dependency on keadm? (I haven't debugged it yet.) Or should I add a separate overall test logic like the previous Cilium e2e tests?

WillardHu:
> For testing, I am currently thinking of adding new test logic using Ginkgo under /tests/e2e/apps/deployment.go, but from the GitHub Actions, it seems that the environment for e2e tests does not depend on keadm. Is it reasonable to add a new dependency on keadm? (I haven't debugged it yet.) Or should I add a separate overall test logic like the previous Cilium e2e tests?

You could consider calling the metadata server API

edge/pkg/metamanager/metaserver/handlerfactory/extend_unhold_upgrade.go
```diff
@@ -0,0 +1,88 @@
...
+			http.Error(w, "pod is not marked with hold-upgrade annotation", http.StatusBadRequest)
+			return
+		}
+
+		if pod.Status.Phase != v1.PodPending {
+			http.Error(w, "pod is not in pending phase", http.StatusBadRequest)
+			return
+		}
+
+		resource := fmt.Sprintf("namespace/%s/id", model.UnholdUpgradeOperation)
```
Shelley-BaoYue: This resource definition may not align well with the common usage in beehive. Typically, we define resources as `namespace/{namespace}/resourcetype/{pod-name}`, and place the verb as the operation. For instance, here we set the operation to `model.UnholdUpgradeOperation`, and separately add a handler for this operation type in the meatamanager.

XmchxUp: From the code perspective, it looks like this format? {namespace}/{resource_type}/{resource_id}

```go
// ParseResourceEdge parses resource at edge and returns namespace, resource_type, resource_id.
// If operation of msg is query list, return namespace, pod.
func ParseResourceEdge(resource string, operation string) (string, string, string, error) {
	resourceSplits := strings.Split(resource, "/")
	if len(resourceSplits) == 3 {
		return resourceSplits[0], resourceSplits[1], resourceSplits[2], nil
	} else if operation == model.QueryOperation || operation == model.ResponseOperation && len(resourceSplits) == 2 {
		return resourceSplits[0], resourceSplits[1], "", nil
	}
	return "", "", "", fmt.Errorf("resource: %s format incorrect, or Operation: %s is not query/response", resource, operation)
}
```

Shelley-BaoYue: Yes! : )  I think this will also make it more convenient for us to expand to do hold/unhold operation for other resources.

keadm/cmd/keadm/app/cmd/ctl/ctl.go
fujitatomoya: > Should we support this? (Add a flag for edge/cloud.)

i do not see any use cases for this. how it would be useful for this even if that is supported?

fujitatomoya (commented):
the fix looks almost promising, @XmchxUp thanks for working on this, really appreciate that.

@WillardHu @Shelley-BaoYue thanks for supporting this as well.

edge/pkg/metamanager/metaserver/handlerfactory/extend_unhold_upgrade.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		pod, err := clientset.CoreV1().Pods(namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			http.Error(w, fmt.Sprintf("pod not found: %v", err), http.StatusNotFound)
+			return
+		}
+
+		// validate pod annotation and status
+		if pod.Annotations["edge.kubeedge.io/hold-upgrade"] != "true" {
```
WillardHu, line 70: Go lint failed. There is a `trueStr` constant in edge/pkg/metamanager/metaserver/handlerfactory/extend_exec_pod.go that we can use. You might consider moving it to edge/pkg/metamanager/metaserver/common/types.go or create a new file consts.go in the handlerfactory package. 
```text
Error: edge/pkg/metamanager/metaserver/handlerfactory/extend_logs_pod.go:72:25: string `true` has 3 occurrences, but such constant `trueStr` already exists (goconst)
```

tangming1996:
Hello, here is a scenario: If a Deployment resource is deployed in the cloud, and its corresponding Pod has been allowed to run on the edge side, but for some reasons, the Deployment resource in the cloud restarts. At this point, the Pod ID will be updated accordingly. Will the new Pod remain in the pending state? And is this situation as expected?

XmchxUp:
> .... Will the new Pod remain in the pending state? And is this situation as expected?

It will remain in the pending state, which is expected, because the newly created Pod from the deployment still has hold-upgrade.

fujitatomoya:
@WillardHu @Shelley-BaoYue really appreciate your support.

@XmchxUp thank you very much for being patient and great work 👍 🚀

Issue #6409, opened 2025-07-14
Standardize and Enhance Logging Across Edge Components


PranjaliBhardwaj:
Description:
Currently, logging practices vary across different edge components such as edge/pkg/edged and edge/pkg/edgehub. Some logs are missing critical context, others use inconsistent formats or levels, and a few messages are vague or unclear. This inconsistency makes it harder for developers, operators, and contributors to debug issues or understand system behavior in real-time.

Why:
Make debugging and troubleshooting more efficient.
Provide clearer knowledge into system behavior for maintainers and users. 
Improve observability for operators in production environments. 
Lay the groundwork for better tooling and log-based automation in the future.

PR #6432, opened 2025-08-03
/feat Enhance and standardize logging in edge components

PranjaliBhardwaj:
What type of PR is this?

More consistent, informative, and structured logs.
Easier debugging and monitoring.
No risk to runtime behavior or performance.
issue: #6409 
DCO signed off

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -214,7 +214,7 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	var kubeletFlags kubeletoptions.KubeletFlags
 	err = edgedconfig.ConvertEdgedKubeletConfigurationToConfigKubeletConfiguration(edgedconfig.Config.TailoredKubeletConfig, &kubeletConfig, nil)
 	if err != nil {
-		klog.ErrorS(err, "Failed to convert kubelet config")
+		klog.ErrorS(err, "Failed to convert kubelet config in newEdged")
 		return nil, fmt.Errorf("failed to construct kubelet configuration")
 	}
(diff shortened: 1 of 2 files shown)


edge/pkg/edged/edged.go
```diff
@@ -214,7 +214,7 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	var kubeletFlags kubeletoptions.KubeletFlags
 	err = edgedconfig.ConvertEdgedKubeletConfigurationToConfigKubeletConfiguration(edgedconfig.Config.TailoredKubeletConfig, &kubeletConfig, nil)
 	if err != nil {
-		klog.ErrorS(err, "Failed to convert kubelet config")
+		klog.ErrorS(err, "Failed to convert kubelet config in newEdged")
 		return nil, fmt.Errorf("failed to construct kubelet configuration")
```
fujitatomoya, line 218: i am not sure if this is the right thing to do.
if the caller function properly handles the error (for example, by logging or reporting it), then logging it here can indeed be redundant and may result in double logging of the same error. so maybe we should remove `klog.ErrorS` instead?

PranjaliBhardwaj: @fujitatomoya  Thank you for the review. Let me explain my initial thought process:

I added the additional logging to improve observability and make debugging easier by providing more context about where exactly the error occurred in the edged component. However, I see now that this approach leads to redundant error logging and violates the "don't log and return" principle.

I agree with your assessment ,  we should let the caller handle the error logging. Here's my proposed change:

````go
// filepath: edge/pkg/edged/edged.go
err = edgedconfig.ConvertEdgedKubeletConfigurationToConfigKubeletConfiguration(edgedconfig.Config.TailoredKubeletConfig, &kubeletConfig, nil)
if err != nil {
    return nil, fmt.Errorf("failed to construct kubelet configuration: %w", err)
}

PR #6438, opened 2025-08-09
Tls cloudhub

PranjaliBhardwaj:
Changes made:
cloud/pkg/cloudhub/servers/server.go:
Dropped explicit CipherSuites list; kept MinVersion: tls.VersionTLS12 and let defaults choose modern AEAD suites.
Switched panic calls when building pool and keypair to klog.Exit(...).
cloud/pkg/cloudhub/servers/server_test.go:
Added two tests: TestCreateTLSConfig_BadCA and TestCreateTLSConfig_MismatchedKey.
Hooked klog.OsExit to panic in tests to assert exit behavior.

Updated cloud/pkg/cloudhub/servers/server.go to remove the CBC cipher and rely on Go’s modern defaults for TLS 1.2+ (AEAD suites like AES-GCM/ChaCha20-Poly1305). Also replaced panic paths in createTLSConfig with klog.Exit(...) for structured fail-fast.
Added tests in cloud/pkg/cloudhub/servers/server_test.go to verify createTLSConfig exits when given a bad CA or mismatched cert/key. Implemented local cert/key generation with crypto/x509 to avoid extra deps.
Ran the tests locally (with CGO disabled) and they passed.

### DIFF
--- a/cloud/pkg/cloudhub/servers/server.go
+++ b/cloud/pkg/cloudhub/servers/server.go
@@ -30,30 +30,26 @@ func StartCloudHub(messageHandler handler.Handler) {
 func createTLSConfig(ca, cert, key []byte) tls.Config {
 	// init certificate
 	pool := x509.NewCertPool()
-	ok := pool.AppendCertsFromPEM(pem.EncodeToMemory(&pem.Block{Type: certutil.CertificateBlockType, Bytes: ca}))
-	if !ok {
-		panic(fmt.Errorf("fail to load ca content"))
-	}
+    ok := pool.AppendCertsFromPEM(pem.EncodeToMemory(&pem.Block{Type: certutil.CertificateBlockType, Bytes: ca}))
+    if !ok {
+        klog.Exit(fmt.Errorf("fail to load ca content"))
+    }
 
-	certificate, err := tls.X509KeyPair(pem.EncodeToMemory(&pem.Block{Type: certutil.CertificateBlockType, Bytes: cert}), pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: key}))
-	if err != nil {
-		panic(err)
-	}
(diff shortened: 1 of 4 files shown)


cloud/pkg/cloudhub/servers/server.go
```diff
@@ -30,30 +30,26 @@ func StartCloudHub(messageHandler handler.Handler) {
 func createTLSConfig(ca, cert, key []byte) tls.Config {
 	// init certificate
 	pool := x509.NewCertPool()
-	ok := pool.AppendCertsFromPEM(pem.EncodeToMemory(&pem.Block{Type: certutil.CertificateBlockType, Bytes: ca}))
-	if !ok {
-		panic(fmt.Errorf("fail to load ca content"))
-	}
+    ok := pool.AppendCertsFromPEM(pem.EncodeToMemory(&pem.Block{Type: certutil.CertificateBlockType, Bytes: ca}))
+    if !ok {
+        klog.Exit(fmt.Errorf("fail to load ca content"))
```
WillardHu, line 35: We can use klog.Exit() to pass errors to top caller, but we need to use klog.OnExit at cmd runner to catch the error.

Exiting directly will destroy the graceful shutdown process.

edge/pkg/edged/edged.go
```diff
@@ -214,7 +214,7 @@ func newEdged(enable bool, nodeName, namespace string) (*edged, error) {
 	var kubeletFlags kubeletoptions.KubeletFlags
 	err = edgedconfig.ConvertEdgedKubeletConfigurationToConfigKubeletConfiguration(edgedconfig.Config.TailoredKubeletConfig, &kubeletConfig, nil)
 	if err != nil {
-		klog.ErrorS(err, "Failed to convert kubelet config")
+		klog.ErrorS(err, "Failed to convert kubelet config in newEdged")
```
WillardHu, line 217: If we plan to gradually replace the log with a structured one, we can define a named logger for edged (`klog.FromContext(ctx).WithName(..)`).

PR #6441, opened 2025-08-15
Use keadm to Manage Windows Nodes

### LINKED ISSUE #6314: [OSPP] Use keadm to manage Windows nodes

- Provide a new comprehensive proposal to explain how keadm can use Windows services to anage EdgeCore.
- Keadm can execute join, upgrade/backup/rollback and reset commands normally in Windows nodes.

**Why is this needed**:

Since kubelet and container runtimes already run on Windows, so EdgeCore can also run on Windows, but it needs to be installed manually. This is because keadm still has some bugs and design flaws in Windows that prevent it from working properly. 

Currently, we use the nssm tool to manage Windows services. Refer to: https://github.com/kubeedge/kubeedge/blob/7b2daa1247deb218a951e26d362b9b0e4caa6631/keadm/cmd/keadm/app/cmd/edge/join_windows.go#L198
But kubeadm uses the [golang.org/x/sys/windows/svc/mgr](golang.org/x/sys/windows/svc/mgr) package. We can investigate  this package. Refer to: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/initsystem/initsystem_windows.go

Believeht029:
**What type of PR is this?**
/kind feature

Since kubelet and container runtimes already run on Windows, so EdgeCore can also run on Windows, but it needs to be installed manually. This is because keadm still has some bugs and design flaws in Windows that prevent it from working properly.

Fixes #6314 

```release-note
NONE
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_windows.go
+++ b/keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_windows.go
@@ -1,7 +1,7 @@
 //go:build windows
 
 /*
-Copyright 2024 The KubeEdge Authors.
+Copyright 2025 The KubeEdge Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


Believeht029:
@gemini-code-assist review

Believeht029:
@gemini-code-assist review

Believeht029:
@WillardHu @Shelley-BaoYue PTAL

keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_windows.go
```diff
@@ -1,7 +1,7 @@
 //go:build windows
 
 /*
-Copyright 2024 The KubeEdge Authors.
+Copyright 2025 The KubeEdge Authors.
```
Shelley-BaoYue, line 4: No need to modify the copyright time for exsit files.

keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_windows.go
```diff
@@ -18,13 +18,154 @@ limitations under the License.
...
+	"fmt"
+	"os/exec"
+	"time"
 
-// GetExtSystem returns an ExtSystem for the current system, or nil
-// if we cannot detect a supported init system.
-// This indicates we will skip init system checks, not an error.
+	"github.com/pkg/errors"
+	"golang.org/x/sys/windows/svc"
+
```
Shelley-BaoYue, line 28: remove this blank line

WillardHu:
Windows services started by `golang.org/x/sys/windows` require edgecore to run based on `golang.org/x/sys/windows/svc#Run()` function, and it has a `handler` argement that needs to be implemented. 
```golang
// Handler is the interface that must be implemented to build Windows service.
type Handler interface {
	// Execute will be called by the package code at the start of
	// the service, and the service will exit once Execute completes.
	// Inside Execute you must read service change requests from r and
	// act accordingly. You must keep service control manager up to date
	// about state of your service by writing into s as required.
	// args contains service name followed by argument strings passed
	// to the service.
	// You can provide service exit code in exitCode return parameter,
	// with 0 being "no error". You can also indicate if exit code,
	// if any, is service specific or not by using svcSpecificEC
	// parameter.
	Execute(args []string, r <-chan ChangeRequest, s chan<- Status) (svcSpecificEC bool, exitCode uint32)
}
```

Otherwise, the service process cannot notify the Windows service of its running status and will generate an error: A slow service does not start due to time-out error.

I'll merge this PR first. Next, we need to solve this issue, we can refer to: 
- https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/init_windows.go
- https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server_windows.go

PR #6458, opened 2025-09-18
[OSPP] Make the device model more meaningful

### LINKED ISSUE #6316: [OSPP]Make the device model more meaningful

- Provide a new comprehensive proposal to explain how to design the device model spec, definition of DMI and adjust the workflow of devicetwin.
- Completed the coding work, no obvious bugs.

**Why is this needed**:

In traditional IOT, the device domain is designed as: model, product, and instance. 

Image

The current KubeEdge device model is relatively simple, and the actual effect is not significant, and its design will cause confusion to users when using it.  Refer to: [Device Management v1beta1 Version](https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/device-crd-v1beta1.md)

However, it would be very constly to split two structures into three structures, and it would be meaningless to do so. Therefore, we can define the model as the concept of real device products(i.e., model and product), which is used to describe the information, connection protocolsute acquisition methods, etc.

In this way, device instances can share the definition of device models, and make the positioning of device models clearer.

cukawang:
**What type of PR is this?**
/kind feature

Currently, KubeEdge's definition of the device model is relatively simplistic, with limited practical effectiveness, and its design is prone to confusing users. In traditional IoT, devices are typically designed with a three-tier structure: thing model, product, and device instance. Due to historical reasons, the cost of splitting the model into three independent objects is high, and the significance of fine-grained abstraction is limited. Therefore, we propose defining the model as the concept of "real device product" (i.e., a combination of the thing model and the product), which is used to describe the specifications, connection protocols, attribute acquisition methods, etc., of a type of device product. In this way, device instances can share these configurations, and only need to be configured with different connection addresses for different devices. This design can realize a certain degree of reuse of configuration information and make the positioning more clear.

Fixes #6316

```release-note

```

### DIFF
--- a/build/crds/devices/devices_v1beta1_devicemodel.yaml
+++ b/build/crds/devices/devices_v1beta1_devicemodel.yaml
@@ -247,6 +247,7 @@ spec:
                 description: 'Optional: Any config data.It can be overridden in the
                   device instance.'
                 type: object
+                x-kubernetes-preserve-unknown-fields: true
             type: object
         type: object
     served: true
--- a/manifests/charts/cloudcore/crds/devices_v1beta1_devicemodel.yaml
+++ b/manifests/charts/cloudcore/crds/devices_v1beta1_devicemodel.yaml
@@ -247,6 +247,7 @@ spec:
                 description: 'Optional: Any config data.It can be overridden in the
                   device instance.'
                 type: object
+                x-kubernetes-preserve-unknown-fields: true
             type: object
         type: object
     served: true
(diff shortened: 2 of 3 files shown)


edge/pkg/devicetwin/dtmanager/dmiworker_test.go
```diff
@@ -0,0 +1,654 @@
+/*
+Copyright 2022 The KubeEdge Authors.
```
Shelley-BaoYue:
```suggestion
Copyright 2025 The KubeEdge Authors.
```

cukawang:
> Because you have modified the device API, you need to run `hack/update-codegen.sh` to autogenerate API related code. Then you can execute `make verify` to make sure everything update is ready. : )

resolved it

Shelley-BaoYue (commented):
/approve  
PTAL @JiaweiGithub @WillardHu

PR #6463, opened 2025-09-25
fix: correct cloudcore.config.kubeedge.io to v1alpha1

Abirdcfly:
Note: changelog should not be modified after release, so its wrong reference remains unchanged.

**What type of PR is this?**

/kind bug

The apiVersion for cloudcore.config.kubeedge.io was incorrectly changed to v1alpha2 in a previous commit, but there is no v1alpha2 implementation in code.
Revert it back to v1alpha1.

Fixes #

```release-note

```

### DIFF
--- a/build/cloud/05-configmap.yaml
+++ b/build/cloud/05-configmap.yaml
@@ -8,7 +8,7 @@ metadata:
     kubeedge: cloudcore
 data:
   cloudcore.yaml: |
-    apiVersion: cloudcore.config.kubeedge.io/v1alpha2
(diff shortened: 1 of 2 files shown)


Shelley-BaoYue (commented):
Indeed the version here is not consistent with the actual cloudcore configuration version.  However, it seems this mistakes has existed since CloudCore support containerized deployment.  I'm concern that if we revert back to v1alpha1, might cause compatibility issue or cause confusion. We can consider upgrading cloudcore version to a new unified version in the future.

fujitatomoya (approved):

PR #6474, opened 2025-09-29
chore: remove debug log for db init

Abirdcfly:
**What type of PR is this?**
/kind cleanup

` I0929 18:00:41.499106  209103 db.go:38] !!!!!!!!!!in db.go, DBAccess = &{{0xc000965200 0xc000d6f740}}` I guess this is debug log, we should remove it.

Fixes #

```release-note
NONE
```

### DIFF
--- a/edge/pkg/common/dbm/db.go
+++ b/edge/pkg/common/dbm/db.go
@@ -35,7 +35,6 @@ func InitDBConfig(driverName, dbName, dataSource string) {
 		}()
 		// create orm
 		DBAccess = orm.NewOrmUsingDB(dbName)
-		klog.Infof("!!!!!!!!!!in db.go, DBAccess = %v", DBAccess)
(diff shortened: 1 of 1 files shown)


edge/pkg/common/dbm/db.go
```diff
@@ -35,7 +35,6 @@ func InitDBConfig(driverName, dbName, dataSource string) {
 		}()
 		// create orm
 		DBAccess = orm.NewOrmUsingDB(dbName)
-		klog.Infof("!!!!!!!!!!in db.go, DBAccess = %v", DBAccess)
```
fujitatomoya, line 38: I do not quite understand this info message, how it could be useful to see Ormer object... i would also say we can just go ahead to remove this. or if necessary, perhaps we can change this into `debug` log at least.

PR #6478, opened 2025-10-09
LFX Mentorship: Device Anomaly Detection Framework proposal

### LINKED ISSUE #6312: Device Anomaly Detection Framework Based on KubeEdge
This project aims to design a device anomaly detection framework for KubeEdge, which uncovers causal relationships between device states to enable lightweight and efficient anomaly detection. 
This issue includes:
- A general-purpose device anomaly detection framework that supports user-defined detection algorithms.
- A complete technical design document including model selection, training procedures, and detailed architecture diagrams for both training and online detection components.
- A machine learning model and corresponding anomaly detection algorithm capable of capturing causal relationships among device states, trained and tested using standard frameworks.
- An online anomaly detection module integrated into the KubeEdge device state reporting workflow, enabling real-time analysis through a model inference hook.

**Why is this needed**:
The current KubeEdge platform represents device states using three status indicators: Desired, ObservedDesired, and Reported. The display of the device state on the platform entirely depends on data collected and reported by the Mapper from the device side. However, due to limitations in Mapper implementation, physical device failures, network latency, and potential cyberattacks, the platform may present device states that do not accurately reflect the actual condition of the devices. In scenarios where applications rely on these device states for decision-making, such inconsistency could lead to adverse consequences.

HaojieZhang6848:
**What type of PR is this?**

/kind documentation

Fixes #6312

```release-note

```


docs/proposals/sig-device-iot/6312-device-anomaly-detection-framework/architecture.mermaid
```diff
@@ -0,0 +1,56 @@
+flowchart LR
```
Shelley-BaoYue: Please move this images to `docs/proposals/sig-device-iot/device-anomaly-detection-framework/` path, same with your proposal README.md

HaojieZhang6848: Done. Thank you.

docs/proposals/sig-device-iot/6312-device-anomaly-detection-framework/README.md
```diff
@@ -0,0 +1,344 @@
...
+- **Algorithm plug-in**: Supports interfaces for rules, statistics, machine learning, deep learning, and other algorithms.  
+- **Model hot loading**: Supports switching models without interrupting services.  
+- **Horizontal scaling**: Can scale according to load, supporting load balancing or device sharding.
+
+### 4.5 Result Writing and Alarm Module
+
+**Location**: Integrated within **DeviceController**.
+
+**Function**:  
+- Updates detection results to the corresponding `status.anomaly` field of the **Device CRD**, including the most recent detection state, score, type, and timestamp.
```
WillardHu: Perhaps `status.extensions`(map[string]any) is more appropriate, and it would be suitable for more scenes.

WillardHu:
The scenario is very great, but not all capabilities need to be coded in KubeEdge.

image

The throughput of the hub tunnel is currently bottlenecked. It will be easier to do these as shown above. This also requires some modifications to KubeEdge:
1. Add a field `status.extensions ` to Device CRD.
1. Consider split the Device CRD to Device and DeviceStatus. It seems that the device status will be updated frequently in the future. This will cause the `resourceVersion` to be incremented continuously, which will cause the device spec update to fail.
1. Optional, CloudCore provides an API for serially updating device status (for the same reason as pont 2).
1. Optional, if necessary, extend other implementations of Mapper's pushmethod.

HaojieZhang6848:
> The scenario is very great, but not all capabilities need to be coded in KubeEdge.
> 
> image
> The throughput of the hub tunnel is currently bottlenecked. It will be easier to do these as shown above. This also requires some modifications to KubeEdge:
> 
> 1. Add a field `status.extensions ` to Device CRD.
> 2. Consider split the Device CRD to Device and DeviceStatus. It seems that the device status will be updated frequently in the future. This will cause the `resourceVersion` to be incremented continuously, which will cause the device spec update to fail.
> 3. Optional, CloudCore provides an API for serially updating device status (for the same reason as pont 2).
> 4. Optional, if necessary, extend other implementations of Mapper's pushmethod.



> The scenario is very great, but not all capabilities need to be coded in KubeEdge.
> 
> image
> The throughput of the hub tunnel is currently bottlenecked. It will be easier to do these as shown above. This also requires some modifications to KubeEdge:
> 
> 1. Add a field `status.extensions ` to Device CRD.
> 2. Consider split the Device CRD to Device and DeviceStatus. It seems that the device status will be updated frequently in the future. This will cause the `resourceVersion` to be incremented continuously, which will cause the device spec update to fail.
> 3. Optional, CloudCore provides an API for serially updating device status (for the same reason as pont 2).
> 4. Optional, if necessary, extend other implementations of Mapper's pushmethod.

Thank you. I will revise the proposal based on your suggestions.

Issue #6465, opened 2025-09-26
Enhance MetaServer to support content-type application/vnd.kubernetes.protobuf


brinker-tbaker:
**What would you like to be added/modified**:
I am requesting that the metaserver be modified to also support application/vnd.kubernetes.protobuf instead of just application/yaml and application/json.
**Why is this needed**:
When metaserver is enabled and an operator using k8s.io/client-go attempts to UPDATE a configmap (there may be other activities, but this is what was tested), client-go sends a Content-Type header of `application/vnd.kubernetes.protobuf` with an accept of `application/vnd.kubernetes.protobuf,application/json`. 
Since metaserver does not accept application/vnd.kubernetes.protobuf the call fails.

This is needed to allow github.com/rook/rook to be able to run on the edge.

brinker-tbaker:
this is also affecting cilium as a L2 load balancer as it is not able to generate a lease.

```
time=2025-10-06T23:35:47.363089018Z level=error msg="error initially creating leader election record: the body of the request was in an unknown format - accepted media types include: application/json, application/yaml" subsys=klog
```

danielrutledgebrinker:
@Shelley-BaoYue This is required for kubernetes 1.32 support with this change https://github.com/kubernetes/client-go/commit/6f44458e5efa4a1d73063e438616bac9f980b67b

PR #6480, opened 2025-10-10
[wip]Add MetaServer to support content-type application/vnd.kubernetes.protobuf

luomengY:
**What type of PR is this?**
/kind feature

Fixes # https://github.com/kubeedge/kubeedge/issues/6465

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/metaserver/kubernetes/serializer/serializer.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/serializer/serializer.go
@@ -7,6 +7,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/apimachinery/pkg/runtime/serializer/protobuf"
 )
 
 // NewUnstructuredNegotiatedSerializer returns a simple, negotiated serializer
@@ -44,6 +45,16 @@ func (f WithoutConversionCodecFactory) SupportedMediaTypes() []runtime.Serialize
 			MediaTypeSubType: "yaml",
 			EncodesAsText:    true,
 			Serializer:       json.NewSerializerWithOptions(json.DefaultMetaFactory, f.creator, f.typer, json.SerializerOptions{Yaml: true}),
+		},
+		{
+			MediaType:        "application/vnd.kubernetes.protobuf",
+			MediaTypeType:    "application",
+			MediaTypeSubType: "vnd.kubernetes.protobuf",
+			Serializer:       protobuf.NewSerializer(f.creator, f.typer),
+			StreamSerializer: &runtime.StreamSerializerInfo{
+				Serializer: protobuf.NewRawSerializer(f.creator, f.typer),
+				Framer:     protobuf.LengthDelimitedFramer,
+			},
 		},
 	}
 }


brinker-tbaker:
hey @luomengY, this wont work as-is due to how storage.go forces items to be a unstructured.Unstructured type. protobuf cant resolve unstructured objects. I've already tried the above patch and ran into the unstructured issue.

luomengY:
> hey @luomengY, this wont work as-is due to how storage.go forces items to be a unstructured.Unstructured type. protobuf cant resolve unstructured objects. I've already tried the above patch and ran into the unstructured issue.

ok， I'll take a look at the deep logic inside and then refine it.

luomengY:
Can you tell me how to test and reproduce it? I will further optimize it. @brinker-tbaker

brinker-tbaker:
@luomengY 

here is a sample app that will attempt to perform various resource actions

```
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"os/signal"
	"syscall"
	"time"

	coordinationv1 "k8s.io/api/coordination/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/client-go/kubernetes"
	"k8s.io/client-go/rest"
)

func main() {
	fmt.Print("starting up")
	sigChan := make(chan os.Signal, 1)
	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
	defer cancel()

	config, err := rest.InClusterConfig()
	if err != nil {
		log.Fatalf("Error getting in-cluster config: %v", err)
	}

	clientset, err := kubernetes.NewForConfig(config)
	if err != nil {
		log.Fatalf("Error creating clientset: %v", err)
	}

	namespace := "test"
	leaseName := "test-lease"

	// Define the Lease object
	lease := &coordinationv1.Lease{
		ObjectMeta: metav1.ObjectMeta{
			Name:      leaseName,
			Namespace: namespace,
		},
		Spec: coordinationv1.LeaseSpec{
			HolderIdentity:       func(s string) *string { return &s }("test-instance-1"),
			LeaseDurationSeconds: func(i int32) *int32 { return &i }(15), // 15 seconds
			AcquireTime:          &metav1.MicroTime{Time: time.Now()},
			RenewTime:            &metav1.MicroTime{Time: time.Now()},
			LeaseTransitions:     func(i int32) *int32 { return &i }(0),
		},
	}

	//create
	createdLease, err := clientset.CoordinationV1().Leases(namespace).Create(ctx, lease, metav1.CreateOptions{})
	if err != nil {
		fmt.Printf("Error creating Lease: %v\n", err)
		return
	}

	fmt.Printf("Lease '%s' created successfully in namespace '%s'. Holder: %s\n",
		createdLease.Name, createdLease.Namespace, *createdLease.Spec.HolderIdentity)

	//get
	getNamespace, err := clientset.CoreV1().Namespaces().Get(ctx, namespace, metav1.GetOptions{})
	if err != nil {
		fmt.Printf("Error getting namespace: %v\n", err)
		return
	}

	fmt.Printf("Namespace '%s' retrieved successfully.\n", getNamespace.Name)

	//list
	listPods, err := clientset.CoreV1().Pods(namespace).List(ctx, metav1.ListOptions{})
	if err != nil {
		fmt.Printf("Error listing pods: %v\n", err)
		return
	}

	fmt.Printf("PodList with length of '%v' retrieved successfully.\n", len(listPods.Items))

	//update
	getNamespace.Labels["updateTest"] = "test"
	updateNamespace, err := clientset.CoreV1().Namespaces().Update(ctx, getNamespace, metav1.UpdateOptions{})

	if err != nil {
		fmt.Printf("Error updating namespace: %v\n", err)
		return
	}

	fmt.Printf("Namespace '%s' updated successfully with labels '%v'.\n", updateNamespace.Name, updateNamespace.Labels)

	//wait for signal to quit
	<-sigChan
	os.Exit(0)
}
// this sample currently does not check activities such as watch or delete

```

the go.mod will need the following requires

```
require (
	k8s.io/api v0.34.1
	k8s.io/apimachinery v0.34.1
	k8s.io/client-go v0.34.1
)
```

For where things need to be changed in storage.go, an easy example is the 'Get' method. if you look at [storage.go#L114](https://github.com/kubeedge/kubeedge/blob/6dff194f9942b4487f770de33d695ca5a47e9b10/edge/pkg/metamanager/metaserver/kubernetes/storage/storage.go#L114) you can see where the response is unmarshaled into an unstructured.Unstructured object. This is fine for storage storing into sqllite, but breaks when the runtime.Object is handed back up to the serializer.go for passing to the client as unstructured.Unstructured cant be marshaled as protobuf. the Unstructured object is for when you dont know the concrete type.

I feel like we have 3 options here, with the first being the 'better' fix, while the second may be easier as a temporary fix, and the 3rd being a hybrid approach

1) in storage.go, after performing the imitator activity of storing, convert to the real concrete type that the object is, so that it can be marshaled as protobuf

2) somehow make metaserver accept protobuf as a content-type coming in, but always return json back to the client. (making sure the content-type returned gets set as application/json)

3) fix number 1, but if we are unable to determine the concrete type, fall back to returning json as unstructured.Unstructured.

As to how to identify the concrete type the object is assigned to, I believe the raw json (app.RespBody) needs to be set as a runtime.RawExtension object, and then that gets passed thru [Convert_runtime_RawExtension_To_runtime_Object](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime@v0.34.1#Convert_runtime_RawExtension_To_runtime_Object) to get to the real type. I wasn't able to get this part working though.

brinker-tbaker:
@luomengY i was able to get it working. I will submit a PR shortly.

PR #6482, opened 2025-10-13
Fix: make environmentCheck process check case-insensitive and configurable

JianChengBlueDot:
## Motivation
The original environmentCheck only checks for "kubelet" and matches process names case-sensitively.
- This may fail if the process name has different casing.
- Some users may want to check additional processes.

## Changes
- Use case-insensitive comparison (strings.EqualFold)
- Support custom process list via environment variable EDGECORE_CHECK_PROCESSES
- Default behavior unchanged (checks kubelet)

## Benefits
- More robust environment validation
- Customizable for different edge node setups

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -1,7 +1,6 @@
 package app
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


edge/cmd/edgecore/app/server.go
```diff
@@ -1,7 +1,7 @@
 package app
 
 import (
-	"errors"
+	//"errors"
```
WillardHu: Unused packages can be deleted.

edge/cmd/edgecore/app/server.go
```diff
@@ -176,30 +176,41 @@ func cleanupToken(config v1alpha2.EdgeCoreConfig, file string) error {
...
-			continue
-		}
-		switch processName {
-		case "kubelet": // if kubelet is running, return error
-			return errors.New("kubelet should not be running on edge node when starting edgecore")
-		}
-	}
-
-	return nil
+    if skipCheck {
```
WillardHu, line 179: Use tabs instead of spaces for indentation.

edge/cmd/edgecore/app/server.go
```diff
@@ -176,30 +176,41 @@ func cleanupToken(config v1alpha2.EdgeCoreConfig, file string) error {
...
+        return nil
+    }
+
+    // 支持自定义进程列表，通过环境变量配置
+    checkProcesses := []string{"kubelet"}
+    if env := os.Getenv("EDGECORE_CHECK_PROCESSES"); env != "" {
+        checkProcesses = strings.Split(env, ",")
```
WillardHu: Checking the kubelet process is necessary, so it is more appropriate to append custom processes. 
```golang
checkProcess = append(checkProcess, strings.Split(env, ",")...)
```

PR #6483, opened 2025-10-13
Fix: unify exit style when getting custom interface IP

JianChengBlueDot:
## Motivation
In server.go, when fetching IP from a custom interface, the code calls os.Exit(1) directly. 
Other errors in edgecore use klog.Exit, so this is inconsistent.

## Changes
- Replace os.Exit(1) with klog.Exitf
- Log detailed error while exiting
- Keep behavior unchanged, unify error handling style

## Benefits
- Consistent error handling across edgecore
- Easier to debug due to unified logs

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -122,8 +122,7 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 			if config.Modules.Edged.CustomInterfaceName != "" {
 				ip, err := netutil.ChooseBindAddressForInterface(config.Modules.Edged.CustomInterfaceName)
 				if err != nil {
-					klog.Errorf("Failed to get IP address by custom interface %s, err: %v", config.Modules.Edged.CustomInterfaceName, err)
-					os.Exit(1)
+					klog.Exitf("Failed to get IP address by custom interface %s, err: %v", config.Modules.Edged.CustomInterfaceName, err)
 				}
 				config.Modules.Edged.NodeIP = ip.String()
 				klog.Infof("Get IP address by custom interface successfully, %s: %s", config.Modules.Edged.CustomInterfaceName, config.Modules.Edged.NodeIP)


edge/cmd/edgecore/app/server.go
```diff
@@ -122,8 +122,7 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 			if config.Modules.Edged.CustomInterfaceName != "" {
 				ip, err := netutil.ChooseBindAddressForInterface(config.Modules.Edged.CustomInterfaceName)
 				if err != nil {
-					klog.Errorf("Failed to get IP address by custom interface %s, err: %v", config.Modules.Edged.CustomInterfaceName, err)
-					os.Exit(1)
+					klog.Exitf("Failed to get IP address by custom interface %s, err: %v", config.Modules.Edged.CustomInterfaceName, err)
```
fujitatomoya, line 125: To keep the consistency (or unify the code base pattern), i would suggest to change else where that uses the same pattern. (you can grep it with `os.Exit(1)`)

e.g)

https://github.com/kubeedge/kubeedge/blob/b18b7248b21f3d0f67713900a54827f4e2116598/cloud/pkg/cloudhub/cloudhub.go#L97-L98

besides, what can be the justification for this change except simplifying?

JianChengBlueDot:
The `container runtime e2e test (isulad)` failure seems unrelated to this PR.  
This change only replaces `os.Exit(1)` with `klog.Exitf` for consistent error handling  
and should not affect runtime behavior.  

Other checks (build, lint, cri-o, docker) have all passed successfully.

edge/cmd/edgecore/app/server.go
JianChengBlueDot: > To keep the consistency (or unify the code base pattern), i would suggest to change else where that uses the same pattern. (you can grep it with `os.Exit(1)`)
> 
> e.g)
> 
> https://github.com/kubeedge/kubeedge/blob/b18b7248b21f3d0f67713900a54827f4e2116598/cloud/pkg/cloudhub/cloudhub.go#L97-L98
> 
> besides, what can be the justification for this change except simplifying?

Thank you for the feedback. Beyond simplification, this change provides:

1. Better debugging - Structured logging with timestamps and error context
2. Consistent monitoring - Unified error reporting for production systems
3. Easier maintenance - Single pattern for all fatal errors

I'll extend this fix to other os.Exit(1) instances as suggested.

PR #6484, opened 2025-10-14
Add protobuf support for metaserver

### LINKED ISSUE #6465: Enhance MetaServer to support content-type application/vnd.kubernetes.protobuf

brinker-tbaker:
**What type of PR is this?**

/kind feature

starting in kubernetes 1.32 the client-go library defaults to sending requests in protobuf format.

Fixes #6465
Replaces #6480

this change accepts incoming protobuf requests and responds back via protobuf for the core kubernetes api. If the concrete type is not found, it falls back to responding via json.

NONE

```release-note
adds application/vnd.kubernetes.protobuf support to metaserver
```

### DIFF
--- a/edge/pkg/metamanager/metaserver/handlerfactory/standard_write.go
+++ b/edge/pkg/metamanager/metaserver/handlerfactory/standard_write.go
@@ -33,23 +33,61 @@ import (
 	"k8s.io/apiserver/pkg/endpoints/handlers"
 	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
 	"k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/registry/rest"
 	"k8s.io/klog/v2"
 
 	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/metaserver/kubernetes/fakers"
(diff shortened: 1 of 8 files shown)


brinker-tbaker:
i've put this back in [WIP] as i've found a slight defect in the list function with this approach.

brinker-tbaker:
this PR is now ready for review @kadisi @vincentgoat @Shelley-BaoYue 

From a testing perspective, I've successfully used this to allow newer versions of cilium and rook (using kubernetes go-client v0.34.0) which default to using protobuf calls, to run correctly.

brinker-tbaker:
gofmt and goimports issues fixed

PR #6489, opened 2025-10-16
Add the capability to query Node resources from the edge.

luomengY:
**What type of PR is this?**
/kind feature

In the underlying layer of edged, multiple modules need to retrieve Node resources from the API server, such as: nodestatusmanager, volumemanager, evictionmanager, and podmanager. Calls from these modules are typically made in a periodic polling manner. In edgecore, queries for Node resources from edged are handled through the metaclient, which forwards the requests via the cloud-edge channel through cloudcore to the API server. The Node resources obtained from the API server are then returned to the edge node's edgecore through the same cloud-edge channel.

Fixes #https://github.com/kubeedge/kubeedge/issues/6491

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -114,7 +114,6 @@ func requireRemoteQuery(resType string) bool {
 		resType == constants.ResourceTypePersistentVolume ||
 		resType == constants.ResourceTypePersistentVolumeClaim ||
 		resType == constants.ResourceTypeVolumeAttachment ||
-		resType == model.ResourceTypeNode ||
 		resType == model.ResourceTypeServiceAccountToken ||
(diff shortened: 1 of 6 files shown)


luomengY:
@Shelley-BaoYue  see

edge/pkg/metamanager/process.go
Shelley-BaoYue, line 257: Maybe we need optimize process here for node update message, otherwise it will be sent to edged. Althought edged won't process node update message, it log the error log `resType is not pod or configmap or secret or volume: resType is node`, it will confuse users.

luomengY: Done

Shelley-BaoYue (commented):
PTAL @kubeedge/maintainers

PR #6508, opened 2025-10-23
Optimize endpoint type validation using map lookup

JianChengBlueDot:
- Replace array iteration with map-based lookup for validating source to target endpoint type mappings
- Add helper function isValidEndpointMapping for cleaner validation logic
- Improve validation performance with O(1) lookup time
- Maintain same validation rules and error messages


fujitatomoya (changes requested):
having `isValidEndpointMapping` would be nice for helper function. Gemini code review comments make sense, so that those need to be addressed.

Issue #6519, opened 2025-10-29
Request to K8s API provided by Kubeedge Metaserver fails with Unauthorized error in case of multiple edge nodes


VladislavZhuravlyov:
**What happened**:

  All HTTPS requiests to k8s API from inside container running on one of Kubeedge edge node fail with 401 error in case of multiple edge nodes cluster. At the same time the same requests work fine on other nodes.
  PODs are controlled by Daemonset. Kubeedge cluster is configured for in-cluster access Kube-APIServer according to https://release-1-20.docs.kubeedge.io/docs/advanced/inclusterconfig.

**What you expected to happen**:
All PODs must provide valid credentials to make request to K8s API on every edge node.

**How to reproduce it (as minimally and precisely as possible)**:

1. You need Kubeedge cluster wich has two edge nodes as minimum
2. Use attached test-daemonset.yaml which runs pods to execute requests to K8s Api. The daemonset's pod template defines init container which 
makes request to K8s API and requires authorisation credentials.
3. Use attached test-script.sh script to run tests which should reproduce the issue.
The script creates namespace 'n1' and in a loop creates Daemonset and waits for pods are running on every edge node. After the pods start successfully the script deletes the daemonset, waits for related to pods cerviceaccountaccess object is deleted and then creates new daemonset and so on.
The pod has init container which makes request to K8s API. If the request to K8s API failes with 401 error the init container 
waits for 10 seconds and tries again. As result pod doesn't start if 401 is returned every attempt.
Run the test script:  
` ./test-script.sh 1000`
1000 is the number of test iterations (default 20)
In our testing the issue is reproduced as rule in 100 iterations.  
For example: 
```
Iteration 22
daemonset.apps "ke-test-401" deleted from n1 namespace
Waiting serviceaccountaccess objects are deleted
Waiting serviceaccountaccess objects are deleted
Waiting serviceaccountaccess objects are deleted
No resources found in n1 namespace.
05:51:33
daemonset.apps/ke-test-401 created
pod/ke-test-401-5jtnt condition met
pod/ke-test-401-qxpml condition met
05:51:38
Iteration 23
daemonset.apps "ke-test-401" deleted from n1 namespace
Waiting serviceaccountaccess objects are deleted
Waiting serviceaccountaccess objects are deleted
Waiting serviceaccountaccess objects are deleted
No resources found in n1 namespace.
05:52:14
daemonset.apps/ke-test-401 created
[...]

PR #6533, opened 2025-11-04
Fix policymanager reconciliation

mkhon:
- Use non-cached reader (returned by manager.GetAPIReader()) instead of
manager.GetClient() which returns cached client which does not always see recent Kubernetes object changes

Using cached client leads to wrong decisions during Reconcile() so sometime ssend2Edge() is not called for some nodes as Reconcile() does not see recently created Pods

Side-effect of using cached client: many Reconcile() errors like:

2025-11-03T15:58:30.713482Z stderr F E1103 15:58:30.713304       1 reconcile.go:218] failed to create serviceaccountaccess, serviceaccountaccesses.policy.kubeedge.io "default" already exists
2025-11-03T15:58:30.796016537Z stderr F E1103 15:58:30.795703       1 reconcile.go:472] failed to update serviceaccountaccess status ns01/default, Operation cannot be fulfilled on serviceaccountaccesses.policy.kubeedge.io "default": the object has been modified; please apply your changes to the latest version and try again

- Fix Reconcile() not triggered by Pods at least in DaemonSet case: DaemonSet Pods are created without the .Spec.NodeName first (and are filtered out by filterObject()), then Pod is patched but Reconcile() is not triggered when Pods are updated (by mapObjectFunc())

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/policycontroller/policycontroller.go
+++ b/cloud/pkg/policycontroller/policycontroller.go
@@ -58,11 +58,11 @@ func NewAccessRoleControllerManager(ctx context.Context, kubeCfg *rest.Config) (
 }
 
 func setupControllers(ctx context.Context, mgr manager.Manager) error {
-	// This returned cli will directly acquire the unstructured objects from API Server which
+	// mgr.GetClient() will directly acquire the unstructured objects from API Server which
 	// have not be registered in the accessScheme.
-	cli := mgr.GetClient()
 	pc := &pm.Controller{
-		Client:       cli,
+		Client:       mgr.GetClient(),
+		Reader:       mgr.GetAPIReader(),
 		MessageLayer: messagelayer.PolicyControllerMessageLayer(),
 	}
 
--- a/cloud/pkg/policycontroller/manager/reconcile.go
+++ b/cloud/pkg/policycontroller/manager/reconcile.go
@@ -33,12 +33,13 @@ import (
 
 type Controller struct {
 	client.Client
+	client.Reader
 	MessageLayer messagelayer.MessageLayer
 }
 
 func (c *Controller) Reconcile(ctx context.Context, request controllerruntime.Request) (controllerruntime.Result, error) {
 	acc := &policyv1alpha1.ServiceAccountAccess{}
-	if err := c.Client.Get(ctx, request.NamespacedName, acc); err != nil {
+	if err := c.Reader.Get(ctx, request.NamespacedName, acc); err != nil {
 		if apierrors.IsNotFound(err) {
 			return controllerruntime.Result{}, nil
 		}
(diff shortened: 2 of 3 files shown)


mkhon:
Fixes #6519

cloud/pkg/policycontroller/policycontroller.go
```diff
@@ -63,6 +63,7 @@ func setupControllers(ctx context.Context, mgr manager.Manager) error {
 	cli := mgr.GetClient()
 	pc := &pm.Controller{
 		Client:       cli,
+		Reader:       mgr.GetAPIReader(),
 		MessageLayer: messagelayer.PolicyControllerMessageLayer(),
 	}
```
WillardHu:
```golang
pc := &pm.Controller{
    Client:       mgr.GetClient(),
    Reader:       mgr.GetAPIReader(),
    MessageLayer: messagelayer.PolicyControllerMessageLayer(),
}
```

Shelley-BaoYue (commented):
Please modify the commit message like `fix policymanager reconciliation with non-cached reader`

Image

cloud/pkg/policycontroller/policycontroller.go
mkhon: @WillardHu I haven't changed the old code (Client: cli is the old code) to keep the changes as minimal as possible.
I would suggest to do the style changes in a separate change

mkhon:
> Please modify the commit message like `fix policymanager reconciliation with non-cached reader`

Done

cloud/pkg/policycontroller/policycontroller.go
mkhon: I went forward and made the suggested change (also fixing a comment to match the new code)

Shelley-BaoYue:
> > Please modify the commit message like `fix policymanager reconciliation with non-cached reader`
> 
> Done

Sorry, it seems the commit message still contains too much.  

image

Shelley-BaoYue:
Perhaps I didn't explain clearly. What I mean is to recommit using `git commit -s -m "fix policymanager reconciliation with non-cached reader"`.

mkhon:
> Perhaps I didn't explain clearly. What I mean is to recommit using `git commit -s -m "fix policymanager reconciliation with non-cached reader"`.

But that was not the only change. There is also this change:

```
    - Fix Reconcile() not triggered by Pods at least in DaemonSet case:
    DaemonSet Pods are created without the .Spec.NodeName first (and are filtered
    out by filterObject()), then Pod is patched but Reconcile() is not triggered
    when Pods are updated (by mapObjectFunc())
```

So please advise.

Also, I thought more detailed commit message won't hurt

Shelley-BaoYue:
> > Perhaps I didn't explain clearly. What I mean is to recommit using `git commit -s -m "fix policymanager reconciliation with non-cached reader"`.
> 
> But that was not the only change. There is also this change:
> 
> ```
>     - Fix Reconcile() not triggered by Pods at least in DaemonSet case:
>     DaemonSet Pods are created without the .Spec.NodeName first (and are filtered
>     out by filterObject()), then Pod is patched but Reconcile() is not triggered
>     when Pods are updated (by mapObjectFunc())
> ```
> 
> So please advise.
> 
> Also, I thought more detailed commit message won't hurt

`git commit -s -m "fix policymanager reconciliation with non-cached reader"` is  just an example; it's intended to be a concise, one-sentence summary of what this commit does, allowing everyone to quickly and clearly grasp the work at a glance. Detailed descriptions are available in PR description. 
As reported by the robot, the current commit message contains invalid keywords and characters, which violates the Code of Conduct, and therefore cannot be merged.

Shelley-BaoYue (commented):
/approve

PR #6536, opened 2025-11-08
bugfix: keadm reset edge failed to remove containers

### LINKED ISSUE #6535: keadm reset edge failed with runtime error: invalid memory address or nil pointer dereference
**What happened**:
when use keadm version v1.22.0, keadm reset edge always failed,

Image

keadm reset edge success and edge containers were clean up

use keadm version v1.22.0, and execute `keadm reset edge`.

`containerRuntime.Connect` need to be called before use.

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

will4j:
**What type of PR is this?**

/kind bug

Fixes #6535

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/reset_test.go
+++ b/keadm/cmd/keadm/app/cmd/util/reset_test.go
@@ -50,6 +50,10 @@ type mockRuntime struct {
 	utilruntime.ContainerRuntime
 }
 
+func (m *mockRuntime) Connect() error {
+	return nil
+}
+
 func (m *mockRuntime) ListKubeContainers() ([]string, error) {
 	return []string{"container1", "container2"}, nil
 }


will4j:
> the fix itself looks good to address the issue.
> 
> but i think gemini makes sense here, it would be really nice to have this under generic implementation agnostic from platform. this is also good for maintenance.



> the fix itself looks good to address the issue.
> 
> but i think gemini makes sense here, it would be really nice to have this under generic implementation agnostic from platform. this is also good for maintenance.

I just want to fix the issue, but as gemini reviewed, use util.RemoveContainers to reduce duplicated code and it do good for maintenance, so i did a little bit refactoring.

keadm/cmd/keadm/app/cmd/util/reset.go
```diff
@@ -43,6 +43,9 @@ func RemoveContainers(criSocketPath string) error {
 	}
 
 	containerRuntime := utilruntime.NewContainerRuntime(criSocketPath)
+	if err := containerRuntime.Connect(); err != nil {
+		return err
+	}
```
will4j, line 48: I noticed that in latest k8s repo, a `Close` method is added to `ContainerRuntime` interface, so if upgrade dependencies in the following version, it may need to add `defer containerRuntime.Close()` as well. Which indicates that the impl now has potential resources leak problem, and it is far beyond this pull request.

will4j: see https://github.com/kubernetes/kubernetes/pull/133893

luomengY:
Resolve the e2e test error first.

keadm/cmd/keadm/app/cmd/util/reset.go
Shelley-BaoYue: Thanks for you remind. : )

Shelley-BaoYue (commented):
/approve

Shelley-BaoYue (commented):
Hi, @will4j Could you please cherry-pick this PR to previous releases(maybe release 1.19-1.22)?  FYI: https://kubeedge.io/docs/developer/cherry_pick

PR #6551, opened 2025-11-26
Allow cloudCore service type to be any valid Kubernetes type instead …

### LINKED ISSUE #6550: keadm init ignores --set cloudCore.service.type=LoadBalancer
https://github.com/kubeedge/kubeedge/blob/49107e9cbc76db5062a35f1432d50842e4481843/manifests/charts/cloudcore/templates/service_cloudcore.yaml#L13

Using `keadm init` with `--set cloudCore.service.type=LoadBalancer` does not change the cloudCore Service type. It remains ClusterIP.

**Steps to reproduce**
1. Install KubeEdge v1.22.0
2. Run:
```
   keadm init --kubeedge-version=v1.22.0 --kube-config-path=/root/.kube/config --advertise-address=<cloud-ip> --set cloudCore.service.type=LoadBalancer
```
3. Check cloudCore Service:
  `kubectl get svc -n kubeedge`

**Observed behavior**
Service type is `ClusterIP`.

**Expected behavior**
Service type should be `LoadBalancer`.

**Additional context**
The Helm template in `cloudcore/templates/service.yaml` has this conditional:
```
spec:
{{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
type: {{ .Values.cloudCore.service.type }}
{{- else }}
type: ClusterIP
{{- end }}
```

denmarkmeralpis:
**What type of PR is this?**

/kind bug

Allows the `cloudCore` Service type to respect any valid Kubernetes Service type (`ClusterIP`, `NodePort`, `LoadBalancer`) instead of forcing `ClusterIP` when using `keadm` or Helm values. This fixes cases where `--set cloudCore.service.type=LoadBalancer` was being ignored.

Fixes #6550 

cloudCore Service now respects the configured service type from Helm values or keadm flags, allowing LoadBalancer or NodePort without being overridden to ClusterIP.

### DIFF
--- a/manifests/charts/cloudcore/templates/service_cloudcore.yaml
+++ b/manifests/charts/cloudcore/templates/service_cloudcore.yaml
@@ -1,54 +1,50 @@
-{{- if and (.Values.cloudCore.service.enable) }}
+{{- if .Values.cloudCore.service.enable }}
 apiVersion: v1
 kind: Service
 metadata:
+  name: cloudcore
   {{- with .Values.cloudCore.service.annotations }}
   annotations: {{- toYaml . | nindent 4 }}
   {{- end }}
   {{- with .Values.cloudCore.labels }}
   labels: {{- toYaml . | nindent 4 }}
   {{- end }}
-  name: cloudcore
 spec:
-  {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
-  type: {{ .Values.cloudCore.service.type }}
-  {{- else }}
-  type: ClusterIP
-  {{- end }}
+  type: {{ .Values.cloudCore.service.type | default "ClusterIP" }}
   ports:
-  - port: 10000
-    targetPort: 10000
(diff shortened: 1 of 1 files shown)


manifests/charts/cloudcore/templates/service_cloudcore.yaml
```diff
@@ -1,54 +1,50 @@
...
-    name: cloudstream
-  - port: 10004
-    targetPort: 10004
-    {{- if and (eq .Values.cloudCore.service.type "NodePort") ( not .Values.cloudCore.hostNetWork) }}
-    nodePort: {{ .Values.cloudCore.service.tunnelNodePort }}
-    {{- end }}
-    name: tunnelport
+    - port: 10000
+      targetPort: 10000
+      {{- if eq .Values.cloudCore.service.type "NodePort" }}
```
WillardHu, line 17: `cloudCore.service.type` and `cloudCore.hostNetWork` are different field configurations. If the `hostNetwork` is `true`, the Service will not take effect. You could consider using `hostNetwork` to determine wheter to create the Service.

PR #6558, opened 2025-11-28
Fix PolicyManager reconciliation:

mkhon:
- Use non-cached reader. Using cached client leads to wrong decisions during Reconcile() so sometime send2Edge() is not called for some nodes as Reconcile() does not see recently created Pods

- Fix Reconcile() not triggered by Pods at least in DaemonSet case: DaemonSet Pods are created without the .Spec.NodeName first (and are filtered out by filterObject()), then Pod is patched but Reconcile() is not triggered when Pods are updated (by mapObjectFunc())

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/cloud/pkg/policycontroller/manager/reconcile.go
+++ b/cloud/pkg/policycontroller/manager/reconcile.go
@@ -33,12 +33,13 @@ import (
 
 type Controller struct {
 	client.Client
+	client.Reader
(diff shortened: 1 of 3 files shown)


Shelley-BaoYue (commented):
/lgtm 
/approve

PR #6560, opened 2025-12-01
refactor: extract dmicache into a seperate module and add unit tests

HaojieZhang6848:
**What type of PR is this?**

/kind design

**What this PR does / why we need it**: In the previous version, the `DMICache` struct is defined in the `dmiserver/server.go`, and the `dmiserver` and `dmiworker` directly operates the internal structure of `DMICache`. This is a historial problem. In this PR, I extracted the `DMICache` struct into a seperate package, excapsulate its fields, add member methods for `DMICache`, and add unit tests for it. Now, the `dmiserver` and `dmiworker` works on public methods of `DMICache`, and `DMICache`'s internal fields are fully encapsulated. This improves the code style and maintainablity.

**Which issue(s) this PR fixes**:

Fixes [WillardHu](https://github.com/WillardHu)'s advice in #6542. https://github.com/kubeedge/kubeedge/pull/6542#discussion_r2555062167

**Special notes for your reviewer**: None

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #6572, opened 2025-12-15
Avoid unnecessary UpdateDevice calls in dealMetaDeviceOperation when device has no change

vincyfli:
**What type of PR is this?**
/kind bug

In `edge/pkg/devicetwin/dtmanager/dmiworker.go`, the `dealMetaDeviceOperation` method handles device operations via a `switch` on `message.GetOperation()`.

When receiving `UpdateOperation`, it always calls:

```
err = dmiclient.DMIClientsImp.UpdateDevice(&device)
```

However, even when the device `spec` has no changes compared to the previous state (only `status` is updated), the code still triggers `UpdateDevice`.

This leads to an unexpected behavior in mapper implementations:
Most default mapper frameworks restart the device as part of their update logic (stop → start).
As a result, the device keeps restarting repeatedly even though the `desired` configuration never changed — only the `status` got updated.

This behavior is not reasonable, as `status` updates should not trigger a device reconfiguration or restart.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/dmiworker.go
+++ b/edge/pkg/devicetwin/dtmanager/dmiworker.go
@@ -134,16 +134,19 @@ func (dw *DMIWorker) dealMetaDeviceOperation(_ *dtcontext.DTContext, _ string, m
 			}
 			dw.dmiCache.RemoveDevice(device.Namespace, device.Name)
 		case model.UpdateOperation:
-			dw.dmiCache.PutDevice(&device)
-			// Override device instance config with model defaults before updating
-			devicePtr, _, err := dw.dmiCache.GetOverriddenDevice(device.Namespace, device.Name)
-			if err != nil {
-				return err
-			}
-			err = dmiclient.DMIClientsImp.UpdateDevice(devicePtr)
-			if err != nil {
-				klog.Errorf("update device %s failed with err: %v", devicePtr.Name, err)
-				return err
+			if dw.dmiCache.CompareDeviceSpecHasChanged(&device) {
+				dw.dmiCache.PutDevice(&device)
+				// Override device instance config with model defaults before updating
+				devicePtr, _, err := dw.dmiCache.GetOverriddenDevice(device.Namespace, device.Name)
+				if err != nil {
+					return err
+				}
+
+				err = dmiclient.DMIClientsImp.UpdateDevice(devicePtr)
+				if err != nil {
+					klog.Errorf("update device %s failed with err: %v", devicePtr.Name, err)
+					return err
+				}
 			}
 		default:
 			klog.Warningf("unsupported operation %s", message.GetOperation())
--- a/edge/pkg/devicetwin/dmicache/dmicache.go
+++ b/edge/pkg/devicetwin/dmicache/dmicache.go
@@ -2,6 +2,8 @@ package dmicache
 
 import (
 	"fmt"
+	"reflect"
+	"sort"
 	"sync"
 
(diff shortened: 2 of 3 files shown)


vincyfli:
@Shelley-BaoYue As discussed earlier, it has been adjusted to be merged into master, can we merge it now?

edge/pkg/devicetwin/dtmanager/dmiworker.go
```diff
@@ -155,6 +157,35 @@ func (dw *DMIWorker) dealMetaDeviceOperation(_ *dtcontext.DTContext, _ string, m
 				return err
 			}
 			dw.dmiCache.DeviceMu.Lock()
+			oldDevice, ok := dw.dmiCache.DeviceList[deviceID]
```
WillardHu: You can use `metadata.generation` field to determine if the spec has changed.
In Kubernetes, `metadata.generation` is an automatically managed sequence number that is incremented by the API server every time the spec (desired state) of a resource is changed. It is a critical field used by controllers to manage updates and ensure the system's actual state matches the desired state.

WillardHu: > You can use metadata.generation field to determine if the spec has changed.

Sorry, this does not apply to situations where the device model changes.

You can complete this fix based on PR #6560, and abstruct the comparison operation as a method into  DMICache, and add some unit test cases.

Otherwise, the conflict would need to be handled separately.

WillardHu: The #6560 PR has been merged, you can rebase the master branch.

vincyfli: > The #6560 PR has been merged, you can rebase the master branch.

The code has been updated based on [#6560 ](https://github.com/kubeedge/kubeedge/pull/6560) PR.

edge/pkg/devicetwin/dtmanager/dmiworker.go
```diff
@@ -134,16 +134,19 @@ func (dw *DMIWorker) dealMetaDeviceOperation(_ *dtcontext.DTContext, _ string, m
...
 			devicePtr, _, err := dw.dmiCache.GetOverriddenDevice(device.Namespace, device.Name)
 			if err != nil {
 				return err
 			}
-			err = dmiclient.DMIClientsImp.UpdateDevice(devicePtr)
-			if err != nil {
-				klog.Errorf("update device %s failed with err: %v", devicePtr.Name, err)
-				return err
+
+			if dw.dmiCache.CompareDeviceSpecHasChanged(&device) {
```
WillardHu, line 137: `GetOverriddenDevice()` needs to be called after `PutDevice()`
```golang
if dw.dmiCache.CompareDeviceSpecHasChanged(&device) {
    dw.dmiCache.PutDevice(&device)
    // Override device instance config with model defaults before updating
    devicePtr, _, err := dw.dmiCache.GetOverriddenDevice(device.Namespace, device.Name)
    if err != nil {
        return err
    }
    if err := dmiclient.DMIClientsImp.UpdateDevice(devicePtr); err != nil {
        klog.Errorf("update device %s failed with err: %v", devicePtr.Name, err)
        return err
    }
}
```

WillardHu (changes requested):

WillardHu:
The new commit lost the `Signed-off-by`

vincyfli:
> The new commit lost the `Signed-off-by`

Signed-off-by has been added.

PR #6591, opened 2025-12-23
fix: remove alsologtostderr args for admission and edgemark template

Abirdcfly:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

**In fact, #5517 already attempted to remove these parameters, but for some reason that PR was not merged at that time.**

Currently, by default, a Helm installation does not enable the `admission` component because `enable` is set to `false` in [values.yaml](https://github.com/kubeedge/kubeedge/blob/e5e12547e29aa344a5c5e3382ce88984d0a50e25/manifests/charts/cloudcore/values.yaml#L147). However, even if it is enabled, admission still fails to start successfully. There are two issues:

1. The `alsologtostderr` flag is no longer supported by the program (this is what the current PR is addressing).
2. A certificate file is strictly required; otherwise, the following error is reported:
   `Error: unable to read cacert file: open : no such file or directory`. This can be generated by running `build/admission/gen-admission-secret.sh`. #6592 try to implement this using Helm’s [genca](https://helm.sh/docs/chart_template_guide/function_list/#genca).

### Details about the missing `alsologtostderr` flag

At commit 5f6b0a59b06ac73ebfd5107c6d6ef278905d1142, the `alsologtostderr` flag was still present:

```bash
go run -v ./cloud/cmd/admission/ --alsologtostderr
...
--alsologtostderr
                log to standard error as well as files (default false) (DEPRECATED: will be removed in a future
                release, see
                https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/2845-deprecate-klog-specific-flags-in-k8s-components)
...
```

The call chain is:

```
cloud/cmd/admission/admission.go
  -> "k8s.io/component-base/logs" init()
    -> klog.InitFlags(packageFlags)
```

In this path, `klog.InitFlags(packageFlags)` adds the `alsologtostderr` flag.

The corresponding dependency versions were:

```
k8s.io/component-base v0.26.0
k8s.io/component-base => github.com/kubeedge/kubernetes/staging/src/k8s.io/component-base v1.24.14-kubeedge1
```

At commit fa9960bdb8902b4aca10f9bd5945826f6771c56e , the `alsologtostderr` flag no longer exists:

```bash
go run -v ./cloud/cmd/admission/ --alsologtostderr
Error: unknown flag: --alsologtostderr
```

The call chain remains the same:

```
cloud/cmd/admission/admission.go
  -> "k8s.io/component-base/logs" init()
    -> klog.InitFlags(packageFlags)
```

[...]

PR #6592, opened 2025-12-23
feat: auto-generate certificates for admission in chart

Abirdcfly:
**What type of PR is this?**
/kind feature

Currently, by default, a Helm installation does not enable the admission component because enable is set to false in [values.yaml](https://github.com/kubeedge/kubeedge/blob/e5e12547e29aa344a5c5e3382ce88984d0a50e25/manifests/charts/cloudcore/values.yaml#L147). However, even if it is enabled, admission still fails to start successfully. There are two issues:

1. The alsologtostderr flag is no longer supported by the program (#6591).
2. A certificate file is strictly required; otherwise, the following error is reported:
`Error: unable to read cacert file: open : no such file or directory`. This can be generated by running build/admission/gen-admission-secret.sh. this PR try to implement this using Helm’s [genca](https://helm.sh/docs/chart_template_guide/function_list/#genca).

Fixes #

```release-note
None
```

### DIFF
--- a/manifests/charts/cloudcore/templates/_helpers.tpl
+++ b/manifests/charts/cloudcore/templates/_helpers.tpl
@@ -34,3 +34,26 @@ streamCA.crt: {{ $ca.Cert | b64enc }}
 stream.crt: {{ $cert.Cert | b64enc }}
 stream.key: {{ $cert.Key | b64enc }}
 {{- end -}}
+
+{{/*
+Return admission cert secret name
+*/}}
+{{- define "kubeedge.admission.certsSecretName" -}}
+{{- if .Values.admission.certsSecretName -}}
+{{ .Values.admission.certsSecretName }}
+{{- else -}}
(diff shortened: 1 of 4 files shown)


WillardHu:
The current kube-builder framework used for controllers and admissions is outdated. We can utilize a newer kube-builder framework to:
- Combining admissions and controller-manager into a single process.
- Using a WebHook service instead of admissions.
- Managing certificates using https://github.com/open-policy-agent/cert-controller.
- Adding LeaderElection settings to the controller-manager.

If you are willing to make these changes, you can write a proposal first.

Shelley-BaoYue:
Hi, @Abirdcfly , as @WillardHu said, would you like to attend KubeEdge community meeting to discuss more about this PR?  Also, we really appreciate your recent contributions to community and look forward to have deeper exchanges with you.  The community meeting time is 16:00 every Wednesday.

PR #6597, opened 2025-12-24
load previously registered mapper info and init DMI Client automatically after restarted edgecore

vincyfli:
**What type of PR is this?**
/kind bug

Once edgecore restarts, the existing device mapper cannot be found using the current code,  the following error will occur:

`fail to get dmi client of protocol modbus`

the direct triggering factors are as follows：

```
dc, ok := dcs.clients[protocol]
if !ok {
	return nil, fmt.Errorf("fail to get dmi client of protocol %s", protocol)
}
```

After restarting edgecore, the previously registered device mapper is no longer found in **dcs.clients**.
The above phenomenon is abnormal for previously registered device mapper app, because the edgecore process is still running at this time, it has only restarted but not stopped.

Fixes #

Fixes the issue of mapper information being lost after edgecore restart.

All current release versions have this issue, therefore the target branch is master. 
Thanks.

```release-note

```

### DIFF
--- a/edge/pkg/devicetwin/dmiclient/client.go
+++ b/edge/pkg/devicetwin/dmiclient/client.go
@@ -78,8 +78,13 @@ func (dc *DMIClient) connect() error {
 }
 
 func (dc *DMIClient) close() {
-	dc.Conn.Close()
-	dc.CancelFunc()
+	if dc.Conn != nil {
+		dc.Conn.Close()
+	}
+	if dc.CancelFunc != nil {
+		dc.CancelFunc()
+	}
+	dc.Client = nil
 }
 
 func createDeviceRequest(device *v1beta1.Device) (*dmiapi.RegisterDeviceRequest, error) {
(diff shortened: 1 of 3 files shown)


Shelley-BaoYue (commented):
/approve

edge/pkg/devicetwin/dmiclient/client.go
```diff
@@ -168,6 +172,40 @@ func (dcs *DMIClients) CreateDMIClient(protocol, sockPath string) {
 	dcs.mutex.Unlock()
 }
 
+// CreateDMIClientAndTryToConnect If the dmi client of the protocol does not exist, create one and try to connect.
+// If a connection to the given sockPath cannot be established (e.g., the mapper app is not running), it indicates
+// that the mapper corresponding to the sockPath does not need to be re-registered; otherwise, the DMIClient
+// for that protocol will be created normally.
+func (dcs *DMIClients) CreateDMIClientAndTryToConnect(protocol, sockPath string) error {
```
WillardHu: The `CreateDMIClient()` method is not called in many places, so it is recommended to extend it to `CreateDMIClient(protocol, sockPath string, tryConnect bool)`.

edge/pkg/devicetwin/dmiclient/client.go
```diff
@@ -168,6 +172,40 @@ func (dcs *DMIClients) CreateDMIClient(protocol, sockPath string) {
...
+		return fmt.Errorf("fail to connect the mapper app for protocol %s with socket %s: %v", protocol, sockPath, err)
+	}
+	dcs.mutex.Lock()
+	defer dcs.mutex.Unlock()
+	c, ok := dcs.clients[protocol]
+	if ok {
+		c.close()
+	}
+	dcs.clients[protocol] = dc
+	klog.Infof("created dmi client for protocol %s with socket %s successfully", protocol, sockPath)
```
WillardHu: If this message will be printed frequently, change it to debug level.

edge/pkg/devicetwin/dtmanager/dmiworker.go
```diff
@@ -244,7 +244,12 @@ func (dw *DMIWorker) initDeviceMapperInfoFromDB() {
 			return
 		}
 
+		err := dmiclient.DMIClientsImp.CreateDMIClientAndTryToConnect(deviceMapper.Protocol, string(deviceMapper.Address))
+		if err != nil {
+			continue
```
WillardHu, line 250: Print error message to a warning log.

edge/pkg/devicetwin/dtmanager/dmiworker.go
```diff
@@ -244,7 +244,12 @@ func (dw *DMIWorker) initDeviceMapperInfoFromDB() {
...
+		}
 		dw.dmiCache.PutMapper(&deviceMapper)
+		klog.Infof("device mapper %s loaded successfully", deviceMapper.Name)
```
WillardHu: If this message will be printed frequently, change it to debug level.

PR #6603, opened 2025-12-29
fix: add get verbs for cloudcore rbac

Abirdcfly:
**What type of PR is this?**
/kind cleanup

Fixes #

```release-note
None
```


Abirdcfly:
> @Abirdcfly can you share why this is required to add? if you have concrete use case, i am interested in that!

~~Deploying kubeedge v1.22.1 and the latest edgemesh in a modified k8s 1.27.15, we encountered this permission problem, I am still investigating...~~
It is actually the business code on the edge that is accessed through the in-cluster that needs these permissions...

PR #6606, opened 2025-12-30
Add unit tests for pkg/viaduct/pkg/fifo

ddc-baiye:
This PR adds unit tests for the `pkg/viaduct/pkg/fifo` package. It covers the following scenarios for `MessageFifo`:
- Initialization with `NewMessageFifo`
- Basic `Put` and `Get` operations
- Overflow handling (discarding old messages)
- Blocking behavior of `Get`
- `Close` operation and error handling on closed fifo

Fixes # 

None

```release-note
None
```

### DIFF
--- a/pkg/viaduct/pkg/fifo/msgfifo_test.go
+++ b/pkg/viaduct/pkg/fifo/msgfifo_test.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2025 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


pkg/viaduct/pkg/fifo/msgfifo_test.go
```diff
@@ -0,0 +1,151 @@
...
+		f.Close()
+	})
+}
+
+func TestMessageFifo_Get_Blocking(t *testing.T) {
+	f := NewMessageFifo()
+	msg := &model.Message{Header: model.MessageHeader{ID: "async"}}
+
+	done := make(chan struct{})
+	go func() {
```
WillardHu, line 126: Why use goroutine here?

ddc-baiye: The Get method blocks when the FIFO is empty. We must run it in a separate goroutine to prevent the main test thread from deadlocking. This allows the main thread to verify that Get is correctly blocking before we Put a message to wake it up

PR #6621, opened 2026-01-30
Fix devicemodel update

ddc-baiye:
**What type of PR is this?**

/kind bug

This PR implements the missing logic in `deviceModelUpdated` function within `cloud/pkg/devicecontroller/controller/downstream.go`.

Previously, when a `DeviceModel` was updated, the controller only updated the local cache but failed to propagate the changes to the edge nodes. This caused inconsistency where edge devices would continue using the old model definition even after a cloud-side update.

This PR adds logic to:
1. Iterate through devices to identify those referencing the updated `DeviceModel`.
2. Send the updated `DeviceModel` to the corresponding edge nodes via message layer.
3. Ensure updates are deduplicated per node to avoid redundant messages.

Fixes #

None.

```release-note
Fixed an issue where DeviceModel updates were not propagated to edge nodes.
```

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -96,9 +96,34 @@ func (dc *DownstreamController) deviceModelAdded(deviceModel *v1beta1.DeviceMode
 
 // deviceModelUpdated is function to process updated deviceModel
 func (dc *DownstreamController) deviceModelUpdated(deviceModel *v1beta1.DeviceModel) {
-	// nothing to do when deviceModel updated, only add in map
 	deviceModelID := util.GetResourceID(deviceModel.Namespace, deviceModel.Name)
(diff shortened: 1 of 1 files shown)


cloud/pkg/devicecontroller/controller/downstream.go
```diff
@@ -96,9 +96,34 @@ func (dc *DownstreamController) deviceModelAdded(deviceModel *v1beta1.DeviceMode
 
 // deviceModelUpdated is function to process updated deviceModel
 func (dc *DownstreamController) deviceModelUpdated(deviceModel *v1beta1.DeviceModel) {
-	// nothing to do when deviceModel updated, only add in map
```
fujitatomoya, line 99: seems like this is intentionally implemented in this way? do we have corresponding issue or report from users? i would highly recommend we should have discussion here before moving on, hold this PR until device model design is confirmed to change the behavior like this?

Issue #6609, opened 2026-01-14
Batch Enable/Disable of EdgeHub on Edge Nodes via Keadm


luomengY:
**What would you like to be added/modified**:
A keadm extension that provides commands or APIs to batch enable/disable the EdgeHub switch on specified edge nodes.

**Why is this needed**:
In bandwidth limited environments, as the cluster size expands, communication between cloud and edge nodes in KubeEdge consumes a significant amount of network resources, which may interfere with upper layer business applications. To alleviate this issue, it is necessary to implement batch enable or disable the EdgeHub edge node feature in keadm. Administrators can temporarily take edge nodes offline by turning off the EdgeHub switch, thereby releasing the bandwidth required for critical services. This feature should support adding subcommands in keadm for implementation.

Mysterio-17:
Hi @luomengY , I’m interested in contributing to KubeEdge for the LFX 2026 Term 1 mentorship program. I wanted to ask whether this issue will also have pretests, similar to the upstream issues [304](https://github.com/kubeedge/ianvs/issues/304) and [230](https://github.com/kubeedge/ianvs/issues/230)

aaryyya:
Hey @luomengY  I'm interested in contributing to KubeEdge for the LFX 2026 Term 1 mentorship program.

sarfraj89:
Hi @luomengY  👋
I'm Sarfraj Khan, a DevOps-focused undergraduate exploring the LFX mentorship with KubeEdge. I’m interested in cloud–edge Kubernetes deployments and improving operational workflows for edge nodes. Excited to contribute and learn from the community!

abhicodes11:
Hi @luomengY and @Shelley-BaoYue,

   I'm very interested in working on this batch EdgeHub control feature 
   for the LFX mentorship program. I've been studying the keadm codebase 
   and EdgeHub architecture.

   I've created a node selector utility (PR #6631 ) that could serve as 
   a foundation for this feature. I'd love to get your feedback on the 
   approach.

   Some questions about the implementation:
   1. For the EdgeHub control mechanism, would modifying the config file 
      and restarting edgecore be the preferred approach?
   2. Should we persist the disabled state so nodes stay disabled after 
      reboot?
   3. What's the expected behavior if cloud-edge connection is lost 
      during batch operations?

   Looking forward to contributing to this project!

sarfraj89:
Hi @luomengY and @Shelley-BaoYue , 

[...]

PR #6631, opened 2026-02-07
keadm: add node selector utility for batch operations

abhicodes11:
## What type of PR is this?
/kind feature

This PR adds a reusable NodeSelector utility to support batch operations on multiple edge nodes. It provides the foundation for implementing batch EdgeHub control (#6609) and other multi-node operations in keadm.

## Which issue(s) this PR fixes:
Related to #6609

## Features:
- ✅ Select nodes by explicit names (`--nodes node1,node2,node3`)
- ✅ Select nodes by label selector (`--selector key=value`)
- ✅ Input validation to prevent conflicting selections
- ✅ Comprehensive unit tests with edge case coverage
- ✅ Clear error messages for invalid inputs

## Usage Example:
```go
// Example 1: Select specific nodes
ns := nodeselect.NewNodeSelector()
ns.AddNodes("edge-node-1,edge-node-2,edge-node-3")
if err := ns.Validate(); err != nil {
    return err
}

// Example 2: Select by label
ns := nodeselect.NewNodeSelector()
ns.AddSelector("region", "us-west")
if err := ns.Validate(); err != nil {
    return err
}
```

## Testing:
```bash
go test -v ./keadm/cmd/keadm/app/cmd/util/nodeselect/
```

All tests pass with 100% coverage of the core logic.

## Future Use:
This utility will be used in the upcoming batch EdgeHub control feature and can be extended for other batch operations like:
- Batch node upgrades
- Batch configuration updates
- Batch health checks

## Special notes for your reviewer:
This is a foundational utility with no external dependencies. It's designed to be simple, testable, and reusable across different keadm commands.

I'm working on the batch EdgeHub control feature for the LFX mentorship program and wanted to get this foundational piece reviewed early.

cc @luomengY @Shelley-BaoYue

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/nodeselect/selector.go
+++ b/keadm/cmd/keadm/app/cmd/util/nodeselect/selector.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodeselect
+
+import (
+	"fmt"
+	"strings"
+)
+
+// NodeSelector provides methods to select multiple nodes for batch operations.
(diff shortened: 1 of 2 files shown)


abhicodes11:
After receiving feedback from @luomengY on the implementation approach (#6609), I understand that this node selector utility is not needed for the batch EdgeHub control project.

The correct approach is to extend the existing `keadm batch` framework using  configuration files with explicit node lists (SSH-based), rather than creating new subcommands with Kubernetes label selectors.

I will close this PR as it's not aligned with the project requirements. The code may still be useful for other keadm features in the future, but it's not the right solution for batch EdgeHub control.

Thank you for the guidance!

PR #6635, opened 2026-02-08
keadm: add EdgeHub status check command

abhicodes11:
**What type of PR is this?**

/kind feature

This PR adds a new EdgeHub status check command to the keadm ctl interface. It provides operators with a simple way to verify EdgeCore process status and EdgeHub connectivity on edge nodes through `keadm ctl status edgehub --node <nodename>`. The command checks node readiness conditions and running pods to determine EdgeHub health, following existing KubeEdge command patterns.

Related to #6609 (batch EdgeHub control)

- This follows the established command structure patterns in keadm/ctl
- Uses Kubernetes client API for status verification instead of direct HTTP calls
- Includes comprehensive unit tests with 100% coverage
- Clean, professional output matching existing KubeEdge command style

keadm now supports EdgeHub status checking with the new `keadm ctl status edgehub --node <nodename>` command. This allows operators to verify EdgeCore process status and EdgeHub connectivity on edge nodes.

### DIFF
--- a/keadm/cmd/keadm/app/cmd/ctl/status/edgehub.go
+++ b/keadm/cmd/keadm/app/cmd/ctl/status/edgehub.go
@@ -28,6 +28,15 @@ import (
 	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/util/metaclient"
 )
 
+const (
+	StatusRunning      = "Running"
+	StatusNotReady     = "Not Ready"
+	StatusConnected    = "Connected"
+	StatusNotConnected = "Not Connected"
+	StatusHealthy      = "Healthy"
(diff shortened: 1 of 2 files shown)


luomengY:
I don't think this feature is very necessary; there's no need to add a dedicated command just for checking the status of edgehub.

abhicodes11:
After receiving feedback from @luomengY on the implementation approach (#6609), I understand that the batch EdgeHub control should be implemented by extending the existing `keadm batch` framework, not by creating new standalone commands.

The status checking functionality would be integrated into the batch framework's reporting mechanism rather than as a separate command.

Closing this PR to align with the project requirements.

PR #6638, opened 2026-02-09
fix: update outdated error log message in edged

Abirdcfly:
**What type of PR is this?**
/kind cleanup

the log show `edged.go:362] resType is not pod or configmap or secret or volume: resType is secret` This kind of log is too confusing.

Check the code history and find that the code above has been updated, but the log has not. So update the log.

Fixes #

```release-note

```


Shelley-BaoYue (commented):
/approve

PR #6655, opened 2026-03-04
fix: avoid modifying original object in StrategicMergeMapPatch

Abirdcfly:
**What type of PR is this?**
/kind bug

Prevent the original object from being modified when calling strategicpatch.StrategicMergeMapPatch by:
1. Creating a deep copy of the original object before processing
2. Correctly updating the original object via pointer dereference

This avoids unexpected side effects on the input object.

Fixes #

```release-note

```

### DIFF
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/util/patch.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/util/patch.go
@@ -48,7 +48,7 @@ func StrategicPatchObject(
 	schemaReferenceObj runtime.Object,
 	validationDirective string,
 ) error {
-	originalObjMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(originalObject)
+	originalObjMapCopy, err := runtime.DefaultUnstructuredConverter.ToUnstructured(originalObject.DeepCopyObject())
 	if err != nil {
 		return err
 	}
@@ -66,7 +66,7 @@ func StrategicPatchObject(
 		}
 	}
 
-	return applyPatchToObject(requestContext, defaulter, originalObjMap, patchMap, objToUpdate, schemaReferenceObj, strictErrs, validationDirective)
+	return applyPatchToObject(requestContext, defaulter, originalObjMapCopy, patchMap, objToUpdate, schemaReferenceObj, strictErrs, validationDirective)
 }
 
 // applyPatchToObject applies a strategic merge patch of <patchMap> to
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
@@ -184,7 +184,7 @@ func MergePatchedResource(ctx context.Context, originalObj *unstructured.Unstruc
 		if err = patchutil.StrategicPatchObject(ctx, defaulter, originalObj, []byte((*metas)[0]), updatedResource, schemaReferenceObj, ""); err != nil {
 			return err
 		}
-		originalObj = updatedResource
+		*originalObj = *updatedResource
 	}
 	return nil
 }


luomengY:
The purpose here is to modify the original object. If err occurs, the original object will not be modified. I don't think this is a bug.

Abirdcfly:
> The purpose here is to modify the original object. If err occurs, the original object will not be modified. I don't think this is a bug.

If the purpose is to modify the original object, then this line of code `originalObj = updatedResource` is redundant, and many assignments to the update object in the code are also redundant, which is why I think it might be a bug.

luomengY:
> > The purpose here is to modify the original object. If err occurs, the original object will not be modified. I don't think this is a bug.
> 
> If the purpose is to modify the original object, then this line of code `originalObj = updatedResource` is redundant, and many assignments to the update object in the code are also redundant, which is why I think it might be a bug.

The `StrategicPatchObject` function does not change the `originalObject`, so modifying the deep copy here is meaningless.

edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/util/patch.go
```diff
@@ -66,7 +66,7 @@ func StrategicPatchObject(
 		}
 	}
 
-	return applyPatchToObject(requestContext, defaulter, originalObjMap, patchMap, objToUpdate, schemaReferenceObj, strictErrs, validationDirective)
+	return applyPatchToObject(requestContext, defaulter, originalObjMapCopy, patchMap, objToUpdate, schemaReferenceObj, strictErrs, validationDirective)
```
luomengY: This has no impact on `originalObject`.

Abirdcfly:
> > > The purpose here is to modify the original object. If err occurs, the original object will not be modified. I don't think this is a bug.
> > 
> > 
> > If the purpose is to modify the original object, then this line of code `originalObj = updatedResource` is redundant, and many assignments to the update object in the code are also redundant, which is why I think it might be a bug.
> 
> The `StrategicPatchObject` function does not change the `originalObject`, so modifying the deep copy here is meaningless.

https://github.com/kubeedge/kubernetes/blob/f178d6f24807dea1b11a4b310008ac096ebfe796/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/patch.go#L855
```go
// Warning: the original and patch JSONMap objects are mutated by this function and should not be reused.
```

In fact, I also tried to add logs before and after the code `originalObj = updatedResource`, and my tests show that the original value changes without needing to assign the update value to original...

luomengY:
> > > > The purpose here is to modify the original object. If err occurs, the original object will not be modified. I don't think this is a bug.
> > > 
> > > 
> > > If the purpose is to modify the original object, then this line of code `originalObj = updatedResource` is redundant, and many assignments to the update object in the code are also redundant, which is why I think it might be a bug.
> > 
> > 
> > The `StrategicPatchObject` function does not change the `originalObject`, so modifying the deep copy here is meaningless.
> 
> https://github.com/kubeedge/kubernetes/blob/f178d6f24807dea1b11a4b310008ac096ebfe796/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/patch.go#L855
> 
> ```go
> // Warning: the original and patch JSONMap objects are mutated by this function and should not be reused.
> ```
> 
> In fact, I also tried to add logs before and after the code `originalObj = updatedResource`, and my tests show that the original value changes without needing to assign the update value to original...

`originalObjMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(originalObject)`：
Even if the `originalObjMap` changes later, will it affect the `originalObject`?

Abirdcfly:
> > > > > The purpose here is to modify the original object. If err occurs, the original object will not be modified. I don't think this is a bug.
> > > > 
> > > > 
> > > > If the purpose is to modify the original object, then this line of code `originalObj = updatedResource` is redundant, and many assignments to the update object in the code are also redundant, which is why I think it might be a bug.
> > > 
> > > 
> > > The `StrategicPatchObject` function does not change the `originalObject`, so modifying the deep copy here is meaningless.
> > 
> > 
> > [kubeedge/kubernetes@`f178d6f`/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/patch.go#L855](https://github.com/kubeedge/kubernetes/blob/f178d6f24807dea1b11a4b310008ac096ebfe796/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/patch.go#L855)
> > ```go
> > // Warning: the original and patch JSONMap objects are mutated by this function and should not be reused.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In fact, I also tried to add logs before and after the code `originalObj = updatedResource`, and my tests show that the original value changes without needing to assign the update value to original...
> 
> `originalObjMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(originalObject)`： Even if the `originalObjMap` changes later, will it affect the `originalObject`?

yes... one example 

```go
package main

import (
	"encoding/json"
	"fmt"
	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
	"k8s.io/apimachinery/pkg/runtime"
)

func main() {
	originalObj := &unstructured.Unstructured{
		Object: map[string]interface{}{
			"apiVersion": "v1",
			"kind":       "Pod",
			"metadata": map[string]interface{}{
				"name":      "test-pod",
				"namespace": "default",
			},
			"spec": map[string]interface{}{
				"containers": []interface{}{
					map[string]interface{}{
						"name":  "nginx",
						"image": "nginx:latest",
					},
				},
			},
		},
	}

	fmt.Println("=== before StrategicPatchObject ===")
	printObj(originalObj)

	StrategicPatchObject(originalObj)

	fmt.Println("\n=== after StrategicPatchObject ===")
	printObj(originalObj)
}

func StrategicPatchObject(originalObject runtime.Object) {
	originalObjMap, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(originalObject)
	originalObjMap["key"] = "value"
}

func printObj(obj *unstructured.Unstructured) {
	jsonBytes, err := json.MarshalIndent(obj.Object, "", "  ")
	if err != nil {
		fmt.Printf("Error marshaling: %v\n", err)
		return
	}
	fmt.Println(string(jsonBytes))
}

```

go.mod:
```
module tmpobj

go 1.26

require k8s.io/apimachinery v0.32.10

require (
	github.com/fxamacker/cbor/v2 v2.7.0 // indirect
	github.com/go-logr/logr v1.4.2 // indirect
	github.com/gogo/protobuf v1.3.2 // indirect
	github.com/google/gofuzz v1.2.0 // indirect
	github.com/json-iterator/go v1.1.12 // indirect
	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
	github.com/modern-go/reflect2 v1.0.2 // indirect
	github.com/x448/float16 v0.8.4 // indirect
	golang.org/x/net v0.30.0 // indirect
	golang.org/x/text v0.19.0 // indirect
	gopkg.in/inf.v0 v0.9.1 // indirect
	k8s.io/klog/v2 v2.130.1 // indirect
	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738 // indirect
	sigs.k8s.io/json v0.0.0-20241010143419-9aa6b5e7a4b3 // indirect
	sigs.k8s.io/structured-merge-diff/v4 v4.4.2 // indirect
	sigs.k8s.io/yaml v1.4.0 // indirect
)

replace k8s.io/apimachinery => github.com/kubeedge/kubernetes/staging/src/k8s.io/apimachinery v1.32.10-kubeedge1

```

luomengY:
Okay, just remove `originalObj=updatedResoource` and eliminate redundancy.

WillardHu:
You could try rebase the latest master branch to resolve the unit test errors, and add a signature to the commit.

PR #6702, opened 2026-03-30
fix serviceaccountaccess status sync

### LINKED ISSUE #6581: ServiceAccountAccess is missing from node
**What happened**:
The serviceaccountaccess will disappear from a node randomly

serviceaccountaccess replication works

We tried this process to restore the token
1. delete the service account access
2. delete the pod that was failing due to the missing access 
3. let kubeedge recreate the service account access
4. observe that the nodeList on the serviceaccountaccess included the node that was missing
5. still observe that the pod is unable to auth because it doesn't actually have the token
6. look in edgecore.db and see that the serviceaccountaccess hasn't been replicated

**Environment**:
Client Version: v1.32.3
Kustomize Version: v5.5.0
Server Version: v1.30.2
WARNING: version difference between client (1.32) and server (1.30) exceeds the supported minor version skew of +/-1
docker.io/kubeedge/cloudcore:v1.21.0
edgecore 1.21.0 + https://github.com/kubeedge/kubeedge/issues/6465 +  https://github.com/kubeedge/kubeedge/pull/6533 cherry-picked

  - Kernel (e.g. `uname -a`):  
  - Others:

  </details>

  - OS (e.g. `cat /etc/os-release`): Fedora 42
  - Kernel (e.g. `uname -a`): Linux 6.4.11
  - Go version (e.g. `go version`): 1.24.2
  - Others:

  </details>

ShaanveerS:
**What type of PR is this?**
/kind bug

When the reconciled spec is unchanged, the controller already sends `DeleteOperation` for nodes that no longer need the access object, but it only updates `status.nodeList` when there are added nodes. That leaves stale nodes in status.
A stale `status.nodeList` can later suppress the re-insert path, so the controller thinks the edge node already has the `ServiceAccountAccess` object even though it was deleted from edge storage.

This PR updates `status.nodeList` whenever the computed node set changes, including delete-only reconciliations.

Fixes #6581

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/policycontroller/manager/reconcile.go
+++ b/cloud/pkg/policycontroller/manager/reconcile.go
@@ -440,12 +440,14 @@ func (c *Controller) syncRules(ctx context.Context, acc *policyv1alpha1.ServiceA
 	} else {
 		addNodes := subtractSlice(acc.Status.NodeList, nodes)
 		klog.V(4).Infof("serviceaccountaccess spec %s/%s is up to date", acc.Namespace, acc.Name)
-		if len(addNodes) != 0 {
+		if !equality.Semantic.DeepEqual(acc.Status.NodeList, nodes) {
 			acc.Status.NodeList = append([]string{}, nodes...)
 			if err := c.Client.Status().Update(ctx, acc); err != nil {
 				klog.Errorf("failed to update serviceaccountaccess status %s/%s, %v", acc.Namespace, acc.Name, err)
 				return controllerruntime.Result{Requeue: true}, err
 			}
+		}
+		if len(addNodes) != 0 {
 			c.send2Edge(acc, addNodes, model.InsertOperation)
 		}
 	}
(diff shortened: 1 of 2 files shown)


cloud/pkg/policycontroller/manager/reconcile.go
```diff
@@ -440,12 +440,14 @@ func (c *Controller) syncRules(ctx context.Context, acc *policyv1alpha1.ServiceA
 	} else {
 		addNodes := subtractSlice(acc.Status.NodeList, nodes)
 		klog.V(4).Infof("serviceaccountaccess spec %s/%s is up to date", acc.Namespace, acc.Name)
-		if len(addNodes) != 0 {
+		if !equality.Semantic.DeepEqual(acc.Status.NodeList, nodes) {
```
WillardHu, line 443: The else block here appears to be handle scene creation. Therefore, it shouldn't be triggered a second time, right?

ShaanveerS: iiuc, this is the reconcile path when the `ServiceAccountAccess` spec is unchanged, so it can run multiple times for the same object on pod add/delete events.
The problem is that we can send the delete here and keep `status.nodeList` stale. Then the next reconcile still sees that node in status and assumes the object is already there, so it does not send it again.

PR #6722, opened 2026-04-08
fix: keadm config-update --set report err:`Modules.Edged.: No such field: in Config`

luomengY:
**What type of PR is this?**
/kind bug

`--set` did not consider anonymous referencing of structures. for example:

type Edged struct {
......
	TailoredKubeletConfig *TailoredKubeletConfiguration `json:"tailoredKubeletConfig"`
......
}

Unified use of the helm interface for processing sets, abandoning set.go in KubeEdge.

Fixes https://github.com/kubeedge/kubeedge/issues/6721

```release-note

```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/configupdate.go
+++ b/keadm/cmd/keadm/app/cmd/edge/configupdate.go
@@ -17,16 +17,20 @@ limitations under the License.
 package edge
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
 
 	"github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2"
-	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/util"
+	"github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/validation"
+	"github.com/kubeedge/kubeedge/keadm/cmd/keadm/app/cmd/helm"
 	upgrdeedge "github.com/kubeedge/kubeedge/pkg/upgrade/edge"
+	pkgutil "github.com/kubeedge/kubeedge/pkg/util"
 	"github.com/kubeedge/kubeedge/pkg/util/execs"
 )
 
@@ -90,20 +94,21 @@ func (executor *configUpdateExecutor) configUpdate(opts ConfigUpdateOptions) err
 	if err != nil {
 		return fmt.Errorf("failed to read configfile %s, err: %v", opts.Config, err)
 	}
+	sets := strings.Split(opts.Sets, ",")
+	mergedData, err := helm.MergeSetsToBytes(data, sets)
+	if err != nil {
+		return fmt.Errorf("failed to merge sets to edgecore's config with err:%v", err)
+	}
 	edgeConfigure := &v1alpha2.EdgeCoreConfig{}
-	err = yaml.Unmarshal(data, edgeConfigure)
-	if err != nil {
-		return fmt.Errorf("failed to unmarshal configfile %s, err: %v", opts.Config, err)
+	//Check if there are any unknown fields in the set.
+	if err = yaml.UnmarshalStrict(mergedData, edgeConfigure); err != nil {
+		return err
 	}
-
-	err = util.ParseSet(edgeConfigure, opts.Sets)
-	if err != nil {
-		return fmt.Errorf("failed to parse sets value to config file: %v", err)
+	if errs := validation.ValidateEdgeCoreConfiguration(edgeConfigure); len(errs) > 0 {
+		return errors.New(pkgutil.SpliceErrors(errs.ToAggregate().Errors()))
 	}
-
-	err = edgeConfigure.WriteTo(opts.Config)
-	if err != nil {
-		return fmt.Errorf("failed to write new edgecore config: %v", err)
+	if err = writeFile(opts.Config, mergedData); err != nil {
+		return err
 	}
 
 	cmd := execs.NewCommand("sudo systemctl restart edgecore.service")
@@ -111,6 +116,15 @@ func (executor *configUpdateExecutor) configUpdate(opts ConfigUpdateOptions) err
 	if err != nil {
 		return fmt.Errorf("failed restart edgecore %v", err)
 	}
-
 	return nil
+}
+
+func writeFile(filename string, data []byte) error {
+	fileInfo, err := os.Stat(filename)
+	if err != nil {
+		// If the file does not exist, the default permissions are 0644.
+		return os.WriteFile(filename, data, 0644)
+	}
+	// Write the file using the original file permissions.
+	return os.WriteFile(filename, data, fileInfo.Mode())
 }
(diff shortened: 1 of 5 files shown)


keadm/cmd/keadm/app/cmd/util/set.go
```diff
@@ -475,44 +475,65 @@ func getIndex(field string) (int, error) {
...
-		return string(upperFirstChar) + s[1:]
+		// handling anonymous fields
+		if field.Anonymous {
+			fieldVal := obj.Field(i)
+			// if it is a pointer and nil, automatically initialize
+			if fieldVal.Kind() == reflect.Ptr && fieldVal.IsNil() {
+				if fieldVal.CanSet() {
+					newVal := reflect.New(field.Type.Elem())
+					fieldVal.Set(newVal)
+					fieldVal = newVal
```
utafrali: `findFieldPath` is supposed to be a pure read operation, but `fieldVal.Set(newVal)` mutates the actual struct when an anonymous pointer field is nil. This means simply resolving a field path can silently allocate and set a nil embedded pointer on the caller's config object — a surprising side effect.

The fix is straightforward: use the temporary value for recursion without calling `Set`:
```go
if fieldVal.Kind() == reflect.Ptr && fieldVal.IsNil() {
    newVal := reflect.New(field.Type.Elem())
    // Use newVal for traversal only, don't mutate the original struct
    subPath, err := findFieldPath(newVal, tagNames)
    if err == nil {
        return append([]string{field.Name}, subPath...), nil
    }
    continue
}
```
This handles the nil case without touching the caller's data.

keadm/cmd/keadm/app/cmd/util/set.go
```diff
@@ -475,44 +475,65 @@ func getIndex(field string) (int, error) {
...
+	tagName := tagNames[0]
+	typ := obj.Type()
+
+	for i := 0; i < typ.NumField(); i++ {
+		field := typ.Field(i)
+		jsonTag := field.Tag.Get("json")
+		tagVals := strings.Split(jsonTag, ",")
+
+		matched := tagVals[0] == tagName || strings.EqualFold(field.Name, tagName)
+
```
utafrali: The `matched` condition combines a JSON tag exact match with a case-insensitive field name match via `||`. This means if a struct has a field `Foo` with json tag `bar`, a user query for `foo` would match via `EqualFold` even though the intended JSON key is `bar`. In practice this is unlikely to cause issues with KubeEdge's config structs, but it could silently accept incorrect input. Consider prioritizing the JSON tag match and only falling back to `EqualFold` on the field name when there is no JSON tag set:
```go
var matched bool
if tagVals[0] != "" && tagVals[0] != "-" {
    matched = tagVals[0] == tagName
} else {
    matched = strings.EqualFold(field.Name, tagName)
}
```

keadm/cmd/keadm/app/cmd/util/set.go
```diff
@@ -526,21 +547,18 @@ func isFirstLetterUpper(s string) bool {
 
 func parseTag(cfg interface{}, name string) string {
```
utafrali, line 527: `parseTag` dereferences only one level of pointer before calling `findFieldPath`, which itself loops over all pointer levels. The single-level deref here is harmless since `findFieldPath` handles the rest, but it is redundant. Either remove it from `parseTag` and let `findFieldPath` own all pointer handling, or add a comment explaining the intent, so future readers don't add another single-level deref elsewhere.

keadm/cmd/keadm/app/cmd/util/set.go
```diff
@@ -475,44 +475,65 @@ func getIndex(field string) (int, error) {
...
+			subPath, err := findFieldPath(fieldVal, tagNames)
+			if err == nil {
+				// include anonymous field names to ensure complete path
+				return append([]string{field.Name}, subPath...), nil
+			}
+		}
 	}
-	return ""
+	return nil, fmt.Errorf("field %s not found", tagName)
 }
```
utafrali, line 526: No unit tests are added for the new `findFieldPath` function or the fixed `parseTag` behavior. The original bug (anonymous/embedded struct traversal) is exactly the kind of case that regresses silently. A table-driven test covering: (1) a plain json-tagged field, (2) a field accessed by struct name, (3) a field nested inside an anonymous embedded struct, and (4) a field nested inside a nil anonymous pointer struct, would prevent regressions and document the expected behavior.

utafrali (commented):
The core bug fix is correct: switching to a recursive `findFieldPath` that traverses anonymous/embedded struct fields properly addresses the reported issue. However, the anonymous-field handling introduces a side-effect bug where `findFieldPath` mutates the caller's struct by initializing nil pointer fields during what should be a read-only path resolution, and no tests are added to guard the new behavior.

WillardHu:
The ultimate goal is to modify the YAML file. Can we directly use the Helm package to modify the edgecore.yaml?

keadm/cmd/keadm/app/cmd/util/set.go
luomengY: > `findFieldPath` is supposed to be a pure read operation, but `fieldVal.Set(newVal)` mutates the actual struct when an anonymous pointer field is nil. This means simply resolving a field path can silently allocate and set a nil embedded pointer on the caller's config object — a surprising side effect.
> 
> The fix is straightforward: use the temporary value for recursion without calling `Set`:
> 
> ```go
> if fieldVal.Kind() == reflect.Ptr && fieldVal.IsNil() {
>     newVal := reflect.New(field.Type.Elem())
>     // Use newVal for traversal only, don't mutate the original struct
>     subPath, err := findFieldPath(newVal, tagNames)
>     if err == nil {
>         return append([]string{field.Name}, subPath...), nil
>     }
>     continue
> }
> ```
> 
> This handles the nil case without touching the caller's data.

In the configuration structure, some fields may be pointer types. The pointer can be nil, indicating that the field has not been initialized. When a user attempts to set a subfield inside a pointer through -- set, if the pointer is nil, they cannot directly access its internal field because nil pointers do not have addressable memory space.

luomengY:
> The ultimate goal is to modify the YAML file. Can we directly use the Helm package to modify the edgecore.yaml?

If we can reference Helm's set function, it would be the best choice. Currently, there may be many issues with the -- set function here.

WillardHu:
> If we can reference Helm's set function, it would be the best choice. Currently, there may be many issues with the -- set function here.

We could consider referencing Helm's function for implementing the `--set` flag. 

Previously, the flag `--set` of `keadm join` need modified the struct before configuration generation, so a simpler implementation was created. 

The config-update job can directly modify the YAML file, consistent with the behavior of updating values.yaml in Helm.

luomengY:
We can directly use the functions of Helm to complete it, and join can also be used.
@WillardHu

luomengY:
@WillardHu  We can review this PR.

WillardHu:
```
--- FAIL: TestMergeSetsToBytes (0.00s)
    --- FAIL: TestMergeSetsToBytes/nested_key_merge (0.00s)
        values_test.go:321: MergeSetsToBytes() gotMap = map[a:map[b:1 c:2]], want map[a:map[b:1 c:2]]
FAIL
```

luomengY:
> ```
> --- FAIL: TestMergeSetsToBytes (0.00s)
>     --- FAIL: TestMergeSetsToBytes/nested_key_merge (0.00s)
>         values_test.go:321: MergeSetsToBytes() gotMap = map[a:map[b:1 c:2]], want map[a:map[b:1 c:2]]
> FAIL
> ```

e2e tests have passed.

PR #6725, opened 2026-04-09
overridemanager: fix non-deterministic env ordering and unhandled errors in envoverrider

tushar743-ui:
**What type of PR is this?**
/kind bug

Two fixes in envoverrider.go.

First, replaceEnv was using a map to build the result slice which caused non-deterministic ordering of env vars on every reconciliation. This meant unnecessary patches were being generated and applied to edge nodes even when nothing actually changed. Also in Kubernetes, env var order matters for variable expansion so this could cause subtle breakage. Fixed by iterating the original slice to preserve order, replacing in place, and appending genuinely new vars at the end in declaration order.

Second, the error returns from `acquireAddEnvOverrideOption` and `acquireReplaceEnvOverrideOption` were being silently discarded with blank identifier. If either failed, a zero value patch would be appended and applied as a malformed JSON patch with no indication of what went wrong. Fixed by properly handling and returning the errors.

Fixes #

The ordering fix is the more impactful one. The original map iteration in replaceEnv produced a different slice order on every call which would trigger reconciliation loops on edge nodes unnecessarily.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/controllermanager/edgeapplication/overridemanager/envoverrider.go
+++ b/cloud/pkg/controllermanager/edgeapplication/overridemanager/envoverrider.go
@@ -178,8 +178,9 @@ func replaceEnv(curEnv []corev1.EnvVar, replaceValues []corev1.EnvVar) []corev1.
 	}
 
 	for _, replaceVar := range replaceValues {
-		if _, remaining := replaceMap[replaceVar.Name]; remaining {
-			newEnv = append(newEnv, replaceVar)
+		if v, remaining := replaceMap[replaceVar.Name]; remaining {
+			newEnv = append(newEnv, v)
+			delete(replaceMap, replaceVar.Name)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


cloud/pkg/controllermanager/edgeapplication/overridemanager/envoverrider.go
```diff
@@ -156,22 +162,26 @@ func overrideEnv(curEnv []corev1.EnvVar, envOverrider *v1alpha1.EnvOverrider) ([
 }
 
 func replaceEnv(curEnv []corev1.EnvVar, replaceValues []corev1.EnvVar) []corev1.EnvVar {
-	newEnv := make([]corev1.EnvVar, 0, len(curEnv))
-	currentMap := make(map[string]corev1.EnvVar)
-
-	// Populate current map with existing environment variables
-	for _, envVar := range curEnv {
-		currentMap[envVar.Name] = envVar
+	replaceMap := make(map[string]corev1.EnvVar, len(replaceValues))
```
WillardHu: In reality, there aren't many environment variables, so using a combination of `for` and `slices` would be simpler and easier to read.
```go
newEnv := make([]corev1.EnvVar, 0, len(curEnv))
copy(newEnv, curEnv)
for _, rplVal := range replaceValues {
    if idx := slices.IndexFunc(newEnv, func(it corev1.EnvVar) bool {
        return rplVal.Name == it.Name
    }); idx > -1 {
        newEnv[idx] = rplVal
    } else {
        newEnv = append(newEnv, rplVal)
    }
}
return newEnv
```

WillardHu (changes requested):

PR #6726, opened 2026-04-09
fix(overridemanager): apply image override to init containers

tushar743-ui:
**What type of PR is this?**
/kind bug

`extractPatchesBy` was only iterating over `podSpec.Containers` and completely skipping InitContainers. This means any pod with init containers would not get the image override applied to them even when the override is intended for all containers. On edge nodes that only have access to a specific registry, this would cause init containers to fail pulling images from the original registry since the override was never applied.

Fixed by iterating over InitContainers first and generating the correct patch path using initContainers index, then continuing with the regular containers as before.

Fixes #

The fix follows the same pattern already used for regular containers, just with the correct initContainers path prefix.

```release-note
NONE
```


WillardHu:
Please add unit tests for the new content.

PR #6727, opened 2026-04-10
statusmanager: fix deadlock in watchControllersGC and data race on started field

tushar743-ui:
**What type of PR is this?**
/kind bug

Two fixes in statusmanager.go.

First, `watchControllersGC` was holding s.mtx.Lock and then calling s.CancelWatch which sends to s.cancelCh. The `cancelWatchWorker` reads from s.cancelCh and calls s.unmarkWatching which also tries to acquire s.mtx.Lock. This creates a deadlock when `cancelWatchWorker` processes the message while `watchControllersGC` still holds the lock. Fixed by directly cancelling and deleting from s.watching inside the already held lock instead of going through CancelWatch.

Second, the started field was a plain bool being read and written from multiple goroutines without any synchronization. This is a data race that would be caught by the Go race detector. Fixed by replacing it with sync/atomic.Bool so reads and writes are safe across goroutines.

Fixes #

The deadlock in watchControllersGC is the more critical fix. The GC loop runs every 5 minutes and a deadlock there would freeze the entire status manager silently.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/controllermanager/edgeapplication/statusmanager/statusmanager.go
+++ b/cloud/pkg/controllermanager/edgeapplication/statusmanager/statusmanager.go
@@ -63,6 +63,8 @@ func (s *statusManager) WatchStatus(info utils.ResourceInfo) error {
 	}
 
 	select {
+	case <-s.ctx.Done():
+		return fmt.Errorf("status manager is shutting down")
 	case s.watchCh <- infoToGVK(info):
 	default:
 		return fmt.Errorf("the watchCh of status manager is full, drop the info %s", info.String())
@@ -77,6 +79,8 @@ func (s *statusManager) CancelWatch(info utils.ResourceInfo) error {
 	}
 
 	select {
+	case <-s.ctx.Done():
+		return fmt.Errorf("status manager is shutting down")
 	case s.cancelCh <- infoToGVK(info):
 	default:
 		return fmt.Errorf("the cancelCh of status manager is full, drop the info %s", info.String())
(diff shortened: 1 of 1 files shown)


WillardHu:
The `statusManager` struct seems to have some historical design flaws. The methods `watchStatusWorker`, `cancelWatchWorker`, and `waitForTerminatingWorkers` called in `Start()` are asynchronous, so the `started` field being true does not necessarily mean the startup was successful.

Adding `ctx.Done` to the select block in `WatchStatus` method seems to conflict with the started field.

You could try redesigning it to improve code readability and maintainability.

tushar743-ui:
Thanks for the feedback.

The deadlock and data race fixes are independent of the broader design issues, so I kept them scoped to those two problems. You are right that started being true does not mean the workers are fully up, that is a deeper design problem that probably deserves its own PR rather than being patched here.

The ctx.Done in WatchStatus came from a Copilot suggestion to avoid sending on a closed channel, happy to drop it if it is adding noise.

Would you be open to merging the bug fixes as-is and I can follow up with a redesign PR? @WillardHu

cloud/pkg/controllermanager/edgeapplication/statusmanager/statusmanager.go
```diff
@@ -71,11 +74,13 @@ func (s *statusManager) WatchStatus(info utils.ResourceInfo) error {
 }
 
 func (s *statusManager) CancelWatch(info utils.ResourceInfo) error {
-	if !s.started {
+	if !s.started.Load() {
 		return fmt.Errorf("status manager has not started")
 	}
 
 	select {
+	case <-s.ctx.Done():
```
WillardHu, line 82: 1. The `select{}` option uses the default path,and does not need to receive the cancellation event.
2. `s.cancelCh <- infoToGVK(info)` The scenario where the channel is closed needs to be considered.

cloud/pkg/controllermanager/edgeapplication/statusmanager/statusmanager.go
```diff
@@ -57,11 +58,13 @@ func NewStatusManager(ctx context.Context, mgr manager.Manager, client client.Cl
 }
 
 func (s *statusManager) WatchStatus(info utils.ResourceInfo) error {
-	if !s.started {
+	if !s.started.Load() {
 		return fmt.Errorf("status manager has not started")
 	}
 
 	select {
+	case <-s.ctx.Done():
```
WillardHu, line 66: 1. The `select{}` option uses the default path,and does not need to receive the cancellation event.
2. `s.cancelCh <- infoToGVK(info)` The scenario where the channel is closed needs to be considered.

cloud/pkg/controllermanager/edgeapplication/statusmanager/statusmanager.go
```diff
@@ -101,46 +108,53 @@ func (s *statusManager) SetReconcileTriggerChan(ch chan event.GenericEvent) {
...
+			s.unmarkWatching(info)
 		}
 	}
-	klog.Info("cancelWatchWorker exited")
 }
 
 func (s *statusManager) waitForTerminatingWorkers() {
 	<-s.ctx.Done()
-	close(s.watchCh)
-	close(s.cancelCh)
```
WillardHu, line 143: The logic for closing a channel should not be removed, as buffered channels may cause memory leaks. We can handle the case where the channel is closed.

WillardHu (changes requested):

tushar743-ui:
the requested changes are in the PR #6734

PR #6730, opened 2026-04-11
fix(authorization): return error from splitResource on malformed input

tushar743-ui:
**What type of PR is this?**
/kind bug

`splitResource` previously returned silently incorrect values (empty strings) on malformed or empty resource input, causing authorization decisions to be made on invalid data with no error surfaced at the split boundary. This fix adds an error return to `splitResource` and updates all call sites to handle it, returning a proper error in `getBuiltinResourceAttributes` and safely
discarding it in `isKubeedgeResourceMessage` where a bool return is expected.

Fixes #

All existing tests pass. The error log line visible in test output:
`invalid resource "": empty resource string`
is expected — it is produced by an existing test case that deliberately
passes empty input to verify the error path.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/authorization/resource_attributes.go
+++ b/cloud/pkg/cloudhub/authorization/resource_attributes.go
@@ -172,14 +172,6 @@ func splitResource(resource string) (namespace string, resourceType string, reso
 	return
 }
 
-func isKubeedgeResourceAttributes(attrs authorizer.Attributes) bool {
-	if attrs == nil || attrs.GetUser() == nil {
-		return false
-	}
-	_, ok := attrs.GetUser().GetExtra()[kubeedgeResourceKey]
-	return ok
-}
-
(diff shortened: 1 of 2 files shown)


WillardHu:
The newly added validation logic is too strict, the original logic allowed three empty strings to be returned(Or it can be any combination of an empty string and a valid string). 
Furthermore, only errors during development could cause resource field parsing failed. 
Therefore, do not modify this function without understanding its scope of impact.

You could consider defining a new function with clearly defined rules to parses resource, allowing callers to gradually replace them.

tushar743-ui:
added a new function `parseResourceStrict`   with clearly defined validation rules, and update getBuiltinResourceAttributes to use it since that is the security-sensitive call site where strict validation matters most, `splitResource` stays unchanged for `isKubeedgeResourceMessage` and any other lenient callers.

Issue #6434, opened 2025-08-05
Improving read timeout handling


PulkitDadwal:
bug #6412 has been resolved

**What happened**:
 WebSocket connections between edge nodes and the cloud server would hang indefinitely when the server failed to respond. The read timeout was not being enforced correctly, causing connections to remain open forever without any timeout mechanism triggering recovery actions.

**What you expected to happen**:
When the cloud server doesn't respond within the configured read timeout period, the WebSocket connection should timeout and trigger appropriate error recovery mechanisms (such as reconnection attempts) rather than hanging indefinitely.

**How to reproduce it (as minimally and precisely as possible)**: 
1. Set up a KubeEdge environment with edge nodes connected to cloud server via WebSocket
2. Configure a WebSocket read timeout value
3. Simulate a scenario where the cloud server becomes unresponsive (e.g., network partition, server hang, or deliberate packet dropping)
4. Observe that the edge node WebSocket connection hangs indefinitely instead of timing out after the configured duration
5. The connection remains stuck without triggering any recovery mechanisms

**Anything else we need to know?**:
This is a critical reliability issue that affects the fault tolerance of edge-cloud communication. The bug prevents proper error recovery and can lead to resource leaks. The fix ensures that read timeout deadlines are properly set and enforced, allowing edge nodes to detect unresponsive cloud servers and recover gracefully through reconnection mechanisms.

eshentials:
Can I work on this? /assign

PR #6733, opened 2026-04-14
fix: enforce WebSocket read deadline to prevent indefinite connection hang

ssam18:
The configured read deadline was never actually applied to the underlying WebSocket connection `SetReadDeadline` on `WSConnection` stored the value but didn't call through to `wsConn.SetReadDeadline`, so connections would hang forever when the cloud went silent. On top of that, when the read goroutine exited on error it left the message fifo open, meaning `Receive()` would stay blocked even after the timeout fired. This fixes both issues and mirrors the existing write deadline pattern in `Send()`. Fix for the issue #6434

### DIFF
--- a/pkg/viaduct/pkg/conn/ws.go
+++ b/pkg/viaduct/pkg/conn/ws.go
@@ -2,6 +2,7 @@ package conn
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"sync"
@@ -109,6 +110,7 @@ func (conn *WSConnection) handleMessage() {
 			}
 			conn.state.State = api.StatDisconnected
 			_ = conn.wsConn.Close()
+			conn.messageFifo.Close()
 
 			if conn.OnReadTransportErr != nil {
 				conn.OnReadTransportErr(conn.state.Headers.Get("node_id"),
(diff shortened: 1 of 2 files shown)


pkg/viaduct/pkg/conn/ws.go
```diff
@@ -151,12 +152,12 @@ func (conn *WSConnection) handleMessage() {
 
 func (conn *WSConnection) SetReadDeadline(t time.Time) error {
 	conn.ReadDeadline = t
-	return nil
+	return conn.wsConn.SetReadDeadline(t)
```
fujitatomoya, line 159: should we check the connection has been established before? the same fix should be applied to SetWriteDeadline as well, if this is taken.

```suggestion
    if conn.wsConn == nil {
        return fmt.Errorf("connection not established")
    }
	return conn.wsConn.SetReadDeadline(t)
```

fujitatomoya (approved):
lgtm with minor comment.

pkg/viaduct/pkg/conn/ws.go
ssam18: yeah, I missed it. just now fixed.

WillardHu:
Thanks for your contribution!

WillardHu:
PTAL @fujitatomoya

PR #6735, opened 2026-04-15
cloudhub: register node message pool before connect event

Sanchit2662:
/kind bug                                                                                                                                                                   

I found a race condition in how CloudHub registers node message pools when edge nodes connect. Basically, when a new edge node connects, CloudHub publishes a "node connected" event before actually registering the node's message pool in the dispatcher. During that tiny window between the event being published and the pool being registered, if any downstream messages get dispatched for that node, they end up in a silently auto-created pool that gets immediately overwritten. So those messages just disappear.      

The fix is simple: register the message pool first, then publish the connect event. That way any messages that come in during the connect flow go to the right pool. I also   hardened the `GetNodeMessagePool` function to stop silently creating throwaway pools. Now if a pool isn't found, it logs an error and returns nil instead. This makes any similar lifecycle bugs obvious instead of silently dropping messages.                                                                                                       
                                                                                                                                                             
  
  ## Special notes for your reviewer

  The changes are in two files:
  - `message_handler.go`: moved pool registration before the connect event publish, and added rollback if the connect fails
  - `message_dispatcher.go`: changed `GetNodeMessagePool` to fail loudly on miss instead of auto-creating, and updated both enqueue functions to handle nil pools             
This is a pretty small change but it fixes a real edge case that could cause workloads on edge nodes to silently get stale config after reconnects.                                                                                                                                                                                                       
 
  ```release-note
Fixed a race condition in CloudHub where edge node message pools could be silently orphaned during node reconnection, potentially causing loss of downstream messages like pod specs and config updates.

### DIFF
--- a/cloud/pkg/cloudhub/dispatcher/message_dispatcher.go
+++ b/cloud/pkg/cloudhub/dispatcher/message_dispatcher.go
@@ -215,6 +215,10 @@ func (md *messageDispatcher) PubToController(info *model.HubInfo, msg *beehivemo
 
 func (md *messageDispatcher) enqueueNoAckMessage(nodeID string, msg *beehivemodel.Message) {
 	nodeMessagePool := md.GetNodeMessagePool(nodeID)
+	if nodeMessagePool == nil {
+		klog.Errorf("drop no-ack message for node %s, no message pool registered: %s", nodeID, msg.String())
+		return
+	}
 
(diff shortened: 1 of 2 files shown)


Sanchit2662:
Hi @WillardHu , ready for review whenever you have time. Thanks!

PR #6739, opened 2026-04-17
keadm: run pre-flight checks on Kubernetes cluster health before cloudcore installation

### LINKED ISSUE #6696: Installed KubeEdge without verifying Kubernetes cluster health first
**What happened**:
I ran keadm init on a cloud node that had Kubernetes installed, but the Kubernetes cluster wasn't fully healthy. The cloudcore installation seemed to complete, but later I ran into issues with pod scheduling, API server connectivity, and certificate handling. It took me a while to realize the root cause was the underlying Kubernetes cluster being unhealthy from the start.

- keadm init to check Kubernetes cluster health before proceeding and give a clear error if something was wrong, or
- The installation to work even if the Kubernetes cluster wasn't perfect, since KubeEdge might not need a fully healthy cluster for initial setup

**How to reproduce it**:

1. Set up a Kubernetes cluster that is not fully healthy (e.g., control plane components not all running, API server unreachable, or missing CNI)
2. Verify issues: kubectl get nodes shows nodes NotReady, or kubectl get pods -n kube-system shows some control plane pods CrashLoopBackOff
3. Run keadm init --version=v1.20.0 on the cloud node
4. Observe that keadm init completes successfully
5. Later check cloudcore status: pod fails to start, or service has connectivity issues with the API server

- Cloudcore pod couldn't register itself with the API server because the API server was unreachable
- The mutating webhook for edge nodes failed to load because the API server's admission control wasn't working
- Certificates couldn't be generated properly because the cluster wasn't healthy
- I wasted time troubleshooting cloudcore when the real issue was upstream in the Kubernetes cluster

KubeEdge is tightly integrated with Kubernetes on the cloud side, so a healthy Kubernetes cluster is essential. However, keadm init doesn't perform any health checks before installation. It would be helpful if:

- keadm init included a pre-flight check that verifies:
-- API server is reachable
-- CoreDNS is running (or at least DNS resolution works)
-- At least one node is ready
-- The user has sufficient permissions (cluster-admin or equivalent) 

- The error messages clearly indicated when the problem was with the underlying Kubernetes cluster rather than with KubeEdge itself

**Environment**:
Kubernetes version (use kubectl version): v1.28.13
KubeEdge version(e.g. cloudcore --version and edgecore --version): v1.20.0
- <details><summary>Cloud nodes Environment:</summary> - Hardware configuration: Intel Xeon CPU, 4 cores - OS: Ubuntu 22.04 - Kernel: 5.15.0-generic - Go version: N/A - Kubernetes cluster status: Not healthy (control plane issues) </details>
- <details><summary>Edge nodes Environment:</summary> - edgecore version: Not configured yet - Hardware configuration: N/A - OS: N/A - Kernel: N/A - Go version: N/A </details>

ssam18:
Before this, `keadm init` would happily proceed through the entire helm installation even when the underlying Kubernetes cluster was broken, unhealthy nodes, missing CoreDNS, or insufficient RBAC permissions only surfaced as cryptic cloudcore failures later. This adds upfront checks for node readiness, CoreDNS availability, and required permissions, with a `--skip-preflight-checks` flag for cases where you knowingly want to bypass them. Fixes #6696.

### DIFF
--- a/keadm/cmd/keadm/app/cmd/cloud/init.go
+++ b/keadm/cmd/keadm/app/cmd/cloud/init.go
@@ -96,6 +96,9 @@ func addInitOtherFlags(cmd *cobra.Command, initOpts *types.InitOptions) {
 
 	cmd.Flags().StringVar(&initOpts.ImageRepository, types.FlagNameImageRepository, initOpts.ImageRepository,
 		"Choose a container image repository to pull the image of the kubedge component.")
+
+	cmd.Flags().BoolVar(&initOpts.SkipPreflightChecks, types.FlagNameSkipPreflightChecks, initOpts.SkipPreflightChecks,
+		"Skip pre-flight checks on the Kubernetes cluster health before installing cloudcore.")
 }
 
 func addHelmValueOptionsFlags(cmd *cobra.Command, initOpts *types.InitOptions) {
--- a/keadm/cmd/keadm/app/cmd/common/constant.go
+++ b/keadm/cmd/keadm/app/cmd/common/constant.go
@@ -61,7 +61,8 @@ const (
 // Cloud init flag names
 const (
 	// FlagNameSkipCRDs skip CRDs
-	FlagNameSkipCRDs = "skip-crds"
+	FlagNameSkipCRDs            = "skip-crds"
+	FlagNameSkipPreflightChecks = "skip-preflight-checks"
 
 	// FlagNameMaster sets the address of K8s master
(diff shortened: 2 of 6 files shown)


WillardHu:
Thanks for your contribution!

WillardHu:
Please add some unit-tests for preflight.go

ssam18:
> Please add some unit-tests for preflight.go



> Please add some unit-tests for preflight.go

Done

PR #6744, opened 2026-04-19
improved test coverage of jsonpatch.go in pkg

tushar743-ui:
**What type of PR is this?**
/kind test

Improves unit test coverage for the `pkg/jsonpatch` package from 82% to 100%.

Two test cases were added to cover previously untested branches:
- `TestAdd_NilValue`: covers the `value == nil` early return in `setValue()`
- `TestAdd_UnmarshalableValue`: covers the `klog.Warningf` + early return in `Add()` when `json.Marshal` fails on an unmarshallable type (e.g. `chan int`)

Fixes #

No logic changes were made. Only test coverage improvements.
The `chan int` type is used intentionally to trigger a `json.Marshal` error,
as channels have no JSON representation in Go.

```release-note
NONE
```

### DIFF
--- a/pkg/jsonpatch/jsonpatch_test.go
+++ b/pkg/jsonpatch/jsonpatch_test.go
@@ -1,11 +1,10 @@
 /*
 Copyright 2025 The KubeEdge Authors.
-
 Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


tushar743-ui:
would appreciate a review @WillardHu, whenever you are free
Thankyou

PR #6759, opened 2026-05-02
test: add 100% test coverage for pkg/util/execs

tushar743-ui:
**What type of PR is this?**
/kind test

Increases test coverage for `pkg/util/execs` from 80% to 100%.

The following changes were made:
- Refactored `exec.go` by extracting `handleWaitError()` from `Exec()` to make the non-`*exec.ExitError` branch directly testable
- Added `TestMain` helper-process pattern to precisely control subprocess exit codes and output without relying on external binaries
- Added missing tests to cover all previously uncovered branches:
  - `Start()` failure when binary does not exist at the OS level
  - `*exec.ExitError` with non-empty stderr (stderr used as error message)
  - `*exec.ExitError` with empty stderr (fallback to stdout as error message)
  - Non-`*exec.ExitError` from `Wait()` via direct call to `handleWaitError()`
  - `GetCommand()` method

Fixes #

- `NewCommand("fake_executable_name")` wraps the command in `bash -c`, so bash starts successfully and exits 127, 
it does **not** trigger the `Start()` error branch. The new `TestExec_StartError` uses `exec.Command` directly with an invalid binary path to properly trigger `Start()` failure at the OS level.
- The `else` branch in `handleWaitError()` (non-`*exec.ExitError`) is unreachable through `Exec()` under normal OS conditions since `bytes.Buffer` I/O never fails. Extracting it into `handleWaitError()` was necessary to unit test it directly.

```release-note
NONE
```

### DIFF
--- a/pkg/util/execs/exec.go
+++ b/pkg/util/execs/exec.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
-	"syscall"
 )
 
 // Command defines commands to be executed and captures std out and std error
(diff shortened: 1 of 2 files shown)


WillardHu:
--- FAIL: TestGetCommand (0.00s)
    exec_test.go:164: GetCommand() = "bash -c echo hello world", want "sh -c echo hello world"
FAIL
coverage: 100.0% of statements
FAIL	github.com/kubeedge/kubeedge/pkg/util/execs	0.017s

PR #6762, opened 2026-05-05
cloudstream: add updateNodeCloudCoreAddress to support direct apiserv…

### LINKED ISSUE #6754: Exploring Alternatives to iptableManager: Optimizing Edge-Cloud Request Forwarding via Apiserver Redirection to cloudcore

**Why is this needed**:
Currently, in KubeEdge, the `cloudstream` component utilizes `iptables` to redirect requests from the `Apiserver` intended for the edge node's `KubeletEndpoint `to port `10003` of the `cloudcore` to which that edge node is connected. The request is then routed through `cloudstream` to `edgestream`, ultimately accessing the `KubeletEndpoint` port on the `edged`. This requires the `iptableManager` to use `iptables` for interception and forwarding every 10 seconds. Since this mechanism depends on `iptables`, a proposal has been made to explore whether it's possible to have the `Apiserver` redirect to cloudcore directly, thereby replacing the `iptableManager`.

tushar743-ui:
**What type of PR is this?**
/kind feature

Adds updateNodeCloudCoreAddress to TunnelServer as groundwork for replacing the iptableManager with direct apiserver-to-cloudcore routing. Currently the iptableManager intercepts apiserver requests via iptables DNAT rules every 10 seconds and redirects them to cloudcore. This new method enables updating the node InternalIP address to point directly to CloudCore so the apiserver can route requests without iptables interception. This is the first step toward the design proposed in the iptableManager replacement issue.

Fixes #6754 

This is purely additive and does not change any existing behavior. The method is not yet called in the connect flow, that integration will follow after the design proposal is reviewed and approved.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudstream/tunnelserver.go
+++ b/cloud/pkg/cloudstream/tunnelserver.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	corev1 "k8s.io/api/core/v1"
 	"net/http"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


unchangedraman:
Hi @tushar743-ui ,
I've been digging into this approach on a local 2-VM setup (v1.22) and ran into a pretty fundamental issue with the IP mutation. Even after patching the InternalIP to point to CloudCore, the edge-side edged agent overwrites it back to the original IP every 10 seconds.

Did you test this approach and saw that agent wasn't overwriting it back to the original IP.

I think we should pause this mutation pattern rather we can use this kubelet-preferred-address-types or eBPF based direct(what calido and cilium use for pod to pod communication they just skip IP table matching for native communication).

PR #6772, opened 2026-05-08
fix(synccontroller): replace panic with error return in CompareResourceVersion

dev-aditya-hub:
## Summary

- Changes `CompareResourceVersion` in `cloud/pkg/synccontroller/objectsync.go` to return `(int, error)` instead of calling `panic(err)` on a parse failure; the comment `// coder error` misclassified a data-validation failure as a programmer mistake, ensuring no one added proper handling
- Updates `sendEvents` and `sendClusterObjectSyncEvent` to handle the error with a fail-safe: log a warning and treat an unparseable version as "unknown — send the update", so the edge stays consistent rather than silently stalling forever
- Explicitly types `compareResourceVersionFunc` in `clusterobjectsync.go` as `func(string, string) (int, error)` to cover the same panic path in the cluster-scoped reconcile loop
- Updates `TestCompareResourceVersion` in `objectsync_test.go` for the new two-value return and adds error-path cases for empty and non-numeric inputs; updates mock signatures in `clusterobjectsync_test.go` accordingly

**Why fail-safe toward delivery**

The Kubernetes API spec defines `resourceVersion` as an opaque string — the integer-string form is an implementation detail of standard `kube-apiserver` that non-standard API servers (common in IoT/edge KubeEdge deployments) are not required to follow. When the version cannot be compared, sending one extra idempotent update message is always safer than a permanent reconcile stall that silently leaves the edge node out of date with no signal to `kubectl`.

## Test plan

- [ ] `go test ./cloud/pkg/synccontroller/...` passes after this change
- [ ] `TestCompareResourceVersion` covers the new error-path cases (`""`, `"abc"`, non-numeric second argument) and all three return values exit cleanly with `err == nil` for valid inputs
- [ ] `TestReconcileClusterObjectSync` mock updated — `compareResourceVersionFunc("1000", "500")` returns `(1, nil)` as expected

### DIFF
--- a/cloud/pkg/synccontroller/clusterobjectsync.go
+++ b/cloud/pkg/synccontroller/clusterobjectsync.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2021 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package synccontroller
 
 import (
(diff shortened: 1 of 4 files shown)


dev-aditya-hub:
Hi maintainers, could someone please run /ok-to-test to trigger the presubmit CI? Thank you!

PR #6780, opened 2026-05-09
docs: add troubleshooting guide for network vs cloud-edge sync issues

### LINKED ISSUE #6701: Add guidance for distinguishing network connectivity issues from cloud-edge state synchronization issues
### What would you like to be added?

I would like a troubleshooting guide that helps users distinguish between network connectivity issues and cloud-edge state synchronization issues in KubeEdge.

It would be helpful to include:
- typical symptoms of network failures,
- typical symptoms of cloud-edge state synchronization problems,
- recommended commands and logs to check,
- a simple checklist for narrowing down the root cause.

### Why is this needed?

Many failures look similar from the user perspective. For example, workloads may not behave as expected even though the edge node appears healthy. Without guidance, it is hard to tell whether the root cause is network-related or caused by cloud-edge state inconsistency.

darshankerkar:
**What type of PR is this?**
/kind feature

Adds `docs/troubleshooting/network-vs-sync-issues.md` to help users distinguish
between network connectivity failures and cloud-edge state synchronization issues.

Addresses the exact requirements from #6701:
1. Typical symptoms of network failures vs sync problems
2. Recommended commands and log patterns to check
3. A decision tree checklist for narrowing down the root cause
4. Common fixes for each category

Fixes #6701

**Special notes for reviewer:**
1 new file added under `docs/troubleshooting/`. No existing files modified.

**Does this PR introduce a user-facing change?:**
Yes - adds a troubleshooting guide at `docs/troubleshooting/network-vs-sync-issues.md`
that helps users diagnose and resolve cloud-edge connectivity and sync failures.

### DIFF
--- a/docs/troubleshooting/network-vs-sync-issues.md
+++ b/docs/troubleshooting/network-vs-sync-issues.md
@@ -1,0 +1,139 @@
+# Troubleshooting: Network Connectivity vs Cloud-Edge State Synchronization
+
+Many KubeEdge failures present similar symptoms — workloads misbehave or edge
+nodes appear unhealthy — even though the root cause is different. This guide
+helps you quickly determine whether the problem is a **network connectivity
+failure** or a **cloud-edge state synchronization issue**.
+
+## Quick Checklist
+
+```
+1. Can the edge node reach CloudCore?        → ping / curl CloudCore IP
+2. Is EdgeHub connected?                     → check edgecore logs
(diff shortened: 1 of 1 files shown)


darshankerkar:
@gemini-code-assist good catch, added a warning that deleting edgecore.db will clear all local state and the node will need to re-sync from CloudCore after reconnect. Also fixed the broken links in the Related section. Updated in 3d9f36b, suggest more fixes if any. Thank You!

PR #6795, opened 2026-05-15
Kubernetes native service routing alt iptable

### LINKED ISSUE #6754: Exploring Alternatives to iptableManager: Optimizing Edge-Cloud Request Forwarding via Apiserver Redirection to cloudcore

tushar743-ui:
**What type of PR is this?**

/kind feature

Currently, the `cloudstream` component relies on `iptableManager` to intercept requests from the API server intended for an edge node's `KubeletEndpoint` and redirect them to port `10003` of the connected `cloudcore` instance via DNAT rules. This mechanism runs every 10 seconds, requires root-level network privileges, fails silently on eBPF-based CNIs like Cilium where iptables manipulation is restricted, and introduces a periodic reconciliation dependency that is unnecessary if the API server can be directed to cloudcore through Kubernetes-native mechanisms.

This PR introduces an iptables-free alternative by:

- Updating `node.Status.DaemonEndpoints.KubeletEndpoint.Port` to the cloudstream port (`10003)` when an edge node connects, so the API server knows to send kubelet requests to the correct port on cloudcore directly
- Creating a per-node Kubernetes `Service` and `Endpoints` object in the `kubeedge` namespace when an edge node connects, with the Endpoints pointing to `cloudCoreIP:streamPort`
- Patching `node.Status.Addresses` InternalIP to the Service ClusterIP so API server traffic routes through kube-proxy to cloudcore
- Running an informer-based Node Address Reconciler inside cloudcore that watches node objects and reconciles the InternalIP back to the Service ClusterIP whenever edgecore heartbeat overwrites it
- Adding a `DisableIptablesManager` feature flag to the CloudStream config that gates iptableManager startup, preserving full backward compatibility for existing deployments
- Cleaning up the per-node Service and Endpoints when the edge node disconnects

With these changes, `kubectl logs` and `kubectl`exec on edge pods work without any iptables DNAT rules present, verified on a kind cluster with real cloudcore and edgecore running simultaneously.

Fixes #6754 

- The `DisableIptablesManager` flag is automatically set to `true` when `cloudCoreIP` is non-empty in the advertise address config. Existing deployments with no advertise address configured will continue using iptableManager unchanged.
- The per-node Service is named e`dge-node-<nodeName>` in the `kubeedge` namespace and is cleaned up automatically on tunnel session disconnect.
- The Node Address Reconciler uses an informer `UpdateFunc` handler and only triggers `updateNodeAddressToClusterIP` when it detects that the node's InternalIP has drifted from the Service ClusterIP, avoiding unnecessary API calls.
- All new functions have unit test coverage using fake Kubernetes clients. 77 tests pass with zero lint issues introduced (in the new code added).
- The pre-existing failure in `TestNegotiateTunnelPort` exists on the main branch before this PR and is unrelated to these changes.

```release-note
Add iptables-free edge communication support in cloudstream. When cloudCoreIP is configured via advertiseAddress, cloudcore now creates a per-node Kubernetes Service and Endpoints for each connected edge node, updates the node KubeletEndpoint port to the stream port, and runs an address reconciler to maintain stable routing from the API server to cloudcore without iptables DNAT rules. The existing iptableManager behavior is preserved by default for backward compatibility and can be disabled via the DisableIptablesManager config flag.
```

Screenshot from 2026-05-10 17-41-26
Screenshot from 2026-05-12 01-42-46
Screenshot from 2026-05-12 01-42-23

### DIFF
--- a/cloud/cmd/cloudcore/app/server.go
+++ b/cloud/cmd/cloudcore/app/server.go
@@ -41,6 +41,7 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/cmd/cloudcore/app/options"
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub"
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudstream"
+	streamconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudstream/config"
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudstream/iptables"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/informers"
@@ -125,7 +126,7 @@ kubernetes controller which manages devices so that the device metadata/status d
 
 			registerModules(config)
 
-			if config.Modules.IptablesManager == nil || config.Modules.IptablesManager.Enable && config.Modules.IptablesManager.Mode == v1alpha1.InternalMode {
+			if (config.Modules.IptablesManager == nil || config.Modules.IptablesManager.Enable && config.Modules.IptablesManager.Mode == v1alpha1.InternalMode) && !streamconfig.Config.DisableIptablesManager {
 				// By default, IptablesManager manages tunnel port related iptables rules
 				// The internal mode will share the host network, forward to the stream port.
 				streamPort := int(config.Modules.CloudStream.StreamPort)
--- a/cloud/cmd/cloudcore/app/server_test.go
+++ b/cloud/cmd/cloudcore/app/server_test.go
@@ -86,3 +86,27 @@ func TestNegotiateTunnelPort(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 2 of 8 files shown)


luomengY:
Please submit the proposal design document first, instead of developing it first.

tushar743-ui:
> Please submit the proposal design document first, instead of developing it first.

Thankyou @luomengY,
On it.

PR #6798, opened 2026-05-17
docs: clarify edge pod networking behaviour

### LINKED ISSUE #6682: Clarify whether Pod networking on edge nodes is automatically available after cloud-side CNI is installed
### What would you like to be added?

I would like the documentation to clearly explain whether Pod networking on edge nodes is automatically available once a CNI plugin is installed and working on cloud-side Kubernetes nodes.

For new users, it is easy to assume that if the cloud-side CNI is already set up correctly, Pod networking on edge nodes should also work by default. However, this assumption does not seem to always hold true in KubeEdge environments.

It would be helpful to explicitly document:
- whether cloud-side CNI setup is sufficient for Pod networking on edge nodes,
- whether edge nodes have additional networking requirements,
- whether any extra KubeEdge-specific configuration is needed.

### Why is this needed?

This is a common source of confusion during initial deployment. Without explicit documentation, it is difficult to determine whether missing Pod connectivity on edge nodes is caused by incomplete CNI setup, KubeEdge-specific behavior, or some edge-side limitation.

IITI-tushar:
**What type of PR is this?**

/kind documentation

This PR improves the KubeEdge documentation by clarifying how Pod networking works on edge nodes in relation to cloud-side CNI installation.

The updated documentation explains:

- whether installing a CNI plugin on cloud-side Kubernetes nodes is sufficient for edge Pod networking,
- additional networking requirements for edge nodes,
- KubeEdge-specific considerations and limitations related to Pod connectivity.

This helps reduce confusion for new users during initial deployment and troubleshooting.

Fixes #6682 

This is a documentation-only change. No code or configuration behavior was modified.

Clarified documentation around Pod networking behavior on edge nodes and the relationship between cloud-side CNI setup and edge-side networking requirements.

### DIFF
--- a/build/edge/kubernetes/README.md
+++ b/build/edge/kubernetes/README.md
@@ -4,6 +4,10 @@ This method will guide you to deploy the edge part into a k8s cluster,
 so you need to login to the k8s master node (or where else if you can
 operate the cluster with `kubectl`).
 
+### Networking note (CNI on edge nodes)
+
+Pod networking on edge nodes is not automatically provided by the cloud-side CNI. If your edge workloads require container networking, install and configure a CNI plugin on each edge node (binaries + config). `hostNetwork` pods do not require CNI. KubeEdge does not install CNI on edge nodes by default.
+
 The manifests and scripts in `github.com/kubeedge/kubeedge/build/edge/kubernetes`
 will be used, so place these files to somewhere you can kubectl with.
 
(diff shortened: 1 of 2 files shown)


DoisLONG:
Thanks for working on this. The direction looks useful to me, especially for new users who may assume that installing a CNI plugin on the cloud-side Kubernetes nodes is enough for edge Pod networking.

A few small suggestions before maintainer review:

1. The wording suggested by Gemini looks reasonable to me. It would be better to use “binaries and configuration” instead of “binaries + config”, and keep hostNetwork formatted consistently.
2. The PR currently looks not mergeable, so it may need a rebase against the latest master.
3. For a documentation-only change, please also keep the release-note section in the standard format, usually:NONE

PR #6800, opened 2026-05-17
fix (nodetask) : correct misleading default failureTolerate log value

charantejguniganti:
## What type of PR is this?

/kind bug

The `parseFailureTolerate()` helper in `cloud/pkg/controllermanager/nodetask/` logs `use default value 1` when it cannot parse the `failureTolerate` field, while the actual runtime behavior falls back to the zero value (`0.0`), which represents strict zero tolerance (0% failure allowed).

This creates an observability issue because operators reading logs may believe that one failure is tolerated when the system will actually fail immediately upon any node failure.

This PR:

- Corrects the log message from `use default value 1` → `use default value 0`
- Adds regression tests covering malformed and boundary inputs
- Keeps runtime behavior unchanged

**Before**

```text
failed to parse failureTolerate, use default value 1
```

Logged value did not match runtime behavior and could cause debugging confusion.

**After**

```text
failed to parse failureTolerate, use default value 0
```

Log output now accurately reflects the existing runtime behavior.

## Special notes for reviewers

No runtime behavior was changed; this PR only aligns the log output with existing behavior.

Investigation confirmed that the existing runtime behavior was already correct and only the log message was inconsistent and only the log message was inconsistent. The `parseFailureTolerate()` helper is shared across node job types, so this fix ensures  consistent and accurate logging.

```release-note
Operators monitoring logs will now see an accurate default value in the
failureTolerate warning message.
```

## Testing

```bash
go test ./cloud/pkg/controllermanager/nodetask/...
```

## Test coverage added

Added regression coverage for:

- Empty string input
- Valid float input
- Boundary value `0`
- Boundary value `1`
- Non-numeric strings (`abc`)
- Malformed float strings (`1.x`)

### DIFF
--- a/cloud/pkg/controllermanager/nodetask/calculation.go
+++ b/cloud/pkg/controllermanager/nodetask/calculation.go
@@ -43,7 +43,7 @@ func CalculatePhaseWithCounts(total, proc, fail int64,
 	if failureTolerateSpec != "" {
 		parsed, err := strconv.ParseFloat(failureTolerateSpec, 64)
 		if err != nil {
-			klog.Errorf("failed to parse failureTolerate, use default value 1, err: %v", err)
+			klog.Errorf("failed to parse failureTolerate, use default value 0, err: %v", err)
 		} else {
 			failureTolerate = parsed
 		}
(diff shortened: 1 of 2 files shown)


charantejguniganti:
Investigation confirmed that runtime behavior was already correct and only the log output was inconsistent. Added regression tests for malformed and boundary inputs as well.

PR #6801, opened 2026-05-17
fix(keadm): validate required RBAC resources and verbs during preflight checks

charantejguniganti:

keadm currently verifies Kubernetes connectivity during preflight checks, but missing RBAC permissions may not be detected early in the installation flow. As a result, installation can proceed and later fail during deployment or resource creation, requiring users to troubleshoot permission issues after partial setup has already started.

This PR adds RBAC validation during the preflight stage so that required permissions are checked before installation begins.

Changes included in this PR:

- Validate required resources and verbs before installation
- Run `SelfSubjectAccessReview` checks during preflight
- Return grouped and actionable error messages for missing permissions
- Respect existing installer execution paths and existing configuration behavior
- Add regression tests covering permission validation scenarios

## Special notes for reviewers

- The implementation follows existing `keadm` installer behavior and remains aligned with the current namespace handling approach in KubeEdge.
- Modern and legacy installer paths were verified independently to ensure permission validation executes correctly for each installation flow.
- The change is intended to fail early with clear feedback while avoiding changes to existing installation behavior.

Yes.

Users with missing RBAC permissions will now receive earlier and more actionable validation errors during preflight checks instead of encountering failures later in the installation process.

## Testing

```bash
go test ./keadm/...
```

## What type of PR is this?

/kind bug

### DIFF
--- a/keadm/cmd/keadm/app/cmd/helm/cloudcore.go
+++ b/keadm/cmd/keadm/app/cmd/helm/cloudcore.go
@@ -106,7 +106,7 @@ func (c *CloudCoreHelmTool) Install(opts *types.InitOptions) error {
 	if err := c.Common.OSTypeInstaller.IsK8SComponentInstalled(c.Common.KubeConfig, c.Common.Master); err != nil {
 		return fmt.Errorf("failed to verify k8s component installed, err: %v", err)
 	}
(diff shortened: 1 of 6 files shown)


charantejguniganti:
Opening this for review. This change adds RBAC validation during preflight checks so missing permissions are detected before installation starts. Tests were added for permission validation paths and installer flow behavior.

PR #6811, opened 2026-05-19
router: preserve REST timeout status

### LINKED ISSUE #5428: ServiceBus REST returns inaccurate status code on timeout
**What happened**:

When sending a cloud-to-edge REST request via the cloudcore router, [a timeout can occur](https://github.com/kubeedge/kubeedge/blob/802f3bb202ee9705d293601cee53371299de60b7/cloud/pkg/router/provider/rest/rest.go#L174-L180) waiting to retrieve the result from the edge node. If this happens, status code of the response is set to `http. StatusRequestTimeout` (408).

I did not expect to receive a client side (4xx) status code, which would indicate that something is wrong with the client issuing the request. Rather, I'd have expected a service side (5xx) error status such as 504 (Gateway Timeout) to indicate that the issue occurred upstream of the cloudcore router.

**Environment**:
- Kubernetes version: v1.24
- KubeEdge version: v1.16.0 (cloudcore and edgecore)
  </details>

vyagh:
**What type of PR is this?**
/kind bug

REST router timeout responses now use `504 Gateway Timeout` instead of `408 Request Timeout`
The cloud router listener also preserves the final non 200 status, headers, and body after retry
  
Fixes #5428 

NONE

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/router/listener/http.go
+++ b/cloud/pkg/router/listener/http.go
@@ -3,7 +3,6 @@ package listener
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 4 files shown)


vyagh:
@WillardHu gentle reminder.

PR #6818, opened 2026-05-22
edgecontroller: keep assigned pod deletes after node cache removal

vyagh:
**What type of PR is this?**
/kind bug

Pod delete events can arrive after the edge node has already been removed from `LocationCache`. In this case `syncPod` should still send the delete to the node recorded in `Spec.NodeName`. This keeps the existing filter for add/update events and skips pod events without a node name.

None

The test covers delete after node cache removal, plus the add/update and empty-node filters.

```release-note
Fixed edgecontroller pod delete delivery after edge node cache removal.
```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/downstream.go
+++ b/cloud/pkg/edgecontroller/controller/downstream.go
@@ -62,7 +62,8 @@ func (dc *DownstreamController) syncPod() {
 				klog.Warningf("object type: %T unsupported", e.Object)
 				continue
 			}
-			if !dc.lc.IsEdgeNode(pod.Spec.NodeName) {
+			if !dc.lc.IsEdgeNode(pod.Spec.NodeName) &&
+				(e.Type != watch.Deleted || !dc.lc.IsPodOnNode(pod.UID, pod.Spec.NodeName)) {
 				continue
 			}
 			resource, err := messagelayer.BuildResource(pod.Spec.NodeName, pod.Namespace, model.ResourceTypePod, pod.Name)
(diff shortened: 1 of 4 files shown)


cloud/pkg/edgecontroller/controller/downstream.go
```diff
@@ -62,7 +62,7 @@ func (dc *DownstreamController) syncPod() {
 				klog.Warningf("object type: %T unsupported", e.Object)
 				continue
 			}
-			if !dc.lc.IsEdgeNode(pod.Spec.NodeName) {
+			if pod.Spec.NodeName == "" || (e.Type != watch.Deleted && !dc.lc.IsEdgeNode(pod.Spec.NodeName)) {
```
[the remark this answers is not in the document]
vyagh: updated to keep cloud node pod deletes filtered now

cloud/pkg/edgecontroller/controller/downstream.go
```diff
@@ -62,7 +62,11 @@ func (dc *DownstreamController) syncPod() {
 				klog.Warningf("object type: %T unsupported", e.Object)
 				continue
 			}
-			if !dc.lc.IsEdgeNode(pod.Spec.NodeName) {
+			if pod.Spec.NodeName == "" {
+				continue
+			}
+			isEdgeNode := dc.lc.IsEdgeNode(pod.Spec.NodeName)
+			if !isEdgeNode && (e.Type != watch.Deleted || !dc.lc.IsDeletedEdgeNode(pod.Spec.NodeName)) {
```
WillardHu: This logic seems unusual. If a node is deleted or becomes invalid, the Pod's NodeName should change to the new node.

vyagh: i'll take a look at this, thanks for reviewing :)

PR #6819, opened 2026-05-23
taskmanager: handle downstream tombstone deletes

vyagh:
**What type of PR is this?**
/kind bug

Kubernetes informers may send a tombstone object on delete when the final object state is not available. 
Taskmanager now unwraps that object before looking up the downstream handler, so deleting a node job still stops its running executor.

None

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/taskmanager/downstream/event_handler.go
+++ b/cloud/pkg/taskmanager/downstream/event_handler.go
@@ -79,10 +79,19 @@ func (h *NodeJobEventHandler) OnUpdate(_oldObj, newObj any) {
 // OnDelete gets the watched node job deletion event, and uses InterruptExecutor
 // method to interrupt the downstream executor.
 func (h *NodeJobEventHandler) OnDelete(obj any) {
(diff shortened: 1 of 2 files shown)


cloud/pkg/taskmanager/downstream/event_handler.go
```diff
@@ -79,10 +79,19 @@ func (h *NodeJobEventHandler) OnUpdate(_oldObj, newObj any) {
 // OnDelete gets the watched node job deletion event, and uses InterruptExecutor
 // method to interrupt the downstream executor.
 func (h *NodeJobEventHandler) OnDelete(obj any) {
+	obj = unwrapDeletedEventObj(obj)
```
WillardHu: It can be similar to https://github.com/kubeedge/kubeedge/pull/6817, without using abstract functions.

vyagh: I'll take a look at this

PR #6820, opened 2026-05-24
devicecontroller: handle informer tombstone deletes

vyagh:
**What type of PR is this?**
/kind bug

DeviceController delete events should still be handled when the informer sends a tombstone object.
This unwraps the deleted object before passing it into the existing watch event path

None

```release-note
NONE
```


Spiral-Memory (approved):
Good work

PR #6828, opened 2026-05-28
edgecore: Fix SA token handling in MetaServer

mkhon:
/kind bug

MetaServer keeps a copy of last SA token. However, when the tokens expire (and re-mounted to the container, for example), the application may have a copy of preivous SA token which is still valid because tokens are rotated at 80% expiration, but since MetaServer keeps only the latest copy, previous (still valid) SA tokens are not found and application receives Unauthorized.

Sample app log (prometheus):

```
ts=2026-05-27T04:51:33.174Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/
mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1.Endpoints: Unauthorized"        
ts=2026-05-27T04:51:33.174Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T04:51:35.078Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T04:51:35.078Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T04:51:38.564Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T04:51:38.564Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T12:57:15.445Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T12:57:15.445Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T12:57:18.253Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T12:57:18.253Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="p
kg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1.Endpoints: failed to list *v
1.Endpoints: Unauthorized"
ts=2026-05-27T12:57:22.194Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/
mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T12:57:22.194Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="p
kg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1.Endpoints: failed to list *v
1.Endpoints: Unauthorized"
ts=2026-05-27T12:57:31.864Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/
mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: failed to list *v1.Endpoints: Unauthorized"
ts=2026-05-27T12:57:31.864Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="p
kg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: Failed to watch *v1.Endpoints: failed to list *v
1.Endpoints: Unauthorized"
```

Sample edgecore log:
```
May 27 04:51:31 ip-10-100-121-202 edgecore[2399364]: E0527 04:51:31.602825 2399364 authentication.go:73] "Unab
le to authenticate the request" err="tokenData not found when authenticating"                                 
May 27 04:51:33 ip-10-100-121-202 edgecore[2399364]: E0527 04:51:33.173704 2399364 authentication.go:73] "Unab
le to authenticate the request" err="tokenData not found when authenticating"                                 
May 27 04:51:35 ip-10-100-121-202 edgecore[2399364]: E0527 04:51:35.077079 2399364 authentication.go:73] "Unab
le to authenticate the request" err="tokenData not found when authenticating"                                 
May 27 04:51:38 ip-10-100-121-202 edgecore[2399364]: E0527 04:51:38.563202 2399364 authentication.go:73] "Unab
le to authenticate the request" err="tokenData not found when authenticating"  
May 27 12:57:14 ip-10-100-121-202 edgecore[2399364]: E0527 12:57:14.138065 2399364 authentication.go:73] "Unable to authenticate the request" err="tokenData not found when authenticating"
May 27 12:57:15 ip-10-100-121-202 edgecore[2399364]: E0527 12:57:15.444101 2399364 authentication.go:73] "Unable to authenticate the request" err="tokenData not found when authenticating"
May 27 12:57:18 ip-10-100-121-202 edgecore[2399364]: E0527 12:57:18.252789 2399364 authentication.go:73] "Unable to authenticate the request" err="tokenData not found when authenticating"
May 27 12:57:22 ip-10-100-121-202 edgecore[2399364]: E0527 12:57:22.193603 2399364 authentication.go:73] "Unable to authenticate the request" err="tokenData not found when authenticating"
May 27 12:57:31 ip-10-100-121-202 edgecore[2399364]: E0527 12:57:31.863779 2399364 authentication.go:73] "Unable to authenticate the request" err="tokenData not found when authenticating"
```

```release-note
NONE
```

### DIFF
--- a/edge/pkg/metamanager/client/serviceaccount.go
+++ b/edge/pkg/metamanager/client/serviceaccount.go
@@ -87,8 +87,10 @@ func requiresRefresh(tr *authenticationv1.TokenRequest) bool {
 	return false
 }
 
-// KeyFunc keys should be nonconfidential and safe to log
-func KeyFunc(name, namespace string, tr *authenticationv1.TokenRequest) string {
+// baseKey is the spec-derived key prefix that identifies the logical token
+// (independent of which generation). All generations of the same logical token
+// share this prefix. Used for prefix lookup in metaDB.
+func baseKey(name, namespace string, tr *authenticationv1.TokenRequest) string {
 	var exp int64
 	if tr.Spec.ExpirationSeconds != nil {
 		exp = *tr.Spec.ExpirationSeconds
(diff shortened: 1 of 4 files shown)


harshrajdebug:
@mkhon I opened #7307 before I had read this properly, and the two overlap. So that a maintainer does not have to untangle them, here is what I think the relationship is.

We reached the same conclusion about the deletion independently. `getTokenLocally` calling `DeleteMetaByKey` as soon as `requiresRefresh` fires destroys the only local copy before any replacement exists, and on a disconnected node that is unrecoverable. Both PRs stop that.

**What this PR covers that mine does not.** The per-generation keys are solving a real problem I missed. My change keeps one row per logical token and lets the refreshed token overwrite it, so during the window between a refresh and the pod re-reading its projected volume, a request still carrying the previous token finds nothing in metaDB and `CheckTokenExist` rejects it. Your `KeyFunc` suffix keeps both generations alive across that window. I have no answer to that in #7307.

**What mine covers that this does not.** Two things, both in `edge/pkg/metamanager/client/serviceaccount.go`:

1. `GetServiceAccountToken` still has no fallback here. When `getTokenLocally` reports that a refresh is due, it goes to `getTokenRemotely`, and offline that fails, so the call returns an error even though a usable cached token exists. Upstream kubelet separates "no cached token", "cached token expired" and "cached token stale but usable" in `pkg/kubelet/token/token_manager.go` and serves the cached one in the third case. That third case is where an offline node always sits.

2. `CheckTokenExist` still treats presence in metaDB as proof a token is usable. `RunExpiredTokenGC` bounds how long an expired row survives, but there is a window between a token expiring and the next sweep, and in that window an expired token authenticates. It matters because `edge/pkg/metamanager/metaserver/auth/auth.go` falls back to `CheckTokenExist` as the only check when no public key is available, which is the offline case. #7307 adds an expiry test at that point, which closes the window regardless of when the GC last ran.

There is a third defect in this area that neither PR touches, over in the metamanager read path. `parseResource` returns an empty resID for service account tokens, so `processQuery` discards the key it computed one line earlier and queries cached tokens by type, and the caller then rejects any result that is not exactly one row. A node holding more than one cached token cannot read any of them back while offline. That is #7306.

On sequencing, I would rather this landed than raced. This PR is the older and the larger, and it answers the propagation window. If you are still working on it, I am happy to rebase #7307 down to the fallback and the `CheckTokenExist` check and drop the part that duplicates yours. The `CheckTokenExist` change is independent and would apply as is. The fallback needs reshaping against your version, since your `getTokenLocally` returns an error when a refresh is due where mine returns the cached token and leaves the decision to the caller, so the three way branch would have to move or take its input differently. If you have moved on, say so and I can carry the per-generation keys across with attribution.

Either way it would help to know whether you are still on this, since it has been open since May.

mkhon:
@harshrajdebug you can copy missing functionality from this PR to yours. Absolutely no problems with this as soon as this gets landed into main/master somehow

harshrajdebug:
Thanks @mkhon. I have taken you up on it and #7307 now carries your commit, cherry-picked so it keeps your authorship, with nothing copied in under mine. Reviewing that half of #7307 is reviewing your work.

What I added on top is the fallback in `GetServiceAccountToken` and the expiry test in `CheckTokenExist`. Combining them changed my half: I split `newestCachedGeneration` out of your `getTokenLocally` so the caller can see the cached token and decide, and since your loop already drops expired generations, anything it returns is un-expired by construction. The three way branch I described earlier collapses to two, which is simpler than what I had. `getTokenLocally` keeps its signature, error strings and behaviour, so your tests still describe it accurately.

Your per-generation keys fixed the propagation window my version had wrong, and there is now a test for it. `TestGenerationsCoexistAcrossRefresh` writes two generations to a real metaDB through `dao.Init`, asserts both rows survive a refresh, asserts both still authenticate through `CheckTokenExist`, and asserts the newest is the one handed back.

One thing worth passing on, since it concerns your test and leaves your code untouched. `TestGetTokenLocally` fails for me, and it fails the same way on this branch unmodified, so it is not something I introduced. The cause is that gomonkey does not re-patch a method within a single test function on darwin/arm64 with Go 1.26. I checked it with a probe: after `patches.Reset()`, and even with a fresh `gomonkey.NewPatches()`, a second `ApplyMethodFunc` on the same method still returned the first stub. Your test patches `QueryMetaByKeyPrefix` several times in one function, so every case after the DB error one reads that first stub, and the assertion for "no cached token" sees "db error" instead.

It very likely passes on the Linux runners, so this may never have shown up for you. If you want it to hold on arm64 as well, splitting those cases into separate test functions is enough, since patching once per function works fine. I have deliberately left your test as it is, since changing it inside my PR would be presumptuous.

If any of this looks wrong to you, say so and I will change it. The aim is that the fix lands, which is what you asked for.

PR #6838, opened 2026-06-03
added proposal documentation for iptableManager alternative

tushar743-ui:
**What type of PR is this?**

/kind documentation

This PR adds a design proposal for Alternatives to iptableManager via Apiserver Redirection (LFX 2026 Term 2)

Fixes #

This PR adds a Proposal design documentation for LFX term 2 mentorship program

```release-note

```


tushar743-ui:
Would appreciate a review, @luomengY, whenever you're free.

luomengY:
When `CloudCore` is deployed with high availability, multiple `CloudCore` instances exist, each potentially managing a portion of edge nodes. Has your solution already considered the compatibility of the high-availability `CloudCore` deployment scheme?

luomengY:
You can refer to https://github.com/kubeedge/kubeedge/blob/master/docs/proposals/sig-scalability/ha_router.md By obtaining the managed `cloudcore` from the annotation of the edge node: `cloudcore`, it can solve the problem of correctly routing the deployed `Apiserver` to the `cloudcore` connected to the edge node when the `cloudcore` is highly available.

luomengY:
Directly modifying the `status.daemonEndpoints.kubeletEndpoint.Port` and `status.addresses.InternalIP` of edge nodes here may pose other risks. If these property values are used by others, it may cause unexpected errors. Have you considered this?

tushar743-ui:
> When `CloudCore` is deployed with high availability, multiple `CloudCore` instances exist, each potentially managing a portion of edge nodes. Has your solution already considered the compatibility of the high-availability `CloudCore` deployment scheme?

 After reading ha_router.md, I have updated the proposal to adopt the existing `cloudcore: <cloudCoreIP>` annotation convention. A new `setCloudCoreAnnotation` function sets this annotation when an edge node connects. The Node Address Reconciler checks it before acting, skipping nodes managed by a different CloudCore instance. `ensureNodeEndpoints` uses the annotation value as the Endpoint IP. On failover the reconnecting CloudCore 
updates both the annotation and Endpoints to point to itself. See section 6.5 in the updated proposal.

tushar743-ui:
> Directly modifying the `status.daemonEndpoints.kubeletEndpoint.Port` and `status.addresses.InternalIP` of edge nodes here may pose other risks. If these property values are used by others, it may cause unexpected errors. Have you considered this?

Both fields are already exclusively managed by KubeEdge in edge node deployments, edgecore overwrites InternalIP on every heartbeat and iptableManager already depends on KubeletEndpoint.Port being set to a specific value. This proposal substitutes one KubeEdge-managed value with another within the same trust boundary. Additionally the feature only activates when `advertiseAddress` is explicitly configured via the 
`DisableIptablesManager` flag, leaving all other deployments untouched. See section 6.6 in the updated proposal for the full risk assessment.

docs/proposals/sig-networking/iptables-free-apiserver-redirection-to-cloudcore.md
```diff
@@ -0,0 +1,421 @@
...
+
+In HA CloudCore deployments multiple CloudCore instances run simultaneously,
+each managing a subset of edge nodes via load balancing (e.g. HAProxy).
+
+This solution adopts the annotation convention established in the router HA
+design (`docs/proposals/sig-scalability/ha_router.md`):
+
+**Annotation:** `cloudcore: <cloudCoreIP>`
+
+This annotation is not currently set in the cloudstream code path. This
```
luomengY: When `EdgeCore` connects to `CloudCore` via `CloudHub`, this annotation will be updated; therefore, `CloudStream` does not need to check the annotation again.

docs/proposals/sig-networking/iptables-free-apiserver-redirection-to-cloudcore.md
```diff
@@ -0,0 +1,421 @@
...
+metadata:
+  name: edge-node-<nodeName>
+  namespace: kubeedge
+subsets:
+  - addresses:
+      - ip: <cloudCoreIP>
+    ports:
+      - port: 10003
+        protocol: TCP
+```
```
luomengY, line 247: Additionally, you are creating a `Service` and `Endpoints` for each edge node here. I don’t think this approach is ideal, because in edge scenarios, there can be tens of thousands of edge nodes—this would result in a massive number of Services. Why is it necessary to use Services and Endpoints here? In your design, since the apiserver can directly obtain the IP and port of the `CloudCore` to connect to, could we avoid using `Service` and `Endpoints` altogether? Are there any other considerations behind this design choice?

tushar743-ui: > Additionally, you are creating a `Service` and `Endpoints` for each edge node here. I don’t think this approach is ideal, because in edge scenarios, there can be tens of thousands of edge nodes—this would result in a massive number of Services. Why is it necessary to use Services and Endpoints here? In your design, since the apiserver can directly obtain the IP and port of the `CloudCore` to connect to, could we avoid using `Service` and `Endpoints` altogether? Are there any other considerations behind this design choice?

You are right. Creating a Service and Endpoints per edge node does not scale. I have redesigned the proposal to set InternalIP directly to cloudCoreIP, read from the existing cloudcore annotation. The API server connects directly to cloudCoreIP:10003 without any Service or kube-proxy involvement. Zero additional Kubernetes objects are created per edge node. The Node Address Reconciler corrects heartbeat drift using the same annotation as the HA guard. Proposal updated accordingly. And also removed setCloudCoreAnnotation from the proposal entirely. The reconciler will read the existing annotation directly.

luomengY:
In this Thursday's KubeEdge community call, we discussed this proposal. Directly modifying `status.addresses.InternalIP` may not be the best approach, as it would result in the node's IP being a pseudo-IP. For example, running `kubectl get nodes -o wide -A` would show the edge node's `INTERNAL-IP` as the `CloudCore` IP. The main issue with the current iptables-based routing approach is that it creates a separate chain for every edge node, leading to a proliferation of iptables chains as edge nodes scale up. You could look into whether it's possible to handle forwarding for all edge nodes using a single iptables chain, or explore alternatives such as eBPF or other solutions.

tushar743-ui:
> In this Thursday's KubeEdge community call, we discussed this proposal. Directly modifying `status.addresses.InternalIP` may not be the best approach, as it would result in the node's IP being a pseudo-IP. For example, running `kubectl get nodes -o wide -A` would show the edge node's `INTERNAL-IP` as the `CloudCore` IP. The main issue with the current iptables-based routing approach is that it creates a separate chain for every edge node, leading to a proliferation of iptables chains as edge nodes scale up. You could look into whether it's possible to handle forwarding for all edge nodes using a single iptables chain, or explore alternatives such as eBPF or other solutions.

The pseudo-IP concern is addressed as follows:

kubectl get nodes -o wide will show cloudCoreIP as INTERNAL-IP because that is the actual routing target the API server uses. This is intentional and correct — it reflects where kubectl logs and kubectl exec traffic goes.

The real edge node IP is preserved and accessible via:

1. Annotation:
   kubectl get node <name> \
     -o jsonpath='{.metadata.annotations.node\.kubeedge\.io/internal-ip}'

2. New keadm ctl get node command showing both columns:
   NAME     REAL-IP    CLOUD-IP       STATUS
   ubuntu   10.0.3.5   192.168.1.100  NotReady

Setting ExternalIP = cloudCoreIP was investigated as a way to keep INTERNAL-IP as the real edge IP while routing via EXTERNAL-IP. This requires --kubelet-preferred-address-types=ExternalIP,InternalIP,Hostname on kube-apiserver which is not possible on managed Kubernetes clusters (EKS, GKE, AKS). Since KubeEdge is commonly deployed on managed Kubernetes this was rejected.

If the community considers the INTERNAL-IP column showing cloudCoreIP unacceptable, I welcome guidance on the preferred approach. Our current implementation preserves the real IP and provides operator tooling to access it without requiring any cluster configuration changes.

luomengY:
> > In this Thursday's KubeEdge community call, we discussed this proposal. Directly modifying `status.addresses.InternalIP` may not be the best approach, as it would result in the node's IP being a pseudo-IP. For example, running `kubectl get nodes -o wide -A` would show the edge node's `INTERNAL-IP` as the `CloudCore` IP. The main issue with the current iptables-based routing approach is that it creates a separate chain for every edge node, leading to a proliferation of iptables chains as edge nodes scale up. You could look into whether it's possible to handle forwarding for all edge nodes using a single iptables chain, or explore alternatives such as eBPF or other solutions.
> 
> The pseudo-IP concern is addressed as follows:
> 
> kubectl get nodes -o wide will show cloudCoreIP as INTERNAL-IP because that is the actual routing target the API server uses. This is intentional and correct — it reflects where kubectl logs and kubectl exec traffic goes.
> 
> The real edge node IP is preserved and accessible via:
> 
> 1. Annotation:
>    kubectl get node  
>    -o jsonpath='{.metadata.annotations.node.kubeedge.io/internal-ip}'
> 2. New keadm ctl get node command showing both columns:
>    NAME     REAL-IP    CLOUD-IP       STATUS
>    ubuntu   10.0.3.5   192.168.1.100  NotReady
> 
> Setting ExternalIP = cloudCoreIP was investigated as a way to keep INTERNAL-IP as the real edge IP while routing via EXTERNAL-IP. This requires --kubelet-preferred-address-types=ExternalIP,InternalIP,Hostname on kube-apiserver which is not possible on managed Kubernetes clusters (EKS, GKE, AKS). Since KubeEdge is commonly deployed on managed Kubernetes this was rejected.
> 
> If the community considers the INTERNAL-IP column showing cloudCoreIP unacceptable, I welcome guidance on the preferred approach. Our current implementation preserves the real IP and provides operator tooling to access it without requiring any cluster configuration changes.

I previously considered adding `EdgeTunnelIP` to the `kubelet-preferred-address-types` in the `API-Server`, and even thought about adding `EdgeTunnelIP` on edge nodes, for example: `--kubelet-preferred-address-types=EdgeTunnelIP,Hostname,InternalIP,ExternalIP`. However, this approach requires modifying the `API-Server`, which is also an intrusive modification. I shared this idea at the community meeting last Thursday, but it might still not be quite appropriate.

tushar743-ui:
> > > In this Thursday's KubeEdge community call, we discussed this proposal. Directly modifying `status.addresses.InternalIP` may not be the best approach, as it would result in the node's IP being a pseudo-IP. For example, running `kubectl get nodes -o wide -A` would show the edge node's `INTERNAL-IP` as the `CloudCore` IP. The main issue with the current iptables-based routing approach is that it creates a separate chain for every edge node, leading to a proliferation of iptables chains as edge nodes scale up. You could look into whether it's possible to handle forwarding for all edge nodes using a single iptables chain, or explore alternatives such as eBPF or other solutions.
> > 
> > 
> > The pseudo-IP concern is addressed as follows:
> > kubectl get nodes -o wide will show cloudCoreIP as INTERNAL-IP because that is the actual routing target the API server uses. This is intentional and correct — it reflects where kubectl logs and kubectl exec traffic goes.
> > The real edge node IP is preserved and accessible via:
> > 
> > 1. Annotation:
> >    kubectl get node
> >    -o jsonpath='{.metadata.annotations.node.kubeedge.io/internal-ip}'
> > 2. New keadm ctl get node command showing both columns:
> >    NAME     REAL-IP    CLOUD-IP       STATUS
> >    ubuntu   10.0.3.5   192.168.1.100  NotReady
> > 
> > Setting ExternalIP = cloudCoreIP was investigated as a way to keep INTERNAL-IP as the real edge IP while routing via EXTERNAL-IP. This requires --kubelet-preferred-address-types=ExternalIP,InternalIP,Hostname on kube-apiserver which is not possible on managed Kubernetes clusters (EKS, GKE, AKS). Since KubeEdge is commonly deployed on managed Kubernetes this was rejected.
> > If the community considers the INTERNAL-IP column showing cloudCoreIP unacceptable, I welcome guidance on the preferred approach. Our current implementation preserves the real IP and provides operator tooling to access it without requiring any cluster configuration changes.
> 
> I previously considered adding `EdgeTunnelIP` to the `kubelet-preferred-address-types` in the `API-Server`, and even thought about adding `EdgeTunnelIP` on edge nodes, for example: `--kubelet-preferred-address-types=EdgeTunnelIP,Hostname,InternalIP,ExternalIP`. However, this approach requires modifying the `API-Server`, which is also an intrusive modification. I shared this idea at the community meeting last Thursday, but it might still not be quite appropriate.

I also investigated the eBPF approach as a potential solution to the pseudo-IP concern. A TC eBPF program attached to the cloudcore host network interface could intercept packets destined for any edge node IP on port 10003 and rewrite the destination to cloudCoreIP:10003 at the kernel level, keeping InternalIP as the real edge node IP with no
pseudo-IP at all.

The concerns with this approach:

1. Implementation complexity -> requires writing a TC eBPF C program, compiling to BPF bytecode, loading via Go (cilium/ebpf library), attaching to network interfaces, and managing lifecycle across cloudcore restarts. Zero existing eBPF infrastructure in the KubeEdge codebase to build on.

2. Kernel version dependency -> TC eBPF requires kernel 4.14+. While modern edge hardware (Raspberry Pi 4, recent industrial controllers) meets this, older deployed edge devices may not. This creates an uneven deployment story.

3. CNI interaction -> on Cilium clusters where eBPF is already managing the dataplane, attaching additional TC programs requires careful coordination to avoid conflicts.

Should we explore this further or focus on the current InternalIP approach with keadm ctl get node for operator visibility?

luomengY:
Indeed, I have reviewed the iptablesmanager code and confirmed that iptables here only handles port forwarding. The iptables rules do not increase with the number of edge nodes; the number of rules equals the number of cloudcore replicas, and the tables and chains are also unrelated to the number of edge nodes. So the concern we discussed last week no longer exists. If adopting eBPF would introduce too many modules, we should continue to focus on properly handling `INTERNAL-IP`.

tushar743-ui:
> Indeed, I have reviewed the iptablesmanager code and confirmed that iptables here only handles port forwarding. The iptables rules do not increase with the number of edge nodes; the number of rules equals the number of cloudcore replicas, and the tables and chains are also unrelated to the number of edge nodes. So the concern we discussed last week no longer exists. If adopting eBPF would introduce too many modules, we should continue to focus on properly handling `INTERNAL-IP`.

Based on your confirmation that iptables rules do not scale with edge node count, I believe the cleanest solution is to:

1. Keep iptableManager as-is for routing
2. Set KubeletEndpoint.Port = negotiated tunnelPort (e.g. 10351) instead of streamPort (10003)
3. Keep INTERNAL-IP as the real edge node IP — no pseudo-IP
4. API server sends to realEdgeIP:10351, iptableManager DNAT on the cloudcore/apiserver node redirects to cloudCoreIP:10003

This requires minimal changes, preserves real INTERNAL-IP, and works with the existing iptableManager architecture exactly as designed.

The only remaining concern is internal iptableManager mode where DNAT rules only exist on cloudcore's node. If cloudcore and API server run on different nodes in internal mode, the redirect does not fire. External DaemonSet mode (the default in Helm) handles this correctly since it runs on all non-edge nodes.

Does this direction align with your thinking?Then please let me know we will implement and test it.

luomengY:
> > Indeed, I have reviewed the iptablesmanager code and confirmed that iptables here only handles port forwarding. The iptables rules do not increase with the number of edge nodes; the number of rules equals the number of cloudcore replicas, and the tables and chains are also unrelated to the number of edge nodes. So the concern we discussed last week no longer exists. If adopting eBPF would introduce too many modules, we should continue to focus on properly handling `INTERNAL-IP`.
> 
> Based on your confirmation that iptables rules do not scale with edge node count, I believe the cleanest solution is to:
> 
> 1. Keep iptableManager as-is for routing
> 2. Set KubeletEndpoint.Port = negotiated tunnelPort (e.g. 10351) instead of streamPort (10003)
> 3. Keep INTERNAL-IP as the real edge node IP — no pseudo-IP
> 4. API server sends to realEdgeIP:10351, iptableManager DNAT on the cloudcore/apiserver node redirects to cloudCoreIP:10003
> 
> This requires minimal changes, preserves real INTERNAL-IP, and works with the existing iptableManager architecture exactly as designed.
> 
> The only remaining concern is internal iptableManager mode where DNAT rules only exist on cloudcore's node. If cloudcore and API server run on different nodes in internal mode, the redirect does not fire. External DaemonSet mode (the default in Helm) handles this correctly since it runs on all non-edge nodes.
> 
> Does this direction align with your thinking?Then please let me know we will implement and test it.

The points 1 to 4 you mentioned have already been implemented in the current iptablesManager and are also applicable to the high‑availability CloudCore deployment scenario, so they should not need further modification. However, regarding the issue you raised that when CloudCore and the API Server are not on the same node, traffic cannot be intercepted by iptables and forwarded to CloudCore — could you look into possible solutions for this problem?

tushar743-ui:
> > > Indeed, I have reviewed the iptablesmanager code and confirmed that iptables here only handles port forwarding. The iptables rules do not increase with the number of edge nodes; the number of rules equals the number of cloudcore replicas, and the tables and chains are also unrelated to the number of edge nodes. So the concern we discussed last week no longer exists. If adopting eBPF would introduce too many modules, we should continue to focus on properly handling `INTERNAL-IP`.
> > 
> > 
> > Based on your confirmation that iptables rules do not scale with edge node count, I believe the cleanest solution is to:
> > 
> > 1. Keep iptableManager as-is for routing
> > 2. Set KubeletEndpoint.Port = negotiated tunnelPort (e.g. 10351) instead of streamPort (10003)
> > 3. Keep INTERNAL-IP as the real edge node IP — no pseudo-IP
> > 4. API server sends to realEdgeIP:10351, iptableManager DNAT on the cloudcore/apiserver node redirects to cloudCoreIP:10003
> > 
> > This requires minimal changes, preserves real INTERNAL-IP, and works with the existing iptableManager architecture exactly as designed.
> > The only remaining concern is internal iptableManager mode where DNAT rules only exist on cloudcore's node. If cloudcore and API server run on different nodes in internal mode, the redirect does not fire. External DaemonSet mode (the default in Helm) handles this correctly since it runs on all non-edge nodes.
> > Does this direction align with your thinking?Then please let me know we will implement and test it.
> 
> The points 1 to 4 you mentioned have already been implemented in the current iptablesManager and are also applicable to the high‑availability CloudCore deployment scenario, so they should not need further modification. However, regarding the issue you raised that when CloudCore and the API Server are not on the same node, traffic cannot be intercepted by iptables and forwarded to CloudCore — could you look into possible solutions for this problem?

Thank you for the clarification. 

For internal mode with cloudcore and API server on different nodes, I propose an auto-detection fallback: when cloudcore runs in internal iptableManager mode and advertiseAddress is configured, automatically enable direct routing (set KubeletEndpoint.Port = streamPort, InternalIP = cloudCoreIP) so the API server connects to cloudCoreIP:10003 directly without requiring any iptables interception. This handles the separated-node scenario gracefully without forcing operators to switch deployment modes.

In summary:
- External mode (Helm default): iptableManager on all nodes, works correctly
- Internal mode + co-located: iptableManager on cloudcore node, works correctly  
- Internal mode + separated + advertiseAddress set: auto-fallback to direct 
  routing, works correctly
- Internal mode + separated + no advertiseAddress: document as unsupported, 
  operator must use external mode or set advertiseAddress

tushar743-ui:
The previous approach had a few production gaps: Mode A had no fallback when CloudCore's tunnelPort changed after a restart, internal mode and external mode were treated as interchangeable when internal mode was never designed for separated nodes, and a misconfigured separated-node deployment would fail silently with no warning.

This is a redesign to close those gaps:

- Scope internal mode strictly to co-located CloudCore + API server, matching its original host-network design intent.
- Rely on external mode (Helm default) for separated-node deployments, since its DaemonSet already runs on every non-edge node.
- Keep default routing as KubeletEndpoint.Port = tunnelPort with InternalIP untouched -> real edge IP, no pseudo-IP.
- Reserve InternalIP = cloudCoreIP strictly for Cilium/nftables-incompatible environments, activated explicitly via advertiseAddress.
- Add a startup warning when internal mode runs without hostNetwork, so separated-node misconfiguration is never silent.
- Add a narrow reconciler that only corrects tunnelPort drift after CloudCore's IP changes, with no general InternalIP reconciliation needed in default mode.

Please let me know your thoughts on this.

luomengY:
Here, only when `iptablesmanager` is deployed in `Internal` mode and `cloudcore` and `apiserver` are not running on the same node, the traffic from `apiserver` cannot be forwarded to `cloudcore` via `iptablesmanager`. For this situation, the solution you adopted is still to use `InternalIP = cloudCoreIP`, but this approach was considered inappropriate in community meetings. Please consider whether there are any other solutions.

tushar743-ui:
> Here, only when `iptablesmanager` is deployed in `Internal` mode and `cloudcore` and `apiserver` are not running on the same node, the traffic from `apiserver` cannot be forwarded to `cloudcore` via `iptablesmanager`. For this situation, the solution you adopted is still to use `InternalIP = cloudCoreIP`, but this approach was considered inappropriate in community meetings. Please consider whether there are any other solutions.


**Approach**:
For the internal mode + separated-node scenario, I propose using the existing cloudcore Kubernetes Service ClusterIP. The core issue is that in internal mode, DNAT rules only exist on cloudcore's node, so when cloudcore and the API server run on different nodes the redirect never fires. Using the cloudcore Service ClusterIP solves this because kube-proxy already maintains routing rules for the Service on every node in the cluster including the API server's node, so no cloudcore- managed iptables rules are needed at all.

- KubeletEndpoint.Port = 10003 (already exposed on cloudcore Service)
- InternalIP = cloudcore Service ClusterIP (e.g. 10.96.108.249)

API server sends to ClusterIP:10003. kube-proxy routes to cloudcore without any iptables manipulation by cloudcore, works regardless of which node cloudcore runs on, survives restarts and rescheduling. Zero new Kubernetes objects needed.

For HA, when kube-proxy routes to a cloudcore instance that does not own the tunnel session, cloudcore reads the cloudcore annotation (constants.EdgeMappingCloudKey) and forwards to the correct instance, following the pattern in ha_router.md. This inter-instance forwarding is not yet implemented in cloudstream and would be part of this
proposal's scope.

**Remaining concern**:
 InternalIP still shows the Service ClusterIP rather than the real edge node IP in kubectl get nodes -o wide. The ClusterIP is stable and Kubernetes-native unlike a pod IP, but it is still not the physical edge device IP. I consider this more defensible than setting InternalIP to cloudCoreIP directly, but wanted to flag it explicitly -  is a stable virtual ClusterIP acceptable here, or does the community require the real edge IP to remain untouched in InternalIP?

luomengY:
> > Here, only when `iptablesmanager` is deployed in `Internal` mode and `cloudcore` and `apiserver` are not running on the same node, the traffic from `apiserver` cannot be forwarded to `cloudcore` via `iptablesmanager`. For this situation, the solution you adopted is still to use `InternalIP = cloudCoreIP`, but this approach was considered inappropriate in community meetings. Please consider whether there are any other solutions.
> 
> **Approach**: For the internal mode + separated-node scenario, I propose using the existing cloudcore Kubernetes Service ClusterIP. The core issue is that in internal mode, DNAT rules only exist on cloudcore's node, so when cloudcore and the API server run on different nodes the redirect never fires. Using the cloudcore Service ClusterIP solves this because kube-proxy already maintains routing rules for the Service on every node in the cluster including the API server's node, so no cloudcore- managed iptables rules are needed at all.
> 
> * KubeletEndpoint.Port = 10003 (already exposed on cloudcore Service)
> * InternalIP = cloudcore Service ClusterIP (e.g. 10.96.108.249)
> 
> API server sends to ClusterIP:10003. kube-proxy routes to cloudcore without any iptables manipulation by cloudcore, works regardless of which node cloudcore runs on, survives restarts and rescheduling. Zero new Kubernetes objects needed.
> 
> For HA, when kube-proxy routes to a cloudcore instance that does not own the tunnel session, cloudcore reads the cloudcore annotation (constants.EdgeMappingCloudKey) and forwards to the correct instance, following the pattern in ha_router.md. This inter-instance forwarding is not yet implemented in cloudstream and would be part of this proposal's scope.
> 
> **Remaining concern**: InternalIP still shows the Service ClusterIP rather than the real edge node IP in kubectl get nodes -o wide. The ClusterIP is stable and Kubernetes-native unlike a pod IP, but it is still not the physical edge device IP. I consider this more defensible than setting InternalIP to cloudCoreIP directly, but wanted to flag it explicitly - is a stable virtual ClusterIP acceptable here, or does the community require the real edge IP to remain untouched in InternalIP?

Adding `ClusterIP` in this way is still a pseudo IP, so there is no need to add a `Service`. The issue here is not to provide a stable IP for the `InternalIP`, but rather that the `InternalIP` of edge nodes should preferably not be modified.

tushar743-ui:
> > > Here, only when `iptablesmanager` is deployed in `Internal` mode and `cloudcore` and `apiserver` are not running on the same node, the traffic from `apiserver` cannot be forwarded to `cloudcore` via `iptablesmanager`. For this situation, the solution you adopted is still to use `InternalIP = cloudCoreIP`, but this approach was considered inappropriate in community meetings. Please consider whether there are any other solutions.
> > 
> > 
> > **Approach**: For the internal mode + separated-node scenario, I propose using the existing cloudcore Kubernetes Service ClusterIP. The core issue is that in internal mode, DNAT rules only exist on cloudcore's node, so when cloudcore and the API server run on different nodes the redirect never fires. Using the cloudcore Service ClusterIP solves this because kube-proxy already maintains routing rules for the Service on every node in the cluster including the API server's node, so no cloudcore- managed iptables rules are needed at all.
> > 
> > * KubeletEndpoint.Port = 10003 (already exposed on cloudcore Service)
> > * InternalIP = cloudcore Service ClusterIP (e.g. 10.96.108.249)
> > 
> > API server sends to ClusterIP:10003. kube-proxy routes to cloudcore without any iptables manipulation by cloudcore, works regardless of which node cloudcore runs on, survives restarts and rescheduling. Zero new Kubernetes objects needed.
> > For HA, when kube-proxy routes to a cloudcore instance that does not own the tunnel session, cloudcore reads the cloudcore annotation (constants.EdgeMappingCloudKey) and forwards to the correct instance, following the pattern in ha_router.md. This inter-instance forwarding is not yet implemented in cloudstream and would be part of this proposal's scope.
> > **Remaining concern**: InternalIP still shows the Service ClusterIP rather than the real edge node IP in kubectl get nodes -o wide. The ClusterIP is stable and Kubernetes-native unlike a pod IP, but it is still not the physical edge device IP. I consider this more defensible than setting InternalIP to cloudCoreIP directly, but wanted to flag it explicitly - is a stable virtual ClusterIP acceptable here, or does the community require the real edge IP to remain untouched in InternalIP?
> 
> Adding `ClusterIP` in this way is still a pseudo IP, so there is no need to add a `Service`. The issue here is not to provide a stable IP for the `InternalIP`, but rather that the `InternalIP` of edge nodes should preferably not be modified.

Ok if InternalIP should not be modified at all. Given this constraint, for internal mode with cloudcore and API server on different nodes, I do not see a clean solution that avoids both InternalIP modification and kube-apiserver flag changes.

My conclusion is that internal mode should be explicitly documented as requiring cloudcore and API server to be co-located (matching its original host-network design intent), and external DaemonSet mode (the Helm default) should be the recommended deployment for all production multi-node clusters.

For the standard case if I set KubeletEndpoint.Port = tunnelPort and leave InternalIP completely untouched. This works correctly in external mode and co-located internal mode which covers all supported topologies.

Is documenting internal mode + separated nodes as unsupported acceptable to the community, or is there a specific approach I have not considered?

tushar743-ui:
My recommendation:

For the default case (external DaemonSet mode, Helm default): Set KubeletEndpoint.Port = tunnelPort, InternalIP untouched.
iptableManager DaemonSet on every node handles routing correctly. This covers the vast majority of production deployments.

For internal mode with separated nodes: document as requiring external mode or co-location. Internal mode was designed for co-located deployments per its original hostNetwork intent.

The EdgeTunnelIP custom address type idea from the community meeting is architecturally the cleanest long-term solution as it eliminates the InternalIP conflict entirely. I suggest pursuing that upstream in Kubernetes if the community agrees.

Please let me know further directions.

luomengY:
> My recommendation:
> 
> For the default case (external DaemonSet mode, Helm default): Set KubeletEndpoint.Port = tunnelPort, InternalIP untouched. iptableManager DaemonSet on every node handles routing correctly. This covers the vast majority of production deployments.
> 
> For internal mode with separated nodes: document as requiring external mode or co-location. Internal mode was designed for co-located deployments per its original hostNetwork intent.
> 
> The EdgeTunnelIP custom address type idea from the community meeting is architecturally the cleanest long-term solution as it eliminates the InternalIP conflict entirely. I suggest pursuing that upstream in Kubernetes if the community agrees.
> 
> Please let me know further directions.


Currently, the `internal` mode of `iptablesManager` cannot address the scenario where cloudcore and the API server are deployed on different nodes—traffic to the API server cannot be intercepted by iptables. The existing proposals are:

1.The proposal previously raised by @tushar743-ui  sets the InternalIP to the cloudcore IP that the edge node connects to. We discussed this in the community meeting and believe it carries risks; moreover, InternalIP is a pseudo IP.
2.By adding an EdgeTunnelIP in the API server's parameters, with `--kubelet-preferred-address-types=EdgeTunnelIP,Hostname,InternalIP,ExternalIP`, and defining `EdgeTunnelIP` on the edge node as the cloudcore's IP it connects to. This approach requires modifying the `--kubelet-preferred-address-types` parameter in the API-Server's startup arguments, i.e., modifying `/etc/kubernetes/manifests/kube-apiserver.yaml`. This solution avoids modifying the edge node's `InternalI`P, but it does change the API server's startup parameters.

Previously, @tushar743-ui  experimented with eBPF and concluded that the adaptation is too complex and introduces too many additional dependencies. @WillardHu , do you think the above two approaches are feasible?

luomengY:
Today we discussed this issue again in the community meeting. We are more inclined to accept the solution of modifying the API-Server startup parameter `--kubelet-preferred-address-types`. Next, you need to read the API-Server source code to determine whether the API server traverses the address types in the order specified by the `--kubelet-preferred-address-types` configuration when requesting the kubelet. For example, suppose the default API-Server startup parameter is `--kubelet-preferred-address-types=InternalIP,ExternalIP,Hostname`. Does it automatically fall back to `ExternalIP` if `InternalIP` is unreachable, and then to `Hostname` if `ExternalIP` is unreachable? If that's the case, we can set `ExternalIP` to the `CloudcoreIP` that the edge node connects to, because `ExternalIP` can by definition serve as a pseudo IP, and we would not need to modify `InternalIP`. If the API-Server does not request the kubelet in this fallback manner, but always uses the first configured type as the IP to contact the kubelet, then we can introduce `EdgeTunnelIP`.  @tushar743-ui

tushar743-ui:
After reading the GetPreferredNodeAddress source code in k8s.io/kubernetes/pkg/util/node/node.go I noticed that -

The API server does NOT fall back on connection failure. It iterates through the preferredAddressTypes list and returns the address of the first type that EXISTS on the node, regardless of whether that address is reachable. There is no retry or fallback to the next type.

This means:

Option 1 -> ExternalIP = cloudCoreIP:
Only works if ExternalIP appears BEFORE InternalIP in --kubelet-preferred-address-types. With the default order
(InternalIP,ExternalIP,Hostname), InternalIP is found first and ExternalIP is never reached. So this requires changing the flag to --kubelet-preferred-address-types=ExternalIP,InternalIP,Hostname. InternalIP remains untouched and shows real edge IP. ExternalIP carries cloudCoreIP as the routing target -> semantically valid since ExternalIP by definition represents an externally routable address, which cloudCoreIP is for edge traffic.

Option 2 -> EdgeTunnelIP custom address type:
Set --kubelet-preferred-address-types=EdgeTunnelIP,InternalIP, ExternalIP,Hostname. API server finds EdgeTunnelIP first and uses cloudCoreIP. InternalIP completely untouched. Semantically cleaner since the type name explicitly describes its purpose.

Both options require the --kubelet-preferred-address-types flag change. ExternalIP is immediately implementable with no Kubernetes upstream changes needed. EdgeTunnelIP requires adding a new address type to
Kubernetes upstream first.

I suggest proceeding with ExternalIP = cloudCoreIP and --kubelet-preferred-address-types=ExternalIP,InternalIP,Hostname
as the near-term solution, with EdgeTunnelIP as the long-term clean approach. What do you think about this ?

luomengY:
> After reading the GetPreferredNodeAddress source code in k8s.io/kubernetes/pkg/util/node/node.go I noticed that -
> 
> The API server does NOT fall back on connection failure. It iterates through the preferredAddressTypes list and returns the address of the first type that EXISTS on the node, regardless of whether that address is reachable. There is no retry or fallback to the next type.
> 
> This means:
> 
> Option 1 -> ExternalIP = cloudCoreIP: Only works if ExternalIP appears BEFORE InternalIP in --kubelet-preferred-address-types. With the default order (InternalIP,ExternalIP,Hostname), InternalIP is found first and ExternalIP is never reached. So this requires changing the flag to --kubelet-preferred-address-types=ExternalIP,InternalIP,Hostname. InternalIP remains untouched and shows real edge IP. ExternalIP carries cloudCoreIP as the routing target -> semantically valid since ExternalIP by definition represents an externally routable address, which cloudCoreIP is for edge traffic.
> 
> Option 2 -> EdgeTunnelIP custom address type: Set --kubelet-preferred-address-types=EdgeTunnelIP,InternalIP, ExternalIP,Hostname. API server finds EdgeTunnelIP first and uses cloudCoreIP. InternalIP completely untouched. Semantically cleaner since the type name explicitly describes its purpose.
> 
> Both options require the --kubelet-preferred-address-types flag change. ExternalIP is immediately implementable with no Kubernetes upstream changes needed. EdgeTunnelIP requires adding a new address type to Kubernetes upstream first.
> 
> I suggest proceeding with ExternalIP = cloudCoreIP and --kubelet-preferred-address-types=ExternalIP,InternalIP,Hostname as the near-term solution, with EdgeTunnelIP as the long-term clean approach. What do you think about this ?

Then let's adopt the `EdgeTunnelIP` solution. When an edge node connects to `cloudcore`, define `EdgeTunnelIP` directly in the node's status. When deploying `cloudcore`, user can manually put `EdgeTunnelIP` first in the API-Server's startup parameters. This part of the functionality can be implemented as a featureGate initially. 
What part of the upstream Kubernetes adaptation were you referring to?

tushar743-ui:
> Then let's adopt the `EdgeTunnelIP` solution. When an edge node connects to `cloudcore`, define `EdgeTunnelIP` directly in the node's status. When deploying `cloudcore`, user can manually put `EdgeTunnelIP` first in the API-Server's startup parameters. This part of the functionality can be implemented as a featureGate initially. What part of the upstream Kubernetes adaptation were you referring to?

After checking the Kubernetes source, NodeAddressType is a plain string type with the explicit comment: "A cloud provider may set a type not listed here." No upstream Kubernetes changes needed. KubeEdge can define
EdgeTunnelIP as its own constant and set it directly on node.Status.Addresses. When I mentioned upstream adaptation earlier I was mistaken, I assumed NodeAddressType was a sealed enum. It is not.

One implementation detail found during cross-checking: upstream.go contains                                            getNode.Status = nodeStatusRequest.Status which overwrites the entire Addresses slice on every edgecore heartbeat. Since edgecore never sends EdgeTunnelIP, it gets removed on every heartbeat. The fix mirrors
the existing KubeletEndpoint.Port preservation at upstream.go:694  --> preserve EdgeTunnelIP addresses before the overwrite and re-append them after.

Implementation plan:
1. Define NodeEdgeTunnelIP corev1.NodeAddressType = "EdgeTunnelIP"
   in KubeEdge constants
2. Add EdgeTunnelIP preservation in upstream.go alongside the existing
   KubeletEndpoint.Port preservation logic
3. In tunnelserver.go connect(), append EdgeTunnelIP = cloudCoreIP to
   node.Status.Addresses when edge node connects
4. Remove EdgeTunnelIP after session.Serve() returns on disconnect
5. Gate behind FeatureGates in cloudcore config as suggested
6. Document that operators must configure kube-apiserver:
   --kubelet-preferred-address-types=EdgeTunnelIP,InternalIP,
   ExternalIP,Hostname

Shall we proceed with this implementation?

luomengY:
> > Then let's adopt the `EdgeTunnelIP` solution. When an edge node connects to `cloudcore`, define `EdgeTunnelIP` directly in the node's status. When deploying `cloudcore`, user can manually put `EdgeTunnelIP` first in the API-Server's startup parameters. This part of the functionality can be implemented as a featureGate initially. What part of the upstream Kubernetes adaptation were you referring to?
> 
> After checking the Kubernetes source, NodeAddressType is a plain string type with the explicit comment: "A cloud provider may set a type not listed here." No upstream Kubernetes changes needed. KubeEdge can define EdgeTunnelIP as its own constant and set it directly on node.Status.Addresses. When I mentioned upstream adaptation earlier I was mistaken, I assumed NodeAddressType was a sealed enum. It is not.
> 
> One implementation detail found during cross-checking: upstream.go contains getNode.Status = nodeStatusRequest.Status which overwrites the entire Addresses slice on every edgecore heartbeat. Since edgecore never sends EdgeTunnelIP, it gets removed on every heartbeat. The fix mirrors the existing KubeletEndpoint.Port preservation at upstream.go:694 --> preserve EdgeTunnelIP addresses before the overwrite and re-append them after.
> 
> Implementation plan:
> 
> 1. Define NodeEdgeTunnelIP corev1.NodeAddressType = "EdgeTunnelIP"
>    in KubeEdge constants
> 2. Add EdgeTunnelIP preservation in upstream.go alongside the existing
>    KubeletEndpoint.Port preservation logic
> 3. In tunnelserver.go connect(), append EdgeTunnelIP = cloudCoreIP to
>    node.Status.Addresses when edge node connects
> 4. Remove EdgeTunnelIP after session.Serve() returns on disconnect
> 5. Gate behind FeatureGates in cloudcore config as suggested
> 6. Document that operators must configure kube-apiserver:
>    --kubelet-preferred-address-types=EdgeTunnelIP,InternalIP,
>    ExternalIP,Hostname
> 
> Shall we proceed with this implementation?

Okay, we can now proceed with the implementation according to the plan, and also update the proposal with the new solution.

tushar743-ui:
Updated the proposal and opened the Phase-1 implementation PR -> #6912

luomengY:
The `EdgeTunnelIP` mode is employed only under the condition that iptablesManager is running in internal mode and cloudcore and apiserver reside on different nodes.

docs/proposals/sig-networking/iptables-free-apiserver-redirection-to-cloudcore.md
```diff
@@ -0,0 +1,464 @@
...
+```go
+klog.Errorf("iptables unavailable: %v", err)
+return
+```
+
+CloudCore remains operational when iptables is unavailable.
+
+### 6.3 Request Flow
+
+![edgetunnelip-flow](../../images/proposals/KubeEdge-EdgeTunnelIP-Connection-and-Request-Routing-Flow.png)
```
luomengY, line 377: "In `connect()` (as illustrated in `KubeEdge-EdgeTunnelIP-Connection-and-Request-Routing-Flow.png`), a check should be added so that the `EdgeTunnelIP` is only set when cloudcore and the apiserver are not on the same node."

tushar743-ui:
Updated the proposal accordingly

luomengY:
/lgtm

@WillardHu  Review

luomengY:
cc @WillardHu

PR #6841, opened 2026-06-06
fix(containers): stop container before remove in CopyResources

Norway-02:
**What type of PR is this?**
/kind bug

RemoveContainer is called via defer immediately after CreateContainer succeeds, before StartContainer is invoked. When CopyResources completes normally the container is still in Running state. Calling RemoveContainer on a running container violates the CRI API contract; containerd returns an error which is only logged at V(3), leaving the container and its sandbox orphaned on the node.

This PR adds `StopContainer(gracePeriod=0)` before `RemoveContainer` in the defer to ensure proper cleanup, and tests to verify the CRI sequence.

NONE

(Note: The git branch history was heavily contaminated with 33 unrelated file modifications from an accidental submodule pointer branch. This has been resolved and the PR is now atomic).

```release-note
NONE
```

### DIFF
--- a/pkg/containers/container_runtime.go
+++ b/pkg/containers/container_runtime.go
@@ -137,6 +137,9 @@ func (runtime *ContainerRuntimeImpl) CopyResources(
 		return fmt.Errorf("create container failed: %v", err)
 	}
 	defer func() {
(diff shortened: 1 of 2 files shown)


Norway-02:
Hi @doislong and @WillardHu,

PR #6842, opened 2026-06-06
test(upgrade): fix assert/require misuse and handle unchecked errors in task event reporter

Norway-02:
**What type of PR is this?**
/kind test

This PR cleans up the unit tests and internal error handling for the upgrade `TaskEventReporter`:
1. Replaces misuse of `assert` with `require` to prevent cascading panics in test execution.
2. Checks the error return from `tls.LoadX509KeyPair` to ensure we gracefully log rather than silently ignoring invalid certificates.
3. Fixes a test issue by generating valid ephemeral `x509` client certificates instead of passing empty dummy files, proving the full cert loading flow works.
4. Uses `http.NewRequestWithContext` rather than `client.Post` to satisfy static analysis context propagation rules.

*(Note: The previous accidental commit regarding `StopContainer` has been rebased out of this branch, as it belongs in #6841).*

NONE

This branch has been cleaned up and bot comments regarding security, context, and error checking have been fully addressed.

```release-note
NONE
```

### DIFF
--- a/pkg/upgrade/edge/upgrade_reporter_taskevent.go
+++ b/pkg/upgrade/edge/upgrade_reporter_taskevent.go
@@ -93,11 +93,12 @@ func ReportTaskResult(config *edgeconfig.EdgeCoreConfig, taskType, taskID string
 	certFile := edgeHub.TLSCertFile
 	keyFile := edgeHub.TLSPrivateKeyFile
 	var certs []tls.Certificate
-	cliCrt, err := tls.LoadX509KeyPair(certFile, keyFile)
-	if err == nil {
+	if certFile != "" || keyFile != "" {
+		cliCrt, err := tls.LoadX509KeyPair(certFile, keyFile)
+		if err != nil {
+			return fmt.Errorf("failed to load client certificate: %v", err)
+		}
 		certs = []tls.Certificate{cliCrt}
-	} else {
-		klog.Warningf("failed to load client certificate: %v", err)
 	}
 
 	transport := &http.Transport{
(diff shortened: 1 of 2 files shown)


Norway-02:
Hi @doislong and @WillardHu,

DoisLONG (commented):
Thanks for cleaning this up. The PR is much more focused now, and moving the unrelated `StopContainer` change out of this branch makes sense.

The test cleanup looks good to me: using `require` for setup failures and generating a real client certificate/key is much better than using empty dummy files. The `NewRequestWithContext` change also looks fine.

One thing I think still needs to be fixed before this can move forward: `tls.LoadX509KeyPair` now checks the error, but it only logs a warning and continues without a client certificate.

If `TLSCertFile` and `TLSPrivateKeyFile` are configured but invalid, that should probably be treated as a configuration error and returned immediately. Otherwise the request may be sent without a client certificate, and in environments where the server does not require client auth, this misconfiguration could be silently hidden.

Could you change this so invalid configured client cert/key files return an error? If client certificates are intended to be optional, we can skip loading only when both paths are empty. It would also be good to add a small regression test for the invalid cert/key path.

Norway-02:
@DoisLONG Thank you for the review! I have addressed your feedback regarding client certificate error handling:

1. **Enforced Errors on Invalid Configured Certs:** If `TLSCertFile` or `TLSPrivateKeyFile` is configured (not empty string) but invalid, `tls.LoadX509KeyPair` will now return an error immediately and fail the request rather than silently falling back to no client auth. Optional client certificates remain supported (by leaving both fields empty).
2. **Added Regression Test:** I added `TestReportTaskResult_InvalidClientCert` which provides a valid CA but mock invalid paths for the client cert/key, verifying that the function correctly bubbles up the 'failed to load client certificate' error.

All tests are passing. Please let me know if there's anything else!

PR #6844, opened 2026-06-07
fix(pkg/metaserver): close data race in Application Wait and Cancel methods

Norway-02:
**What type of PR is this?**
/kind bug

This PR resolves data races in `pkg/metaserver/application.go` when `Wait()`, `Cancel()`, and `Reset()` are invoked concurrently on an `Application`:

1. **Mutex Value Type**: Changes `countLock` from `*sync.Mutex` to `sync.Mutex` value type to eliminate nil pointer risks.
2. **Synchronized Field Access**: Acquires `countLock` in `Wait()` and `Cancel()` when reading `a.ctx` and `a.cancel`.
3. **Deadlock-Free Reset**: Acquires `countLock` in `Reset()` when re-assigning `a.ctx` and `a.cancel`, and executes the previous context cancel function *after* releasing the lock to avoid lock-holding during callbacks.
4. **Deterministic Regression Test**: Adds `TestConcurrentResetWaitCancel` running heavy concurrent invocations under `go test -race` with a guaranteed final cancel so `Wait()` never blocks indefinitely.

NONE

- Formatted with `gofmt -s`.
- Exactly 1 squashed commit with DCO sign-off.
- Verified with `go test -race ./pkg/metaserver/...`.

```release-note
NONE
```

### DIFF
--- a/pkg/metaserver/application.go
+++ b/pkg/metaserver/application.go
@@ -62,7 +62,7 @@ type Application struct {
 
 	// count the number of current citations
 	count     uint64
-	countLock *sync.Mutex
+	countLock sync.Mutex
 	// Timestamp record the last closing time of application, only make sense when count == 0
 	Timestamp time.Time
 }
@@ -89,7 +89,6 @@ func NewApplication(ctx context.Context, key string, verb ApplicationVerb, noden
 		ctx:         ctx2,
 		cancel:      cancel,
 		count:       0,
-		countLock:   &sync.Mutex{},
 		Timestamp:   time.Time{},
 	}
 	app.Add()
@@ -158,8 +157,12 @@ func (a *Application) Namespace() string {
 }
 
 func (a *Application) Cancel() {
-	if a.cancel != nil {
-		a.cancel()
+	a.countLock.Lock()
+	cancel := a.cancel
+	a.countLock.Unlock()
+
+	if cancel != nil {
+		cancel()
 	}
 }
 
@@ -169,8 +172,12 @@ func (a *Application) GetStatus() ApplicationStatus {
 
 // Wait the result of application after it is applied by application agent
 func (a *Application) Wait() {
-	if a.ctx != nil {
-		<-a.ctx.Done()
+	a.countLock.Lock()
+	ctx := a.ctx
+	a.countLock.Unlock()
+
+	if ctx != nil {
+		<-ctx.Done()
 	}
 }
 
--- a/pkg/metaserver/application.go
+++ b/pkg/metaserver/application.go
@@ -182,12 +182,16 @@ func (a *Application) Wait() {
 }
 
 func (a *Application) Reset() {
-	if a.ctx != nil && a.cancel != nil {
-		a.cancel()
-	}
+	a.countLock.Lock()
+	oldCancel := a.cancel
 	a.ctx, a.cancel = context.WithCancel(beehiveContext.GetContext())
 	a.Reason = ""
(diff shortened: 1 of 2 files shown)


Norway-02:
Hi @doislong and @WillardHu,

DoisLONG (commented):
Thanks for cleaning up the PR scope.

Changing `countLock` from `*sync.Mutex` to `sync.Mutex` looks good, and adding locking in `Wait()` / `Cancel()` is the right direction.

However, I do not think the data race is fully fixed yet. In the current version, `Reset()` still writes to `a.ctx` and `a.cancel` without holding `countLock`, while `Wait()` and `Cancel()` read those fields under the lock. Since the writer side is still unlocked, concurrent `Reset()` with `Wait()` / `Cancel()` can still race.

Could you also protect the `Reset()` updates to `ctx` / `cancel` with the same lock? One possible pattern is to copy the old cancel func, create and assign the new context under the lock, then call the old cancel func after releasing the lock.

It would also be useful to add a small regression test that exercises concurrent `Reset()` with `Wait()` / `Cancel()` under the race detector, since this PR is specifically fixing a data race.

Norway-02:
@DoisLONG Thanks for pointing that out! I have pushed a new commit to fully resolve the remaining data race. 

Here is what was addressed:
1. **Protected `Reset()` context updates:** The `Reset()` method now correctly acquires the `countLock` before re-assigning the `ctx` and `cancel` fields. 
2. **Safe cancellation:** The old cancel function is cached under the lock but executed *after* the lock is released to guarantee no deadlocks.
3. **Concurrent regression test:** Added `TestConcurrentResetWaitCancel` which heavily spans concurrent calls to `Reset`, `Wait`, and `Cancel` in a loop to ensure the race detector validates it.

All tests passed successfully under `go test -race`. Let me know if everything looks good to you now!

DoisLONG (commented):
Thanks for the update.

The production code looks much better now. `Reset()` now updates `ctx` / `cancel` under the same `countLock`, and calling the old cancel function after releasing the lock avoids holding the mutex while invoking the callback. That addresses the main race concern from my previous comment.

One issue remains in the new regression test: `TestConcurrentResetWaitCancel` can potentially hang. Since `Reset()`, `Wait()`, and `Cancel()` run independently, it is possible for the `Cancel()` goroutine to finish before the last `Reset()` creates a new context. In that case, a later `Wait()` call may read the final uncancelled context and block forever.

Could you make the test deterministic, for example by ensuring there is a final cancel after the reset loop, or by structuring the test so every context that `Wait()` can observe is guaranteed to be cancelled?

Also, please run gofmt on `application_test.go`; the import order currently looks off.

After that, this should be close.

Norway-02:
@doislong Excellent catch! I have added a final `app.Cancel()` at the end of the `Reset()` goroutine to guarantee that the final context created is cancelled and `Wait()` never hangs. I also ran `gofmt` on the test file to correct the import order. Everything is pushed now, let me know if it looks solid!

Norway-02:
I have:
1. Made the concurrent regression test (`TestConcurrentResetWaitCancel`) deterministic by ensuring the final context is cancelled so `Wait()` never hangs.
2. Ran `gofmt` on `application_test.go` to fix the import ordering.
3. Squashed all commits into a single clean, DCO-signed commit.

All tests pass cleanly under `go test -race`. Please take a look when you have a chance!

DoisLONG (commented):
Thanks for the update.

/lgtm

Norway-02:

PR #6847, opened 2026-06-08
fix(metaserver): remove unreachable code and fix IndexCheck signature

Norway-02:
**What type of PR is this?**
/kind bug
/kind cleanup

1. Fixes an issue where `ParseKey` would not gracefully handle malformed keys (missing leading slash) and returns proper zero values instead of propagating invalid data.
2. Changes the signature of `IndexCheck` from `*int` to `int` since the function was previously modified to strictly read bounds and no longer mutates the indexes. This removes confusion and silences unused pointer mutation warnings.

NONE

The `IndexCheck` function is actively used in `ParseKey` — this PR just refines its signature to be safer.

```release-note
NONE
```

### DIFF
--- a/pkg/metaserver/key.go
+++ b/pkg/metaserver/key.go
@@ -129,10 +129,11 @@ func ParseKey(key string) (gvr schema.GroupVersionResource, namespace string, na
 	}
 	slices := strings.Split(sl, "/")
 	length := len(slices)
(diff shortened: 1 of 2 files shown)


Norway-02:
Hi @doislong and @WillardHu,

PR #6849, opened 2026-06-09
fix(nodetask): prevent nil panics in actionflow.Find

Norway-02:
**What type of PR is this?**
/kind bug

This PR adds a `nil` receiver guard for `sf *Flow` and a `nil` action guard in `doFind()` to prevent panics if an empty or nil flow is searched.

(Note: An accidental submodule pointer commit in the root `kubeedge` directory has been successfully rebased out of this branch).

NONE

Bot concerns regarding the `nil` receiver have been addressed, and tests verifying the behavior of `Find` on a nil/empty flow have been added.

```release-note
NONE
```

### DIFF
--- a/pkg/nodetask/actionflow/action_flow.go
+++ b/pkg/nodetask/actionflow/action_flow.go
@@ -40,26 +40,27 @@ type Flow struct {
 
 // Find returns the found action by name.
 func (sf *Flow) Find(name string) *Action {
-	if sf.First.Name == name {
-		return sf.First
+	// Guard against nil receiver
+	if sf == nil {
+		return nil
 	}
+	
 	return doFind(name, sf.First)
 }
 
-// Using recursion to find a action by name.
+// doFind recursively searches for an action by name across both
+// NextSuccessful and NextFailure branches.
 func doFind(name string, act *Action) *Action {
+	if act == nil {
+		return nil
+	}
 	if act.Name == name {
 		return act
 	}
-	if act.NextSuccessful != nil {
-		if next := doFind(name, act.NextSuccessful); next != nil {
-			return next
-		}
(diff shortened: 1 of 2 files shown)


Norway-02:
Hi @doislong and @WillardHu,

DoisLONG (commented):
Thanks for the update.

The nil receiver guard in `Find()` and the nil `act` guard in `doFind()` look reasonable to me.

One thing I am not fully convinced about is the “incomplete DFS traversal” part. From the old code in this diff, `doFind()` already searched `NextSuccessful` first and then searched `NextFailure` if the successful branch did not contain the target. So the new `ConfigUpdateJobActionRollBack` case looks like it may already pass with the old implementation.

Could you clarify the exact traversal case that was broken, or adjust the PR description/title to focus on the nil receiver / nil action safety improvement?

It would also be useful to add tests for the actual new behavior, for example `var f *Flow; f.Find(...)` and `(&Flow{}).Find(...)`, since those are the cases fixed by the new nil guards.

Also, please run gofmt on `action_flow.go`; there seems to be whitespace on the blank line after the nil receiver check.

Norway-02:
@doislong Thanks for the review! You are completely correct, `doFind()` was already searching the `NextFailure` branch during the recursive step, so the DFS traversal wasn't actually broken as I originally stated! I have updated the PR title and description to accurately reflect that the sole purpose of this PR is adding the nil receiver/action guards to prevent panics.

I've also:
1. Run `gofmt` on `action_flow.go` to fix the stray whitespace on line 47.
2. Added the two new tests (`TestFindWithNilFlow` and `TestFindWithEmptyFlow`) in `action_flow_test.go` to explicitly verify the new nil-guard behavior.

Everything is pushed and the PR text is updated! Let me know if everything looks good now.

PR #6852, opened 2026-06-11
docs: explain keadm ctl and kubectl differences

### LINKED ISSUE #6705: Unclear what keadm ctl can do and how it differs from kubectl
I know kubectl is used to manage Kubernetes resources, but I saw there is also keadm ctl. I wasn't sure what keadm ctl can do, whether it replaces kubectl for edge operations, or how the two tools work together. I expected clear documentation explaining the scope of keadm ctl, what it can and cannot do, and when I should use it versus kubectl.

1. Install KubeEdge and have both cloud and edge nodes running
2. Run keadm ctl --help to see available commands
3. Compare with kubectl --help - the command sets look similar but the scope is unclear
4. Try to use keadm ctl get nodes vs kubectl get nodes - both work but show different information
5. Try to use keadm ctl describe pod on an edge pod - unsure if this should work or not

From what I gathered through trial and error:

- kubectl operates on the Kubernetes API server and sees cloud + edge resources as part of the cluster
- keadm ctl seems to be an extension that can interact directly with edge nodes, possibly for operations that don't go through the API server
- Some keadm ctl commands like logs, exec, describe appear to work on edge pods without needing direct access to the edge node
- But the documentation doesn't clearly explain the use cases for each tool

It would be helpful if:

- The documentation included a comparison table showing kubectl vs keadm ctl - what each can do, limitations, and when to use which
- The help output (keadm ctl --help) provided clearer categorization of commands
- There were examples showing common scenarios and which tool to use

**Environment**:
- Kubernetes version (use kubectl version): v1.28.13
- KubeEdge version(e.g. cloudcore --version and edgecore --version): v1.20.0

DoisLONG:
### What type of PR is this?

/kind documentation

### What this PR does / why we need it

This PR adds documentation to explain the difference between `keadm ctl` and `kubectl`.

New KubeEdge users may be confused about whether `keadm ctl` replaces `kubectl`, or when each tool should be used. This document clarifies their different scopes:

* `kubectl` works from the Kubernetes control-plane perspective.
* `keadm ctl` is used for KubeEdge-specific edge-side operations and troubleshooting.

The document also includes a comparison table, common examples, typical scenarios, and a basic troubleshooting workflow to help users choose the right tool.

### Which issue(s) this PR fixes

Fixes #6705

### Special notes for reviewers

This PR focuses on documentation only. It does not change the behavior of `keadm ctl` or `kubectl`.

I added this as a new document under `docs/keadm/`. If maintainers prefer another documentation location, I am happy to adjust it.

### DIFF
--- a/docs/keadm/keadm-ctl-vs-kubectl.md
+++ b/docs/keadm/keadm-ctl-vs-kubectl.md
@@ -1,0 +1,95 @@
+# keadm ctl and kubectl
+
+This document explains the difference between `keadm ctl` and `kubectl`, and
+helps users choose the right command-line tool for common KubeEdge operations.
+
+## Overview
+
+`kubectl` is the standard Kubernetes command-line tool. It communicates with
+the Kubernetes API server and shows resources from the Kubernetes control-plane
+perspective.
+
+`keadm ctl` is a KubeEdge command group for edge-side operations. It is designed
+to help users inspect or operate resources from the edge data-plane perspective,
+especially in scenarios where edge autonomy or edge-side troubleshooting is
+needed.
+
+`keadm ctl` does not replace `kubectl`. In most cases, users should still use
+`kubectl` to manage Kubernetes resources from the cloud side. Use `keadm ctl`
+when the operation is related to KubeEdge-specific edge-side behavior.
+
+## Comparison
+
+| Tool | Scope | Typical use cases |
+| --- | --- | --- |
+| `kubectl` | Kubernetes control plane | Manage Kubernetes resources such as Nodes, Pods, Deployments, Services, ConfigMaps, and Secrets through the API server. |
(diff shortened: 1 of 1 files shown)


DoisLONG:
Updated the document according to the review suggestions: changed the title to match the filename, replaced indented code blocks with fenced `bash` code blocks, and clarified the offline edge-side diagnosis behavior of `keadm ctl`.

WillardHu:
The comparison section can be supplemented based on the original documentation after merging from https://github.com/kubeedge/kubeedge/pull/6855.

Using two documents is not suitable, and there will be a lot of overlap.

DoisLONG:
Thanks for the suggestion. I agree that maintaining two separate documents would introduce unnecessary overlap.

I will wait for #6855 to be merged, then rebase this PR and fold the useful comparison content into the document introduced by #6855 instead of keeping a separate document.

PR #6861, opened 2026-06-12
Cleanup/remove stale disablenodetaskv1alpha2 featuregate

Norway-02:
**What type of PR is this?**
/kind cleanup

The DisableNodeTaskV1alpha2 feature gate had an explicit TODO comment
stating it should be removed after v1.23. The module now targets
go 1.23.12, meaning the deadline has passed.

This PR removes the stale feature gate constant and map entry from
features.go, removes all conditional branches gated on it in
task_manager.go and controllermanager.go, and deletes the dead
v1alpha1 init/start functions. The v1alpha2 code path is now
unconditionally active, which was already the default behavior since
the gate defaulted to false, making !Enabled() always true.

NONE

The v1alpha2 path was already the default. DisableNodeTaskV1alpha2
defaulted to false, so !Enabled() evaluated to true in all 3 call
sites. This change makes that explicit and removes the dead v1alpha1
code paths including initV1alpha1, startV1alpha1, and related struct
fields and imports.

```release-note
Removed the deprecated DisableNodeTaskV1alpha2 feature gate.
The v1alpha2 node task API is now always active. Users who
explicitly set disableNodeTaskV1alpha2=true in their configuration
should remove that setting.
```

### DIFF
--- a/build/edge/kubernetes/README.md
+++ b/build/edge/kubernetes/README.md
@@ -4,6 +4,15 @@ This method will guide you to deploy the edge part into a k8s cluster,
 so you need to login to the k8s master node (or where else if you can
 operate the cluster with `kubectl`).
 
+### Flannel on edge nodes
+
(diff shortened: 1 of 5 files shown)


Norway-02:
Hi @doislong and @WillardHu,

PR #6864, opened 2026-06-13
fix(pkg/metaserver): use pointer map in MsgToApplications to avoid copylocks

Norway-02:
**What type of PR is this?**
/kind bug

Adding `sync.Once` as a value field in the `Application` struct made it a non-copyable type. `MsgToApplications` was returning `map[string]Application` (copy-by-value), which triggered a `go vet copylocks` failure because it copies the lock value.

This PR changes `MsgToApplications` to return `map[string]*Application` and updates all downstream callers (in `dynamiccontroller` and `metamanager`) to use pointers. It also fixes `json.Marshal(app)` to `json.Marshal(&app)` in tests to resolve identical copylocks warnings.

NONE

This addresses the linter/copylocks failure flagged by the bot.

```release-note
NONE
```

### DIFF
--- a/pkg/metaserver/application.go
+++ b/pkg/metaserver/application.go
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package metaserver provides meta server functionality.
 package metaserver
 
 import (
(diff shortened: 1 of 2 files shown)


Norway-02:
Hi @doislong and @WillardHu,

PR #6867, opened 2026-06-13
Fix processVolume sending garbage response to cloud when edged call fails

### LINKED ISSUE #6865: processVolume sends garbage response to cloud when edged call fails
In processVolume, when SendSync to edged fails or times out, the error
is logged but execution falls through and still calls sendToCloud with
whatever zero-value back.GetContent() returns:

```
    back, err := beehiveContext.SendSync(modules.EdgedModuleName, message, ...)
    if err != nil {
        klog.Errorf("process volume send to edged failed: %v", err)
        // missing return here
    }
    resp := message.NewRespByMessage(&message, back.GetContent())
    sendToCloud(resp)
```

So if edged is slow, crashed, or the context deadline expires, the cloud
CSI controller gets a nil/empty response and treats the volume operation
as completed. The operation never actually ran on the edge. No retry, no
error propagation, nothing.

This is particularly bad for volume attach/detach because there's no
reconciliation loop on the edge side, if the cloud thinks it's done,
it won't retry.

Relevant code: edge/pkg/metamanager/process.go, processVolume(), 

/kind bug

avinxshKD:
**What type of PR is this?**

/kind bug

Stops sending empty/garbage success responses to the cloud CSI controller when the `SendSync` to `edged` times out or fails. 

Added the missing error handling and `return` in `processVolume`. Without this, the cloud controller assumes the volume operation succeeded and never retries, leaving pods permanently stuck.

Fixes #6865

Just a missing return that was causing silent CSI state desyncs between cloud and edge. Used the existing `feedbackError` helper to propagate it properly.

### DIFF
--- a/edge/pkg/metamanager/process.go
+++ b/edge/pkg/metamanager/process.go
@@ -424,6 +424,8 @@ func (m *metaManager) processVolume(message model.Message) {
 	klog.Infof("process volume get: req[%+v], back[%+v], err[%+v]", message, back, err)
 	if err != nil {
 		klog.Errorf("process volume send to edged failed: %v", err)
+		feedbackError(fmt.Errorf("process volume send to edged failed: %v", err), message)
+		return
 	}
 
 	resp := message.NewRespByMessage(&message, back.GetContent())


avinxshKD:
@luomengY @vincentgoat pls take a look when free. minimal addition, ofthe missing error handling and return in processVolume.

DoisLONG:
The error-handling direction looks correct.

When `SendSync` fails, returning immediately prevents `processVolume` from constructing a normal response from an invalid `back` message. Using `feedbackError` is also consistent with the cloud CSI controller, which handles `ResponseErrorOperation` as a failed CSI request.

There are two items to address before this is ready:

1. The branch is currently behind master and the PR is not mergeable. Please rebase it onto the latest master and confirm that the final diff remains limited to this fix.

2. Please add a regression test for the `SendSync` failure path. The test should verify that:

   * an error response is sent back to the cloud;
   * its operation is `ResponseErrorOperation`;
   * the response remains correlated with the original request;
   * no normal success response is sent after the error.

It would also be useful to retain coverage for the existing successful path.

Non-blocking: `%w` would be preferable to `%v` when wrapping the `SendSync` error, although the error is eventually serialized as text.

The implementation itself is small and reasonable, but this failure path should be locked in with a test before merging.

PR #6874, opened 2026-06-15
fix: handle nil deploy.Spec.Replicas in EdgeApplication status reconciler

### LINKED ISSUE #6872: Nil pointer dereference on deploy.Spec.Replicas in EdgeApplication status reconciler causes controller-manager panic
**What happened**:

In `cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler.go`, the `deploymentAvailable.IsAvailable()` function dereferences `deploy.Spec.Replicas` without a nil check on [line 203](https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler.go#L203):

`deploy.Status.ReadyReplicas == *deploy.Spec.Replicas`

`DeploymentSpec.Replicas` is `*int32` (`+optional`). When a Deployment is created without explicitly setting `.spec.replicas` (very common — Kubernetes defaults it to 1), the field is `nil`. Dereferencing it causes a nil pointer panic, crashing the cloudcore controller-manager.

Additionally, on [line 195](https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler.go#L195), `return false, err` should be `return false, nil` since `err` is guaranteed to be nil at that point.

The status reconciler should handle nil `Spec.Replicas` by defaulting to 1 (matching Kubernetes behavior), instead of panicking.

1. Create an EdgeApplication with a Deployment template that does not set `.spec.replicas`.
2. The status reconciler panics at `*deploy.Spec.Replicas`.

Affected file: `cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler.go`
Severity: High — crashes the controller-manager

I would like to work on this fix. Please assign this issue to me.

/kind bug

**Environment**:
- Kubernetes version (use `kubectl version`): N/A (code-level bug)
- KubeEdge version: master (latest)
- Cloud nodes Environment: N/A (found via code review)
- Edge nodes Environment: N/A

rahulkr182:
## What type of PR is this?

/kind bug

Fixes a nil pointer dereference in `deploymentAvailable.IsAvailable()` in `cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler.go` that crashes the cloudcore controller-manager.

### Changes:

1. **Nil pointer dereference on `deploy.Spec.Replicas` (line 203):**
   `DeploymentSpec.Replicas` is `*int32` and is optional. When a Deployment is created without explicitly setting `.spec.replicas`, Kubernetes defaults it to `1`, but the pointer remains `nil` in the spec. The existing code dereferences it directly (`*deploy.Spec.Replicas`), causing a panic. This fix checks for nil and defaults to `1`, matching Kubernetes behavior.

2. **Incorrect error return (line 195):**
   `return false, err` was returning `err` which is guaranteed to be `nil` at that point (the `err != nil` check on line 191 already returned). Changed to `return false, nil` for correctness and clarity.

## Which issue(s) this PR fixes:

Fixes #6872

## Special notes for your reviewer:

Minimal, safe change — only touches the `deploymentAvailable.IsAvailable()` function. No new dependencies or behavioral changes beyond the nil-safety fix.

```release-note
Fixed nil pointer dereference on deploy.Spec.Replicas in EdgeApplication status reconciler that caused controller-manager panic when Deployment did not explicitly set .spec.replicas.
```

### DIFF
--- a/cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler.go
+++ b/cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler.go
@@ -192,7 +192,7 @@ func (d deploymentAvailable) IsAvailable(ctx context.Context, client client.Clie
 		return false, err
 	}
 	if obj == nil {
-		return false, err
+		return false, nil
 	}
 
 	deploy := &appsv1.Deployment{}
@@ -200,7 +200,13 @@ func (d deploymentAvailable) IsAvailable(ctx context.Context, client client.Clie
 		return false, fmt.Errorf("failed to convert unstructured to deployment for %s/%s, %v", info.Namespace, info.Name, err)
 	}
 
-	return deploy.Status.ReadyReplicas == *deploy.Spec.Replicas, nil
+	// Kubernetes defaults Spec.Replicas to 1 when not explicitly set,
+	// so we must handle the nil case to avoid a nil pointer dereference.
+	var desiredReplicas int32 = 1
+	if deploy.Spec.Replicas != nil {
+		desiredReplicas = *deploy.Spec.Replicas
+	}
+	return deploy.Status.ReadyReplicas == desiredReplicas, nil
 }
 
 func getObjAccordingToResourceInfo(ctx context.Context, client client.Client, info utils.ResourceInfo) (*unstructured.Unstructured, error) {
--- a/cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler_test.go
+++ b/cloud/pkg/controllermanager/edgeapplication/statusmanager/reconciler_test.go
@@ -1,0 +1,231 @@
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


DoisLONG:
Thanks for fixing this.

The nil check for deploy.Spec.Replicas makes sense. Since Replicas is a pointer field, directly dereferencing it may panic when the field is nil. Defaulting to 1 in that case is also consistent with Kubernetes Deployment behavior.

One suggestion: could you add a small unit test for this reconciler logic if there is already a suitable test structure nearby? The test could cover at least:

deploy.Spec.Replicas == nil and ReadyReplicas == 1, expected available;
deploy.Spec.Replicas == nil and ReadyReplicas == 0, expected unavailable;
getObjAccordingToResourceInfo returns nil object, expected unavailable without error.

The code change itself looks straightforward; adding a test would make this safer for future refactoring.

DoisLONG:
Thanks for adding the tests.

The new test cases cover the scenarios discussed in the previous review:

- nil Spec.Replicas with ReadyReplicas equal to 1;
- nil Spec.Replicas with ReadyReplicas equal to 0;
- the Deployment object not being found;
- explicit replica values for both available and unavailable cases.

The implementation and the focused regression coverage look good to me now. No further concerns from my side, pending CI and maintainer review.

PR #6879, opened 2026-06-16
Fix/keyrootfunc panic

Norway-02:
**What type of PR is this?**
/kind bug

KeyRootFunc called panic on any context missing request info.
This path is reachable at runtime when middleware is absent or
request is malformed. panic crashes the handler goroutine.
Replace panic with klog.Errorf and return empty string.

NONE

KeyRootFunc is assigned as KeyRootFunc field in storage.go:62.
It is called by vendor store during GetList and Watch (store.go:423, 1447).
Both are hot paths. Test added: TestKeyRootFuncNopanic verifies
empty context returns "" without panic.

```release-note
Fix panic in metaserver KeyRootFunc when request context is missing info.
Edge node no longer crashes during list/watch on malformed requests.
```

### DIFF
--- a/pkg/metaserver/key.go
+++ b/pkg/metaserver/key.go
@@ -98,7 +98,8 @@ func KeyFuncReq(ctx context.Context, _ string) (string, error) {
 func KeyRootFunc(ctx context.Context) string {
 	key, err := KeyFuncReq(ctx, "")
 	if err != nil {
-		panic("fail to get list key!")
+		klog.Errorf("failed to get list key: %v", err)
+		return ""
 	}
 	return key
 }
--- a/pkg/metaserver/key_test.go
+++ b/pkg/metaserver/key_test.go
@@ -293,3 +293,11 @@ func TestParseKey(t *testing.T) {
 		})
 	}
 }
+
+func TestKeyRootFuncNopanic(t *testing.T) {
+	// empty context has no request info, must return "" not panic
+	result := KeyRootFunc(context.TODO())
+	if result != "" {
+		t.Errorf("expected empty string, got %v", result)
+	}
+}


DoisLONG:
Thanks for the PR.

The intended fix for KeyRootFunc looks useful: avoiding a panic in the storage/list/watch path and returning an empty key with an error log is easier to handle than crashing the process.

However, this PR currently includes many unrelated changes outside the KeyRootFunc fix, such as controller/taskmanager changes, container runtime cleanup, stream changes, and metaserver application changes. These make the PR much harder to review and increase the risk of introducing unrelated behavior changes.

Could you please rebase or cherry-pick this PR so that it only contains the minimal changes for the KeyRootFunc panic issue, ideally:

- pkg/metaserver/key.go;
- the related unit test in pkg/metaserver/key_test.go;
- any directly required supporting change.

If the metaserver Application.Identifier() concurrency changes are intended, I suggest moving them to a separate PR. Also, the current atomic state logic still appears to write a.ID without synchronization, so it may not fully avoid data races under concurrent calls. If that behavior needs to be made concurrency-safe, a sync.Once or a mutex-based approach may be safer and easier to reason about.

For the KeyRootFunc behavior itself, it may also be useful to add a short comment explaining why returning an empty string is safe for the callers.

pkg/metaserver/application.go
```diff
@@ -96,28 +101,43 @@ func NewApplication(ctx context.Context, key string, verb ApplicationVerb, noden
 	return app, nil
 }
 
+// Identifier returns the unique SHA256-based ID of the application.
 func (a *Application) Identifier() string {
-	if a.ID != "" {
+	// Already computed? Return fast.
```
WillardHu: You can use `golang.org/x/sync/singleflight` without new field `idState` here. The `atomic` package only guarantees atomic operations within a single function.

cloud/pkg/controllermanager/controllermanager.go
```diff
@@ -109,13 +108,9 @@ func setupControllers(ctx context.Context, mgr manager.Manager, che cache.Cache)
 		nodegroup.NewController(cli),
 		edgeapplication.NewController(ctx, cli, serializer, mgr),
 	}
-	if !features.DefaultFeatureGate.Enabled(features.DisableNodeTaskV1alpha2) {
```
WillardHu, line 112: The cleanup of the v1alpha1 node task needs to be tracked by another PR.

pkg/metaserver/application.go
Norway-02: Thanks for the suggestion. I've moved the KeyRootFunc fix to a focused PR #6881 without the Identifier() changes. Will address the Identifier() concurrency issue using singleflight in a separate PR.

cloud/pkg/controllermanager/controllermanager.go
Norway-02: Understood. The DisableNodeTaskV1alpha2 feature gate cleanup will be tracked in a separate focused PR. Thank you for the guidance.

Norway-02:
@WillardHu @doislong The branch has been fully cleaned up and is ready for your review whenever you have a moment!

PR #6881, opened 2026-06-17
fix(metaserver): replace panic with error log in KeyRootFunc

Norway-02:
**What type of PR is this?**
/kind bug

`KeyRootFunc` previously panicked if an invalid or unexpected key format was encountered. This caused the entire metaserver to crash.

This PR replaces the panic with a `klog.Errorf` and returns an empty string. It also adds a guard in the storage backend to prevent a nil `RequestInfo` from causing nil pointer dereferences.

NONE

This resolves the same panic bug that was attempted in #6879. However, this PR is focused solely on the panic fix, making it much safer to merge without side-effects.

```release-note
NONE
```

### DIFF
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/storage.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/storage.go
@@ -174,7 +174,10 @@ func (r *REST) PassThrough(ctx context.Context, options *metav1.GetOptions) ([]b
 }
 
 func (r *REST) List(ctx context.Context, options *metainternalversion.ListOptions) (runtime.Object, error) {
-	info, _ := apirequest.RequestInfoFrom(ctx)
+	info, ok := apirequest.RequestInfoFrom(ctx)
+	if !ok || info == nil {
+		return nil, fmt.Errorf("no request info in context, cannot list")
+	}
 	// First try to list the object from remote cloud
 	list, err := func() (runtime.Object, error) {
 		app, err := r.Agent.Generate(ctx, metaserver.List, *options, nil)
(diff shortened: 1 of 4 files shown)


DoisLONG:
Returning an empty key does not appear to be a safe no-op in the current storage path.

`genericregistry.Store` passes the result of `KeyRootFunc(ctx)` directly to `Storage.GetList` and `Storage.Watch`. In the SQLite imitator, an empty key is parsed into an empty GVR, namespace, and name. `RawMetaByGVRNN` then applies no filters, so a malformed list request may read records across all cached resource types. An empty watch key may similarly create a hook that matches all events.

Could we avoid returning an empty root key here? A safer approach would be to validate the request context before delegating to the generic store and return an explicit request error, or otherwise ensure that the failure path cannot broaden the query/watch scope.

Please also add regression coverage at the storage List/Watch level, since testing `KeyRootFunc` alone would not catch this behavior.

edge/pkg/metamanager/metaserver/kubernetes/storage/storage.go
```diff
@@ -213,6 +216,9 @@ func (r *REST) List(ctx context.Context, options *metainternalversion.ListOption
 
 func (r *REST) Watch(ctx context.Context, options *metainternalversion.ListOptions) (watch.Interface, error) {
 	info, _ := apirequest.RequestInfoFrom(ctx)
+	if info == nil {
```
DoisLONG: Thanks for adding the storage-level guards and regression tests. This addresses the original concern that an empty key could broaden the local SQLite List or Watch scope.

One remaining suggestion: could we validate the context and `RequestInfo` once at the beginning of `List()` and `Watch()`, before attempting either the remote or local path?

The current `Watch()` implementation creates an empty `RequestInfo` and continues remote processing with an empty path, then checks the original context again before the local fallback. `List()` similarly validates only after the remote request has failed.

An early validation would avoid processing a malformed context, remove the need for the placeholder `RequestInfo`, and avoid repeated calls to `RequestInfoFrom(ctx)`.

Please also handle a truly nil context and add corresponding test coverage. `KeyRootFunc(nil)` can still panic before reaching the current error handling, and the existing test only covers `context.TODO()`.

Norway-02:
Hi @doislong and @WillardHu,

PR #6888, opened 2026-06-20
fix: edge/metaserver: replace Count panic with sentinel error in sqlite store

### LINKED ISSUE #6887: edge/metaserver: store.Count panic("implement me") will crash edgecore on first call
description:

**What happened**:

`edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go:141`:

```go
func (s *store) Count(string) (int64, error) {
    panic("implement me")
}
```

Other unsupported methods in the same file use a different sentinel (`panic("Do not call this function")`) to mark *intentional* refusal. `Count` is marked `"implement me"`, i.e. an unfinished TODO. `Count` is part of `storage.Interface`, which is called by the upstream Kubernetes apiserver machinery; any path that exercises it (or a future k8s.io/apiserver bump that adds a new caller of `Count`) crashes the entire edgecore process.

Either implement against the SQLite backend or return a typed "not supported" error so callers
can degrade gracefully:

```go
func (s *store) Count(key string) (int64, error) {
    return 0, fmt.Errorf("Count is not supported by the edge SQLite store")
}
```

Affected file: `edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go:141`. The "Do not call this function" methods (`GuaranteedUpdate`, `RequestWatchProgress`) are arguably also worth converting to typed errors but `Count` is the most urgent because the comment indicates it's incomplete rather than intentionally rejected.

---

Priyanshubhartistm:
**What type of PR is this?**

/kind bug

The `Count` method in the SQLite-backed `storage.Interface` implementation was left as an unfinished stub with `panic("implement me")` — unlike sibling methods (`Create`, `Delete`, `GuaranteedUpdate`, `RequestWatchProgress`) which intentionally refuse with `panic("Do not call this function")`.

This makes `Count` a crash hazard: any code path in the upstream `k8s.io/apiserver` machinery that calls it, or a future `k8s.io/apiserver` bump that adds a new caller, would crash the entire edgecore process.

fix #6887 
This PR:
- Replaces the panic with a package-level **sentinel error** `ErrCountUnsupported` so callers can degrade gracefully and use `errors.Is` to detect the unsupported operation
- Makes the unused `key` parameter anonymous (`_`) to avoid implying the value affects the result
- Adds a focused unit test that verifies `Count()` returns `0`, a non-nil error matching
  `ErrCountUnsupported`, and never panics

```go
// ErrCountUnsupported is returned by Count to signal that the operation is
// not implemented by this SQLite-backed store. Callers can test for it with
// errors.Is.
var ErrCountUnsupported = errors.New("Count is not supported by the edge SQLite store")

func (s *store) Count(_ string) (int64, error) {
    return 0, ErrCountUnsupported
}

### DIFF
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
@@ -138,8 +138,8 @@ func (s *store) GuaranteedUpdate(context.Context, string, runtime.Object, bool, 
 	panic("Do not call this function")
 }
 
-func (s *store) Count(string) (int64, error) {
-	panic("implement me")
+func (s *store) Count(key string) (int64, error) {
+	return 0, fmt.Errorf("Count is not supported by the edge SQLite store")
 }
 
 func (s *store) RequestWatchProgress(context.Context) error {
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store.go
@@ -2,6 +2,7 @@ package sqlite
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -24,6 +25,11 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/metaserver"
 	"github.com/kubeedge/kubeedge/pkg/metaserver/util"
 )
+
+// ErrCountUnsupported is returned by Count to signal that the operation is
+// not implemented by this SQLite-backed store. Callers can test for it with
+// errors.Is.
+var ErrCountUnsupported = errors.New("Count is not supported by the edge SQLite store")
 
 /*
 This file is designed to encapsulate the Imitator as Store.Interface,
@@ -138,8 +144,8 @@ func (s *store) GuaranteedUpdate(context.Context, string, runtime.Object, bool, 
 	panic("Do not call this function")
 }
 
-func (s *store) Count(key string) (int64, error) {
-	return 0, fmt.Errorf("Count is not supported by the edge SQLite store")
+func (s *store) Count(_ string) (int64, error) {
+	return 0, ErrCountUnsupported
 }
 
 func (s *store) RequestWatchProgress(context.Context) error {
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store_test.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/store_test.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


DoisLONG:
Thanks for replacing the unfinished panic with a regular error return. This is safer for callers and avoids turning an unexpected `Count()` invocation into an EdgeCore process crash.

A couple of small suggestions:

1. Could we add a focused unit test for `Count()` that verifies it returns `0` with a non-nil error and does not panic? Since this PR specifically changes the failure behavior, a regression test would help prevent the panic from being reintroduced later.

2. The PR description calls this a “typed error”, but `fmt.Errorf(...)` returns a regular error value. Either:

   * introduce a package-level sentinel error such as `ErrCountUnsupported`, so callers can use `errors.Is`; or
   * update the PR wording to say that `Count()` now returns an error instead of a typed error.

Also, since the `key` parameter is not used, keeping it anonymous would avoid implying that the value is relevant to the result.

Overall, the direction looks good.

Priyanshubhartistm:
Thanks for the review, @DoisLONG!

I've addressed all three points in the latest commit
1. **Unit test** - added `TestCountReturnsErrorAndDoesNotPanic` in `store_test.go` that verifies
   `Count()` returns `0`, a non-nil error, satisfies `errors.Is(err, ErrCountUnsupported)`,
   and never panics.

2. **Sentinel error** - replaced `fmt.Errorf(...)` with a package-level
   `var ErrCountUnsupported = errors.New(...)` so callers can reliably use `errors.Is`.

3. **Anonymous parameter** - changed `key string` to `_ string` since the value is not used.

Let me know if anything else needs adjustment. Thanks again for the feedback!

Priyanshubhartistm:
Hii @DoisLONG could you please take a look at this PR? I would really appreciate your review and any feedback or suggestions you may have.

Thank you for your time!

DoisLONG:
Thanks for the update.

The sentinel error is appropriate here: it removes the process-level panic, and the current Kubernetes storage-count observer already handles `Count()` errors by reporting an unavailable count rather than terminating the server.

The regression test also covers the intended behavior: it fails if `Count()` panics and verifies the zero result, non-nil error, and `errors.Is` compatibility.

Two non-blocking cleanup notes:

* The PR title still says “typed error”, while the implementation uses a sentinel error.
* Since `store_test.go` is a newly added file, please check whether its copyright year should be 2026 rather than 2024.

I do not see a remaining blocking issue.

Priyanshubhartistm:
Hi @DoisLONG, thank you for your feedback. I've addressed the suggested changes by updating the PR title and the copyright year in store_test.go. Please let me know if anything else needs to be improved. Thanks!

DoisLONG:
Thanks for the update. The latest version addresses the previous review points:

* `Count` now returns a package-level sentinel error instead of panicking;
* callers can identify the unsupported operation with `errors.Is`;
* the unused key parameter is anonymous;
* the regression test verifies the zero result, non-nil sentinel error, and guards against the panic being reintroduced;
* the PR title and new-file copyright year have been corrected.

I do not see any remaining blocking issues.

/lgtm

Priyanshubhartistm:
Thanks for the review, @DoisLONG , @WillardHu

Priyanshubhartistm:
Hii @WillardHu , If everything looks good now, could you please /approve this PR when you have a chance? I'd really appreciate it.

Thanks

Shelley-BaoYue (commented):
/approve

PR #6889, opened 2026-06-20
Fix: delayed session shutdown and ackMessageCache leaks in sendMessageWithRetry

ashnaaseth2325-oss:
# SUMMARY

This PR improves the reliability of CloudHub's ACK delivery path by ensuring `sendMessageWithRetry()` respects session cancellation and properly cleans up cached ACK tracking entries on all non-success exit paths. The changes primarily affect `cloud/pkg/cloudhub/session/node_session.go`, specifically the retry logic used by `NodeSession` when delivering ACK-required messages.

# FIX

**Before**

```go
//Before
select {
case <-ackChan:
    ns.saveSuccessPoint(msg)
    return nil

case <-ticker.C:
    if retryCount == 4 {
        return ErrWaitTimeout
    }
}



//After
select {
case <-ackChan:
    ns.saveSuccessPoint(msg)
    return nil

case <-ns.ctx.Done():
    ns.ackMessageCache.Delete(copyMsg.GetID())
    return fmt.Errorf("session closed while waiting for ack")

case <-ticker.C:
    if retryCount == 4 {
        ns.ackMessageCache.Delete(copyMsg.GetID())
        return ErrWaitTimeout
    }
}
```

### DIFF
--- a/cloud/pkg/cloudhub/session/node_session.go
+++ b/cloud/pkg/cloudhub/session/node_session.go
@@ -345,6 +345,7 @@ func (ns *NodeSession) sendMessageWithRetry(copyMsg, msg *beehivemodel.Message) 
 
 	err := ns.connection.WriteMessageAsync(copyMsg)
 	if err != nil {
+		ns.ackMessageCache.Delete(copyMsg.GetID())
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


ashnaaseth2325-oss:
Hello @kadisi @ls889 
Kindly review this PR whenever you have time.
Thanks!

DoisLONG:
Thanks for improving the ACK retry cleanup and making the wait path respond to session cancellation.

I have two suggestions before maintainer review:

1. The cache cleanup is currently repeated in each failure branch, and the timer is not stopped on every exit path. It would be safer to centralize both with deferred cleanup, for example:

```go
defer ticker.Stop()
defer ns.ackMessageCache.Delete(copyMsg.GetID())
```

This avoids missing cleanup when new return paths are added later. Repeated deletion from `sync.Map` is safe even if the ACK handling path has already removed the entry.

2. Since this change affects retry, timeout, context cancellation, and shared cache state, could we add focused regression tests covering:

   * initial write failure;
   * retry write failure;
   * retry timeout;
   * session cancellation;
   * successful ACK delivery.

The tests should verify both the returned error and that the ACK cache entry is removed on every exit path.

For the cancellation case, it would also be useful to wrap `ns.ctx.Err()` with `%w`, so callers can detect `context.Canceled` using `errors.Is`.

The direction looks good, but test coverage is important for this concurrent retry path.

PR #6891, opened 2026-06-21
Fix: cloud/edgecontroller: handle json.Marshal error for GPU status annotation

### LINKED ISSUE #6890: cloud/edgecontroller: upstream drops json.Marshal error and overwrites NvidiaGPUStatus annotation with ""
### What happened:

cloud/pkg/edgecontroller/controller/upstream.go:679–681:

```go
if len(gpuStatus) > 0 {
    data, _ := json.Marshal(gpuStatus)
    getNode.Annotations[constants.NvidiaGPUStatusAnnotationKey] = string(data)
}
```

The error from json.Marshal is discarded. If the call ever fails, data is nil and the GPU status annotation is silently overwritten with the empty string "", erasing previous GPU information from the Node object. No log line is emitted.

### What you expected to happen:

Don't overwrite the annotation if marshalling fails; log and skip:

```go
data, err := json.Marshal(gpuStatus)
if err != nil {
    klog.Warningf("message: %s, marshal GPU status failed: %v", msg.GetID(), err)
    continue
}
getNode.Annotations[constants.NvidiaGPUStatusAnnotationKey] = string(data)
```

### Anything else we need to know?:

json.Marshal on the local []types.NvidiaGPUStatus would not realistically fail in the current code, but the same file does check the error on the very next json.Marshal call (line 683–685) - fixing this for consistency hardens the path against future schema changes.

### Affected file:

cloud/pkg/edgecontroller/controller/upstream.go:679.

Priyanshubhartistm:
## What type of PR is this?

/kind bug

In `updateNodeStatus` (`cloud/pkg/edgecontroller/controller/upstream.go`), the error return from `json.Marshal(gpuStatus)` was silently discarded with the blank identifier `_`. If marshalling were to fail, `data` would be `nil` and `string(nil)` evaluates to `""`, causing the existing `huawei.com/gpu-status` Node annotation to be silently overwritten with an empty string - erasing all GPU health information with no log output.

This PR fixes the discarded error by:

1. Capturing the `err` return value.
2. Emitting a `klog.Warningf` with the message ID so operators can trace the
   event.
3. Skipping the annotation write via `continue` so the pre-existing value is
   preserved.

The change also makes the first `json.Marshal` call consistent with the immediately following `json.Marshal(v)` call at line 687, which already checks and logs its error. A unit test covering the happy-path annotation
write is added in `upstream_test.go`.

Fixes #6890 

## Special notes for your reviewer

`json.Marshal` on `[]types.NvidiaGPUStatus` (a slice of a struct with only `string` and `bool` fields) cannot realistically fail in the current code.

The fix is a defensive hardening change for consistency and correctness against future schema changes, not a regression fix.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -572,6 +572,24 @@ func (uc *UpstreamController) createNode(nodeID, name string, node *v1.Node) (*v
 	return node, err
 }
 
+// marshalGPUStatus marshals GPU status entries to JSON. It is a package-level
+// variable, rather than a direct call to json.Marshal, so tests can substitute
+// a failing implementation: []types.NvidiaGPUStatus only has string/bool
+// fields and can never actually fail to marshal.
+var marshalGPUStatus = json.Marshal
+
+// setGPUStatusAnnotation marshals gpuStatus and stores it under
+// constants.NvidiaGPUStatusAnnotationKey on node. If marshalling fails, it
+// logs a warning and leaves the existing annotation value untouched.
+func setGPUStatusAnnotation(node *v1.Node, msgID string, gpuStatus []types.NvidiaGPUStatus) {
+	data, err := marshalGPUStatus(gpuStatus)
+	if err != nil {
+		klog.Warningf("message: %s, marshal GPU status failed: %v", msgID, err)
+		return
+	}
+	node.Annotations[constants.NvidiaGPUStatusAnnotationKey] = string(data)
+}
+
 // updateNodeStatus update node status
 // Deprecated: updateNodeStatus will be deleted in subsequent versions, use patchNode instead.
 func (uc *UpstreamController) updateNodeStatus() {
@@ -676,8 +694,7 @@ func (uc *UpstreamController) updateNodeStatus() {
 							gpuStatus = append(gpuStatus, types.NvidiaGPUStatus{ID: er.Name, Healthy: true})
 						}
 						if len(gpuStatus) > 0 {
-							data, _ := json.Marshal(gpuStatus)
-							getNode.Annotations[constants.NvidiaGPUStatusAnnotationKey] = string(data)
+							setGPUStatusAnnotation(getNode, msg.GetID(), gpuStatus)
 						}
 					}
 					data, err := json.Marshal(v)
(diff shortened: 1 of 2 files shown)


DoisLONG:
Thanks for handling the marshal error without skipping the remaining extend-resource annotation update. The revised `if/else` flow addresses the previous review concern.

A few test-related suggestions:

1. The new test currently uses a fixed one-second sleep before reading the updated Node. Since the status processing is asynchronous, it would be more reliable and usually faster to poll with `require.Eventually` or an equivalent retry loop until the expected annotation appears.

2. Please use `constants.NvidiaGPUStatusAnnotationKey` in the test instead of the literal `"huawei.com/gpu-status"`, so the test stays aligned if the annotation key changes later.

3. The added test verifies the existing successful annotation path, but it does not exercise the newly added marshal-error handling. Since `json.Marshal([]types.NvidiaGPUStatus)` cannot realistically fail with the current schema, that is understandable, but the test should be described as happy-path regression coverage rather than coverage of the new error branch.

No further concerns with the production-code change from my side, pending CI and maintainer review.

Priyanshubhartistm:
Thanks for the review, @DoisLONG!
All three points addressed - Swapped the `time.Sleep` out for `require.Eventually` with a 50ms poll, so the test no longer burns a full second unconditionally. Also pulled in `constants.NvidiaGPUStatusAnnotationKey` everywhere the literal was used - should've done that from the start. Added a doc comment on the test too, making it clear this is happy-path regression coverage and that the marshal-error path isn't tested (and why it can't realistically be).

Appreciate the review!

WillardHu:
cc @DoisLONG

DoisLONG (commented):
This PR currently has multiple commits. Could you please squash them into a single logical commit?

Priyanshubhartistm:
Hi @DoisLONG, I've addressed the previous feedback and squashed the commits. Could you please take another look at this PR? If there are any further changes needed, please let me know.
 Thank you!

DoisLONG (commented):
Thanks for the update.

I rechecked the latest version. The PR is now squashed into a single commit, and the code still looks good to me.

The GPU status annotation update is now handled through `setGPUStatusAnnotation()`, which keeps the existing annotation untouched when marshalling fails. The normal extend-resource annotation update is not skipped by that failure path.

The updated tests also cover both sides now: the async `updateNodeStatus` happy path uses `require.Eventually`, and the direct unit test covers both successful marshal and injected marshal-failure behavior.


/lgtm

PR #6892, opened 2026-06-21
Update keadm ctl command descriptions and help

### LINKED ISSUE #6708: Command behavior doesn't match expectation - unsure if it's a bug, missing docs, or user error
I tried using keadm ctl logs to view logs from a pod running on an edge node, but the command either returned no output, hung indefinitely, or gave an error I didn't understand. I expected it to show the pod logs similarly to kubectl logs. I wasn't sure whether I was using it wrong, whether the feature isn't fully implemented, whether I needed to configure something first, or if this was actually a bug that should be reported.

1. Deploy a pod on an edge node that produces logs (e.g., a simple nginx pod)
2. Try keadm ctl logs <pod-name> -n <namespace> from the cloud side
3. Observe one of the following:
    - Command hangs with no output
    - Returns "error: not found" even though the pod exists
    - Returns logs but only partial or incomplete
    - Returns logs but they appear to be from the wrong pod
4. Check keadm ctl logs --help - minimal information available
5. Search documentation for "keadm ctl logs" - no clear guidance found

I encountered similar uncertainty with other commands:

- `keadm ctl exec` sometimes works, sometimes gives cryptic errors about WebSocket connections
- `keadm ctl describe `works for some resources but not others without explanation
- `keadm ctl top node` isn't available at all, but I wasn't sure if it's supposed to be

**Environment**:
- Kubernetes version (use `kubectl version`): v1.28.13
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): v1.20.0

SurbhiAgarwal1:
**What type of PR is this?**

/kind documentation

Updates the CLI help documentation for `keadm ctl logs`, `exec`, `describe`, and the root `ctl` commands. Previously, the minimal `--help` text caused confusion about their expected behavior and environment (e.g., users trying to run them from the cloud side). 

This PR adds comprehensive `Long` descriptions and `Example` blocks to clarify that these commands are exclusively intended to be run on edge nodes and require the `MetaServer` module to be enabled in `edgecore.yaml`. It also points cloud-side users to the appropriate `kubectl` commands.

Fixes #6708

The updates are purely text-based documentation enhancements to the cobra commands. No functional logic was changed.

```release-note
Add detailed help descriptions to `keadm ctl` commands to clarify edge node usage and MetaServer prerequisites

### DIFF
--- a/keadm/cmd/keadm/app/cmd/ctl/ctl.go
+++ b/keadm/cmd/keadm/app/cmd/ctl/ctl.go
@@ -30,13 +30,20 @@ import (
 )
 
 var ctlShortDescription = `Commands operating on the data plane at edge`
+var ctlLongDescription = `Commands operating on the data plane at the edge.
+
+The 'keadm ctl' commands are designed to be executed directly on edge nodes where EdgeCore is installed and running. They interact with the local MetaServer API to perform operations such as retrieving logs, executing commands, and querying resource statuses locally, even when disconnected from the cloud.
+
+Note:
+- Ensure the MetaServer module is enabled in your edgecore.yaml configuration.
+- To manage edge nodes and pods from the cloud side, please use standard 'kubectl' commands instead.`
 
 // NewCtl returns KubeEdge edge pod command.
 func NewCtl() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "ctl",
 		Short: ctlShortDescription,
-		Long:  ctlShortDescription,
+		Long:  ctlLongDescription,
 	}
 
 	cmd.AddCommand(get.NewEdgeGet())
--- a/keadm/cmd/keadm/app/cmd/ctl/ctl.go
+++ b/keadm/cmd/keadm/app/cmd/ctl/ctl.go
@@ -32,7 +32,7 @@ import (
 var ctlShortDescription = `Commands operating on the data plane at edge`
 var ctlLongDescription = `Commands operating on the data plane at the edge.
 
-The 'keadm ctl' commands are designed to be executed directly on edge nodes where EdgeCore is installed and running. They interact with the local MetaServer API to perform operations such as retrieving logs, executing commands, and querying resource statuses locally, even when disconnected from the cloud.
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by DoisLONG** [comment]
Thanks for improving the `keadm ctl` help text. The added descriptions and examples make the intended edge-side usage and MetaServer prerequisite much clearer.

I have a few suggestions before maintainer review:

1. Please consider updating the Cobra `Use` fields so the required positional arguments are visible directly in the generated usage text. For example:

   * `logs <pod-name>`
   * `exec <pod-name> -- <command> [args...]`

2. The root description currently suggests that the commands can operate even when disconnected from the cloud. This may be too broad, since offline behavior still depends on the required resource data being available locally and the relevant edge-side operation being supported. It would be safer to qualify this as “some operations” rather than implying that all `keadm ctl` commands are guaranteed to work offline.

3. Please also close the release-note fenced code block in the PR description.

Overall, this is a useful documentation improvement for `keadm ctl`.

**Comment by SurbhiAgarwal1** [comment]
@DoisLONG Thanks for the review! I've just pushed a commit that addresses your feedback:

1. Updated the Cobra `Use` fields for both `logs` and `exec` commands so the positional arguments are visible in the usage text.
2. Updated the root description for `keadm ctl` to clarify that only "some operations" are guaranteed to work offline when disconnected from the cloud, rather than implying all commands do.
3. Closed the `release-note` fenced code block in the PR description to fix the formatting issue.

Let me know if anything else is needed!

**Comment by DoisLONG** [comment]
Thanks for the update. The previous feedback on the long descriptions and example formatting has been addressed.

One small consistency issue remains: `ctlShortDescription` still uses `at edge`, while the new long description uses the corrected phrase `at the edge`.

Could you please update it as well?

```go
var ctlShortDescription = `Commands operating on the data plane at the edge`
```

This short description is also visible in the parent command help output, so keeping both descriptions consistent would be better.

SurbhiAgarwal1:
@DoisLONG Thanks for catching that! I've just pushed a commit to update ctlShortDescription to "Commands operating on the data plane at the edge", keeping it perfectly consistent with the long description. Let me know if anything else is needed!

PR #6893, opened 2026-06-22
mapper-framework: reposition to SDK

### LINKED ISSUE #6564: Reposition the mapper-framework project as an SDK

Develop mapper-framework into an SDK that users can depend on.

Image

**Why is this needed**:

The mapper-framework project provides as a template for a mapper project, copying all files from the `_template` directory to a new mapper project. This approach presents two problems:
1. The source code in the `_template` directory becomes difficult to maintain within the KubeEge project due to nested Go projects;
2. After the project code is updated, the generated mapper project is difficult to update;

However, we can see that in mapper-framework, only the `devicetype.go` and `driver.go` files under the `driver` package actually need modification, other files generally don't require further modification. Therefore, mapper-framework is better suited as an SDK.

SurbhiAgarwal1:
**What type of PR is this?**

/kind feature
/kind cleanup

Repositions the `mapper-framework` as a Go SDK to resolve maintainability issues with nested Go modules and to make it easier for developers to upgrade their mappers.

Specifically, this PR:
- Extracts the boilerplate code from `_template/mapper` into reusable SDK packages: `pkg/mapper/device` and `pkg/mapper/data`.
- Abstracts the concrete `CustomizedClient` into a generic `driver.Adapter` interface that developers can implement for their protocols.
- Condenses application bootstrapping into a single `mapper.Run(factory func(protocol []byte) (driver.Adapter, error))` function, significantly reducing the boilerplate needed in a user's `main.go`.
- Deletes the `_template` folder and `generate.sh` scripts, as mappers no longer need to be copied from templates.

Fixes #6564

This is an architectural shift that removes the `_template` directory entirely. Existing mappers generated using the template will need to be refactored to use the new SDK pattern when upgrading to this version of the framework.

```release-note
Repositioned the mapper-framework from a template-generation approach into a Go SDK. Developers must now implement the `driver.Adapter` interface and use `mapper.Run()` to start their custom mappers.


DoisLONG:
Thanks for working on the mapper-framework SDK direction. Moving reusable mapper logic out of the generated template can reduce duplicated code and make future framework upgrades easier.

However, this PR is a breaking architectural change and is not ready to merge in its current form.

The PR removes the complete mapper template, generation scripts, Dockerfiles, configuration example, and deployment manifests, but does not provide an equivalent SDK-based example or migration guide. Developers need at least one minimal, buildable mapper showing:

* how to implement `driver.Adapter`;
* how to parse protocol and visitor configuration;
* how to call `mapper.Run`;
* the expected `config.yaml`;
* Docker build and Kubernetes deployment manifests;
* how an existing generated mapper should migrate to the SDK.

There are also several API and lifecycle concerns:

1. `mapper.Run()` is now a public SDK entry point, but it still calls `klog.Fatal` internally. A reusable SDK function should preferably return an error and let the application decide whether to exit.
2. The adapter factory is not validated. A nil factory, or a factory returning `(nil, nil)`, will result in a runtime panic when the adapter is used.
3. `DevPanel` remains a package-level singleton initialized through `sync.Once`. Since `adapterFactory` is now constructor input, only the first factory will ever take effect. This makes isolated tests and multiple mapper instances in one process difficult.
4. This PR adds no unit tests for the new `Adapter` abstraction, factory failure paths, mapper startup, device initialization, or migration behavior.
5. Please address and test the existing concurrency concerns around `serviceMutex`, nested calls, and concurrent access to the device maps.

Before removing the template-based workflow, please add:

* a buildable example mapper;
* a migration document;
* focused unit tests for the new SDK API;
* a clear compatibility and versioning statement for existing mapper users.

Minor cleanup:

* fix the bot-reported invalid commit message;
* close the `release-note` code block in the PR description.

The SDK direction may be valuable, but the replacement developer experience and regression coverage need to be complete before the existing template workflow is removed.

PR #6896, opened 2026-06-22
synccontroller: guard nil message on object sync update path

### LINKED ISSUE #6895: synccontroller: nil-pointer panic in sendEvents when BuildResource fails on update path
**What happened**:

In `cloud/pkg/synccontroller`, `sendEvents` (and its cluster-scoped sibling `sendClusterObjectSyncEvent`) dereference the message returned by `buildEdgeControllerMessage` without a nil check on the update path:

```go
// cloud/pkg/synccontroller/objectsync.go
msg := buildEdgeControllerMessage(nodeName, sync.Namespace, resourceType, sync.Spec.ObjectName, model.UpdateOperation, obj)
```

`buildEdgeControllerMessage` returns `nil` when `messagelayer.BuildResource` fails (it logs a warning and returns nil):

```go
resource, err := messagelayer.BuildResource(nodeName, namespace, resourceType, resourceName)
if err != nil {
    klog.Warningf("build message resource failed with error: %s", err)
    return nil
}
```

So when `BuildResource` fails, `*msg` panics the syncController goroutine in cloudcore.

The delete path in the same file already guards against this correctly:

```go
if msg := buildEdgeControllerMessage(...); msg != nil {
} else {
    ...
}
```

The update path is missing the equivalent guard. The same asymmetry exists in `clusterobjectsync.go`: the delete path checks `msg != nil`, the update path (`sendClusterObjectSyncEvent`) does not.

When `buildEdgeControllerMessage` returns nil, the update path should log and skip sending instead of dereferencing nil, the same way the delete path does.

`BuildResource` returns an error when `nodeID`, `namespace`, or `resourceType` is empty. A unit test calling `sendEvents` with an empty node name reaches the nil dereference and panics on master (`1a3f39fdd`).

Affected lines on master:
- `cloud/pkg/synccontroller/objectsync.go` — `sendEvents`, update path
- `cloud/pkg/synccontroller/clusterobjectsync.go` — `sendClusterObjectSyncEvent`, update path

I have a fix that mirrors the existing nil-guard from the delete path, plus unit tests, and can open a PR.

**Environment**:
- KubeEdge version: master (`1a3f39fdd`)

santhil-cyber:
**What type of PR is this?**

/kind bug

`buildEdgeControllerMessage` returns `nil` when `messagelayer.BuildResource` fails. The update path in `sendEvents` (`objectsync.go`) and `sendClusterObjectSyncEvent` (`clusterobjectsync.go`) dereferenced the returned message without a nil check, so a failed `BuildResource` panics the syncController goroutine in cloudcore.

The delete path in both files already guards this with `if msg != nil`. This PR mirrors that guard on the update path: log and skip sending when the message is nil, matching the existing pattern.

Adds unit tests asserting neither function panics when message build returns nil.

Fixes #6895

`TestSendEventsNilMessage` triggers the nil return naturally by passing an empty node name (`BuildResource` errors on empty `nodeID`/`namespace`/`resourceType`). `TestSendClusterObjectSyncEventNilMessage` uses the existing `buildEdgeControllerMessageFunc` test seam to return nil. Verified the tests fail (panic) without the fix and pass with it.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/synccontroller/clusterobjectsync.go
+++ b/cloud/pkg/synccontroller/clusterobjectsync.go
@@ -126,6 +126,10 @@ func sendClusterObjectSyncEvent(nodeName string, sync *v1alpha1.ClusterObjectSyn
 		// trigger the update event
 		klog.V(4).Infof("The resourceVersion: %s of %s in K8s is greater than in edgenode: %s, send the update event", objectResourceVersion, resourceType, sync.Status.ObjectResourceVersion)
 		msg := buildEdgeControllerMessageFunc(nodeName, models.NullNamespace, resourceType, sync.Spec.ObjectName, model.UpdateOperation, obj)
+		if msg == nil {
+			klog.Warningf("failed to build update message for %s %s, skip sending update event", resourceType, sync.Spec.ObjectName)
+			return
+		}
 		sendToEdge(commonconst.DefaultContextSendModuleName, *msg)
 	}
 }
--- a/cloud/pkg/synccontroller/clusterobjectsync_test.go
+++ b/cloud/pkg/synccontroller/clusterobjectsync_test.go
@@ -429,6 +429,37 @@ func TestSendClusterObjectSyncEventDirect(t *testing.T) {
 	assert.True(t, sendCalled, "Send should have been called")
 }
 
+func TestSendClusterObjectSyncEventNilMessage(t *testing.T) {
+	backup := setupTest()
+	defer backup.restore()
+
+	compareResourceVersionFunc = func(rv1, rv2 string) int {
+		if rv1 == "1000" && rv2 == "500" {
+			return 1
+		}
+		return 0
+	}
+
+	sendCalled := false
+	sendToEdge = func(module string, msg model.Message) {
+		sendCalled = true
+	}
+
+	// buildEdgeControllerMessageFunc returns nil when BuildResource fails;
+	// the update path must skip sending instead of dereferencing nil.
+	buildEdgeControllerMessageFunc = func(nodeID, namespace, resource, resourceID string, operation string, content interface{}) *model.Message {
+		return nil
+	}
+
+	obj := createTestPod("test-pod", "12345", "1000")
+	sync := createTestSync("node1-pod-12345", "test-pod", "Pod", "v1", "500")
+
+	assert.NotPanics(t, func() {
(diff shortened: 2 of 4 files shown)


DoisLONG:
Thanks for the fix.

The nil checks added to both update paths look appropriate and make the behavior consistent with the existing delete paths. Skipping the send operation when message construction fails is preferable to dereferencing a nil message and panicking the syncController goroutine.

I noticed one test robustness issue in `TestSendEventsNilMessage`.

The ClusterObjectSync test directly mocks `buildEdgeControllerMessageFunc` to return `nil` and also verifies that the send function is not called. However, the ObjectSync test relies on passing an empty `nodeName` and assumes that the current implementation of `messagelayer.BuildResource` will reject it.

This makes the test dependent on the internal validation behavior of `BuildResource`. If that validation changes in the future and an empty node name no longer produces a nil message, the test could still pass without exercising the new nil-guard branch.

Since `buildEdgeControllerMessageFunc` already exists as a package-level test seam, could `sendEvents` use it consistently as well? The test could then explicitly mock it to return `nil`, restore it after the test, and verify that the send operation is skipped.

For example, the test should explicitly cover these two expectations:

1. `sendEvents` does not panic when the message builder returns `nil`;
2. no message is sent when message construction fails.

This would make the regression test deterministic and consistent with `TestSendClusterObjectSyncEventNilMessage`.

santhil-cyber:
Good point, thanks. You're right that keying the test off an empty nodeName ties it to BuildResource's validation rather than the nil-guard itself.

I've switched the update path in sendEvents to call buildEdgeControllerMessageFunc, the same seam clusterobjectsync.go already uses, and reworked TestSendEventsNilMessage to mock it to return nil and restore it afterward. The status/object resourceVersions ("1" < "2") drive the update branch so the builder is actually invoked. With the nil-guard the function returns before beehiveContext.Send, so it neither dereferences nor sends; I confirmed the test panics without the guard and passes with it.

santhil-cyber:
Thanks for the review @DoisLONG ! Let me know if there's anything else needed, or if this is ready to get it merged.

santhil-cyber:
/assign @fisherxu

@DoisLONG lgtm’d this on Jun 25. DCO passes, there are no conflicts with master, and tide reports the `approved` label as the only thing outstanding. Could you take a look when you have a moment?

PR #6897, opened 2026-06-22
fix(edgecore): correct token expiry duration unit to seconds

Norway-02:
**What type of PR is this?**
/kind bug

The token expiry was incorrectly calculated or evaluated using the wrong time unit, potentially leading to tokens expiring much earlier or later than intended. This PR explicitly enforces the expected duration unit in seconds to ensure consistent token lifecycle management.

NONE

Squashed the original commits into a single atomic commit.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/servers/httpserver/pre_server.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/pre_server.go
@@ -168,15 +168,20 @@ func createCertsToSecret(ctx context.Context) error {
 
 // GenerateAndRefreshToken creates a token and save it to secret, then create a timer to refresh the token.
 func GenerateAndRefreshToken(ctx context.Context) error {
-	if err := createNewToken(ctx); err != nil {
+	tokenDuration := hubconfig.Config.CloudHub.TokenRefreshDuration
+	if tokenDuration < time.Minute {
+		tokenDuration *= time.Hour
+	}
+
+	if err := createNewToken(ctx, tokenDuration); err != nil {
 		return err
 	}
-	t := time.NewTicker(time.Hour * hubconfig.Config.CloudHub.TokenRefreshDuration)
+	t := time.NewTicker(tokenDuration)
 	go func() {
 		for {
 			select {
 			case <-t.C:
-				if err := createNewToken(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


DoisLONG:
Thanks for working on this issue. I reviewed the current diff and noticed a few points that may need to be addressed before the token-duration change can be evaluated clearly.

First, the PR currently contains changes that appear unrelated to token expiry:

* `pkg/security/certs/factory.go` changes `HanndlerType` / `HandlerType`;
* `pkg/viaduct/pkg/conn/quic.go` changes error handling in `WriteMessageSync`.

Since the stated purpose of this PR is to fix token expiry duration handling, could these unrelated changes be removed or submitted separately? Keeping this PR limited to the token-related files would make the behavior and regression risk much easier to review.

Second, the current token test changes the input from `1` to `time.Hour`, but it only verifies that a non-empty token is returned. It does not verify the behavior being changed here. It would be useful to decode the generated JWT claims and assert that `ExpiresAt` is approximately:

```go
startTime.Add(2 * interval)
```

This would provide a direct regression test for the expiration calculation.

I also think the expected unit of `TokenRefreshDuration` needs to be clarified before moving the `time.Hour` multiplication to the call site. If the configuration accepts a standard duration value such as `12h`, multiplying it by `time.Hour` again would be incorrect. Rather than inferring the unit from the duration's magnitude, it may be safer to define one canonical configuration format, normalize it once at the configuration boundary, and use the resulting `time.Duration` consistently for both the refresh ticker and `token.Create`.

Minor PR metadata items:

* `Fixes #` is currently incomplete;
* the `release-note` code block appears to be missing its closing triple backticks.

Norway-02:
Thanks for the review.

1. Removed pkg/security/certs/factory.go 
    and pkg/viaduct/pkg/conn/quic.go from this PR.

2. Added regression test that decodes JWT claims and asserts ExpiresAt ≈ startTime.Add(2 * interval).

3. TokenRefreshDuration unit ambiguity predates this PR — field is time.Duration but documented and used as hours at call site. Fixing that is a separate change.

4. Fixed Fixes # and release-note backticks.

Norway-02:
Hi @doislong and @WillardHu,

PR #6863, opened 2026-06-12
fix: add devicestatuses permission to cloudcore rbac

RedZapdos123:
## Description:

Issue #6862 reported that `cloudcore` RBAC was missing permission for `devicestatuses.devices.kubeedge.io`, which caused repeated forbidden errors when the `DeviceStatus` informer started.

This PR updates the cloudcore RBAC manifests to include `devicestatuses` in all affected manifest variants:

- `build/cloud/03-clusterrole.yaml`
- `build/cloud/ha/01-ha-prepare.yaml`
- `manifests/charts/cloudcore/templates/rbac_cloudcore.yaml`

The fix is minimal and limited to the affected RBAC resource lists.

## Validation:

- `helm lint manifests/charts/cloudcore`
- rendered chart inspection to confirm `devicestatuses` is present in cloudcore RBAC
- fresh `kind` cluster deployment on latest `master`, where:
  - `kubectl auth can-i --as=system:serviceaccount:kubeedge:cloudcore list devicestatuses.devices.kubeedge.io` returned `yes`
  - `cloudcore` no longer logged `devicestatuses.devices.kubeedge.io is forbidden`
- `kubectl apply --dry-run=client -f build/cloud/03-clusterrole.yaml`
- `kubectl apply --dry-run=client -f build/cloud/ha/01-ha-prepare.yaml`

Closes #6862.

## Checklist:

- [x] I have run validation using `helm lint manifests/charts/cloudcore`.
- [x] I have validated the fix in a fresh `kind` cluster on latest `master`.

## Release note:

```release-note
CloudCore RBAC now includes permission for devicestatuses.devices.kubeedge.io to prevent informer forbidden errors.
```

PR #6898, opened 2026-06-22
fix: add devicestatuses resource to cloudcore RBAC ClusterRole

zgfh:
## Summary
- Add `devicestatuses` to the `devices.kubeedge.io` resources in cloudcore ClusterRole
- The DeviceStatus CRD was extracted from `devices/status` as a separate CRD, but RBAC rules were not updated accordingly
- This causes cloudcore to fail with: `cannot list resource "devicestatuses" in API group "devices.kubeedge.io"`

## Test plan
- [ ] Deploy cloudcore with the updated RBAC and verify no `devicestatuses` permission errors in logs

### DIFF
--- a/manifests/charts/cloudcore/templates/rbac_cloudcore.yaml
+++ b/manifests/charts/cloudcore/templates/rbac_cloudcore.yaml
@@ -24,7 +24,7 @@ rules:
     resources: ["leases"]
     verbs: ["get", "list", "watch", "create", "update"]
   - apiGroups: ["devices.kubeedge.io"]
-    resources: ["devices", "devicemodels", "devices/status", "devicemodels/status"]
+    resources: ["devices", "devicemodels","devicestatuses", "devices/status", "devicemodels/status"]
     verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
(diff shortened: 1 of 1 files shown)


DoisLONG:
Thanks for working on this issue.

This PR appears to overlap with #6863, which is already addressing the same missing `devicestatuses` permission.

One important difference is that the RBAC rule exists in multiple installation manifests. This PR currently updates only the Helm chart template, while #6863 updates all three affected locations:

* `build/cloud/03-clusterrole.yaml`
* `build/cloud/ha/01-ha-prepare.yaml`
* `manifests/charts/cloudcore/templates/rbac_cloudcore.yaml`

#6863 also includes validation with Helm lint, manifest dry-runs, `kubectl auth can-i`, and a CloudCore deployment check.

To avoid maintaining two overlapping fixes, it may be better to continue the work in #6863 and close this PR as a duplicate, unless there is a specific reason why #6863 should not be used.

There is also a minor formatting issue in the current diff: a space is missing between `"devicemodels",` and `"devicestatuses"`, but the duplicate and incomplete manifest coverage are the more important concerns.

PR #6900, opened 2026-06-22
fix: pkg/viaduct: WSConnection.handleRawData uses ProtocolTypeWS instead of ProtocolTypeQuic

### LINKED ISSUE #6899: pkg/viaduct: WSConnection.handleRawData uses ProtocolTypeQuic, causing nil lane and panic
## What happened

`pkg/viaduct/pkg/conn/ws.go` is the WebSocket connection implementation. Every call in the file constructs a lane with `api.ProtocolTypeWS` - except `handleRawData`, which mistakenly passes`api.ProtocolTypeQuic`:

```go
// pkg/viaduct/pkg/conn/ws.go  line 93
_, err := io.Copy(conn.consumer, lane.NewLane(api.ProtocolTypeQuic, conn.wsConn))
```

`lane.NewLane(api.ProtocolTypeQuic, van)` calls `NewQuicLane(van)`, which does a type assertion to `quic.Stream`. Since `conn.wsConn` is a `*websocket.Conn`, the assertion fails, `klog.Error("oops! bad type of van")` is emitted, and `nil` is returned.

`io.Copy(conn.consumer, nil)` then panics with a nil interface dereference.

`handleRawData` should use `api.ProtocolTypeWS`, consistent with every other call in the same file, so `NewLane` builds a `WSLane` over `conn.wsConn` instead of returning `nil`.

## How to reproduce

Create a `WSConnection` with `connUse = api.UseTypeStream`, `autoRoute = true`, and a
non-nil `consumer`, then call `ServeConn()`. `handleRawData` is reached and panics.

## Root cause

Almost certainly a copy/paste from the QUIC equivalent in `pkg/viaduct/pkg/conn/quic.go`. All sibling methods in `ws.go` use `ProtocolTypeWS` correctly - only `handleRawData` carries the wrong constant.

## Environment

- Component: `pkg/viaduct`
- File: `pkg/viaduct/pkg/conn/ws.go`

Priyanshubhartistm:
## What type of PR is this?
/kind bug
fix #6899 

In `pkg/viaduct/pkg/conn/ws.go`, `handleRawData` was passing `api.ProtocolTypeQuic` to `lane.NewLane` instead of `api.ProtocolTypeWS`:

```go
// before (buggy)
_, err := io.Copy(conn.consumer, lane.NewLane(api.ProtocolTypeQuic, conn.wsConn))

// after (fixed)
_, err := io.Copy(conn.consumer, lane.NewLane(api.ProtocolTypeWS, conn.wsConn))

### DIFF
--- a/pkg/viaduct/pkg/conn/ws.go
+++ b/pkg/viaduct/pkg/conn/ws.go
@@ -90,7 +90,7 @@ func (conn *WSConnection) handleRawData() {
 	}
 
 	// TODO: support control message processing in raw data mode
-	_, err := io.Copy(conn.consumer, lane.NewLane(api.ProtocolTypeQuic, conn.wsConn))
+	_, err := io.Copy(conn.consumer, lane.NewLane(api.ProtocolTypeWS, conn.wsConn))
 	if err != nil {
 		klog.Errorf("failed to copy data, error: %+v", err)
 		conn.state.State = api.StatDisconnected
--- a/pkg/viaduct/pkg/conn/ws_test.go
+++ b/pkg/viaduct/pkg/conn/ws_test.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2024 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package conn
+
+import (
+	"bytes"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/gorilla/websocket"
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by DoisLONG** [comment]
Thanks for the update.

The production change looks correct. `handleRawData()` is processing a WebSocket connection, so using `api.ProtocolTypeWS` instead of `api.ProtocolTypeQuic` prevents `lane.NewLane` from returning an invalid lane for the underlying `*websocket.Conn`.

The updated `wsTestPair` helper also addresses the previously reported data race and removes the timing dependency on `time.Sleep` by passing the server-side connection through a channel.

I noticed one small resource-cleanup issue in the current test. The client connection and the `httptest.Server` are registered with `t.Cleanup`, but the server-side `*websocket.Conn` returned from `serverChan` is not explicitly closed. Since an upgraded WebSocket connection is hijacked from the HTTP server, it would be safer to register cleanup for it as well:

```go
case s := <-serverChan:
    t.Cleanup(func() {
        _ = s.Close()
    })
    return s, c
```

One additional non-blocking test improvement: instead of only checking `consumer.Len() > 0`, the test could compare the received bytes with the expected payload. That would verify that the complete and correct data was forwarded, rather than only verifying that some data was received.

For example:

```go
if !bytes.Equal(consumer.Bytes(), payload) {
    t.Errorf("unexpected payload: got %q, want %q", consumer.Bytes(), payload)
}
```

Priyanshubhartistm:

WillardHu:
Please fix the lint error.

PR #6902, opened 2026-06-22
fix: cloud/cloudhub: correct "Node founded" typo to "Node found" in checknode response

### LINKED ISSUE #6901: loud/cloudhub: checknode endpoint returns "Node founded" instead of "Node found"
## What happened

`cloud/pkg/cloudhub/servers/httpserver/node/checknode.go` returns the literal string `"Node founded"` on the success path when a node exists:

```go
// checknode.go
err = response.WriteErrorString(http.StatusOK, "Node founded")
```

This is a user-visible response body sent to every caller of the `/checknode/{nodename}` endpoint.

The corresponding test (`checknode_test.go` ) correctly expects
`"Node found"`:

```go
expectedBody: "Node found",
...
assert.Contains(t, httpWriter.Body.String(), tt.expectedBody)
```

The test passes **only because `assert.Contains` is a substring check** and `"Node founded"` happens to contain `"Node found"`. The intent of the test confirms the production string is wrong.

The response body on the success path should be `"Node found"`.

## How to reproduce

```bash
curl http://<cloudhub-host>/node/checknode/<existing-node-name>
# Response body: "Node founded"   ← wrong
# Expected:      "Node found"
```

## Root cause

Typo in `checknode.go` - `"Node founded"` should be `"Node found"`.

The bug is invisible to the existing test suite because `assert.Contains` performs a substring match, not an exact equality check. Changing the assertion to `assert.Equal` would have caught this at the time it was
introduced.

## Suggested fix

**`cloud/pkg/cloudhub/servers/httpserver/node/checknode.go`** — fix the string:

```go
// before
err = response.WriteErrorString(http.StatusOK, "Node founded")

// after
err = response.WriteErrorString(http.StatusOK, "Node found")
```

**`checknode_test.go`** — strengthen the assertion so this cannot regress:

```go
// before
assert.Contains(t, httpWriter.Body.String(), tt.expectedBody)

// after
assert.Equal(t, tt.expectedBody, httpWriter.Body.String())
```

## Secondary note

`response.WriteErrorString(http.StatusOK, ...)` on the success path is semantically odd — using an "error" writer for a 200 OK response. Replacing it with `response.WriteHeader(http.StatusOK)` followed by `response.Write([]byte("Node found"))` would be cleaner, but the typo fix is the user-visible change that matters most.

## Environment

- Component: `cloud/pkg/cloudhub`
- File: `cloud/pkg/cloudhub/servers/httpserver/node/checknode.go` 
- Test: `cloud/pkg/cloudhub/servers/httpserver/node/checknode_test.go`

Priyanshubhartistm:
## What type of PR is this?
/kind bug

fix #6901

`checknode.go` returned `"Node founded"` on the success path — a grammatically wrong, user-visible string sent to every caller of the `/checknode/{nodename}` endpoint:

```go
// before (buggy)
err = response.WriteErrorString(http.StatusOK, "Node founded")

// after (fixed)
err = response.WriteErrorString(http.StatusOK, "Node found")

### DIFF
--- a/cloud/pkg/cloudhub/servers/httpserver/node/checknode.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/node/checknode.go
@@ -59,7 +59,7 @@ func CheckNode(request *restful.Request, response *restful.Response) {
 	}
 
 	// node exists return success
-	err = response.WriteErrorString(http.StatusOK, "Node founded")
+	err = response.WriteErrorString(http.StatusOK, "Node found")
 	if err != nil {
 		klog.Errorf("failed to send the task resp to edge , err: %v", err)
 	}
--- a/cloud/pkg/cloudhub/servers/httpserver/node/checknode_test.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/node/checknode_test.go
@@ -93,7 +93,7 @@ func TestCheckNode(t *testing.T) {
 			container.ServeHTTP(httpWriter, httpReq)
 
 			assert.Equal(t, tt.expectedStatus, httpWriter.Code)
-			assert.Contains(t, httpWriter.Body.String(), tt.expectedBody)
+			assert.Equal(t, tt.expectedBody, httpWriter.Body.String())
 		})
 	}
 }


DoisLONG:
Thanks for the fix.

I checked both the response text change and the corresponding test update.

Changing `Node founded` to `Node found` fixes the user-visible typo in the `/checknode/{nodename}` response.

The update from `assert.Contains` to `assert.Equal` also makes the test more precise, since it now validates the exact response body and can better prevent this typo from being introduced again.

From my review, the change is focused and consistent with the reported issue. I did not notice any additional problems in the current diff.

One minor non-blocking suggestion: the Go code block in the PR description appears to be missing the closing triple backticks, so it may be worth fixing for readability.

PR #6903, opened 2026-06-23
keadm: fix segmentation fault in init/install commands

### LINKED ISSUE #5003: keadm e2e test always report Segmentation fault error
In `keadm e2e` test always report `Segmentation fault` error, seems there's something wrong in `keadm init`

```
Starting kubeedge...
tests/scripts/keadm_e2e.sh: line 69: 28517 Segmentation fault      /usr/local/bin/keadm init --advertise-address=$MASTER_IP --profile version=$KUBEEDGE_VERSION --set cloudCore.service.enable=false --kube-config=$KUBECONFIG --force
```  

error job: https://github.com/kubeedge/kubeedge/actions/runs/6117072877/job/16603162420?pr=4998

KUSHAL-BANKHEDE:
**What type of PR is this?**
/kind bug
Fixes #5003

Fixes two crash causes in `keadm init` that produce a segmentation fault (Issue #5003):

1. **Removed** `discoveryClient.RESTClient().Post()` — a no-op call that can nil-dereference if the REST client is nil, causing SIGSEGV.

2. **Replaced** `panic()` in `GetOSInterface()` with a graceful `unsupportedOSType{}` struct. On unsupported OS (non-apt/yum/pacman), the old code panicked, which manifests as a segfault. The new implementation returns descriptive errors from all methods instead.

The e2e test at `tests/scripts/keadm_e2e.sh:60` was already updated to use `--kubeedge-version` instead of `--profile version=`, removing that trigger, but the underlying code bugs remained.

Fixes #5003

**Special notes for your reviewer:**
The `unsupportedOSType` struct implements all 7 methods of the `OSTypeInstaller` interface, each returning an appropriate error. This ensures no nil interface dereferences can occur on unsupported platforms.

```release-note
Fixed segmentation fault in keadm init/install on unsupported OS or when K8s REST client is unavailable
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/common.go
+++ b/keadm/cmd/keadm/app/cmd/util/common.go
@@ -124,6 +124,30 @@ func GetPackageManager() string {
 	}
 }
 
+// unsupportedOSType implements OSTypeInstaller for unsupported operating systems.
+// All methods return errors instead of panicking or causing segfaults.
+type unsupportedOSType struct{}
+
+func (u *unsupportedOSType) InstallMQTT() error {
+	return fmt.Errorf("unsupported operating system")
+}
+func (u *unsupportedOSType) IsK8SComponentInstalled(string, string) error {
+	return fmt.Errorf("unsupported operating system")
+}
+func (u *unsupportedOSType) SetKubeEdgeVersion(version semver.Version) {}
+func (u *unsupportedOSType) InstallKubeEdge(types.InstallOptions) error {
+	return fmt.Errorf("unsupported operating system")
+}
+func (u *unsupportedOSType) RunEdgeCore() error {
+	return fmt.Errorf("unsupported operating system")
+}
+func (u *unsupportedOSType) KillKubeEdgeBinary(string) error {
+	return fmt.Errorf("unsupported operating system")
+}
+func (u *unsupportedOSType) IsKubeEdgeProcessRunning(string) (bool, error) {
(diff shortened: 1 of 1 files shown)


DoisLONG:
Thanks for working on this crash issue.

The two production changes are directionally reasonable:

1. removing the unused `discoveryClient.RESTClient().Post()` call avoids dereferencing the REST client for a result that is never used;
2. replacing the panic in `GetOSInterface()` with an `OSTypeInstaller` implementation allows callers to receive an error instead of crashing the process.

However, I do not think this is ready to merge yet because the PR currently changes two failure paths without adding regression tests.

Please add focused tests covering at least the following behavior:

1. unsupported package-manager detection must not panic;
2. the returned installer must produce a clear error for unsupported operations;
3. `isK8SComponentInstalled` must continue to call `ServerVersion()` successfully without the removed `RESTClient().Post()` call;
4. errors returned by the discovery client must still be propagated correctly.

For the unsupported OS path, a test should explicitly demonstrate that the old panic is replaced by an error-returning flow. Since `GetPackageManager()` currently executes a shell command directly, it may be necessary to introduce a small test seam for package-manager detection rather than depending on the host running the unit test.

There are also two smaller cleanup items already visible in the current diff:

* the message still says `exit`, although the function no longer exits or panics;
* the repeated `fmt.Errorf("unsupported operating system")` values should preferably be replaced with a shared sentinel error.

The implementation direction looks useful, but regression coverage is important here because this code controls installation and process-management failure behavior.

PR #6906, opened 2026-06-23
cloud/policycontroller: wire health probes; run policy controller on every CloudCore replica

### LINKED ISSUE #6905: cloud/policycontroller: add leader election and health endpoint to the controller-runtime manager
Issue description:

`cloud/pkg/policycontroller/policycontroller.go:47–48`:

```go
// TODO: leader election
// TODO: /healthz
```

The `controller-runtime` manager is constructed without `LeaderElection: true` or a `HealthProbeBindAddress`. In HA CloudCore deployments every replica runs the policy reconciliation loop concurrently, which can produce duplicate writes and contention with the apiserver. Liveness probes also have nothing to read.

**Why is this needed**:

Standard controller-manager behavior expected by anyone running multi-replica CloudCore. `controller-runtime` exposes both as `ctrl.Options{LeaderElection, LeaderElectionID, HealthProbeBindAddress}` — small wiring change.

---

Priyanshubhartistm:
**What type of PR is this?**
/kind feature

The `controller-runtime` manager in `cloud/pkg/policycontroller` was constructed without a `HealthProbeBindAddress`, so liveness/readiness probes had nothing to serve.

This PR wires `/healthz` and `/readyz` on `:9002` for the policy controller manager.

Leader election was considered during review but is not used: each CloudCore replica only holds live CloudHub sessions for the edge nodes routed to it (session affinity is handled by the Service in front of CloudCore), so a single elected leader would be unable to deliver policy updates to edge nodes connected to the other replicas. Instead, every replica runs its own policy controller and its own health endpoint, and `Controller.send2Edge` (in the `manager` package) narrows delivery to the edge nodes actually managed by the local instance.

Manager construction is deferred from `Register()` to `Start()`, so the health-probe listener is only bound when the `RequireAuthorization` feature gate is enabled. The Helm chart and `build/cloud/07-deployment.yaml` liveness/readiness probes on `:9002` are likewise conditional on `cloudCore.featureGates.requireAuthorization`.

Fixes #6905

**Test command and output:**
go test -race ./cloud/pkg/policycontroller/...

```release-note
cloud/policycontroller: expose /healthz and /readyz on port 9002 for the policy controller manager when CloudCore runs with RequireAuthorization enabled. Each CloudCore replica runs the policy controller independently (no leader election) and scopes edge delivery to its own CloudHub sessions. Configure liveness and readiness probes in the Helm chart and build manifests.
```


DoisLONG:
Thanks for working on the policy controller HA behavior.

Enabling leader election for the policy reconciler is a reasonable direction, but the current implementation introduces a startup compatibility issue.

`Register()` creates the controller-runtime manager before the Beehive module's `Enable()` result is evaluated. Therefore, the new manager options are processed even when the `RequireAuthorization` feature gate is disabled.

There are two consequences:

1. `LeaderElection` is always enabled during manager construction, but `LeaderElectionNamespace` is not configured. Controller-runtime attempts to infer the namespace from the in-cluster service account file. For CloudCore installations running through keadm or as a standalone binary outside Kubernetes, manager creation will fail with an error similar to:

   `not running in-cluster, please specify LeaderElectionNamespace`

2. `HealthProbeBindAddress: ":9002"` creates the listener during `NewManager()`, not when the policy controller becomes enabled. This means CloudCore may bind or fail to bind port 9002 even when the policy controller module is disabled.

Please avoid making leader election and the probe listener unconditional during registration.

The implementation should define how these settings work for both deployment modes:

* CloudCore running in a Kubernetes Pod;
* CloudCore running through keadm or as a standalone binary.

Possible approaches include making leader election and the health probe address configurable, explicitly providing a leader-election namespace, and only constructing or starting these facilities when the policy controller module is enabled.

Please also add tests covering:

1. manager creation with an out-of-cluster REST config;
2. the policy controller feature gate disabled;
3. the health probe address disabled or already occupied;
4. the configured leader-election namespace.

Using `healthz.Ping` instead of a custom always-successful function would also be cleaner, but the deployment and initialization behavior should be resolved first.

Minor item: the `release-note` block in the PR description is missing its closing triple backticks.

Priyanshubhartistm:
Hi @DoisLONG,

I have addressed the requested feedback and updated the PR accordingly. The PR is now ready for review.

Thank you!

DoisLONG:
Thanks for the substantial update. Deferring manager construction from `Register` to `Start`, using `healthz.Ping`, and introducing a testable options layer are all improvements.

However, after following the actual production path, the deployment-compatibility concern is not fully resolved yet.

### 1. The production path always enables leader election

`Options` allows callers to use:

```go
LeaderElection:         false
HealthProbeBindAddress: ""
```

for standalone or keadm deployments, but `policyController.Start` always passes:

```go
Options{
    LeaderElection:          true,
    LeaderElectionNamespace: "kubeedge",
    HealthProbeBindAddress:  ":9002",
}
```

Therefore, when `RequireAuthorization` is enabled, every deployment mode still enables leader election. The `LeaderElection=false` path currently exists only in unit tests and has no production caller.

Please make the production options derive from an explicit CloudCore configuration or from a clearly defined and testable deployment-mode decision. Standalone compatibility cannot be established by testing an option that `Start` never uses.

### 2. Hard-coding the `kubeedge` namespace does not by itself make standalone startup safe

An explicit namespace avoids controller-runtime's in-cluster namespace auto-detection, but it does not guarantee that:

* the `kubeedge` namespace exists;
* the current credentials can create and update a Lease there;
* CloudCore is deployed in that namespace;
* a standalone user intended to enable leader election.

If the namespace or RBAC is unavailable, `mgr.Start` will fail and `klog.Fatalf` will terminate CloudCore.

Please use the actual configured/deployment namespace, or make the leader-election namespace and enablement explicit CloudCore settings with appropriate defaults for Pod and standalone deployments.

### 3. The new tests do not verify the claimed production behavior

`TestRegisterDoesNotConstructManager` does not call `Register`; it manually constructs a `policyController` and inspects its fields. This verifies the struct shape, not the behavior of `Register`.

`TestNewAccessRoleControllerManagerLeaderElectionNamespace` only calls `newManager`. As its own comment notes, the namespace and resource lock are not validated until `mgr.Start`, so this test does not verify Lease acquisition, namespace existence, RBAC, or leader election.

Please add tests around the actual options used by `Start`, preferably through an injected manager factory or a dedicated options-building function. The test should prove that each supported deployment mode receives the intended leader-election settings.

An integration or envtest-based check should also cover actual leader-election startup, or the PR should provide equivalent manual validation showing two replicas competing for the same Lease and failover working.

### 4. Restore the original feature-gate state

`TestPolicyControllerDisabled` sets `RequireAuthorization` to false and its cleanup also sets it to false. That does not restore the original global state and can make the package tests order-dependent.

Please capture the original value before changing it and restore that value in `t.Cleanup`.

### 5. Clarify readiness semantics

Both `/healthz` and `/readyz` use `healthz.Ping`. This is acceptable as a basic process-level check, but it does not prove cache synchronization, controller startup, or lease ownership. Please ensure the PR description and release note do not overstate the readiness guarantee.

Finally, please provide the text output for:

```bash
go test -race ./cloud/pkg/policycontroller
```

Priyanshubhartistm:
Hi @DoisLONG , could you please take a look at this PR when you have time? I’ve addressed the review feedback and would really appreciate your feedback.

DoisLONG:
Thanks for iterating on this. The direction looks good to me, and I noticed the previous `healthz.Ping` suggestion has already been addressed.

One thing I’m still not fully sure about is whether the production path actually enables leader election and health probes now.

From what I can see, `registerModules()` still calls `policycontroller.Register(client.CrdConfig)`, and `Register()` defaults to `DeploymentModeStandalone`. In that mode, `buildOptions()` returns `LeaderElection=false` and an empty `HealthProbeBindAddress`.

So this PR seems to add an opt-in in-cluster path, but the current CloudCore registration path may still run policycontroller without leader election or health probes. Could you clarify how `DeploymentModeInCluster` is expected to be used in the real CloudCore deployment path, or wire it from the CloudCore registration/config path if that is intended?

Other than that, the standalone/in-cluster separation looks like a good direction.

DoisLONG:
Thanks for the updates.

This version looks much closer to me. The normal CloudCore registration path now goes through `registerPolicyController()`, and `Register()` no longer constructs the controller-runtime manager eagerly. Deferring manager creation to `Start()` also makes sense, especially when the policy controller module is disabled by the feature gate or CloudCore is running in standalone / keadm mode.

The leader-election RBAC also looks okay to me: the existing CloudCore ClusterRole already includes permissions for `coordination.k8s.io` `leases`.

A couple of things still need cleanup before LGTM:

1. The PR says it wires health probes, but the current diff only exposes the healthz/readyz endpoint in the manager. I do not see a Helm chart or deployment manifest update that actually configures CloudCore liveness/readiness probes against `:9002`. Could you either add the manifest wiring, or clarify in the PR description that this PR only exposes the probe endpoints and does not configure Kubernetes probes yet?

2. The current Main CI workflow is still failing in the edge integration test step. This may be unrelated to this PR, but please confirm or rerun it.

3. The release-note block in the PR description also looks incomplete. Since this is a feature/behavior change for HA CloudCore deployments, it should probably have a short release note instead of an empty block.

Priyanshubhartistm:
Hi @DoisLONG , could you please take another look at this PR when you have time? I’ve addressed the review feedback and would really appreciate your feedback.

DoisLONG (commented):
Thanks for the follow-up.

I already reviewed the current version, and my main concern is still the interaction between the newly added CloudCore probes and the policycontroller feature gate.

The manifests add liveness/readiness probes on `:9002`, but that endpoint is served by the policycontroller manager. Since the policycontroller module only starts when `RequireAuthorization` is enabled, a CloudCore Pod with that feature gate disabled may not have anything listening on `:9002`, which could make the new probes fail even if CloudCore itself is running normally.

I do not see a new commit or update addressing this point yet. Could you please update the PR to clarify or fix this behavior first? For example, make the probes conditional, move the health endpoint to a component that is always started, or clearly show that the affected manifests always enable `RequireAuthorization`.

Priyanshubhartistm:
Hi @DoisLONG , could you please take another look at this PR when you have time? 
thanks

DoisLONG (commented):
Thanks for the follow-up update.

This version addresses most of the deployment-compatibility concerns from the previous review. Deferring manager construction from `Register()` to `Start()`, deriving the manager options from a deployment mode, using `healthz.Ping`, restoring the feature-gate state in tests, and adding tests around the option-building/start path are all good improvements.

I have one remaining concern from the same compatibility angle.

The Helm chart now adds the policycontroller `:9002` port and `/healthz` / `/readyz` probes only when `cloudCore.featureGates.requireAuthorization` is enabled. However, `build/cloud/07-deployment.yaml` still adds the same port and probes unconditionally.

Since the policycontroller manager and its health-probe listener are only started when the policycontroller module is enabled, the static deployment manifest may make CloudCore unhealthy when `RequireAuthorization` is disabled.

Could you please align `build/cloud/07-deployment.yaml` with the Helm chart behavior, or clarify why this static manifest path always enables `RequireAuthorization`?

For CI, I noticed that the core checks such as build, unit/integration test, lint, and docker e2e are passing, while the current Main CI failure is in the CRI-O e2e job. Please also check whether that failure is related to this health/readiness probe change or just an unrelated e2e flake.

DoisLONG (commented):
Thanks for the update.

This version addresses most of my previous concern. In particular, the Helm chart now makes the CloudCore probes conditional on `cloudCore.featureGates.requireAuthorization`, and the policycontroller manager is only created in `Start()`, so it should only start the health endpoint when the module is actually enabled. The note about `/readyz` only being a process-level ping is also helpful.

One remaining point: `build/cloud/07-deployment.yaml` still adds the `:9002` container port and liveness/readiness probes unconditionally. Could you please either make this generated/static manifest consistent with the Helm chart behavior, or clearly confirm that this manifest is only used with `RequireAuthorization` enabled?

DoisLONG (commented):

The main policycontroller change looks good to me now. Before LGTM, could you please do two final cleanups:

1. Remove the unrelated whitespace-only changes in the release note / documentation files.
2. Squash the commits into a single logical commit for this PR.

DoisLONG (commented):
Thanks for the continued updates. The PR is much cleaner now, and the main policycontroller change looks good to me.

The previous compatibility concerns around feature gates, probes, and deployment mode have been addressed, and the commits are now squashed into a single logical commit.

I noticed a couple of minor cleanup points in the release note / codespell change, but they are not blockers from my side.

/lgtm

Priyanshubhartistm:

WillardHu:
This should require each instance to execute the policy-controller. 
Each CloudCore instance connects to different edge nodes via service routing, and filters out meaningless reconcile events by interacting with its associated edge nodes.

Priyanshubhartistm:
@DoisLONG, could you please take another look when you have a chance? Thanks!

Priyanshubhartistm:
Hi @DoisLONG, just a gentle reminder could you please review this and let me know if any changes are needed? 
Thank you!

DoisLONG (commented):
Thanks for the update.

I think the previous issues around feature gate, probe wiring, and manager creation have been addressed.

One remaining cleanup: the latest code is no longer using leader election. It now lets each CloudCore replica run the policy controller and filters the target nodes by the local CloudHub sessions.

This direction is fine to me, but the PR title/body/release note still say that leader election is enabled. That is a bit confusing.

Could you please update the PR description and release note to match the current implementation? After that, I think this should be good from my side again.

Priyanshubhartistm:
Thnaks @DoisLONG  updated the PR title/body/release note to match the current implementation.

DoisLONG (commented):
Thanks for the update.

/lgtm

Priyanshubhartistm:
Hii @WillardHu , If everything looks good now, could you please /approve this PR when you have a chance? I'd really appreciate it.

cloud/pkg/policycontroller/policycontroller.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2026 The KubeEdge Authors.
```
Shelley-BaoYue:
```suggestion
Copyright 2023 The KubeEdge Authors.
```

cloud/pkg/policycontroller/policycontroller_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2025 The KubeEdge Authors.
```
Shelley-BaoYue, line 2: Don't need change the date  for exist file.

cloud/pkg/policycontroller/policycontroller.go
```diff
@@ -73,14 +136,20 @@ func setupControllers(ctx context.Context, mgr manager.Manager) error {
 	return nil
 }
 
+// Register stores the REST config on the policyController and registers the
+// module with the Beehive runtime. The controller-runtime manager is NOT
+// constructed here; that happens in Start(), which Beehive only calls when
+// Enable() returns true. This ensures the health-probe listener is never
+// bound when the RequireAuthorization feature gate is disabled.
```
Shelley-BaoYue, line 143: Would it still be ok to keep newManager in Register? Because  even after `register` in beehive is complete, as long as it is not enabled, we won't proceed with `start` or `run`. I think keeping `newManager` in `Register` would be more appropriate for consistency with other modules.

cloud/pkg/policycontroller/policycontroller_test.go
Priyanshubhartistm: Done

cloud/pkg/policycontroller/policycontroller.go
Shelley-BaoYue: > Would it still be ok to keep newManager in Register? Because even after `register` in beehive is complete, as long as it is not enabled, we won't proceed with `start` or `run`. I think keeping `newManager` in `Register` would be more appropriate for consistency with other modules.

@Priyanshubhartistm  Could you answer this question for me? Thanks.

Priyanshubhartistm: @Shelley-BaoYue , The reason newManager is called in Start() instead of Register() is that controllerruntime.NewManager() immediately binds to the health-probe port (:9002) as soon as it's called  it doesn't wait for Start(). So if we put it in Register(), the port would be bound even when RequireAuthorization is disabled and beehive skips Start() entirely. Moving it to Start() keeps the side effect (port binding) tied to when the module is actually running.

PR #6911, opened 2026-06-25
fix: prevent command injection via spec.image in NodeUpgradeJob

### LINKED ISSUE #6909: Unsanitized spec.image in NodeUpgradeJob allows command injection on edge nodes
**What happened**:

`spec.image` in `NodeUpgradeJob` is concatenated into a `bash -c` command without sanitization, allowing command injection on edge nodes. The admission webhook validates `spec.version` (semver check) but never checks `spec.image`.

Affected code:
- `edge/pkg/taskmanager/actions/nodeupgradejob.` (v1alpha2)
- `edge/pkg/taskmanager/v1alpha1/taskexecutor/node_upgrade.` (v1alpha1)

Both build a shell string with the image value and pass it to `exec.Command("bash", "-c", ...)` via `execs.NewCommand`.

The command should use `exec.Command("keadm", args...)` with separate arguments so shell metacharacters aren't interpreted.

Set `spec.image` to something like `"legit-image; touch /tmp/pwned; #"` in a `NodeUpgradeJob` so the injected command runs as root on the edge node.

Happy to open a PR for this. Fix is just switching from string concat to arg list:

```go
args := []string{"upgrade", "edge", "--force", "--toVersion", spec.Version}
if spec.Image != "" {
    args = append(args, "--image", spec.Image)
}
cmd := exec.Command("keadm", args...)
```

**Environment**:
- KubeEdge version: latest main (both v1alpha1 and v1alpha2 affected)

Anshumancanrock:
## What this PR does

Fixes command injection in `NodeUpgradeJob` where `spec.image` was concatenated directly into a `bash -c` command string on edge nodes.

Both the v1alpha2 (`nodeupgradejob.go`) and v1alpha1 (`node_upgrade.go`) upgrade handlers built a shell command by string concatenation and passed it to `exec.Command("bash", "-c", ...)`. Since `spec.image` has no validation in the admission webhook or CRD schema, shell metacharacters in the image field would be interpreted.

## Changes

- Added `NewCommandDirect(name, args...)` to the `execs` package (creates commands without shell wrapping).
- **v1alpha2**: `upgrade()` now uses `NewCommandDirect("keadm", args...)` with direct arg passing. Stdout/stderr redirect to `/tmp/keadm.log` via `os.OpenFile` (needed to avoid SIGPIPE when edgecore is killed during upgrade).
- **v1alpha1**: `keadmUpgrade()` now uses `exec.Command("keadm", args...)` with `cmd.Start()` instead of `nohup bash -c`.
- Updated tests accordingly.

## Which issue this PR fixes

Fixes #6909

### DIFF
--- a/edge/pkg/taskmanager/actions/nodeupgradejob.go
+++ b/edge/pkg/taskmanager/actions/nodeupgradejob.go
@@ -20,9 +20,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"runtime"
-	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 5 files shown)


DoisLONG:
Thanks for reporting and working on this issue.

The KubeEdge maintainers are already handling this problem through a coordinated fix, which is planned to be included in the upcoming v1.23.1 patch release.

To avoid maintaining duplicate or conflicting implementations for the same issue, we will proceed with the coordinated fix and close this PR.

Thank you again for identifying the problem and contributing a solution.

PR #6912, opened 2026-06-25
Implementation of Phase-1 for iptableManager alternatives,  LFX'26 term-2

tushar743-ui:
**What type of PR is this?**
/kind feature
/kind design

This PR marks the starting of the Phase-1 implementation to explore alternatives for iptableManager via API server redirection. It follows the implementation plan from #6838.

**Implementation plan:**

1. Define `NodeEdgeTunnelIP corev1.NodeAddressType = "EdgeTunnelIP"` in KubeEdge constants
2. Add `EdgeTunnelIP` preservation in `upstream.go` alongside the existing `KubeletEndpoint.Port` preservation logic
3. In `tunnelserver.go` `connect()`, append `EdgeTunnelIP = cloudCoreIP` to `node.Status.Addresses` when edge node connects
4. Remove `EdgeTunnelIP` after `session.Serve()` returns on disconnect
5. Gate behind `FeatureGates` in cloudcore config as suggested
6. Document that operators must configure kube-apiserver:
   `--kubelet-preferred-address-types=EdgeTunnelIP,InternalIP,ExternalIP,Hostname`
   

**Additional pre-existing fix included:**
`edge/test/integration/utils/helpers/helpers.go` — the CI integration test `TC_TEST_EBUS_14` was failing due to the mock HTTP server at `:12345` (started by edgecore testManager with no restart policy) becoming unavailable between test cases. A single connection refused immediately triggered `common.Fatalf` with no recovery path. Added a 3-attempt retry loop with 100ms sleep. No test assertions or expected behavior changed. Included here to unblock CI rather than open a separate PR.

Fixes #

**Special notes for your reviewer:**
Would appreciate any reviews on this approach. And also please let me know if these changes needed separate PR each as this PR included changes in many files in a single PR.

```release-note

```


tushar743-ui:
Would appreciate a review @luomengY whenever you are free.

tushar743-ui:
@luomengY, would appreciate further directions from your side

edge/test/integration/utils/helpers/helpers.go
```diff
@@ -247,20 +247,29 @@ func HandleAddAndDeleteDevice(operation, testMgrEndPoint string, device dttype.D
...
 	t := time.Now()
-	resp, err := client.Do(req)
 
+	const maxAttempts = 3
+	var req *http.Request
+	var resp *http.Response
+	for attempt := 0; attempt < maxAttempts; attempt++ {
+		if attempt > 0 {
+			time.Sleep(100 * time.Millisecond)
+		}
```
luomengY: Do not put proposals in code‑implementation pull requests.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -695,8 +694,24 @@ func (uc *UpstreamController) updateNodeStatus() {
 					nodeStatusRequest.Status.DaemonEndpoints.KubeletEndpoint.Port = getNode.Status.DaemonEndpoints.KubeletEndpoint.Port
 				}
 
+				// Preserve EdgeTunnelIP address set by CloudCore.
+				// edgecore heartbeat does not include EdgeTunnelIP in its address
+				// list, so it must be re-applied after status overwrite, following
+				// the same pattern as KubeletEndpoint.Port preservation above.
+				var edgeTunnelAddrs []v1.NodeAddress
+				for _, addr := range getNode.Status.Addresses {
+					if addr.Type == common.NodeEdgeTunnelIP {
```
luomengY: updateNodeStatus() is a deprecated method. Have you tested it here?

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -1011,11 +1026,56 @@ func (uc *UpstreamController) patchNode() {
 				continue
 			}
 
-			node, err := uc.kubeClient.CoreV1().Nodes().Patch(utilcontext.FromMessage(context.TODO(), msg), name, apimachineryType.StrategicMergePatchType, patchBytes, metaV1.PatchOptions{}, "status")
+			node, err := uc.kubeClient.CoreV1().Nodes().Patch(utilcontext.FromMessage(context.TODO(), msg), name, patchtypes.StrategicMergePatchType, patchBytes, metaV1.PatchOptions{}, "status")
```
luomengY: Why is this being changed? Please do not modify parts that are unrelated to your feature.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -1590,15 +1650,15 @@ func UpdateAnnotation(ctx context.Context, nodeName string) error {
 	if err != nil {
 		return fmt.Errorf("failed to get cloudcore localIP with err:%v", err)
 	}
-	if value, ok := node.Annotations[comconstants.EdgeMappingCloudKey]; ok {
```
luomengY, line 1593: Why is this being changed? Please do not modify parts that are unrelated to your feature.

edge/test/integration/utils/helpers/helpers.go
tushar743-ui: Removed the proposal files

cloud/pkg/edgecontroller/controller/upstream.go
tushar743-ui: The preservation in `updateNodeStatus()` is necessary because that function does a full UpdateStatus overwrite (getNode.Status = nodeStatusRequest.Status) which wipes EdgeTunnelIP on every heartbeat - that's the exact problem we're solving. patchNode uses StrategicMergePatch which preserves unmentioned addresses, but updateNodeStatus doesn't. We acknowledge the function is deprecated; I've also added coverage in patchNode as the forward path. Once updateNodeStatus is removed, only the patchNode guard remains.

cloud/pkg/edgecontroller/controller/upstream.go
tushar743-ui: removed

cloud/pkg/edgecontroller/controller/upstream.go
tushar743-ui: removed

luomengY:
It seems that the `updateNodeStatus` functionality is not called in the current KubeEdge code logic.

tushar743-ui:
The `updateNodeStatus()` is marked deprecated and its `nodeStatusChan` is never dispatched to in dispatchMessage() - edgecore now sends `ResourceTypeNodePatch` messages which route to `patchNode()` instead. The preservation block we added inside `updateNodeStatus()` was dead code that never executed during edgecore heartbeats.

Removed the dead preservation block. The single active preservation path is the patchNode() re-injection block (around lines 1020-1059) which fires on every `ResourceTypeNodePatch` message from edgecore and correctly restores EdgeTunnelIP after the strategic-merge patch wipes it.

updateNode() does not need preservation either - it only patches node metadata (labels and annotations) via a full node Patch, never touching status.addresses.

tushar743-ui:
Reverted `comconstants.EdgeMappingCloudKey ` back to `common.EdgeMappingCloudKey` in `UpdateAnnotation`. The `comconstants` alias was an unrelated lint cleanup change that should not be in this PR. Also removed the now-unused duplicate import alias which was itself causing an ST1019 lint warning.

The upstream.go diff now contains only the `EdgeTunnelIP` re-injection block in `patchNode()` and the features import.

edge/test/integration/utils/helpers/helpers.go
```diff
@@ -247,20 +247,29 @@ func HandleAddAndDeleteDevice(operation, testMgrEndPoint string, device dttype.D
...
-
 	client := &http.Client{}
-	req.Header.Set("Content-Type", "application/json; charset=utf-8")
 	t := time.Now()
-	resp, err := client.Do(req)
 
+	const maxAttempts = 3
+	var req *http.Request
+	var resp *http.Response
+	for attempt := 0; attempt < maxAttempts; attempt++ {
```
luomengY: For retries, please use the `github.com/avast/retry-go` package.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -1011,11 +1026,56 @@ func (uc *UpstreamController) patchNode() {
...
+									break
+								}
+							}
+							if !alreadySet {
+								freshNode.Status.Addresses = append(freshNode.Status.Addresses, v1.NodeAddress{
+									Type:    common.NodeEdgeTunnelIP,
+									Address: cloudCoreIP,
+								})
+								if updated, updateErr := client.GetKubeClient().CoreV1().Nodes().UpdateStatus(
+									context.Background(), freshNode, metaV1.UpdateOptions{}); updateErr != nil {
```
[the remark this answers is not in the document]
luomengY: There is a brief window between the completion of the `Patch` and the `UpdateStatus` call during which the node does not have `EdgeTunnelIP`. If the `Apiserver` issues requests like `logs/exec` to the edge node during this window, it will fail to retrieve `EdgeTunnelIP`, leading to idempotency issues. The correct approach is to merge `EdgeTunnelIP` into the `patchBytes` before performing the `Patch`, thereby eliminating this window entirely.

luomengY:
There is a brief window between the completion of the Patch and the UpdateStatus call during which the node does not have EdgeTunnelIP. If the Apiserver issues requests like logs/exec to the edge node during this window, it will fail to retrieve EdgeTunnelIP, leading to idempotency issues. The correct approach is to merge EdgeTunnelIP into the patchBytes before performing the Patch, thereby eliminating this window entirely.

cloud/cmd/iptablesmanager/app/server.go
```diff
@@ -57,7 +57,9 @@ func NewIptablesManagerCommand() *cobra.Command {
 				Burst:       constants.DefaultKubeBurst,
 				KubeConfig:  opts.KubeConfig,
 			}
-			go iptables.NewIptablesManager(kubeAPIConfig, opts.ForwardPort).Run(ctx)
+			if mgr := iptables.NewIptablesManager(kubeAPIConfig, opts.ForwardPort); mgr != nil {
+				go mgr.Run(ctx)
```
luomengY: Why was this change made here? Can `mgr` ever be nil? I think this modification is unnecessary.

cloud/cmd/cloudcore/app/server.go
```diff
@@ -129,7 +129,9 @@ kubernetes controller which manages devices so that the device metadata/status d
 				// By default, IptablesManager manages tunnel port related iptables rules
 				// The internal mode will share the host network, forward to the stream port.
 				streamPort := int(config.Modules.CloudStream.StreamPort)
-				go iptables.NewIptablesManager(config.KubeAPIConfig, streamPort).Run(ctx)
+				if mgr := iptables.NewIptablesManager(config.KubeAPIConfig, streamPort); mgr != nil {
```
luomengY: Why was this change made here?

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -54,27 +58,31 @@ type TunnelServer struct {
 	sessions      map[string]*Session
 	nodeNameIP    sync.Map
 	tunnelPort    int
+	cloudCoreIP   string
```
luomengY: Here, the definition of `cloudcoreIP` seems redundant. It should be obtained directly from the Node's `annotation`, rather than from the `hubconfig.Config.AdvertiseAddress[0]` configuration.

luomengY:
The `EdgeTunnelIP` scheme should be adopted only when `iptablesManager` is in internal mode and `cloudcore` and the `ApiAerve`r are not on the same node. However, I don't see the relevant conditional logic in your code.

tushar743-ui:
Changes does ->
1. In `(helpers.go)`
Replaced the manual attempt loop with `github.com/avast/retry-go` as suggested - retry.Do() with retry.Attempts(3) and retry.Delay(100ms). 

2. In  `(iptablesmanager/app/server.go) `
   Reverted - this change was unrelated to the EdgeTunnelIP feature and should not have been included.

3. In `(cloudcore/app/server.go)`
 Reverted the nil check. The only remaining change to this file is passing `c.Modules.IptablesManager` to `cloudstream.Register`, which is needed for the new `shouldUseEdgeTunnelIP()` helper.
 
 4.  Done - removed cloudCoreIP from TunnelServer struct and all constructor signatures. `updateNodeEdgeTunnelIP `now reads the cloudcore IP exclusively from the node's annotation `(constants.EdgeMappingCloudKey)`. If the annotation is absent, the function logs a warning and returns nil (graceful skip), rather than falling back to a struct field.

5. Fixed -  EdgeTunnelIP is now merged directly into patchBytes before the Patch call via `mergeEdgeTunnelIPIntoPatch()`. This injects the annotation's cloudcore IP into the status.addresses array of the patch JSON itself, so the Patch atomically includes EdgeTunnelIP. The post-patch `UpdateStatus` re-injection block has been removed entirely.

6.  Added `shouldUseEdgeTunnelIP() `in both tunnelserver.go and upstream.go. In tunnelserver.go it checks two conditions: `iptablesMgrMode == InternalMode` AND `len(hubconfig.Config.AdvertiseAddress) > 0`. External DaemonSet mode (ExternalMode) returns false immediately since the DaemonSet runs on all non-edge nodes including ApiServer's node. A non-empty `advertiseAddress` indicates the operator explicitly configured a separate cloudcore IP, implying a topology where the ApiServer node cannot be covered by cloudcore's local iptables rules. EdgeTunnelIP is only activated when both the feature gate and `shouldUseEdgeTunnelIP()` are true.

cloud/pkg/edgecontroller/controller/upstream.go
```diff
@@ -1030,6 +1040,57 @@ func (uc *UpstreamController) patchNode() {
...
+// (no advertiseAddress), iptablesManager DNAT handles routing directly.
+func shouldUseEdgeTunnelIP() bool {
+	return len(hubconfig.Config.AdvertiseAddress) > 0
+}
+
+// mergeEdgeTunnelIPIntoPatch injects the EdgeTunnelIP address into patchBytes
+// before the Patch call, eliminating the window between Patch and a subsequent
+// UpdateStatus where EdgeTunnelIP would be absent from the node.
+func mergeEdgeTunnelIPIntoPatch(ctx context.Context, nodeName string, patchBytes []byte) ([]byte, error) {
+	node, err := client.GetKubeClient().CoreV1().Nodes().Get(ctx, nodeName, metaV1.GetOptions{})
```
luomengY: For the merge function, please refer to the implementation of the `fixupPatchForNodeStatusAddresses` function in `k8s.io/component-helpers/node/util/status.go`.

tushar743-ui: made the changes

luomengY:
image
Please keep only one implementation PR and one design proposal, and close the rest.

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -218,24 +236,125 @@ func (s *TunnelServer) updateNodeKubeletEndpoint(nodeName string) error {
...
+}
+
+// shouldUseEdgeTunnelIP returns true only when iptablesManager is in internal mode
+// AND advertiseAddress is configured (indicating a potentially separated topology where
+// the API server cannot reach edge nodes via iptables DNAT on cloudcore's node).
+func (s *TunnelServer) shouldUseEdgeTunnelIP() bool {
+	if s.iptablesMgrMode != v1alpha1.InternalMode {
+		return false
+	}
+	return len(hubconfig.Config.AdvertiseAddress) > 0
```
luomengY: Can this judgment determine whether the API server and cloudcore are running on the same node?

luomengY:
You need to set up a local environment, test it repeatedly, and ideally make a demo video.

cloud/pkg/cloudstream/tunnelserver.go
tushar743-ui: No, it couldn't -> AdvertiseAddress is mandatory and auto-defaulted, so that check was always true regardless of topology. Replaced it with a real check comparing CloudCore's own node (NODE_NAME downward API) against the actual apiserver address(es) on the default/kubernetes Endpoints object; falls back safely to "separated" when it can't be determined. (in the latest commit)

tushar743-ui:
**What this change is about**

While doing repeated live testing , I found a real bug in how the code decides whether to add the EdgeTunnelIP address to a node. There were two separate copies of this decision logic, and only one had been fixed in an earlier review round — the other was still broken, and it was winning.

**Why this fix belongs in this PR (not a scope violation)**

I know unrelated changes have been flagged before, and one earlier attempt to touch this same file (upstream.go) for this same reason was reverted for being out of scope — that comment was specifically about a different file, so fixing upstream.go at the same time was scope creep at that point.

This time is different: live testing showed this bug actively breaks the feature this PR exists to deliver. The whole point of EdgeTunnelIP is that it should turn off automatically when CloudCore and the API server share a node. Because of this bug, it never actually turned off — it kept coming back every 15 seconds regardless. That's not a side issue or cleanup; it's the main goal not working. Leaving it unfixed would mean shipping a feature that doesn't do what it's supposed to do in one of its two core cases. So this fix is necessary to actually complete the goal, not an unrelated change bundled in.

The bug, in plain terms

- EdgeTunnelIP should only be added when CloudCore and the Kubernetes API server run on different nodes. Same node → it should be left off.
- tunnelserver.go had the correct check (fixed earlier).
- upstream.go had its own separate copy of the same check — never updated, always assumed "different nodes," no real check at all.
- upstream.go's copy runs on every heartbeat (every 15 seconds). So even when tunnelserver.go's correct check said "same node, leave it off," the next heartbeat added it right back.
- This only showed up because I built a real 2-node test cluster and moved CloudCore between nodes — a single-node setup can't reveal this, which is likely why it went unnoticed until now.

**The fix**

Pull the "are these on the same node?" check into one shared piece of code, and make both places use it instead of each keeping a private copy. One implementation to keep correct, not two that can silently drift apart — which is exactly what happened.

**File-by-file summary (8 files)**
1.` cloud/pkg/common/nodetopology/nodetopology.go` (new file) — The shared "same node?" check, moved here so both places can use it. Logic itself is unchanged, just relocated.
2. `cloud/pkg/common/nodetopology/nodetopology_test.go` (new file) — Tests for that check, moved out of tunnelserver_test.go since the logic now lives here.
3. `cloud/pkg/cloudstream/tunnelserver.go `— Now calls the shared package instead of its own private copy. No behavior change, just removes duplication.
4. `cloud/pkg/cloudstream/tunnelserver_test.go `— Updated to match that small refactor.
5. `cloud/pkg/edgecontroller/controller/upstream.go` — The actual bug fix: the broken, always-true check is replaced with a real one, using the shared logic and the same "check once, remember the answer" pattern already used in tunnelserver.go.
6. `cloud/pkg/edgecontroller/controller/upstream_test.go` — New tests for the fix above, including a helper function that had zero test coverage before this.
7. `cloud/pkg/edgecontroller/edgecontroller.go` — Small plumbing so the code in (5) knows which iptables mode (internal vs. external) is active — needed for the same-node check to be meaningful.
8.` cloud/cmd/cloudcore/app/server.go `— One-line change passing that iptables-mode information through at startup.

tushar743-ui:
https://github.com/user-attachments/assets/80b03a07-52c6-4456-a1ab-93854c2a7bda

This is the Phase 1 demo video. Let me know if you have any feedback.

luomengY:
In your demo, did you finally show the final implementation of this project? Are there any unresolved issues at present? You can directly demonstrate the functionality of the PR and skip the environment setup process. I noticed you mentioned that EdgeTunnelIP gets overwritten periodically. Can merging the node status with EdgeTunnelIP in upstream and patching it to the APIServer solve this problem?

tushar743-ui:
Yes, this demo shows the final implementation on the branch - the fix I describe below  was already merged in before this recording, so what's in the video is the current state of the PR, not an earlier version.

Unresolved issues: None affecting correctness right now. go build is clean and the full test suite (`cloudstream`, `edgecontroller/controller`, `common/nodetopology`) passes on the current tip. The one open item is deployment UX, not a functional bug: the Helm chart doesn't yet expose the `EdgeTunnelIP` feature gate as a `values.yaml` field - it's only settable via a manual `ConfigMap` patch today, which gets wiped on helm upgrade. I'll follow up with that separately if needed for review.

"Yes - I can also record a short follow-up clip that skips `cluster/env` setup entirely and jumps straight to `kubectl exec/logs` succeeding through the tunnel in both scenarios, if that's more useful for review."

On the periodic-overwrite question - yes, and it turned out to be two separate issues(fixed):

1. There was a timing gap: the code used to call `Patch()` and then a separate `UpdateStatus()` call to re-add `EdgeTunnelIP`. A concurrent heartbeat update could land in between those two calls and wipe the field out. That's fixed - `EdgeTunnelIP` is now merged directly into the patch data before `Patch()` is called, so there's no window where it's briefly missing.

2. The actual overwrite behavior visible in the demo was a second, separate bug - a decision bug, not a timing one. The heartbeat-handling code had its own independent copy of the "should EdgeTunnelIP even be set" check, and that copy was still using an old heuristic that always evaluated true regardless of actual node placement. So in the colocated scenario, where `EdgeTunnelIP` should not be set at all, every heartbeat (~15s) re-added it anyway, because the heartbeat path's own logic disagreed with the connect-time logic, which correctly excluded it. Fixed by moving the real same-node check into one shared function that both the connect path and the heartbeat path now call, instead of two separate copies that could drift apart. Verified on a real 2-node cluster: the colocated case now correctly shows no `EdgeTunnelIP` and stays that way across repeated heartbeats; the separated case still correctly shows it surviving heartbeats.

So to directly answer: merging into the patch was a necessary fix but not sufficient on its own - the periodic overwrite you flagged needed the second fix (the shared same-node check) as well, and both are in this branch now.

luomengY:
> 第一阶段.mp4 
> 这是第一阶段的演示视频。如果您有任何反馈意见，请告诉我。

The demo you recorded has no sound. You could write a step-by-step description in the comments so that other community members can understand it more easily.

tushar743-ui:
Step-by-step walkthrough of the video :

**Setup - Parts 0–4 (Commands 1–10)**: Tear down any leftover state (bare `cloudcore/edgecore` processes, edgecore's local `cert/CA` files, and the kind cluster itself), then create a fresh 2-node kind cluster (1 control-plane + 1 worker). Derive the two nodes' real IPs live into `$CP_IP/$WORKER_IP` (kind assigns them fresh every time, so nothing is hardcoded). Patch the apiserver's static pod manifest to add `--kubelet-preferred-address-types=EdgeTunnelIP,InternalIP,ExternalIP,Hostname` - this is what makes the apiserver actually prefer `EdgeTunnelIP` when it's present, and is a one-time cluster-level prerequisite unrelated to the KubeEdge code itself. Build the cloudcore image from this branch and load it into both kind nodes.

**Part 5 - Separated scenario setup (Commands 11–14)**: helm install CloudCore onto the worker node (`iptablesManager.mode=internal, advertiseAddress=$WORKER_IP`), so CloudCore ends up on a different node than the apiserver. Patch the `EdgeTunnelIP` feature gate into the `ConfigMap` (not yet exposed as a values.yaml field - noted as a follow-up), and delete/recreate the pod so it picks up the gate. Confirm the CloudCore cert's SAN matches the worker's IP.

**Part 6 - Separated scenario verification (Commands 15–25)**: Point edgecore at the worker's IP, fetch a fresh one-time bootstrap token from tokensecret, and start edgecore for a true first-time bootstrap (certs were wiped in step 2). Confirm the connection is actually live via heartbeat recency (not the Ready condition - see note below), then verify:
- `EdgeTunnelIP` = `worker's IP`, InternalIP unchanged (real edge IP)
- `kubeletEndpoint.Port = 10003` (the CloudStream port)
- kubectl logs/kubectl exec against a pod on the edge node succeed through the tunnel
- Zero iptables involvement on the apiserver's own node (no TUNNEL-PORT chain there at all)
- 3 disconnect/reconnect cycles, each correctly restoring `EdgeTunnelIP`
- `EdgeTunnelIP` survives 60s of heartbeats without being dropped

**Part 7 - Colocated scenario (Commands 26–39)**: helm upgrade to move CloudCore's pod onto the control-plane node (same node as the apiserver) via `nodeSelector/toleration + advertiseAddress=$CP_IP`. Re-patch the feature gate (wiped by every helm upgrade), delete cloudcoresecret + the pod to force cert regeneration with the new SAN, then point edgecore at the new address with a fresh token and re-bootstrap (old cert is no longer trusted since CloudCore's cert changed). Verify the opposite expectations:
- EdgeTunnelIP correctly absent (same-node detection working)
- InternalIP unchanged
- kubeletEndpoint.Port = the DNAT tunnel port (not 10003)
- kubectl logs/kubectl exec succeed via the internal-mode DNAT path instead
- iptables DNAT rule is present/hit on CloudCore's own node this time
- 3 disconnect/reconnect cycles + 60s heartbeat survival, same as the separated case

**Part 8 - Wrap-up (Commands 40–41)**: Final kubectl get pods/kubectl get node sanity check, then a closing dump of the current node addresses and `kubeletEndpoint.Port` for both scenarios side by side.

Net result shown: In the separated case, traffic flows through `CloudStream (EdgeTunnelIP + port 10003)` with zero iptables on the apiserver's node. In the colocated case, `EdgeTunnelIP` is correctly not added and traffic flows through the existing internal-mode DNAT path instead - confirming the same-node detection correctly picks the right path in both topologies, with kubectl exec/logs working end-to-end in both.

cloud/pkg/cloudstream/tunnelserver.go
```diff
@@ -218,24 +241,147 @@ func (s *TunnelServer) updateNodeKubeletEndpoint(nodeName string) error {
...
+// shouldUseEdgeTunnelIP returns true only when iptablesManager is in internal
+// mode AND kube-apiserver is not colocated with cloudcore -- i.e. the API
+// server cannot reach edge nodes via the iptables DNAT rule installed on
+// cloudcore's own node. Colocation is determined by nodetopology.IsAPIServerColocated,
+// not inferred from configuration presence: advertiseAddress is mandatory and
+// auto-defaulted by cloudcore, so checking whether it is set can never
+// distinguish colocated from separated deployments.
+//
+// The result is cached for the process lifetime: node placement cannot
+// change without a cloudcore restart.
```
luomengY: Some of the comments in your PR are too long and could be simplified.

tushar743-ui: Simplified comments in several files

luomengY:
Check e2e tests

luomengY:
Please submit a usage tutorial to the repository at https://github.com/kubeedge/website.

tushar743-ui:
> Please submit a usage tutorial to the repository at https://github.com/kubeedge/website.

Could you please clarify whether the tutorial should be provided as:

* A video with an English voice-over,
* A silent video with the tutorial explained in the description below it, or
* A purely text-based tutorial, similar to a blog post?
Thank you!

luomengY:
> > Please submit a usage tutorial to the repository at https://github.com/kubeedge/website.
> 
> Could you please clarify whether the tutorial should be provided as:
> 
> * A video with an English voice-over,
> * A silent video with the tutorial explained in the description below it, or
> * A purely text-based tutorial, similar to a blog post?
>   for example - [Add blog post for KubeEdge deployment and validation on VisionFive2 website#787](https://github.com/kubeedge/website/pull/787) (from a previous mentee)
>   Thank you!

Just submit a PR for the usage docs, no need to record a video.

tushar743-ui:
Can you review this whenever you are free 
https://github.com/kubeedge/website/pull/862
Thankyou

luomengY:
Close and reopen the PR to make the e2e tests pass.

luomengY:
> Can you review this whenever you are free [kubeedge/website#862](https://github.com/kubeedge/website/pull/862) Thankyou

ok ,thank you

tushar743-ui:
would appreciate a review @WillardHu whenever you are free
Thankyou

tushar743-ui:
Hey [@luomengY](https://github.com/luomengY), this PR currently has a merge conflict (Tide). I think this may be because someone recently made changes to /cloud/pkg/edgecontroller/controller/upstream_test.go, which is also modified in this PR to increase test coverage. What do you think would be the best way to proceed? Should I resolve the conflict by rebasing my changes on top of the latest version, or would you prefer another approach?

luomengY:
> Hey [@luomengY](https://github.com/luomengY), this PR currently has a merge conflict (Tide). I think this may be because someone recently made changes to /cloud/pkg/edgecontroller/controller/upstream_test.go, which is also modified in this PR to increase test coverage. What do you think would be the best way to proceed? Should I resolve the conflict by rebasing my changes on top of the latest version, or would you prefer another approach?

I did not find any merge conflicts. Additionally, I cannot access the workflow run linked here: https://github.com/kubeedge/kubeedge/actions/runs/29824527406/job/88614690930?pr=6838
image

luomengY:
cc @WillardHu

PR #6913, opened 2026-06-26
test(containers): add unit tests for container runtime utility functions

Norway-02:
**What type of PR is this?**
/kind test

Adds unit tests for container runtime utilities, ensuring correct parsing and handling of runtime endpoints and socket configurations.

NONE

None

```release-note
NONE
```

### DIFF
--- a/pkg/containers/container_runtime_test.go
+++ b/pkg/containers/container_runtime_test.go
@@ -105,6 +105,58 @@ func TestCopyResources(t *testing.T) {
 				m.EXPECT().RemovePodSandbox(gomock.Any(), "sb-fail").Return(nil)
 			},
 		},
+		{
+			name:    "Failure: CreateContainer fails",
+			image:   "kubeedge/pause:3.1",
+			wantErr: true,
+			setupMock: func(t *testing.T, m *mock.MockRuntimeService) {
+				m.EXPECT().RunPodSandbox(gomock.Any(), gomock.Any(), gomock.Any()).Return("sb-cc-fail", nil)
+				m.EXPECT().CreateContainer(gomock.Any(), "sb-cc-fail", gomock.Any(), gomock.Any()).Return("", fmt.Errorf("create container error"))
+				m.EXPECT().RemovePodSandbox(gomock.Any(), "sb-cc-fail").Return(nil)
+			},
+		},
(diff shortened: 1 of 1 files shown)


DoisLONG:
Thanks for adding the test coverage.

The scope of this PR is clear and useful: it adds coverage for Windows exec helpers and several `CopyResources` error paths in the container runtime. The container runtime tests are generally aligned with the current implementation flow:

`RunPodSandbox -> CreateContainer -> StartContainer -> ExecSync -> deferred cleanup`

Before this can be merged, please address the following details:

1. `TestNewContainerRuntime_BadEndpoint` currently uses `unix:///nonexistent/cri.sock`. Since `NewContainerRuntime` uses a 10-second CRI connection timeout, this can make the test wait for the gRPC dial timeout before failing.

   Since this test only needs to verify invalid endpoint handling, please use an invalid scheme such as `invalid://endpoint` so it fails immediately and remains stable in CI.

2. The test names:

   * `RemovePodSandbox defer error is logged not returned`
   * `RemoveContainer defer error is logged not returned`

   state that the cleanup errors are logged, but the tests only verify that the deferred errors are not returned. They do not verify any log output.

   Please either rename these cases to describe what is actually asserted, for example `cleanup error does not override successful result`, or explicitly capture and verify the expected log output.

3. Since `exec_windows_test.go` is guarded by the Windows build tag, please state whether the Windows package tests were actually run on a Windows environment. If they were run, please provide the command and result, for example:

   `go test ./pkg/util/execs -count=1`

   If they were only reviewed or cross-compiled but not executed on Windows, please state that clearly.

Please keep this PR test-only and avoid unrelated runtime behavior changes.

Norway-02:
Thanks for the review feedback.

1. Fixed `TestNewContainerRuntime_BadEndpoint` to use `invalid://endpoint` — fails immediately without waiting for the 10-second gRPC dial timeout.
2. Renamed both test cases to accurately describe what is asserted: `cleanup error does not override successful result`.
3. `exec_windows_test.go` was cross-compiled with `GOOS=windows` but not executed on a native Windows environment. Tests were not run on Windows.

Norway-02:
Hi @doislong and @WillardHu,

PR #6919, opened 2026-06-26
Sanitize SQL identifiers in mapper-framework database templates

### LINKED ISSUE #6918: Sanitize SQL identifiers in mapper-framework database templates
### Describe the bug

The MySQL and TDEngine database client templates in the mapper-framework construct SQL queries using `fmt.Sprintf` to interpolate user-controlled data (device names, namespaces, and property names) directly into the SQL string without input validation. 

While device and property names are typically controlled via Kubernetes CRDs and are usually safe, this pattern in a template encourages downstream users to build mappers that are vulnerable to SQL injection if any identifier happens to contain SQL metacharacters.

Affected files:
- `staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/mysql/client.go`
- `staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/tdengine/client.go`

### Steps to reproduce

Review the `AddData` and `GetDataByDeviceID` methods in the aforementioned template files. For example, in the MySQL client:
```go
tableName := data.Namespace + "/" + data.DeviceName + "/" + data.PropertyName
createTable := fmt.Sprintf("CREATE TABLE IF NOT EXISTS `%s` ...", tableName)
```

### Expected behavior

The template should validate that identifiers (like table names derived from device names) contain only alphanumeric characters, underscores, dots, and hyphens before executing the SQL statement to ensure a secure-by-default pattern for developers building custom mappers.

santhil-cyber:
### What type of PR is this?
/kind bug

### What this PR does / why we need it:
This PR adds input validation to the MySQL and TDEngine database clients in the mapper-framework templates. 

Previously, these templates constructed SQL table names and identifiers by directly interpolating device and property names using `fmt.Sprintf`. While these fields are generally populated via CRDs, leaving this pattern in template code could encourage downstream developers to build mappers susceptible to SQL injection.

A `sanitizeIdentifier` function is introduced to validate that any user-derived identifier contains only safe characters (alphanumeric, underscores, hyphens, dots, and slashes) before it is used in a SQL query.

This PR also corrects minor typos in error messages (`msyql`, `parament`) and replaces a debug `fmt.Println` with `klog.V(4).Infof` in the TDEngine client.

### Which issue(s) this PR fixes:
Fixes #6918

### Special notes for your reviewer:
This only touches the `staging/src/github.com/kubeedge/mapper-framework/_template` directory.

### Does this PR introduce a user-facing change?
```release-note
Sanitize SQL identifiers in mapper-framework database templates to prevent SQL injection patterns.
```

### DIFF
--- a/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/mysql/client.go
+++ b/staging/src/github.com/kubeedge/mapper-framework/_template/mapper/data/dbmethod/mysql/client.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"regexp"
 	"time"
 
 	_ "github.com/go-sql-driver/mysql"
@@ -32,6 +33,24 @@ import (
 var (
 	DB *sql.DB
 )
+
+// identifierRegexp defines allowed characters in SQL identifiers to prevent
+// SQL injection when table names are derived from user-controlled data such
+// as device names, namespaces, and property names.
+var identifierRegexp = regexp.MustCompile(`^[a-zA-Z0-9_\-/.]+$`)
+
+// sanitizeIdentifier validates that a SQL identifier (table name, column name)
+// contains only safe characters. This prevents SQL injection when identifiers
+// are constructed from user-controlled input.
+func sanitizeIdentifier(name string) error {
+	if name == "" {
+		return fmt.Errorf("identifier must not be empty")
+	}
+	if !identifierRegexp.MatchString(name) {
+		return fmt.Errorf("identifier %q contains invalid characters, only alphanumeric characters, underscores, hyphens, dots, and slashes are allowed", name)
+	}
+	return nil
+}
 
 type DataBaseConfig struct {
 	MySQLClientConfig *MySQLClientConfig `json:"mysqlClientConfig"`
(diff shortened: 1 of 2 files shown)


DoisLONG:
Thanks for working on this. Adding protection around SQL identifiers in the mapper-framework database templates is a useful direction.

However, the current implementation still needs changes before it can be merged.

1. The identifier validation should match the actual identifier rules of each target database.

   The current expression allows hyphens, dots, and slashes:

   regexp.MustCompile(`^[a-zA-Z0-9_\-/.]+$`)

   Merely excluding obvious SQL metacharacters does not guarantee that the resulting value is a valid MySQL or unquoted TDengine identifier.

   Please define a canonical identifier format appropriate for each database, including restrictions such as the first character and maximum length.

2. Please normalize generated names before validating and using them.

   For example, the MySQL table name currently uses / as a separator:

   data.Namespace + "/" + data.DeviceName + "/" + data.PropertyName

   The TDengine path applies a different transformation and only replaces -.

   Please introduce shared, deterministic helpers for generating MySQL and TDengine identifiers rather than applying transformations separately at each call site.

3. All write and read paths must use the same naming rule.

   AddData, GetDataByDeviceID, and GetDataByTimeRange currently apply different transformations. This may cause the read path to query a different table from the one created by the write path.

4. Identifier validation does not protect SQL values.

   The TDengine insert statement still interpolates values such as data.PropertyName, data.Value, and data.Type directly into SQL through fmt.Sprintf.

   A value containing a quote or SQL syntax can therefore still alter the generated statement.

   Please use parameter binding where supported by the TDengine driver, or a driver-supported and well-tested escaping mechanism. Identifier validation and SQL value handling are separate security requirements.

5. Please consider collisions introduced by normalization.

   For example, blindly converting different unsupported characters to _ may cause distinct source names to produce the same database identifier. The normalization strategy should either reject ambiguous inputs or preserve uniqueness deterministically.

6. Please add unit tests covering at least:

   - normal namespace, device, and property names;
   - names containing -, /, and .;
   - identifiers beginning with a digit;
   - backticks, quotes, spaces, and semicolons;
   - MySQL and TDengine name generation;
   - consistency across create, insert, and query paths;
   - collisions after normalization;
   - TDengine values containing quotes or SQL-like content.

The security goal is good, but the implementation needs to cover identifier validity, consistent name generation, collision handling, and SQL value safety.

santhil-cyber:
Thanks for the thorough review, this is helpful.

You're right that identifier validation and value safety are separate problems, and that the PR title overstates what it currently does. The biggest gap is the TDengine insert: it still interpolates data.PropertyName, data.Value, and data.Type straight into the statement with fmt.Sprintf, so a value containing a quote can still break out regardless of the identifier check. I'll fix that path to use parameter binding (or the driver's supported escaping if binding isn't available for that statement shape). MySQL already parameterizes values via stmt.Exec, so that side is fine.

The read/write inconsistency is also a real one, and it predates this PR: AddData and GetDataByTimeRange normalize - to _, but GetDataByDeviceID queries the raw deviceID, so a device name with a hyphen gets written and read from different tables. I'll route all three paths through a single shared helper so the naming rule is the same everywhere, which also addresses your point about transformations being scattered across call sites.

I'll add unit tests covering the cases you listed: plain names, names with - / ., leading digit, quotes/backticks/spaces/semicolons, MySQL vs TDengine generation, create/insert/query consistency, and TDengine values containing quotes.

One scope question before I push: would you prefer the full canonical per-database identifier format (first-character and length rules, deterministic normalization with collision handling) in this PR, or should I keep this PR focused on the value-injection fix plus consistent name generation and do the stricter identifier format as a follow-up? Happy to do either, just want to keep the change reviewable.

PR #6921, opened 2026-06-27
fix(pkg/image): add missing index bounds check for image digest parsing

Norway-02:
**What type of PR is this?**
/kind bug

When parsing image digests, an out-of-bounds array access could occur if the digest format was invalid or missing a part. This PR adds the missing index bounds check to gracefully handle malformed digests and prevent a runtime panic in the edge node.

NONE

None

```release-note
NONE
```

### DIFF
--- a/pkg/image/image_runtime.go
+++ b/pkg/image/image_runtime.go
@@ -82,13 +82,48 @@ func (runtime *RuntimeImpl) GetImageDigest(ctx context.Context, image string) (s
 	if resp.Image == nil {
 		return "", nil
 	}
-	for i := range resp.Image.RepoTags {
-		tag := resp.Image.RepoTags[i]
+
+	// Confirm the requested tag exists in RepoTags.
+	tagFound := false
+	for _, tag := range resp.Image.RepoTags {
 		if tag == image {
-			repoDigest := resp.Image.RepoDigests[i]
-			digestIndex := strings.LastIndex(repoDigest, "@sha256:")
-			return repoDigest[digestIndex+1:], nil
+			tagFound = true
+			break
 		}
+	}
+	if !tagFound {
+		return "", nil
+	}
+
+	// Extract repository name.
+	ref, err := reference.ParseAnyReference(image)
+	if err != nil {
+		return "", fmt.Errorf("invalid image reference %s: %w", image, err)
+	}
+	named, ok := ref.(reference.Named)
+	if !ok {
+		return "", fmt.Errorf("invalid image reference %s: not a named reference", image)
+	}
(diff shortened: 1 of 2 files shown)


DoisLONG:
Thanks for fixing the missing @sha256: handling. The current check addresses one invalid case, but the digest lookup still needs to be made more robust.

1. Please do not assume that RepoTags and RepoDigests have matching indexes.

   The current code iterates over RepoTags and then accesses:

   repoDigest := resp.Image.RepoDigests[i]

   These are independent CRI fields and are not guaranteed to have the same length or ordering. This can return the wrong digest or panic when RepoTags is longer than RepoDigests.

   After confirming that the requested tag exists, please inspect RepoDigests independently and select a digest corresponding to the same repository name.

2. Checking only for the presence of @sha256: is not sufficient.

   Values such as the following still pass the current condition:

   image@sha256:
   image@sha256:123
   image@sha256:not-hex

   Please parse and validate the complete digest. The repository already depends on github.com/opencontainers/go-digest, which can be used instead of duplicating digest parsing. If this function specifically requires SHA-256, please also verify the parsed algorithm.

3. Please add regression tests covering at least:

   - a valid SHA-256 digest;
   - a matching tag with no digest separator;
   - an empty digest after @sha256:;
   - a truncated digest;
   - non-hex digest content;
   - RepoTags longer than RepoDigests, verifying that no panic occurs;
   - multiple tags and digests with different ordering;
   - no matching tag.

4. The current commit is also marked with do-not-merge/invalid-commit-message, apparently because the commit subject contains @sha256:.

   Please rename the commit subject to satisfy the repository commit-message validation and remove the do-not-merge label.

The direction of the fix is correct, but the implementation should avoid index assumptions and validate the complete digest before returning it.

Norway-02:
Thanks for the detailed review. All points addressed:

1. Removed index assumption — `GetImageDigest` now iterates `RepoDigests` independently and matches by repository name prefix, avoiding wrong digest or panic when `RepoTags` and `RepoDigests` have different lengths or ordering.
2. Full digest validation via `github.com/opencontainers/go-digest` — rejects empty, truncated, and non-hex digests.
3. Added regression tests covering all requested cases: valid sha256, tag not in RepoTags, empty digest, truncated digest, non-hex content, RepoTags longer than RepoDigests (no panic), multiple tags+digests with different ordering, no matching repo in RepoDigests.
4. Renamed commit subject to remove `@sha256:` — `do-not-merge/invalid-commit-message` should clear.

Norway-02:
Hi @doislong and @WillardHu,

PR #6922, opened 2026-06-27
servicebus: add transport-layer TLS to the embedded HTTP server

### LINKED ISSUE #6920: servicebus: add transport-layer TLS to the embedded HTTP server
kind/feature

## Problem

The ServiceBus embedded HTTP server currently accepts plain HTTP only. Traffic between local applications and ServiceBus is unencrypted in transit, and the server has no verifiable identity.

## Scope of this issue

**Transport encryption + server authentication only.**

This issue tracks adding optional TLS to the ServiceBus HTTP server so:
1. Traffic between local applications and ServiceBus is encrypted in transit.
2. Clients can verify the server identity via the server certificate.

This issue does **not** cover caller authentication (restricting which local processes may submit ServiceBus requests). TLS alone cannot prevent a local process that can reach the port from sending requests.

A separate follow-up issue should track caller authentication options: mTLS, application-level credentials, or a permission-controlled Unix socket.

## Solution

Addressed by PR #6922:
- Adds `tlsCertFile` / `tlsPrivateKeyFile` fields to `v1alpha2.ServiceBus` config
- Wires config through the real production call site
- Fixes dynamic/delayed startup to use the configured TLS options
- Fixes `inited` flag reset on failed TLS startup
## Status

Resolved by #6922. For caller authentication (mTLS / Unix socket),
please open a separate follow-up issue referencing this one.

Priyanshubhartistm:
/kind feature

Adds optional server-only TLS to the ServiceBus embedded HTTP server. When enabled, all traffic between local applications and the ServiceBus HTTP server is encrypted in transit and clients can verify the server identity via certificate.

**Scope: transport encryption + server authentication only.**
TLS does not restrict which local processes may submit ServiceBus requests.
Any process that can reach the configured port can still send requests.
Caller authentication (mTLS, application credentials, or a Unix socket with permission controls) is a separate concern tracked in a follow-up issue.

## Changes

- `v1alpha2.ServiceBus` gains two optional fields: `tlsCertFile` and `tlsPrivateKeyFile`. Both default to empty string — existing configurations are unaffected (plain HTTP continues to work as before).
- The production call site in `registerModules` (`edge/cmd/edgecore/app/server.go`) reads these fields and passes them to `servicebus.Register`, so TLS can now actually be enabled through real production configuration.
- `Register` stores the TLS options in a package-level `configuredTLSOpts` variable so that both initial and dynamic/delayed server startup (when the URL table is empty at boot) use the same TLS configuration.
- `server()` now resets `inited` on every failed TLS startup path so that an invalid certificate does not permanently block a later valid startup attempt.
- `buildTLSConfig` uses `GetCertificate` so certificates can be rotated on disk without restarting EdgeCore.

## Certificate requirements

The certificate supplied via `tlsCertFile` **must**:
- Have `ExtKeyUsageServerAuth`
- Have an IP/DNS SAN matching `ServiceBus.Server` (default `127.0.0.1`)

The EdgeHub client certificate issued by CloudCore **cannot** be reused: it carries only `ExtKeyUsageClientAuth` and has no ServiceBus SANs.

## Security scope

This PR provides **transport encryption and server authentication only**.
It does not add caller authentication — any local process that can reach the port may still submit ServiceBus requests.

For caller authentication (mTLS, application-level credentials, or a permission-controlled Unix socket), please open a separate follow-up issue.

## How to enable TLS

```yaml
modules:
  serviceBus:
    enable: true
    server: 127.0.0.1
    port: 9060
    timeout: 60
    tlsCertFile: /etc/kubeedge/certs/servicebus-server.crt
    tlsPrivateKeyFile: /etc/kubeedge/certs/servicebus-server.key
```

For a full configuration guide including certificate generation, rotation, and client migration, see [`docs/proposals/sig-node/servicebus-tls.md`](docs/proposals/sig-node/servicebus-tls.md).

Fixes #6920

**Test command and output:**
`go test -race -v ./edge/pkg/servicebus/...`

```release-note
ServiceBus: add optional TLS support for the embedded HTTP server.
Set `tlsCertFile` and `tlsPrivateKeyFile` under `modules.serviceBus`
in edgecore.yaml to enable HTTPS. Leaving both fields empty (the
default) keeps the existing plain-HTTP behaviour — no action required
for existing deployments. The certificate must have ExtKeyUsageServerAuth
and an IP/DNS SAN matching ServiceBus.Server. See
docs/proposals/sig-node/servicebus-tls.md for the full configuration guide.
```

image

### DIFF
--- a/codespell_whitelist.txt
+++ b/codespell_whitelist.txt
@@ -1,3 +1,4 @@
+CAS
 Creater
 Convertor
 deviceC
--- a/docs/proposals/sig-node/servicebus-tls.md
+++ b/docs/proposals/sig-node/servicebus-tls.md
@@ -1,0 +1,137 @@
+# ServiceBus TLS Configuration
+
+## Overview
+
+The ServiceBus embedded HTTP server can optionally be protected with
+transport-layer TLS (HTTPS). When TLS is enabled all traffic between
+local applications and the ServiceBus HTTP server is encrypted in
+transit, and clients can verify the server identity via certificate.
+
+**Scope:** transport encryption and server authentication only.
+TLS does not restrict which local processes may submit ServiceBus
+requests. Any process that can reach the configured port can still send
+requests. Caller authentication (mTLS, application credentials, or a
+Unix-socket with permission controls) is a separate concern.
+
+## Prerequisites
+
+You need a dedicated server certificate and private key for the
+ServiceBus endpoint. The certificate must satisfy:
+
+| Requirement | Details |
+|---|---|
+| Key usage | `ExtKeyUsageServerAuth` |
+| Subject Alternative Name | An IP or DNS SAN matching `ServiceBus.Server` (default `127.0.0.1`) |
+| Format | PEM-encoded X.509 certificate (`.crt` / `.pem`) |
+| Private key | PEM-encoded RSA or ECDSA key matching the certificate (`.key`) |
+
+> **Important:** The EdgeHub client certificate issued by CloudCore
+> **cannot** be reused. It carries only `ExtKeyUsageClientAuth` and has
+> no ServiceBus SANs. Attempting to use it will cause TLS handshake
+> failures in any standard HTTPS client.
+
+### Generating a self-signed certificate (example)
+
+```bash
+# Generate a 4096-bit RSA key
+openssl genrsa -out /etc/kubeedge/certs/servicebus-server.key 4096
+
+# Generate a self-signed certificate valid for 1 year,
+# with IP SAN 127.0.0.1 (the default ServiceBus listen address).
+openssl req -new -x509 -days 365 \
+  -key /etc/kubeedge/certs/servicebus-server.key \
+  -out /etc/kubeedge/certs/servicebus-server.crt \
+  -subj "/CN=servicebus.edge.local" \
+  -addext "subjectAltName=IP:127.0.0.1" \
+  -addext "extendedKeyUsage=serverAuth"
+```
+
+For production deployments, issue the certificate from your internal CA
+or a certificate management tool (e.g. cert-manager) rather than using
+a self-signed certificate.
+
+## Enabling TLS in EdgeCore configuration
+
+Set `tlsCertFile` and `tlsPrivateKeyFile` under `modules.serviceBus` in
+your EdgeCore configuration (`/etc/kubeedge/config/edgecore.yaml`):
+
+```yaml
+apiVersion: edgecore.config.kubeedge.io/v1alpha2
+kind: EdgeCore
+modules:
+  serviceBus:
+    enable: true
+    server: 127.0.0.1
+    port: 9060
+    timeout: 60
+    # Enable TLS by providing a dedicated server certificate.
+    # Leave both fields empty (the default) to use plain HTTP.
+    tlsCertFile: /etc/kubeedge/certs/servicebus-server.crt
+    tlsPrivateKeyFile: /etc/kubeedge/certs/servicebus-server.key
+```
+
+When either `tlsCertFile` or `tlsPrivateKeyFile` is non-empty, EdgeCore
+treats TLS as intended and starts the ServiceBus HTTP server as an HTTPS
+server. When both fields are empty (the default), the server starts in
+plain-HTTP mode, preserving backward compatibility.
+
+If TLS is considered intended (i.e. either field is non-empty) but the
+configuration is incomplete or invalid — for example only one of the two
+fields is set, a file is missing, or the key pair cannot be loaded —
+EdgeCore will **not** silently fall back to HTTP. It will log a clear
+error and refuse to start the ServiceBus HTTPS server. This prevents a
+certificate typo or missing file from silently removing transport security.
+
+Note: EdgeCore validates two things at startup before the HTTPS server
+starts:
+
+1. **Key-pair loadability** — the certificate and key files must form a
+   valid PEM key pair; otherwise EdgeCore logs an error and refuses to
+   start the HTTPS server.
+2. **`ExtKeyUsageServerAuth`** — the certificate must contain the
+   `serverAuth` extended key usage. EdgeCore checks this at startup via
+   `validateServerAuthEKU()`. A ClientAuth-only certificate (such as the
+   EdgeHub client certificate issued by CloudCore) is rejected
+   immediately with a clear error message, before the server starts.
+   The same check is applied again during certificate rotation: a
+   rotated certificate that lacks `ExtKeyUsageServerAuth` is also
+   rejected, so startup validation cannot be bypassed by a later
+   certificate swap.
+
+**SAN matching is not checked by EdgeCore.** A certificate whose Subject
+Alternative Names do not include the ServiceBus listen address (e.g.
+`127.0.0.1`) will start the server successfully, but HTTPS clients will
+reject the TLS handshake because they enforce SAN matching during
+certificate verification. Always include the correct IP or DNS SAN in
+your certificate (see the `openssl` example above).
+
+## Updating local application clients
+
+When TLS is enabled, local applications that previously connected to
+`http://127.0.0.1:9060` must switch to `https://127.0.0.1:9060` and
+trust the ServiceBus server certificate (or its CA). For example, using
+`curl`:
+
+```bash
+# Plain HTTP (TLS disabled — default)
+curl http://127.0.0.1:9060/
+
+# HTTPS (TLS enabled) — trust the self-signed cert or your CA bundle
+curl --cacert /etc/kubeedge/certs/servicebus-server.crt \
+     https://127.0.0.1:9060/
+```
+
+## Certificate rotation
+
+The ServiceBus TLS implementation uses `GetCertificate`, which re-reads
+the certificate and key files on every new TLS handshake. You can rotate
+the certificate by replacing the files on disk; no EdgeCore restart is
+required.
+
+## Plain-HTTP default (backward compatibility)
+
+Leaving **both** `tlsCertFile` and `tlsPrivateKeyFile` empty (or absent
+from the configuration) keeps the ServiceBus server in plain-HTTP mode,
+exactly as before this feature was introduced. Existing deployments are
+unaffected. Setting either field without the other signals TLS intent and
+will result in a startup error if the pair is incomplete.
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -213,8 +213,33 @@ func registerModules(c *v1alpha2.EdgeCoreConfig) {
 	edgehub.Register(c.Modules.EdgeHub, c.Modules.Edged.HostnameOverride)
 	eventbus.Register(c.Modules.EventBus, c.Modules.Edged.HostnameOverride)
 	metamanager.Register(c.Modules.MetaManager)
-	servicebus.Register(c.Modules.ServiceBus)
+	servicebus.Register(c.Modules.ServiceBus, buildServiceBusTLSOptions(c.Modules.ServiceBus))
 	edgestream.Register(c.Modules.EdgeStream, c.Modules.Edged.HostnameOverride, c.Modules.Edged.NodeIP)
 	taskmanager.Register(c.Modules.TaskManager)
 	test.Register(c.Modules.DBTest)
+}
+
+// buildServiceBusTLSOptions builds ServiceBus TLS options from the config.
+// TLS is considered intended when either TLSCertFile or TLSPrivateKeyFile is
+// non-empty.  An incomplete pair (one field set, the other empty) is signalled
+// by returning TLSEnabled=true with the partial paths, causing buildTLSConfig
+// to return a clear error rather than silently falling back to plain HTTP.
+// Both fields empty (the default) preserves the plain-HTTP behaviour.
+// The EdgeHub client certificate CANNOT be reused: it carries only
+// ExtKeyUsageClientAuth and has no ServiceBus SANs.
+func buildServiceBusTLSOptions(sb *v1alpha2.ServiceBus) servicebus.TLSOptions {
+	if sb == nil {
+		return servicebus.TLSOptions{}
+	}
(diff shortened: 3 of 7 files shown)


DoisLONG:
Thanks for working on TLS support for ServiceBus. Securing the local ServiceBus endpoint is a useful direction, but the current implementation has several blocking compatibility and certificate issues.

1. The EdgeHub certificate is not suitable for use as a ServiceBus server certificate.

The certificate requested by EdgeCore is issued as a client-auth certificate by default. Its CSR uses a subject such as `system:node:<nodeName>` and does not include DNS or IP SANs for the configured ServiceBus address.

A normal HTTPS client verifying `127.0.0.1`, localhost, or another configured address will therefore not be able to validate this certificate as a server certificate.

Please use a certificate explicitly issued for server authentication with the correct SANs, or introduce dedicated ServiceBus TLS certificate configuration.

2. Please do not implicitly change ServiceBus from HTTP to HTTPS based only on whether EdgeHub certificate paths are populated.

Existing installations normally have EdgeHub certificate paths configured, while existing ServiceBus clients may still use HTTP. This change would silently switch the protocol on the same address and port and break those clients.

Please add an explicit ServiceBus TLS enable/configuration option and preserve HTTP as the default for backward compatibility.

3. The current CA handling also implicitly enables mandatory mTLS:

`tlsCfg.ClientAuth = tls.RequireAndVerifyClientCert`

Local applications using ServiceBus are not necessarily provisioned with EdgeCore client certificates. Please make the client-auth policy explicit and configurable.

Also, please check the return value of `AppendCertsFromPEM`. Enabling `RequireAndVerifyClientCert` with an empty CA pool would reject every client.

4. Certificate rotation is not handled.

The certificate is loaded once into `tls.Config.Certificates` when the server starts. When CertManager rotates the EdgeHub certificate files, ServiceBus will continue serving the old in-memory certificate until EdgeCore restarts.

Please either support certificate reload through `GetCertificate` or define a separate lifecycle for the ServiceBus server certificate.

5. Please avoid `klog.Fatalf` for certificate loading failures.

A bad ServiceBus TLS configuration should return an initialization error or follow a clearly defined fallback policy. It should not unexpectedly terminate the entire EdgeCore process.

6. Please add tests covering:

- TLS disabled and HTTP compatibility;
- a valid server certificate with matching SANs;
- invalid certificate and key handling;
- optional and required client authentication;
- invalid CA data;
- certificate rotation or reload behavior.

The TLS goal is good, but the EdgeHub client certificate cannot be reused directly as an HTTPS server identity.

Priyanshubhartistm:
Hi @DoisLONG,

I have addressed the requested feedback and updated the PR accordingly. The PR is now ready for review.

If any further changes are needed, please let me know. I would be happy to make them.

Thank you!

DoisLONG:
Thanks for the update. The new version improves error handling, checks invalid CA content, avoids `klog.Fatalf`, and introduces `GetCertificate` for reload support.

However, the two main issues from the previous review remain unresolved.

### 1. The EdgeHub certificate is still not a valid ServiceBus server identity

The implementation still reuses `EdgeHub.TLSCertFile` and `EdgeHub.TLSPrivateKeyFile`.

The certificate requested by EdgeCore is generated from a CSR whose subject is `system:node:<nodeName>` and whose SAN argument is nil. CloudCore also signs it with `ExtKeyUsageClientAuth` by default.

ServiceBus listens on addresses such as `127.0.0.1:9060`. A normal HTTPS client verifies both the server-auth extended key usage and the requested hostname/IP. The current EdgeHub certificate therefore cannot be validated as a ServiceBus server certificate.

Please provision a certificate explicitly containing:

* `ExtKeyUsageServerAuth`;
* an IP or DNS SAN matching `ServiceBus.Server`.

A dedicated ServiceBus server certificate is preferable to expanding the trust and usage of the EdgeHub client identity.

### 2. The implementation still changes HTTP to HTTPS implicitly

ServiceBus has no TLS enable field. Since the EdgeHub cert and key paths are populated by default, enabling ServiceBus will automatically switch the existing address and port from HTTP to HTTPS.

Existing callers, including the repository E2E helper, still use:

```go
http://127.0.0.1:9060
```

Please add explicit ServiceBus TLS configuration and keep HTTP as the default for backward compatibility. The protocol must not be inferred from unrelated EdgeHub certificate paths.

### 3. Do not silently downgrade an explicitly enabled TLS endpoint

`buildTLSConfig` currently returns `(nil, nil)` when either the cert or key path is empty, and any key-pair loading error causes `server` to start plain HTTP.

Please distinguish:

* TLS disabled: start HTTP;
* TLS enabled with valid configuration: start HTTPS;
* TLS enabled with incomplete or invalid configuration: return an error and do not start plaintext HTTP.

Otherwise a certificate typo, missing file, or first-start timing issue silently removes transport security.

### 4. Clarify server-only TLS versus mTLS

The current configuration sets `ClientAuth: tls.NoClientCert`, so the loaded `ClientCAs` pool is unused.

If this PR intentionally provides server-only TLS, please remove the unused client CA handling and document that it does not prevent an unauthorized local process from sending requests.

If the goal includes preventing request forgery as described in #6920, client authentication must be explicitly configurable, with a defined client-certificate provisioning workflow.

### 5. The integration test does not test the configured certificate

`httptest.Server.StartTLS` injects its own localhost certificate when `TLS.Certificates` is empty and configures `srv.Client()` to trust that certificate.

Since the current config uses only `GetCertificate`, this test can pass using the `httptest` certificate rather than the certificate loaded from the supplied files. It therefore does not detect the real EdgeHub certificate's ClientAuth-only usage or missing SANs.

Please use a real TLS listener/server and a normal `http.Client` configured with the generated CA. Test:

* a `ServerAuth` certificate with a matching SAN succeeds;
* a ClientAuth-only certificate fails;
* a SAN mismatch fails;
* HTTP remains available when TLS is disabled.

### 6. Test actual certificate reload

`TestBuildTLSConfigGetCertificateReloads` currently calls `GetCertificate` only once. Please replace the certificate and key files with a second key pair and verify that the next call returns the second certificate.

Please also guard `cfg.Modules` and `cfg.Modules.EdgeHub`, or pass the TLS configuration explicitly through `servicebus.Register` instead of reading command-level global options from the package.

After revising the certificate and configuration model, please provide the focused race-enabled test output and an end-to-end request using a normally verifying HTTPS client.

Priyanshubhartistm:
Hi @DoisLONG, Thank you for the review,  I’ve addressed the review feedback. 

Please let me know if there are any other changes or improvements needed.
thnku so much

DoisLONG:
Thanks for the substantial redesign. The explicit TLS options, backward-compatible HTTP default, dedicated ServerAuth certificate requirement, real TLS integration tests, and certificate reload test are all improvements.

There are still several production-path issues to address.

### 1. TLS cannot currently be enabled by any production configuration

The only production call site passes:

```go
servicebus.Register(c.Modules.ServiceBus, servicebus.TLSOptions{})
```

so `TLSEnabled` is always false. The `v1alpha2.ServiceBus` configuration has no TLS fields, and there is no other production caller supplying certificate paths.

Please wire the TLS options into the actual EdgeCore configuration in this PR, or re-scope this as preparatory refactoring and do not mark #6920 as fixed yet.

### 2. Dynamic startup silently discards the configured TLS options

When the URL table is empty at startup, the server is later started from `processMessage` using:

```go
go server(c, TLSOptions{})
```

This always starts plain HTTP, even if the ServiceBus module was configured for TLS.

Please retain the module's TLS options and use the same options for both initial and dynamic server startup. Add a regression test for the “empty table at startup, first rule added later” path.

### 3. TLS startup failure leaves `inited` stuck at 1

The code sets `inited` before starting `server` asynchronously. If `buildTLSConfig` returns an error, `server` returns before installing the stop/shutdown goroutine, and `inited` is never reset.

The result is a server that is not listening but can never be started again.

Please validate TLS configuration before committing the initialized state, or ensure every server exit path resets the state and cleans up its shutdown goroutine. Add a test proving an invalid TLS configuration does not permanently prevent a later successful start.

Also note that the `TLSOptions` comment says `Register` returns an error, but `Register` currently has no return value and TLS validation happens asynchronously.

### 4. Clarify the security scope

Server-only TLS provides encryption and server authentication, but it does not authenticate local clients. Any local process that can reach the port can still submit ServiceBus requests.

If #6920 includes preventing forged local calls, this requires mTLS, an application credential, a permission-controlled Unix socket, or another client-authentication mechanism. Otherwise, please narrow the issue and PR claims to transport encryption only.

Finally, please update the PR title: the current implementation explicitly rejects reusing the EdgeCore/EdgeHub client certificate, so “reusing EdgeCore certificate” is no longer accurate.

Priyanshubhartistm:
Hi @DoisLONG , could you please take a look at this PR when you have time? I’ve addressed the review feedback and would really appreciate your feedback.

DoisLONG:
Thanks for iterating on this. 

One small suggestion: since this PR adds new user-facing ServiceBus config fields (`tlsCertFile` and `tlsPrivateKeyFile`) and changes how the embedded ServiceBus HTTP server can be exposed when enabled, could you also fill in the release note and, if appropriate, add/update a short config example in the docs or sample config? This would make it clearer for users how to enable TLS and what certificate type is expected.

Other than that, the scope and implementation look reasonable to me.

Priyanshubhartistm:
Hi @DoisLONG , could you please take another look at this PR when you have time?
thanks

DoisLONG:
Thanks for the continued updates. The current version looks much more complete to me.

The production path is now wired from the EdgeCore ServiceBus config into `servicebus.Register()`, the scope is clearly server-only TLS, and the release note / configuration guide are useful for users. The previous Gemini comments also look outdated now since this no longer reads EdgeHub TLS config or configures unused `ClientCAs`.

One thing I’d like to clarify before LGTM: the docs and PR description say that the ServiceBus certificate must have `ExtKeyUsageServerAuth` and a SAN matching `ServiceBus.Server`, and the guide also says EdgeCore will refuse to start the HTTPS server if the certificate does not have `ExtKeyUsageServerAuth`.

However, the current `buildTLSConfig()` only validates that the key pair can be loaded with `tls.LoadX509KeyPair()`. It does not parse the certificate and enforce `ExtKeyUsageServerAuth` or SANs at startup. The current test for an EdgeHub-style ClientAuth-only certificate also documents that `buildTLSConfig()` succeeds, while a normal HTTPS client rejects it during verification.

Could you align the implementation and documentation here?

Either:

1. add startup validation for the certificate EKU/SANs so the server really refuses this misconfiguration, or
2. update the docs/release note to say that EdgeCore only validates the key pair is loadable, and that clients will reject the certificate if it lacks the required ServerAuth/SAN properties.

Since this is a security-facing configuration, I think it is important that the docs do not claim enforcement that the code does not actually perform.

Priyanshubhartistm:
Hi @DoisLONG , could you please take another look at this PR when you have time?
thanks

DoisLONG (commented):
Thanks for the continued updates. This version is much clearer than the earlier one.

The overall direction looks reasonable to me now: ServiceBus TLS is optional, plain HTTP remains the backward-compatible default, the implementation no longer tries to reuse the EdgeHub client certificate, and the PR now clearly scopes this to transport encryption plus server authentication only. Keeping mTLS / caller authentication out of scope is also the right call for this PR.

I still see a few remaining issues before LGTM:

1. The docs are inconsistent with the current implementation. The code now validates `ExtKeyUsageServerAuth` at startup through `validateServerAuthEKU()`, but `docs/proposals/sig-node/servicebus-tls.md` still says EdgeCore does not inspect EKU or SANs at startup. Please update the docs to say that EKU is checked by EdgeCore, while SAN matching is enforced by HTTPS clients during the handshake.

2. Please make the certificate-rotation path follow the same EKU rule. The initial startup path validates `ExtKeyUsageServerAuth`, but the `GetCertificate` callback used during rotation only reloads the key pair. If a rotated certificate lacks `ServerAuth`, it should be rejected the same way instead of bypassing the startup validation.

3. `buildServiceBusTLSOptions()` currently treats only `tlsCertFile` as the signal to enable TLS. If an operator sets only `tlsPrivateKeyFile`, ServiceBus silently stays on plain HTTP. I think either TLS field being set should mean TLS was intended, and an incomplete cert/key pair should fail clearly rather than falling back to HTTP.

4. `cloud/pkg/csidriver/uds_test.go` still looks unrelated to ServiceBus TLS. Please remove it from this PR or split it into a separate fix.

The current CI results look good so far, with ARM64 still pending. After these cleanups, this should be much closer from my side.

DoisLONG (commented):
Thanks for the update. This addresses the main issues from my previous review.

The docs now match the implementation better: EdgeCore validates key-pair loadability and `ExtKeyUsageServerAuth`, while SAN matching is left to HTTPS clients. The rotation path also re-validates EKU in `GetCertificate`, so a rotated ClientAuth-only certificate cannot bypass the startup check. The unrelated `csidriver` test change is gone as well.

The ServiceBus TLS direction looks good to me now: TLS remains optional, plain HTTP is still the backward-compatible default, and an explicitly requested TLS config does not silently fall back to HTTP on configuration errors.

One small cleanup before LGTM: the docs still say “When `tlsCertFile` is non-empty...” / “If `tlsCertFile` is set...”, but the code now treats either `tlsCertFile` or `tlsPrivateKeyFile` being non-empty as TLS intent. Could you please align that wording with the current behavior?

It would also be useful to add cert-only and key-only cases to `TestBuildServiceBusTLSOptions()` so the incomplete-pair behavior is covered directly.

DoisLONG (commented):
Thanks for the update. This addresses the remaining cleanup from my previous review.

The docs now match the current behavior: either `tlsCertFile` or `tlsPrivateKeyFile` being non-empty means TLS is intended, while leaving both fields empty preserves the plain-HTTP default. The cert-only and key-only cases are also now covered in `TestBuildServiceBusTLSOptions()`, so the incomplete-pair behavior is tested directly.

The overall optional-TLS direction looks good to me now, and I do not see another code-side blocker from my previous review.

One last thing before LGTM: Main CI is currently failing. Could you please check whether the failure is related to this PR, or retrigger/fix it if needed?

DoisLONG (commented):

/lgtm

Priyanshubhartistm:

PR #6926, opened 2026-06-28
fix(devicetwin): make metadata unmarshal failures explicit

### LINKED ISSUE #6923: Handle invalid metadata unmarshaling explicitly in DeviceTwinToMsgTwin
## Summary

DeviceTwinToMsgTwin currently logs JSON unmarshal failures for metadata fields and continues processing.

The current behavior works, but metadata handling after unmarshal failure is implicit and not clearly expressed.

## Current behavior

When unmarshaling fails for metadata fields:
- the error is logged
- processing continues
- metadata handling is not explicit

## Proposal

Handle invalid metadata explicitly by:

- preserving current graceful degradation behavior
- skipping invalid metadata assignment when unmarshaling fails
- adding unit tests covering malformed metadata inputs

No behavior changes for valid inputs.

## Scope

Files:
- edge/pkg/devicetwin/dttype/types_helper.go
- edge/pkg/devicetwin/dttype/types_helper_test.go

If maintainers agree, I can work on the implementation.

barbierajput378-pixel:
**What type of PR is this?**

/kind bug

---

`DeviceTwinToMsgTwin` currently logs JSON unmarshal failures for metadata/version fields but still assigns zero-value metadata structures implicitly.

This PR makes the behavior explicit and more reliable by:

* logging metadata/version unmarshal failures with warnings
* skipping assignment of invalid metadata/version values
* preserving existing graceful degradation behavior
* adding unit tests covering invalid metadata/version scenarios

Valid metadata behavior remains unchanged.

---

Fixes #6923 

---

* This change intentionally avoids returning errors to prevent interrupting device synchronization flows.
* Invalid metadata/version JSON is treated as non-fatal and skipped.
* Added tests for:

  * valid metadata
  * invalid ExpectedMeta
  * invalid ActualMeta
  * invalid ExpectedVersion
  * invalid ActualVersion

Verification:

```bash
go test ./edge/pkg/devicetwin/dttype -v
```

---

```release-note
NONE
```

### DIFF
--- a/edge/pkg/devicetwin/dttype/types_helper.go
+++ b/edge/pkg/devicetwin/dttype/types_helper.go
@@ -74,10 +74,10 @@ func DeviceTwinToMsgTwin(deviceTwins []models.DeviceTwin) map[string]*MsgTwin {
 			expectedValue := &TwinValue{Value: &expected}
 			if twin.ExpectedMeta != "" {
 				if err := json.Unmarshal([]byte(twin.ExpectedMeta), &expectedMeta); err != nil {
-					// TODO: handle error
-					klog.Error(err)
+					klog.Warningf("skip invalid ExpectedMeta for twin %q: %v", twin.Name, err)
+				} else {
+					expectedValue.Metadata = &expectedMeta
 				}
-				expectedValue.Metadata = &expectedMeta
 			}
 			msgTwin.Expected = expectedValue
 		}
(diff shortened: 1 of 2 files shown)


DoisLONG:
Thanks for improving the metadata and version unmarshal handling.

The malformed JSON cases are handled correctly now, but the valid JSON `null` case still needs to be addressed.

The current implementation unmarshals into value structs and assigns their addresses after a successful unmarshal. For example, unmarshalling `"null"` into `ValueMetadata` succeeds and leaves the value as a zero-value struct. The resulting metadata pointer is therefore still non-nil even though no metadata was provided.

Please unmarshal into pointer variables instead, for example:

    var expectedMeta *ValueMetadata
    if err := json.Unmarshal([]byte(twin.ExpectedMeta), &expectedMeta); err != nil {
        klog.Warningf("skip invalid ExpectedMeta for twin %q: %v", twin.Name, err)
    } else {
        expectedValue.Metadata = expectedMeta
    }

Please apply the same pattern to:

- ExpectedMeta
- ActualMeta
- ExpectedVersion
- ActualVersion

Please also add test cases using `"null"` for all four fields and verify that the corresponding result pointers remain nil.

The existing malformed JSON tests are useful, but the valid JSON null case should be covered before merge.

barbierajput378-pixel:
Fixed! Updated to unmarshal into pointer variables for all four fields 
(ExpectedMeta, ActualMeta, ExpectedVersion, ActualVersion). 
Added test cases with "null" JSON values to verify pointers remain nil. Ready for re-review!

PR #6928, opened 2026-06-28
test: add unit tests for extsystem_unix

Sarvesh-Mishra1981:
### What type of PR is this?
/kind cleanup
/kind testing

### What this PR does / why we need it:
This PR improves the unit test coverage for the `keadm` utility. Currently, the `extsystem` package, which is responsible for managing system service files, lacks test coverage. To make the code testable, I changed the `systemdDir` constant in `extsystem_unix.go` to a package-level variable. This refactor enables unit tests to point the path to a temporary directory using `t.TempDir()`, preventing permission-denied errors when running tests locally or in CI since we no longer write to the root `/etc/systemd/system` folder. I then added unit tests in `extsystem_unix_test.go` to verify that creating and removing service files works correctly.

### Which issue(s) this PR fixes:
This is a proactive improvement to test coverage, so there is no pre-existing issue associated with it.

### Fixes #
None.

### Special notes for your reviewer:
The tests have been written to run in isolation using mock directories so they do not impact any host configuration files during execution.

### Does this PR introduce a user-facing change?:
No, this change only affects internal testing infrastructure and does not modify any user-facing APIs or command-line behavior.

### DIFF
--- a/keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_unix.go
+++ b/keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_unix.go
@@ -56,7 +56,7 @@ type SystemdExtSystem struct {
 	initsystem.SystemdInitSystem
 }
 
-const (
+var (
 	systemdDir = "/etc/systemd/system"
 )
 
--- a/keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_unix_test.go
+++ b/keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_unix_test.go
@@ -1,0 +1,107 @@
+//go:build !windows
+
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


DoisLONG:
Thanks for adding coverage for the Unix init-system helpers.

Using `t.TempDir()` together with the package-level `systemdDir` test seam avoids writing to the real `/etc/systemd/system` directory and exercises the actual service creation and removal paths.

A couple of non-blocking notes:

- `t.Cleanup` would be slightly more idiomatic than `defer` for restoring `systemdDir`.
- Since the test mutates package-level state, please do not mark this test or related tests with `t.Parallel`.
- The repository does not have a `kind/testing` label. Please use `/kind test` instead.

I do not see a blocking issue with the current implementation.

/lgtm

WillardHu:
Please fix the DCO check error, ref: https://github.com/kubeedge/kubeedge/pull/6928/checks?check_run_id=83933955042

keadm/cmd/keadm/app/cmd/util/extsystem/extsystem_unix.go
```diff
@@ -56,7 +56,7 @@ type SystemdExtSystem struct {
 	initsystem.SystemdInitSystem
 }
 
-const (
+var (
```
WillardHu, line 59: Don't modify constant definitions for unit tests. You can create `systemDir` during unit tests and delete the directory when the unit tests end.

PR #6932, opened 2026-06-29
Fix metaserver readyz authentication for discovery paths

### LINKED ISSUE #5961: metaserver readyz：Unable to authenticate the request
**What happened**:
`curl -v -k -H "Authorization: Bearer $TOKEN" https://169.254.30.10:10550/readyz`
`curl -k https://169.254.30.10:10550/readyz`

return Unauthorized

> {
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "Unauthorized",
  "reason": "Unauthorized",
  "code": 401
}

logs:

> Nov 14 12:09:48 k8s-test-edge-02 edgecore[339548]: E1114 12:09:48.010715  339548 authentication.go:73] "Unable to authenticate the request" err="serviceaccount kube-system/spiderpool-agent not found"

return ok，like orig api
`curl -k -H "Authorization: Bearer $TOKEN" https://10.96.0.1:443/readyz`
`curl -k https://10.96.0.1:443/readyz`

return ok

follow by：https://kubeedge.io/zh/docs/advanced/inclusterconfig
then：just `curl -k https://169.254.30.10:10550/readyz`

other api ok，just public api fail：
`curl -v -k -H "Authorization: Bearer $TOKEN" https://169.254.30.10:10550/api/v1/services` is ok
`curl -k -H "Authorization: Bearer $TOKEN" https://169.254.30.10:10550/readyz` is fail

I have read these:
https://github.com/kubeedge/edgemesh/issues/557
https://github.com/kubeedge/kubeedge/pull/5143
https://github.com/kubeedge/kubeedge/issues/5606

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

sh1vam31:
**What type of PR is this?**
/kind bug

This PR fixes an issue where the `metaserver` incorrectly returns a `401 Unauthorised` error when a client requests a public discovery path (like /`readyz`) using a Bearer token that is not synced to the local edge node.

Previously, `authnunion. NewFailOnError` was used directly with the `BearerTokenAuthenticator`. If a token was provided but failed local validation (e.g., service account not found locally), the authenticator would return an error, causing `NewFailOnError` to immediately fail the entire request without falling back to the `AnonymousAuthenticator.`

This PR wraps the `BearerTokenAuthenticator `in a `discoverySafeAuthenticator.` If a token validation error occurs, it checks if the request is targeting a pass-through discovery path (`passthrough.IsPassThroughPath`). If it is, the error is safely ignored, allowing the request to gracefully fall back to the `AnonymousAuthenticator` and successfully pass through to the cloud API server.

Fixes #5961

This approach ensures that discovery paths can be accessed publicly just like native Kubernetes, while preserving strict authentication failures for protected resource paths.

```release-note
Fix: Allow unauthenticated access to metaserver discovery paths like /readyz by ignoring local token validation errors
```

### DIFF
--- a/edge/pkg/metamanager/metaserver/server.go
+++ b/edge/pkg/metamanager/metaserver/server.go
@@ -84,6 +84,21 @@ func (a *AlwaysAllowDiscoveryAuthorizer) Authorize(ctx context.Context, attrs au
 	return authorizer.DecisionNoOpinion, "", nil
 }
 
+// NewDiscoverySafeAuthenticator creates a custom authenticator that ignores errors for discovery paths
+// so they can fall back to the anonymous authenticator instead of failing immediately.
+func NewDiscoverySafeAuthenticator(delegate authenticator.Request) authenticator.Request {
+	return authenticator.RequestFunc(func(req *http.Request) (*authenticator.Response, bool, error) {
+		if req == nil || req.URL == nil {
+			return nil, false, nil
+		}
+		resp, ok, err := delegate.AuthenticateRequest(req)
+		if err != nil && passthrough.IsPassThroughPath(req.URL.Path, strings.ToLower(req.Method)) {
+			return nil, false, nil
+		}
+		return resp, ok, err
+	})
+}
+
 func buildAuth() *metaServerAuth {
 	newAuthorizer := rbac.New(
 		&client.RoleGetter{},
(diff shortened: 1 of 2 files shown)


sh1vam31:
Hi @TianTianBigWang  and @luomengY, I've submitted this fix. Please review this and let me know if any changes are needed.

DoisLONG:
The approach looks reasonable: authentication errors are ignored only for GET requests that match the existing pass-through discovery paths, while protected resource paths and non-GET requests preserve the original authentication error.

Could you please add regression tests for this behavior? At minimum, the tests should cover:

* a token validation error on `GET /readyz` falling back to anonymous authentication;
* the same error on a protected resource path still being returned;
* a non-GET request to a discovery path not bypassing the authentication error.

It may be easier to test this by extracting the wrapper into a small helper that accepts a delegate authenticator.

sh1vam31:
Hi @DoisLONG,
I have extracted the logic into NewDiscoverySafeAuthenticator and added the requested regression tests covering all 3 scenarios. They are passing successfully! Let me know if you need any other changes.

DoisLONG:
Thanks for the update. I reviewed the latest implementation and the requested three scenarios are now covered.

The production logic looks correct: authentication errors are ignored only for GET requests matching the existing pass-through discovery paths, while protected resource paths and non-GET requests preserve the delegate error.

There are two small items to address before this is ready:

1. Please add the standard KubeEdge copyright and Apache license header to the newly added `server_test.go` file, using 2026 as the copyright year.

2. Please assert the complete authenticator result rather than only checking whether `err` is nil. For the `GET /readyz` error case, the important expected result is:

```go
resp == nil
ok == false
err == nil
```

The `ok == false` assertion is important because that is what allows the union authenticator to continue to the anonymous authenticator.

For the protected-path and non-GET cases, please use `errors.Is(err, authErr)` to verify that the original delegate error is preserved.

It would also be useful to add one delegate-success case confirming that `resp`, `ok`, and `err` are returned unchanged.

Non-blocking: since the helper is only used inside this package, please consider naming it `newDiscoverySafeAuthenticator` rather than exporting it.

sh1vam31:
Hi @DoisLONG,

Thanks for the review! I've updated the PR with all the requested changes:

- Added the standard KubeEdge copyright (2026) to the test file.
- Renamed the function to the unexported `newDiscoverySafeAuthenticator.`
- Upgraded the tests to verify the complete authenticator result (`resp`, `ok`, and `err`) using `errors.Is`.
- Added a 4th test scenario to confirm a successful delegate response passes through unchanged.

Everything is passing locally, Let me know if anything else is needed.

PR #6934, opened 2026-06-30
pkg/security/certs: return error from ReadPEMFile when no PEM block is found

### LINKED ISSUE #6933: cloudhub: nil pointer dereference in InitConfigure when a configured TLS file is not valid PEM
**What happened**:

`cloudcore` panics with a nil pointer dereference at startup when a configured CloudHub TLS file (`TLSCAFile`/`TLSCAKeyFile`/`TLSCertFile`/`TLSPrivateKeyFile`) exists but is not valid PEM (empty, DER or corrupt).

`ReadPEMFile` drops the `pem.Decode` result and returns `(nil, nil)` when no block decodes:

Hit the existing `klog.Warningf("failed to load ...")` path (or exit with a clear error) instead of panicking. `ReadPEMFile` should return an error when `pem.Decode` yields no block.

`: > /etc/kubeedge/ca/empty.crt`
Point `hub.TLSCAFile` in `cloudcore.yaml` at that file and start `cloudcore`
CloudCore panics in `config.InitConfigure` (`nil pointer dereference`)

--
**Environment**:
- Kubernetes version (use `kubectl version`): N/A — config-time crash, independent of the cluster
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): master (`91f0943ef`)

  - Others: bug identified by code inspection of `pkg/security/certs/io.go`; reproducible from CloudHub config alone, so it is hardware/OS/kernel independent

  </details>

  - edgecore version (e.g. `edgecore --version`): N/A (CloudHub-side bug)
  - Others:

  </details>

Priyanshu6968:
**What type of PR is this?**
/kind bug

`ReadPEMFile` returned `(nil, nil)` when `pem.Decode` found no PEM block, so cloudhub's `InitConfigure` panicked on `block.Bytes` for an empty/non PEM TLS file instead of taking its existing "failed to load" warning path. It now returns an error in that case.

Fixes #6933

Added `TestReadPEMFileNoBlock` , all four cloudhub callers already handle a non nil error, so no caller changes were needed.

```release-note
NONE
```

### DIFF
--- a/pkg/security/certs/io.go
+++ b/pkg/security/certs/io.go
@@ -28,6 +28,9 @@ func ReadPEMFile(file string) (*pem.Block, error) {
 		return nil, err
 	}
 	p, _ := pem.Decode(bff)
+	if p == nil {
+		return nil, fmt.Errorf("failed to decode PEM block from file %s", file)
+	}
 	return p, nil
 }
 
--- a/pkg/security/certs/io_test.go
+++ b/pkg/security/certs/io_test.go
@@ -22,3 +22,20 @@ func TestReadWrite(t *testing.T) {
 		t.Fatalf("failed to clean testdata, err: %v", err)
 	}
 }
+
+func TestReadPEMFileNoBlock(t *testing.T) {
+	file := "./testdata/ca/invalid.crt"
+	if err := os.MkdirAll("./testdata/ca", 0700); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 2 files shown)


DoisLONG:
The fix is correct, and the regression test using `t.TempDir()` is sufficient. The existing callers already handle the returned error.

/lgtm

PR #6929, opened 2026-06-28
edge/metaserver: add unit tests for auth package JWT issuer validation

Priyanshubhartistm:
/kind test

**What this PR does / why we need it:**

Adds the first unit tests for `edge/pkg/metamanager/metaserver/auth/auth.go`, which had zero test coverage:

- **`hasCorrectIssuer` helper** (white-box, via same-package test):
  - 6 table-driven sub-cases: matching issuer, non-matching issuer, empty issuer, wrong part count, invalid base64 payload, non-JSON payload
  - `TestJWTTokenAuthenticatorMultipleIssuers` — any of the configured issuers is accepted (calls `hasCorrectIssuer` directly)
  - `TestJWTTokenAuthenticatorEmptyIssuers` — no token is accepted when the allowed set is empty (calls `hasCorrectIssuer` directly)

- **`AuthenticateToken` short-circuit** (integration-style, no DB required):
  - `TestJWTTokenAuthenticatorWrongIssuer` — returns `(nil, false, nil)` for a structurally valid 3-part token whose issuer is not in the allowlist
  - `TestJWTTokenAuthenticatorMalformedTokenStructuralFastPath` — returns `(nil, false, nil)` for a token with the wrong number of dot-separated  segments (`len(parts) != 3` guard fires before any issuer comparison)

The `servicebus_test.go` file has been removed from this PR; ServiceBus coverage is handled by PR #6937 which includes the dependency-injection seam and a broader test suite.

**Which issue(s) this PR fixes:**
Fixes #6927

**Test command and output:**
go test -race -v ./edge/pkg/metamanager/metaserver/auth/...

image

**Does this PR introduce a user-facing change?**
NONE

PR #6937, opened 2026-06-30
servicebus: add unit tests to improve coverage

### LINKED ISSUE #6936: test: add unit tests for edge/pkg/servicebus
Add a `servicebus_test.go` for `edge/pkg/servicebus/servicebus.go` to improve test coverage. The following functions have been tested:
- `buildErrorResponse()` — table-driven tests covering various status codes and messages.
- `marshalResult()` — verifies JSON serialization of `serverResponse`.
- `newServicebus()`, `Name()`, `Group()`, `Enable()` — constructor and getter validation.
- `RestartPolicy()` — verified with the `ModuleRestart` feature gate enabled/disabled.
- `buildBasicHandler()` — tested via `httptest.NewRecorder` for request validation and error paths (invalid JSON, empty body, oversized body).

**Why is this needed**:
The `servicebus` package handles HTTP proxying between cloud and edge services, but previously had 0% test coverage and lacked a `_test.go` file. These tests ensure the core configuration, module registration, and fundamental request-handling/error-formatting helper functions behave as expected, catching potential regressions.

ayushsarode:
Added unit tests for `edge/pkg/servicebus` covering initialization, restart policies, HTTP response handling, and the full `buildBasicHandler` request lifecycle via a lightweight dependency-injection seam.

**What type of PR is this?**

/kind test

This PR adds `servicebus_test.go` to the `edge/pkg/servicebus` package, which previously had 0% test coverage.

To enable deterministic testing of `buildBasicHandler` without live beehive or database dependencies, this PR introduces a small production-code change: a `basicHandlerDeps` struct and a `buildBasicHandlerWithDeps` helper. The existing `buildBasicHandler` remains the production entry point and delegates to `buildBasicHandlerWithDeps` with real implementations, so runtime behavior is unchanged.

**Covered test paths:**

- `newServicebus()` — constructor field assignment with various configurations.
- `RestartPolicy()` — behavior with `ModuleRestart` feature gate enabled vs disabled, with proper cleanup.
- `buildErrorResponse()` — various status codes, error messages, and HTTP header population.
- `marshalResult()` — JSON serialization round-trip of `serverResponse`, including special characters.
- `buildBasicHandler` — invalid JSON body, oversized body.
- `buildBasicHandler` — unregistered target URL (verifies `sendSync` is never called).
- `buildBasicHandler` — `SendSync` failure (verifies error propagation).
- `buildBasicHandler` — invalid response content from cloud (verifies `StatusInternalServerError`).
- `buildBasicHandler` — successful path, including full verification of the forwarded message contract (`source`, `group`, `resource`, `operation`, `timeout`, and body content).

Fixes #6936

* Tested locally via `go test -race ./edge/pkg/servicebus`.
* Uses `testify/assert` and `testify/require` to match sibling test patterns (e.g., `devicetwin_test.go`).

```release-note
NONE

### DIFF
--- a/edge/pkg/servicebus/servicebus.go
+++ b/edge/pkg/servicebus/servicebus.go
@@ -21,6 +21,7 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/common/message"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/modules"
 	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/dao/dbclient"
+	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/dao/models"
 	servicebusConfig "github.com/kubeedge/kubeedge/edge/pkg/servicebus/config"
 	"github.com/kubeedge/kubeedge/edge/pkg/servicebus/util"
 	"github.com/kubeedge/kubeedge/pkg/features"
@@ -56,6 +57,11 @@ type serverResponse struct {
 	Code int    `json:"code"`
 	Msg  string `json:"msg"`
 	Body string `json:"body"`
+}
+
+type basicHandlerDeps struct {
+	getURLByKey func(string) (*models.TargetUrls, error)
+	sendSync    func(string, beehiveModel.Message, time.Duration) (beehiveModel.Message, error)
 }
 
 var htc = new(http.Client)
@@ -250,8 +256,14 @@ func server(stopChan <-chan struct{}) {
 	klog.Infof("[servicebus]start to listen and server at %v", s.Addr)
 	utilruntime.HandleError(s.ListenAndServe())
 }
-
 func buildBasicHandler(timeout time.Duration) http.Handler {
+	return buildBasicHandlerWithDeps(timeout, basicHandlerDeps{
+		getURLByKey: dbclient.NewServiceBusService().GetUrlsByKey,
+		sendSync:    beehiveContext.SendSync,
+	})
+}
+
+func buildBasicHandlerWithDeps(timeout time.Duration, deps basicHandlerDeps) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		sReq := &serverRequest{}
 		sResp := &serverResponse{}
@@ -275,7 +287,7 @@ func buildBasicHandler(timeout time.Duration) http.Handler {
 			}
 			return
 		}
-		if targetURL, _ := dbclient.NewServiceBusService().GetUrlsByKey(sReq.TargetURL); targetURL == nil {
+		if targetURL, _ := deps.getURLByKey(sReq.TargetURL); targetURL == nil {
 			sResp.Code = http.StatusBadRequest
 			sResp.Msg = fmt.Sprintf("url %s is not allowed and please make a rule for this url in the cloud", sReq.TargetURL)
 			if _, err := w.Write(marshalResult(sResp)); err != nil {
@@ -286,7 +298,7 @@ func buildBasicHandler(timeout time.Duration) http.Handler {
 		}
 		msg := beehiveModel.NewMessage("").BuildRouter(modules.ServiceBusModuleName, modules.UserGroup,
 			sReq.TargetURL, beehiveModel.UploadOperation).FillBody(byteData)
-		responseMessage, err := beehiveContext.SendSync(modules.EdgeHubModuleName, *msg, timeout)
+		responseMessage, err := deps.sendSync(modules.EdgeHubModuleName, *msg, timeout)
 		if err != nil {
 			sResp.Code = http.StatusBadRequest
 			sResp.Msg = err.Error()
--- a/edge/pkg/servicebus/servicebus_test.go
+++ b/edge/pkg/servicebus/servicebus_test.go
@@ -283,7 +283,6 @@ func TestBuildBasicHandler_OversizedBody(t *testing.T) {
 	assert.Equal(t, "can't read data from body of the http's request", resp.Msg)
 }
 
-
 func TestBuildBasicHandler_UnregisteredURL(t *testing.T) {
 	handler := buildBasicHandlerWithDeps(5*time.Second, basicHandlerDeps{
 		getURLByKey: func(string) (*models.TargetUrls, error) {
--- a/edge/pkg/servicebus/servicebus_test.go
+++ b/edge/pkg/servicebus/servicebus_test.go
@@ -1,0 +1,390 @@
+/*
+Copyright 2025 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package servicebus
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by DoisLONG** [comment]
Thanks for adding the initial ServiceBus test coverage. The previous review comments about subtest assertions and restoring the feature gate state have been addressed.

Before merging, I think the suite should be made more focused and provide stronger behavioral coverage.

At the moment, a large part of the 387-line test file covers fixed-value getters, direct field assignment, and whether `buildBasicHandler` returns a non-nil handler. For example, `Name` and `Group` each have a one-entry table-driven test, while `TestBuildBasicHandler_ReturnsHandler` does not exercise any behavior. Please simplify or remove low-value cases so the suite remains maintainable.

More importantly, the current handler tests only cover JSON decoding and request-size failures. The main handler behavior is still untested:

* a target URL that is not registered;
* a `SendSync` failure;
* invalid response content;
* a successful response path.

Please either add test seams and cover the important paths, or narrow the PR description to accurately state that this is only baseline validation/error-path coverage.

A couple of additional points:

* Use `require.NotNil` before dereferencing `sb` or `got`; `assert.NotNil` does not stop execution and can result in a nil-pointer panic.
* Since this is a newly added file, please check whether the copyright year should be 2026.
* Please provide the exact focused test command and text result, rather than only stating that `make test` was run.

I do not see a production-code risk, but the test suite should be tightened so that its maintenance cost is justified by meaningful regression coverage.

**Comment by ayushsarode** [comment]
> Thanks for adding the initial ServiceBus test coverage. The previous review comments about subtest assertions and restoring the feature gate state have been addressed.
> 
> Before merging, I think the suite should be made more focused and provide stronger behavioral coverage.
> 
> At the moment, a large part of the 387-line test file covers fixed-value getters, direct field assignment, and whether `buildBasicHandler` returns a non-nil handler. For example, `Name` and `Group` each have a one-entry table-driven test, while `TestBuildBasicHandler_ReturnsHandler` does not exercise any behavior. Please simplify or remove low-value cases so the suite remains maintainable.
> 
> More importantly, the current handler tests only cover JSON decoding and request-size failures. The main handler behavior is still untested:
> 
> * a target URL that is not registered;
> * a `SendSync` failure;
> * invalid response content;
> * a successful response path.
> 
> Please either add test seams and cover the important paths, or narrow the PR description to accurately state that this is only baseline validation/error-path coverage.
> 
> A couple of additional points:
> 
> * Use `require.NotNil` before dereferencing `sb` or `got`; `assert.NotNil` does not stop execution and can result in a nil-pointer panic.
> * Since this is a newly added file, please check whether the copyright year should be 2026.
> * Please provide the exact focused test command and text result, rather than only stating that `make test` was run.
> 
> I do not see a production-code risk, but the test suite should be tightened so that its maintenance cost is justified by meaningful regression coverage.

Thanks for the review! To cover the four handler paths (unregistered URL, SendSync failure, invalid response content, success),.

I'll need to introduce two package-level function variables as test seams, one for `dbclient.NewServiceBusService().GetUrlsByKey(...)` and one for `beehiveContext.SendSync(...)`. 
Currently these are hard-coded inside `buildBasicHandler`, so testing those paths would require a live SQLite DB and the full beehive message bus initialized. 

The function variables default to the original calls in production and get swapped in tests via `t.Cleanup`, similar to the existing `htc`/`uc` vars in the same file. Does that approach work?

**Comment by DoisLONG** [comment]
Thanks for checking the design before implementing it.

I would prefer not to introduce additional mutable package-level function variables. Even with `t.Cleanup`, replacing global functions can cause test interference or data races if tests are later run in parallel.

A cleaner approach would be to keep `buildBasicHandler` as the production entry point and add an unexported helper that accepts its dependencies, for example a small dependency struct or two function parameters:

```go
type basicHandlerDeps struct {
    getURLByKey func(string) (*models.TargetUrls, error)
    sendSync    func(string, model.Message, time.Duration) (model.Message, error)
}

func buildBasicHandler(timeout time.Duration) http.Handler {
    return buildBasicHandlerWithDeps(timeout, basicHandlerDeps{
        getURLByKey: dbclient.NewServiceBusService().GetUrlsByKey,
        sendSync:    beehiveContext.SendSync,
    })
}
```

The tests can then call `buildBasicHandlerWithDeps` with deterministic fake functions, without modifying global state or requiring a live database/message bus.

The existing `htc` and `uc` globals are not necessarily a pattern we should extend. Please also retain the other requested cleanup items: simplify the low-value tests, use `require.NotNil` before dereferencing values, update the new-file copyright year, and provide the focused test command and output.

ayushsarode:
@DoisLONG made the necessary changes; please review it.

DoisLONG:
Thanks for the update. I reviewed the latest implementation and the dependency-injection approach is appropriate.

Keeping `buildBasicHandler` as the production entry point while delegating to `buildBasicHandlerWithDeps` avoids mutable package-level test hooks. The requested handler paths are now covered:

* an unregistered target URL;
* a `SendSync` failure;
* invalid response content;
* the successful response path.

The earlier items regarding `require.NotNil`, feature-gate cleanup, and the 2026 license header have also been addressed.

There are a few remaining items before this is ready.

### 1. Verify the forwarded message contract

The successful-path test currently checks only the destination module passed to `sendSync`. Please also verify the message constructed by the handler:

* source is `modules.ServiceBusModuleName`;
* group is `modules.UserGroup`;
* resource is the requested target URL;
* operation is `model.UploadOperation`;
* timeout is the configured timeout;
* content matches the original request body.

The message routing and payload are central parts of the handler behavior, and the new dependency seam makes these assertions straightforward.

### 2. Update the PR description

The description still states that the tests require no refactoring, but the PR now adds `basicHandlerDeps` and `buildBasicHandlerWithDeps` to production code.

It also does not list the newly covered URL lookup, `SendSync`, invalid-response, and success paths. Please update the description to match the final implementation. The description currently mentions an empty-body test, but there is no separate empty-body test in the latest file.

### 3. Coordinate the overlap with #6929

Both this PR and #6929 currently add `edge/pkg/servicebus/servicebus_test.go`, so they cannot merge independently as-is.

Since this PR now provides the more complete ServiceBus coverage and the required test seam, I suggest keeping the ServiceBus changes here. #6929 can retain its unique MetaServer authentication tests and remove the overlapping ServiceBus test file.

Please coordinate the scope with the author of #6929 before merging.

Finally, please provide the focused command and text result for the latest commit:

```bash
go test -race ./edge/pkg/servicebus
```

ayushsarode:
@DoisLONG Thanks for the feedback! All items have been addressed in the latest commit:

* **1. Verified the forwarded message contract:** Added assertions inside the `sendSync` mock to strictly validate that the `source`, `group`, `resource`, `operation`, `timeout`, and `content` are set correctly before the message is forwarded.
* **2. Updated the PR description:** The PR description has been completely rewritten to reflect the final implementation, including the new dependency-injection seam and all the newly covered HTTP execution paths.
* **3. Coordinated with #6929:** I have left a comment on PR #6929 asking the author to drop their overlapping `servicebus_test.go` file so they can focus on their unique MetaServer tests, which will prevent any merge conflicts.

Finally, here is the focused command and text result for the latest commit:
`go test -race ./edge/pkg/servicebus`
```text
ok      github.com/kubeedge/kubeedge/edge/pkg/servicebus        1.454s

DoisLONG:
Thanks for the update. I reviewed the latest implementation and tests again, including the production dependency wiring and the full handler execution paths.

The dependency-injection helper preserves the production behavior, the complete forwarding contract is now verified, the important error paths are covered, and the overlap with #6929 has been resolved.

I do not see any remaining functional or test-correctness blockers.

/lgtm

ayushsarode:
@DoisLONG Thank you for the review and LGTM! I appreciate you taking the time to go through the implementation and tests again.

ayushsarode:
@chendave can you review and approve this PR? if it looks good to you?

WillardHu:
Please fix the lint issue:
Error: edge/pkg/servicebus/servicebus_test.go:286:1: File is not properly formatted (gofmt)

ayushsarode:
Hello @WillardHu, I appreciate for the review! I've made the necessary adjustments. Please review once, and let me know if there are any problems.

ayushsarode:
@WillardHu hey! could you look into this PR, let me know if something is off
thank you!

ayushsarode:
@WillardHu could you please approve this PR if it looks good to you?

PR #6938, opened 2026-06-30
fix(taskmanager/upstream): fix err shadowing in status callback

### LINKED ISSUE #6916: taskmanager upstream handlers: UpdateNodeTaskStatus always returns nil due to shadowed err in status callback
What happened:
In the three task upstream handlers, UpdateNodeTaskStatus declares err in the function's outer var block, but the status-update Callback is defined as func(err error), which shadows the outer variable. The assignment inside the callback writes to the parameter, not the outer err, so the function's return err is always nil. Status-update failures are silently dropped.
Affected files on current master:

cloud/pkg/taskmanager/upstream/node_upgrade.go — outer err L69, callback L100, return err L109
cloud/pkg/taskmanager/upstream/image_prepull.go — same structure
cloud/pkg/taskmanager/upstream/config_update.go — same structure

Additionally, node_upgrade.go:102 and config_update.go:102 carry a copy-pasted error message "failed to update image prepull job status", which is only correct in image_prepull.go.
What you expected to happen:
UpdateNodeTaskStatus should return the error produced when the status update fails, and each handler should report its own job type in the error message.
How to reproduce it (as minimally and precisely as possible):
Source-level defect visible on current master. In any of the three files, the Callback: func(err error) parameter shadows the outer err declared in the var block. The inner assignment err = fmt.Errorf(...) targets only the callback parameter, so the outer err returned after wg.Wait() is unconditionally nil.
Anything else we need to know?:
Proposed fix: rename the callback parameter to callbackErr, assign the wrapped error back to the outer err, and correct the two miscopied messages in node_upgrade.go and config_update.go. Happy to open a PR once this is confirmed.
Environment:

KubeEdge version: current master

g-k-s-03:
## Description

In all three upstream status handlers (`UpdateNodeTaskStatus`), the 
function declares `var err error` in an outer scope, then defines a 
status-update `Callback` with the signature `func(err error)`. Because 
Go allows parameter names to shadow outer variables, the callback's 
`err` parameter is a completely separate variable from the outer one. 
Any assignment inside the callback body only modifies this local 
parameter — the outer `err` is never touched.

As a result, `return err` at the end of each function unconditionally 
returns `nil`, regardless of whether the underlying status update 
actually succeeded or failed. Callers of `UpdateNodeTaskStatus` have 
no way to detect a failed status update; the error is silently 
discarded at the point where the callback fires, well before the 
function returns.

This affects all three node task types:
- NodeUpgradeJob
- ImagePrePullJob  
- ConfigUpdateJob

Since these handlers are responsible for reporting task status back 
to the cluster, a silently swallowed error here means operators have 
no visibility into a failed status write — the task may appear stuck 
or inconsistent with no corresponding error in logs or status 
conditions.

While fixing this, I also corrected a copy-paste error message bug: 
both `node_upgrade.go` and `config_update.go` used the literal string 
"failed to update image prepull job status" regardless of the actual 
job type being handled. Only `image_prepull.go` had the correct 
message.

## Fix

1. Renamed the callback parameter from `err` to `callbackErr` in all 
   three files, eliminating the shadowing.
2. Inside each callback, the wrapped error is now correctly assigned 
   to the outer `err` variable, so it propagates through `wg.Wait()` 
   and is returned to the caller.
3. Corrected the error message in `node_upgrade.go` to reference 
   "node upgrade job status" and in `config_update.go` to reference 
   "config update job status".
4. Changed the error wrapping verb from `%v` to `%w` in all three 
   files, preserving the underlying error in the chain for 
   `errors.Is` / `errors.As` compatibility.

## Files Changed
- `cloud/pkg/taskmanager/upstream/node_upgrade.go`
- `cloud/pkg/taskmanager/upstream/image_prepull.go`
- `cloud/pkg/taskmanager/upstream/config_update.go`

## Testing

This is a pure bug fix with no behavioral change on the success path. 
On the failure path, callers now correctly receive a non-nil error 
instead of nil.

Fixes #6916

### DIFF
--- a/cloud/pkg/taskmanager/upstream/config_update.go
+++ b/cloud/pkg/taskmanager/upstream/config_update.go
@@ -97,9 +97,9 @@ func (h *ConfigUpdateJobHandler) UpdateNodeTaskStatus(
 			ExtendInfo:   upmsg.Extend,
 			ActionStatus: &actoinStatus,
 		},
-		Callback: func(err error) {
-			if err != nil {
-				err = fmt.Errorf("failed to update image prepull job status, err: %v", err)
+		Callback: func(callbackErr error) {
+			if callbackErr != nil {
+				err = fmt.Errorf("failed to update config update job status, err: %w", callbackErr)
 			}
 			wg.Done()
 		},
--- a/cloud/pkg/taskmanager/upstream/config_update_test.go
+++ b/cloud/pkg/taskmanager/upstream/config_update_test.go
@@ -31,7 +31,7 @@ import (
 	taskmsg "github.com/kubeedge/kubeedge/pkg/nodetask/message"
 )
 
-func TestNodeUpgradeJobUpdateNodeTaskStatus(t *testing.T) {
+func TestConfigUpdateJobUpdateNodeTaskStatus(t *testing.T) {
 	var (
 		jobName  = "test-job"
 		nodeName = "node1"
@@ -40,14 +40,14 @@ func TestNodeUpgradeJobUpdateNodeTaskStatus(t *testing.T) {
 		patches := gomonkey.NewPatches()
 		defer patches.Reset()
 
-		gomonkey.ApplyFunc(status.GetNodeUpgradeJobStatusUpdater, func() *status.StatusUpdater {
+		gomonkey.ApplyFunc(status.GetConfigeUpdateJobStatusUpdater, func() *status.StatusUpdater {
 			return &status.StatusUpdater{}
 		})
 		gomonkey.ApplyMethodFunc(reflect.TypeOf(&status.StatusUpdater{}), "UpdateStatus",
 			func(opts status.UpdateStatusOptions) {
-				act, ok := opts.ActionStatus.(*operationsv1alpha2.NodeUpgradeJobActionStatus)
+				act, ok := opts.ActionStatus.(*operationsv1alpha2.ConfigUpdateJobActionStatus)
 				require.True(t, ok)
-				assert.Equal(t, operationsv1alpha2.NodeUpgradeJobActionUpgrade, act.Action)
+				assert.Equal(t, operationsv1alpha2.ConfigUpdateJobActionUpdate, act.Action)
 				assert.Equal(t, operationsv1alpha2.NodeTaskPhaseSuccessful, opts.Phase)
 				assert.Equal(t, jobName, opts.JobName)
 				assert.Equal(t, nodeName, opts.NodeName)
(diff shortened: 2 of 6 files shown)


g-k-s-03:
@daixiang0 , @DoisLONG can you review this

DoisLONG (commented):
Thanks for working on this.

The code fix looks correct to me. The callback parameter no longer shadows the outer error variable, the job-specific error messages are corrected, and using `%w` is better for preserving the underlying error.

One thing I think should be added before this is ready: could you add regression tests for the failure path?

The important behavior here is that when the status updater callback receives a non-nil error, `UpdateNodeTaskStatus()` should return a non-nil error instead of silently returning nil. It would be good to cover this for the affected handlers, especially since this is the core bug fixed by the PR.

Also, there is now another PR, #6991, trying to fix the same issue. Since this PR was opened earlier and is linked to #6916, I think it would be better to continue the fix here and avoid duplicate PRs.

g-k-s-03:
Hi @DoisLONG I've added regression tests for all three handlers covering the failure path. Tests verify that UpdateNodeTaskStatus() returns a non-nil error when the callback receives one, use assert.ErrorIs to verify the wrapped error is preserved via %w, and use behavior-describing test names. config_update_test.go is a new file since no test file existed for that handler previously. Please take another look when you have time."

Paste the push output here.

DoisLONG (commented):
Thanks for the update. The regression tests now cover the core failure path I mentioned earlier: when the status updater callback receives a non-nil error, `UpdateNodeTaskStatus()` returns a non-nil error and preserves the wrapped error. That is the right direction.

I took another look at the tests and noticed one issue with the gomonkey usage.

The tests create a patch set with:

```go
patches := gomonkey.NewPatches()
defer patches.Reset()
```

but then they call the package-level helpers:

```go
gomonkey.ApplyFunc(...)
gomonkey.ApplyMethodFunc(...)
```

instead of applying the patches through the `patches` instance. In that form, I do not think `patches.Reset()` is guaranteed to reset the patches created by the package-level helpers. This can make the monkey patches leak across subtests or later tests in the same package.

Could you please update the touched tests to use the same patch set consistently, for example:

```go
patches := gomonkey.NewPatches()
defer patches.Reset()

patches.ApplyFunc(...)
patches.ApplyMethodFunc(...)
```

or directly keep the returned patch object from `gomonkey.ApplyFunc(...)` and reset that one.

The production fix itself looks reasonable to me: the callback parameter no longer shadows the outer error, `%w` preserves the underlying error, and the job-specific error messages are corrected.

One small metadata point as well: since this fixes a real bug, please add the normal PR metadata such as `/kind bug` and a release-note block, or confirm the intended release note.

DoisLONG (commented):
Thanks for the update.

Two final cleanups:

2. Please add a release-note block to the PR body. Since this is a real bug fix, a short release note would be better than leaving it out.

Suggested release note:

```release-note
Fixed taskmanager upstream status handlers to correctly return status update callback errors for NodeUpgradeJob, ImagePrePullJob, and ConfigUpdateJob.
```

PR #6949, opened 2026-07-01
pkg/viaduct/filter: add unit tests for MessageFilter

### LINKED ISSUE #6947: pkg/viaduct/filter: add unit tests for MessageFilter

pkg/viaduct/pkg/filter/filter.go defines MessageFilter.ProcessFilter, called from MessageMux.processFilter on every received message. There is no test file.

Suggested coverage:

- AddFilterFunc appends and ProcessFilter invokes filters in registration order.
- A filter that returns an error short-circuits subsequent filters and returns that error.
- An empty filter list returns nil without error.

**Why is this needed**:

Filtering runs on every inbound message in the mux. Even a tiny regression (e.g. ordering or early-exit semantics) would change which messages reach handlers. The function is small enough that a test file is straightforward.

Priyanshubhartistm:
**What type of PR is this?**

/kind test

`pkg/viaduct/pkg/filter/filter.go` defines `MessageFilter.ProcessFilter`, which is called by `MessageMux.processFilter` on every received message. The function had no companion test file, leaving two critical behaviours unverified:

1. **Registration order** — filters must be invoked in the order they  were added via `AddFilterFunc`.
2. **Short-circuit semantics** — a filter returning an error must stop  processing and prevent subsequent filters from running.

This PR adds `filter_test.go` with 7 unit tests covering both `AddFilterFunc` and `ProcessFilter`:

- `AddFilterFunc` appends filters and they are invoked in order
- `ProcessFilter` with an empty list returns `nil` without error
- `ProcessFilter` calls all filters when every filter passes
- `ProcessFilter` short-circuits on the first error and skips remaining
  filters
- `ProcessFilter` propagates the exact error returned by the failing
  filter
- `ProcessFilter` passes the same `*model.Message` pointer to all
  filters unchanged

All 7 tests pass under Go's `-race` detector.

Fixes #6947

### Test Evidence

All tests passed successfully.

**Command:**

```bash
go test ./pkg/viaduct/pkg/filter/... -v -race 
```

**Result:**

```text
PASS
ok  	github.com/kubeedge/kubeedge/pkg/viaduct/pkg/filter	1.024s


```

**Screenshot:**

<Attach the screenshot below>
image

No production code was changed. The test file is pure white-box testing using the package's own types — no mocks or external dependencies were added beyond the already-required `github.com/stretchr/testify/assert`.

```release-note
NONE

### DIFF
--- a/pkg/viaduct/pkg/filter/filter_test.go
+++ b/pkg/viaduct/pkg/filter/filter_test.go
@@ -1,0 +1,178 @@
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package filter
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/kubeedge/beehive/pkg/core/model"
+)
+
+// newMsg returns a minimal *model.Message for use in tests.
+func newMsg(id string) *model.Message {
+	return &model.Message{
+		Header: model.MessageHeader{ID: id},
+	}
+}
+
+// ---------------------------------------------------------------------------
+// AddFilterFunc
+// ---------------------------------------------------------------------------
+
+// TestAddFilterFunc_AppendsSingleFilter verifies that AddFilterFunc registers
+// a filter function so that ProcessFilter invokes it.
+func TestAddFilterFunc_AppendsSingleFilter(t *testing.T) {
+	called := false
+	mf := &MessageFilter{}
+
+	mf.AddFilterFunc(func(msg *model.Message) error {
+		called = true
+		return nil
+	})
+
+	err := mf.ProcessFilter(newMsg("m1"))
+
+	assert.NoError(t, err)
+	assert.True(t, called, "expected the registered filter to be called")
+}
+
+// TestAddFilterFunc_AppendsMultipleFilters verifies that multiple calls to
+// AddFilterFunc register all filters and that ProcessFilter invokes them in
+// registration order.
+func TestAddFilterFunc_AppendsMultipleFilters(t *testing.T) {
+	var order []int
+	mf := &MessageFilter{}
+
+	mf.AddFilterFunc(func(msg *model.Message) error {
+		order = append(order, 1)
+		return nil
+	})
+	mf.AddFilterFunc(func(msg *model.Message) error {
+		order = append(order, 2)
+		return nil
+	})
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Priyanshubhartistm** [comment]
Hey @DoisLONG , could you please take a look at this PR? whenever you have sometime, I would really appreciate your review and any feedback or suggestions you may have.

Thank you

**Comment by DoisLONG** [comment]
Thanks for adding the filter coverage. The tests for registration order, the empty-filter case, and short-circuit behavior match the core requirements in #6947.

The current branch needs to be cleaned up before the filter change can be reviewed independently.

### 1. Remove the unrelated `lane_test.go`

This PR description covers only `pkg/viaduct/pkg/filter/filter_test.go`, but the current diff also includes the complete:

```text
pkg/viaduct/pkg/lane/lane_test.go
```

That file belongs to #6948 and still has review items being addressed there. Including it here creates duplicate changes and will cause the two PRs to conflict.

Please rebase or recreate this branch from the latest master and retain only `filter_test.go`.

### 2. Verify exact error identity

The PR states that `ProcessFilter` propagates the exact error returned by the failing filter, but the tests currently use `assert.ErrorIs`.

`ErrorIs` would also pass if `ProcessFilter` wrapped the error. To verify direct propagation of the same error instance, please use:

```go
assert.Same(t, filterErr, err)
```

Apply the same check to the first-error test.

### 3. Clarify the message-pointer test

The test verifies that every filter receives the same `*model.Message` instance, but it does not verify that the message remains unchanged, nor should filters necessarily be prohibited from mutating it.

Please change the test name/comment from “unchanged” to “the same message instance.” Optionally, have the first filter mutate a field and verify that the next filter observes the mutation, which more directly proves the shared-message chain behavior.

After removing the unrelated lane tests, please provide the final text result for:

```bash
go test -race ./pkg/viaduct/pkg/filter/...
```

The filter coverage itself is close; the branch scope and assertions above should be corrected.

Priyanshubhartistm:
Thanks for the  review, @DoisLONG! All three points have been addressed:-

1. Removed lane_test.go — The branch was rebased to drop the lane commit entirely. git diff master --name-only now shows only pkg/viaduct/pkg/filter/filter_test.go.

2. Exact error identity (assert.Same) — Replaced assert.ErrorIs with assert.Same in both TestProcessFilter_ShortCircuitsOnError and TestProcessFilter_ReturnsFirstError. This proves that the exact error pointer is propagated, not just an error that satisfies the Is chain.

3. Message-pointer test clarified — Renamed to TestProcessFilter_PassesSameMessageInstanceToAllFilters and updated the doc comment to say "same instance" instead of "unchanged". Also added a mutation proof: filter 1 mutates Header.ID, and filter 2 asserts it observes that mutation — directly proving the shared-message chain behaviour.

Test output after the rebase:
ok  	github.com/kubeedge/kubeedge/pkg/viaduct/pkg/filter	1.024s

could you please take a look at this PR when you have time?

DoisLONG:
Thanks for the update. I reviewed the cleaned-up diff again.

The previous review items have been addressed:

* the unrelated lane tests have been removed, and the PR now contains only `filter_test.go`;
* the short-circuit tests verify exact error identity with `assert.Same`;
* the message-chain test accurately describes the shared message instance and proves that mutations from an earlier filter are visible to later filters;
* the focused race-enabled test passes.

The current test coverage matches the behavior of `MessageFilter` and I do not see any remaining issues.

/lgtm

PR #6951, opened 2026-07-01
devicetwin: handle all 30 silently swallowed errors in dtmanager

### LINKED ISSUE #6904: devicetwin: 30 errors silently swallowed via '// TODO: handle error' in twin/device/membership managers
**What happened**:

The DeviceTwin module on the edge node has 30 instances where errors are caught, logged with `klog.Error(err)`, and then silently discarded. Each site is marked with a `// TODO: handle error` or `// TODO: handle err` comment that was never resolved.

The affected files and counts:
- `edge/pkg/devicetwin/dtmanager/twin.go` — 19 instances
- `edge/pkg/devicetwin/dtmanager/membership.go` — 6 instances
- `edge/pkg/devicetwin/dtmanager/device.go` — 5 instances

This causes three concrete problems:

1. **State mutation errors are invisible**: Functions like `DealDeviceTwin()` and `UpdateDeviceAttr()` can fail, but their callers (`dealTwinSync`, `Updated`, `dealDeviceAttrUpdate`) swallow the returned error and return `nil` to the worker loop. The in-memory twin state may diverge from SQLite without any indication.

2. **State recovery also silently fails**: When a DB write fails, `SyncDeviceFromSqlite()` is called to restore consistency. But if recovery also fails, that error is swallowed too (e.g., `twin.go` lines 224-227, 244-248), leaving in-memory state inconsistent with both the desired update and the database.

3. **Cloud sync failures are lost**: In `dealDeviceStateUpdate()` (`device.go` line 147-151), a failed `context.Send()` to the cloud is logged but the function returns `nil`. The cloud never learns the device state changed.

The errors can be verified by code inspection:
```bash
grep -rn "TODO: handle err" edge/pkg/devicetwin/dtmanager/
# Returns 30 results
```

- State mutation errors (`DealDeviceTwin`, `UpdateDeviceAttr`, cloud sync sends) should be propagated to callers so they are visible at the worker loop level.
- State recovery errors (`SyncDeviceFromSqlite`) and message-send/JSON-parse errors should be reported via `utilruntime.HandleError()` instead of bare `klog.Error()`.
- The `Updated()` function should return `error` instead of silently discarding the `DealDeviceTwin` result.
- All 30 `// TODO: handle error` comments should be removed after proper handling is added.

1. Deploy an edge node with DeviceTwin enabled and register a device
2. Simulate a SQLite write failure (e.g., fill the disk or make the DB file read-only)
3. Publish a twin update to MQTT topic `$hw/events/device/{device-id}/twin/update`
4. In edgecore logs: the DB error is logged, but `dealTwinSync` returns `nil` — the worker loop reports success
5. The in-memory device twin now differs from what is in SQLite

Alternatively, inspect the code directly — every `// TODO: handle error` site is a location where an error is swallowed.

The 5 highest-risk sites:

| File | Line | Function | Impact |
|------|------|----------|--------|
| `twin.go` | 128-131 | `dealTwinSync` swallows `DealDeviceTwin` error | In-memory twin diverges from DB |
| `twin.go` | 188-191 | `Updated` swallows `DealDeviceTwin` error | Twin update silently lost |
| `twin.go` | 244-248 | `DealDeviceTwin` swallows `SyncDeviceFromSqlite` error | State unrecoverable |
| `device.go` | 147-151 | `dealDeviceStateUpdate` swallows cloud sync send | Cloud never learns device state |
| `device.go` | 170-174 | `dealDeviceAttrUpdate` swallows `UpdateDeviceAttr` error | Attribute update silently lost |

Proposed fix:
- Propagate errors from state-mutating functions to their callers
- Change `Updated()` signature from `func Updated(...)` to `func Updated(...) error`
- Use `utilruntime.HandleError()` for errors that cannot be propagated (e.g., inside `addDevice`/`removeDevice` loops)
- Remove all 30 `// TODO: handle error` / `// TODO: handle err` comments

I am willing to submit a PR for this.

**Environment**:
- Kubernetes version (use `kubectl version`): N/A (code inspection)
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): master branch, commit `864f45eb1`

  - Hardware configuration (e.g. `lscpu`): N/A (code-level bug, not environment-specific)
  - Go version (e.g. `go version`): go1.26.4 darwin/arm64
  - Others: N/A

  </details>

  - edgecore version (e.g. `edgecore --version`): master branch, commit `864f45eb1`
  - Hardware configuration (e.g. `lscpu`): N/A (code-level bug, not environment-specific)
  - Go version (e.g. `go version`): go1.26.4 darwin/arm64
  - Others: N/A

  </details>

rahulkr182:
**What type of PR is this?**

/kind bug
/kind cleanup

The DeviceTwin module on the edge node had 30 instances across `twin.go`, `device.go`, and `membership.go` where errors were caught, logged with `klog.Error(err)`, and then silently discarded. Every site was annotated with a `// TODO: handle error` or `// TODO: handle err` comment that was never resolved.

This caused three concrete problems:
1. **State mutation errors were invisible** — `DealDeviceTwin()` and `UpdateDeviceAttr()` could fail but callers returned `nil` to the worker loop, causing in-memory twin state to silently diverge from SQLite.
2. **State recovery also silently failed** — when a DB write failed, `SyncDeviceFromSqlite()` errors were also discarded, leaving state unrecoverable.
3. **Cloud sync failures were lost** — in `dealDeviceStateUpdate()`, a failed `context.Send()` to the cloud was logged but the function returned `nil`.

This PR resolves all 30 sites using the following strategy:
- **Propagate errors** from state-mutating functions to callers (`dealTwinSync`, `dealTwinGet`, `dealTwinUpdate`, `Updated()`, `dealDeviceStateUpdate`, `dealDeviceAttrUpdate`)
- **Use `utilruntime.HandleError()`** for errors inside recovery paths and per-device iteration loops where propagation is not possible (`SyncDeviceFromSqlite`, JSON unmarshal errors in `dealTwinCompare`, `context.Send` errors in `addDevice`/`removeDevice` loops)
- `Updated()` signature changed from `func Updated(...)` to `func Updated(...) error`
- All 30 `// TODO: handle error` / `// TODO: handle err` comments removed

Fixes #6904

- `Updated()` signature change is intentional to allow error propagation to `dealTwinUpdate`.
- Test expectations updated in `twin_test.go` and `device_test.go` to reflect errors that were previously swallowed are now correctly propagated.
- All existing tests pass: `ok github.com/kubeedge/kubeedge/edge/pkg/devicetwin/dtmanager 7.292s`

```release-note
devicetwin: errors from state-mutating operations (twin updates, device attribute updates, cloud sync sends) are now propagated to the worker loop instead of being silently discarded. Recovery errors use utilruntime.HandleError() for structured reporting. The Updated() function now returns error.

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/device.go
+++ b/edge/pkg/devicetwin/dtmanager/device.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/klog/v2"
 
 	"github.com/kubeedge/beehive/pkg/core/model"
@@ -131,23 +132,19 @@ func dealDeviceStateUpdate(context *dtcontext.DTContext, resource string, msg in
 		return err
 	}
 	topic := dtcommon.DeviceETPrefix + device.ID + dtcommon.DeviceETStateUpdateResultSuffix
-	err = context.Send(device.ID,
+	if err = context.Send(device.ID,
 		dtcommon.SendToEdge,
 		dtcommon.CommModule,
-		context.BuildModelMessage(modules.BusGroup, "", topic, messagepkg.OperationPublish, payload))
-	if err != nil {
-		// TODO: handle error
-		klog.Error(err)
+		context.BuildModelMessage(modules.BusGroup, "", topic, messagepkg.OperationPublish, payload)); err != nil {
+		return err
 	}
 
 	msgResource := "device/" + device.ID + dtcommon.DeviceETStateUpdateSuffix
-	err = context.Send(deviceID,
+	if err = context.Send(deviceID,
 		dtcommon.SendToCloud,
 		dtcommon.CommModule,
-		context.BuildModelMessage("resource", "", msgResource, model.UpdateOperation, string(payload)))
-	if err != nil {
-		// TODO: handle error
-		klog.Error(err)
+		context.BuildModelMessage("resource", "", msgResource, model.UpdateOperation, string(payload))); err != nil {
+		return err
 	}
 	return nil
 }
@@ -167,13 +164,10 @@ func dealDeviceAttrUpdate(context *dtcontext.DTContext, resource string, msg int
 	deviceID := resource
 
 	context.Lock(deviceID)
-	if _, err = UpdateDeviceAttr(context, deviceID, updatedDevice.Attributes,
-		dttype.BaseMessage{EventID: updatedDevice.EventID}, 0); err != nil {
-		// TODO: handle error
-		klog.Error(err)
-	}
+	_, err = UpdateDeviceAttr(context, deviceID, updatedDevice.Attributes,
+		dttype.BaseMessage{EventID: updatedDevice.EventID}, 0)
 	context.Unlock(deviceID)
-	return nil
+	return err
 }
 
 // UpdateDeviceAttr update device attributes
@@ -190,7 +184,7 @@ func UpdateDeviceAttr(context *dtcontext.DTContext, deviceID string, attributes 
 	}
 	dealAttrResult := DealMsgAttr(context, Device.ID, attributes, dealType)
 	if dealAttrResult.Err != nil {
-		return nil, nil
+		return nil, dealAttrResult.Err
 	}
 	add, deviceAttrDelete, update, result := dealAttrResult.Add, dealAttrResult.Delete, dealAttrResult.Update, dealAttrResult.Result
 	if len(add) != 0 || len(deviceAttrDelete) != 0 || len(update) != 0 {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by DoisLONG** [comment]
Thanks for addressing this broad error-handling area. I reviewed the complete diff across `device.go`, `membership.go`, `twin.go`, and the associated tests.

The overall direction is reasonable:

* state-mutation errors are beginning to propagate to the worker callbacks;
* per-device membership loops continue processing while reporting individual failures;
* `Updated` now returns an error;
* the unnecessary JSON round trips used to copy `TypeMetadata` have been removed.

However, there are still several blocking correctness and test-coverage issues.

### 1. The main `dealTwinSync` database failure is still swallowed

`dealTwinSync` now returns the result of `DealDeviceTwin`, but `DealDeviceTwin` only returns a `DeviceTwinTrans` error when `dealType == RestDealType`.

For `SyncDealType`, when `DeviceTwinTrans` fails after all retries, the function:

1. calls `SyncDeviceFromSqlite`;
2. logs the persistence error;
3. continues processing;
4. eventually returns `nil`.

Therefore, the highest-risk path described by the issue is still not fixed: a cloud twin sync can fail to persist while the worker still observes success.

Please return the persistence error for the sync path as well.

### 2. Processing continues with a stale in-memory object after recovery

`DealDeviceTwin` captures the current `device` pointer before calling `DealMsgTwin`, and `DealMsgTwin` mutates that object.

If persistence then fails, `SyncDeviceFromSqlite` restores state by storing a new `*dttype.Device` in `DeviceList`. The local `device` variable in `DealDeviceTwin` still points to the old, mutated object.

For `SyncDealType`, the function then continues to:

* send the twin document;
* calculate and send a delta using the stale `device.Twin`;
* send the sync result;
* return `nil`.

This can publish state derived from a mutation that was not successfully persisted.

After a persistence failure, please attempt recovery, report any recovery error, stop all downstream document/delta/sync-result processing, and return the original persistence error. If a sync response is required after recovery, it should be built explicitly from the restored state rather than from the stale local pointer.

### 3. The new failure semantics are not covered by regression tests

The current test changes mostly adjust the test setup by registering a communication channel and update a few existing expected errors. They do not inject and verify the important failures introduced by this PR.

Please add focused tests for at least the following:

* `SyncDealType` with `DeviceTwinTrans` returning a sentinel error:

  * the same error is returned;
  * document, delta, and cloud sync messages are not emitted after the failed persistence;
  * the in-memory device is restored from the mocked database state;

* recovery failure after a persistence failure:

  * the persistence error remains the returned error;
  * the recovery error is also reported;

* `DeviceAttrTrans` failure:

  * the error is returned by `UpdateDeviceAttr`;
  * the same failure is observable through `dealDeviceAttrUpdate`;

* `dealDeviceStateUpdate`:

  * a database failure is returned;
  * a successful update emits both the edge result and cloud update messages with the expected actions/resources;
  * a missing communication channel returns an error.

Please use sentinel errors and `errors.Is` where possible instead of comparing newly created error values with `reflect.DeepEqual`.

### 4. Do not continue sending after payload construction fails

In `UpdateDeviceAttr`, a `BuildDeviceAttrUpdate` error is logged, but the code still calls `context.Send` with the returned payload.

Please return the build error, or at minimum stop the notification and report it through the chosen best-effort error path. Continuing after payload construction fails is not safe.

### 5. Add operation context to `HandleError` calls

Many new calls pass the raw error directly to `utilruntime.HandleError`. Since this PR is specifically intended to improve error observability, please wrap errors with the operation and device identity, for example:

```go
utilruntime.HandleError(fmt.Errorf(
    "restore device %q from sqlite after twin persistence failure: %w",
    deviceID,
    syncErr,
))
```

The same applies to membership attribute/twin updates and edge notification failures. Raw database or communication errors alone do not identify which operation or device failed.

The use of `HandleError` in membership add/remove loops is otherwise reasonable because a failure for one device should not prevent processing the remaining devices.

Please also provide the focused test output for:

```bash
go test -race ./edge/pkg/devicetwin/dtmanager
```

rahulkr182:
Hi @DoisLONG, PTAL at this PR.

DoisLONG (commented):
Thanks for the update.

The direction looks good to me. Propagating errors from the state mutation paths is the right thing to do, and the added tests cover several important cases.

A few cleanup points:

1. The release-note block looks unfinished. It starts with ```release-note but I do not see the closing ```.

2. I could not find completed workflow runs for the current head SHA, so CI still needs to be triggered/run.

3. Some review threads are still unresolved. A few of them look already addressed by the latest code, so please resolve them if they are fixed.

4. One semantic point: some send failures are now handled with `utilruntime.HandleError()` instead of being returned. That may be fine if those sends are best-effort after local state has already been persisted, but please clarify this distinction a bit: which errors should fail the worker path, and which ones are only reported as best-effort errors.

PR #6959, opened 2026-07-02
pkg/viaduct: add unit tests for mux

boshilin123:
**What type of PR is this?**

/kind test

This PR adds unit tests for `pkg/viaduct/pkg/mux`.

The mux package is responsible for matching viaduct messages against resource patterns, extracting path parameters, and dispatching matched requests to registered handlers. These behaviors are core to request routing in viaduct, but the package previously had no dedicated unit tests.

The new tests cover:
- building message expressions from resource patterns such as `/nodes/{node}/pods/{pod}`
- validating extracted variable names and variable counts
- matching message resources and operations
- matching wildcard operations
- rejecting messages with non-matching operations
- extracting path parameters from matched resources
- dispatching matched requests to handlers
- passing extracted parameters into the message container
- returning an error when no mux entry matches

Fixes #

This PR only adds unit tests and does not change production logic.

Tested with:

```bash
go test ./pkg/viaduct/pkg/mux/...

### DIFF
--- a/pkg/viaduct/pkg/mux/expr.go
+++ b/pkg/viaduct/pkg/mux/expr.go
@@ -10,9 +10,10 @@ import (
 )
 
 type MessageExpression struct {
-	VarNames []string
-	VarCount int
-	Matcher  *regexp.Regexp
+	VarNames   []string
+	VarIndexes []int
+	VarCount   int
+	Matcher    *regexp.Regexp
 }
 
 func NewExpression() *MessageExpression {
@@ -22,7 +23,9 @@ func NewExpression() *MessageExpression {
 func (exp *MessageExpression) GetExpression(resource string) *MessageExpression {
 	var buffer bytes.Buffer
 	var varNames []string
+	var varIndexes []int
 	var varCount int
+	captureIndex := 1
 
 	if resource == "*" {
 		compiled, _ := regexp.Compile("(/.*)?$")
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by DoisLONG** [comment]
Thanks for adding the initial mux test coverage. The current tests cover the basic expression, operation matching, parameter extraction, dispatch, and no-match paths.

After comparing the tests with the production implementation, I found two routing behaviors that need regression coverage because they currently expose real defects.

### 1. Cover and fix `MessagePattern.Res`

`NewPattern` compiles `resExpr` from the initial resource, but `Res` only changes `pattern.resource`:

```go
func (pattern *MessagePattern) Res(resource string) *MessagePattern {
    pattern.resource = resource
    return pattern
}
```

The matcher therefore continues using the old resource expression.

For example:

```go
pattern := NewPattern("/old").Res("/new").Op("get")
```

still matches `/old` and does not match `/new`.

Please add a test for the exported `Res` fluent API and update it to rebuild `resExpr`, or remove/deprecate the setter if changing the resource after construction is not supported.

### 2. Test parameter extraction with custom regular expressions

`GetExpression` wraps a custom parameter expression in a capturing group, while `extractParameters` maps all submatches to variable names by their positional index.

A pattern such as:

```text
/nodes/{node:(foo|bar)}/pods/{pod}
```

produces an additional capture for `(foo|bar)`. For `/nodes/foo/pods/pod-1`, the current extraction can assign `pod=foo` instead of `pod=pod-1`.

Please add a dispatch or extraction test that includes a custom expression with an internal capturing group and verifies every extracted parameter. The implementation should not assume that each custom expression contributes exactly one positional capture; named groups or explicitly recorded group indexes would be safer.

### 3. Verify complete request propagation during dispatch

`wrapMessage` propagates the request Header and PeerCertificates into `MessageContainer`, and production CloudHub handlers depend on the header values for `node_id` and `project_id`.

Please extend `TestMessageMuxDispatch` to set and verify:

* `Header`;
* `PeerCertificates`;
* the original Message pointer;
* the extracted parameters.

### 4. Cover the remaining expression branches

Please also include focused cases for:

* resource wildcard `*`;
* wildcard parameter `{param:*}`;
* a valid custom regular expression;
* an invalid regular expression;
* nil messages;
* trailing sub-resource matching.

It would also be useful to exercise `ServeConn` with a filter that rejects a message, rather than testing only the private `dispatch` method.

Please run the updated package tests with the race detector:

```bash
go test -race ./pkg/viaduct/pkg/mux/...
```

The current happy-path tests are a good start, but the exported pattern mutation and custom-regex extraction behavior should be covered before this test suite is merged.

PR #6962, opened 2026-07-02
synccontroller: harden ObjectSync delivery in HA

fanny7d:
**What type of PR is this?**

/kind bug

This PR hardens reliable resource delivery when CloudCore runs with multiple replicas.

The concrete failures addressed are:

- Multiple replicas can observe the same downstream resource and race while creating the same ObjectSync or ClusterObjectSync. An `AlreadyExists` response is treated as a successful create race, and the existing object is fetched before status initialization continues.
- A delayed ACK can overwrite `status.objectResourceVersion` with an older value. ACK updates are monotonic and ignore resource versions that are not newer than the stored value.
- CloudHub queues and edge sessions are local to each CloudCore Pod. A replica that does not own the target edge session must not create a local message pool or enqueue locally, but it must still persist the shared ObjectSync or ClusterObjectSync. The synccontroller on the session-owning replica can then observe that sync point and deliver the resource.
- Existing ObjectSync records with an empty status are treated as resource version `0`, allowing synccontroller compensation to resend the current object.
- Cleanup ignores `NotFound` when another replica has already deleted a stale ObjectSync.

This deliberately does not add global leader election to synccontroller. Reliable-sync delivery must execute on the CloudHub replica that owns the target edge-node connection; a global leader may not own that session.

Validation performed:

- `go test -count=1 ./cloud/pkg/cloudhub/dispatcher ./cloud/pkg/cloudhub/session ./cloud/pkg/synccontroller`
- `go test -race -count=1 ./cloud/pkg/cloudhub/dispatcher`
- `make verify`
- `git diff --check`
- Reproduced and validated on a three-replica ACK CloudCore deployment where the controller and edge session were on different replicas. Before the fix, the shared ObjectSync was absent and MetaServer requests remained unauthorized. After the fix, the source and acknowledged resource versions converged and the affected reporters succeeded.

The ObjectSync GVK/spec backfill change is intentionally excluded and is submitted separately in #7025.

```release-note
CloudCore reliable sync now handles multi-replica ObjectSync creation races, stale ACKs, and split controller/session ownership without dropping shared sync state.
```

### DIFF
--- a/cloud/pkg/cloudhub/common/testing/test_framework.go
+++ b/cloud/pkg/cloudhub/common/testing/test_framework.go
@@ -39,6 +39,7 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/common"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/messagelayer"
 	"github.com/kubeedge/kubeedge/cloud/pkg/common/modules"
+	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/dao/models"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


fanny7d:
Updated this branch onto the latest upstream `master` (including the v1.23.1 security fixes). The synccontroller and CloudHub dispatcher/session unit tests pass.

/assign @kevin-wangzefeng

fanny7d:
Follow-up from three-replica ACK validation:

- Added `1dad2b73a` so a CloudCore replica without the target node's local session still persists the shared ObjectSync/ClusterObjectSync, while continuing to avoid creating or enqueueing a local message pool.
- Added namespaced and cluster-scoped regression tests for split controller/session ownership.
- Re-ran the dispatcher/session/synccontroller unit tests, the dispatcher race test, `make verify`, and `git diff --check`; all passed.

This updates the existing focused PR rather than reopening the previously closed combined PR.

PR #6964, opened 2026-07-03
fix: use configured address in kubeletHealthCheck and handle ReadOnlyPort=0

### LINKED ISSUE #6952: edged: kubeletHealthCheck hardcodes localhost and breaks initialization when ReadOnlyPort is disabled
**What happened**:
In `edge/pkg/edged/edged.go`, the `kubeletHealthCheck` function hardcodes the health check URL to `http://localhost:%d/healthz/syncloop`. 

This causes `edged` to silently fail and block pod syncing forever in two common scenarios:
1. If the user disables the `ReadOnlyPort` (by setting it to `0` in their configuration for standard security best practices), `kubeletHealthCheck` polls `http://localhost:0` indefinitely and connection is refused.
2. If the user binds the Kubelet to a specific IP address rather than `0.0.0.0`, the `localhost` connection is refused.

Because it polls in an infinite loop, `edgecore` starts but pods are never synced to the edge node.

`kubeletHealthCheck` should use the explicitly configured `HealthzBindAddress` or `Address` from the `KubeletConfiguration` instead of assuming `localhost`. Additionally, if the user has disabled the `ReadOnlyPort` (`port == 0`), it should log a warning and gracefully skip the health check instead of blocking indefinitely.

Configure the KubeEdge edge node with `readOnlyPort: 0` in the configuration. Start `edgecore` and observe that `edged` never begins syncing pods because it is stuck in the `kubeletHealthCheck` loop.

I have already fixed this issue locally by updating the caller to pass `KubeletConfiguration.HealthzBindAddress` to `kubeletHealthCheck`, and adding a check to skip the loop if `port == 0`. I will open a PR shortly!

**Environment**:
- Kubernetes version (use `kubectl version`): (latest)
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): master
- <details> <summary>Cloud nodes Environment:</summary>
  - Kernel (e.g. `uname -a`): 
  - Others:
</details>

- <details> <summary>Edge nodes Environment:</summary>
  - edgecore version (e.g. `edgecore --version`): master
  - Kernel (e.g. `uname -a`): 
  - Others:
</details>

Samurai007AK:
kubeletHealthCheck previously hardcoded `localhost` in the health check URL,
which breaks when:
1. ReadOnlyPort is disabled (set to 0) - causes infinite poll loop
2. Kubelet is bound to a specific IP instead of 0.0.0.0

Now uses KubeletConfiguration.Address for the health check URL and
gracefully skips the check when ReadOnlyPort is 0. Includes IPv6 support
via `net.JoinHostPort`, wildcard address (`0.0.0.0`, `::`) mapping to
loopback for the health check, an HTTP client timeout to prevent
indefinite blocking, and regression tests for the address/URL handling.

Fixes #6952

/kind bug

```release-note
Fix edged kubeletHealthCheck to use the configured address instead of hardcoded localhost, and to gracefully handle ReadOnlyPort=0, wildcard bind addresses, IPv6, and HTTP client timeouts.
```

### DIFF
--- a/edge/pkg/edged/edged.go
+++ b/edge/pkg/edged/edged.go
@@ -28,6 +28,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"reflect"
@@ -158,7 +159,7 @@ func (e *edged) Start() {
 	defer e.cleanupContainers()
 
 	kubeletReadyChan := make(chan struct{}, 1)
-	go kubeletHealthCheck(e.KubeletServer.ReadOnlyPort, kubeletReadyChan)
+	go kubeletHealthCheck(e.KubeletServer.Address, e.KubeletServer.ReadOnlyPort, kubeletReadyChan)
 
 	select {
 	case <-beehiveContext.Done():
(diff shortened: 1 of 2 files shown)


DoisLONG:
Thanks for working on this fix. The current implementation looks much better after handling `ReadOnlyPort=0`, wildcard addresses, IPv6 formatting, and HTTP client timeout.

One small suggestion: could you add a focused regression test for `kubeletHealthCheck`? At least the following cases would be useful:

1. `ReadOnlyPort == 0` should send on `kubeletReadyChan` and return without polling.
2. Empty address / `0.0.0.0` / `::` should be mapped to loopback addresses correctly.
3. A normal configured address should build a valid health check endpoint.

If adding tests is hard because the function currently builds and polls the URL internally, it may be worth extracting the host/URL building logic into a small helper and testing that directly.

Also, please make sure the PR description follows the usual template, including `/kind bug` and release-note if needed.

luomengY:
What is the significance of a configuration like `ReadOnlyPort=0`?Is this an incorrect configuration?

Issue #6965, opened 2026-07-03
edgecore stays NotReady for ~15-20 minutes after a network change (WiFi switchover / link flap)


ryoheimorimoto:
**What happened**:

When an edge node's network changes underneath an established connection — a WiFi switchover, an IP address change, or a link flap — the existing WebSocket connection from edgecore to cloudcore becomes half-open. edgecore does not detect this quickly: it stays blocked on the socket read and the node remains `NotReady` on the control plane for roughly **15-20 minutes** before edgecore finally gives up, reconnects, and the node returns to `Ready`.

**What you expected to happen**:

edgecore should detect the stalled/changed connection and reconnect within seconds to a low number of minutes, returning the node to `Ready` promptly after connectivity is restored.

**How to reproduce it (as minimally and precisely as possible)**:

1. Register an edge node that connects to cloudcore over WebSocket.
2. Optionally set a longer heartbeat to make the effect more visible (`modules.edgeHub.heartbeat`, e.g. `120`).
3. Break the path in a way that leaves the TCP connection half-open (the peer stops receiving ACKs but no RST/FIN is delivered). Either of:
   - Drop packets bidirectionally between the edge node and cloudcore (e.g. `iptables`/`nft` DROP rules on the cloudcore endpoint), or
   - Trigger a real interface change on the edge node (bring the active interface down so traffic fails over to another interface / IP).
4. Watch the node status: `kubectl get node <edge-node> -w`.

Observed: the node transitions to `NotReady` and stays there for ~15-20 minutes before recovering.

**Anything else we need to know? (root cause analysis)**:

There are two independent contributors to the long recovery time:

1. **Half-open connections are not detected (read deadline is a silent no-op).** `WSConnection.SetReadDeadline` (`pkg/viaduct/pkg/conn/ws.go`) stores the deadline in a struct field but never propagates it to the underlying `gorilla/websocket` connection, and `handleMessage` never applies a read deadline at all. As a result the configured `modules.edgeHub.websocket.readDeadline` (default `15`) has no effect on WebSocket connections. A half-open TCP socket therefore blocks in `ReadMessage` until the kernel's TCP retransmission timeout (`net.ipv4.tcp_retries2=15`, ~15 minutes on Linux) — which dominates the ~15-20 minute recovery time.

[...]

PR #6966, opened 2026-07-03
fix(viaduct): respect read deadline on WSConnection

ryoheimorimoto:
**What type of PR is this?**

/kind bug

This PR fixes two related defects in the WebSocket transport (`viaduct`) that cause an edge node to remain `NotReady` for ~15-20 minutes after a network change (WiFi switchover, link flap, IP change):

1. `WSConnection.SetReadDeadline` stored the deadline in a struct field but never propagated it to the underlying `gorilla/websocket` connection, and `handleMessage` never applied a read deadline at all. This made `modules.edgeHub.websocket.readDeadline` (default `15`) a silent no-op: a half-open TCP socket blocked on `ReadMessage` until the kernel's TCP retransmission timeout (`net.ipv4.tcp_retries2=15`, ~15 minutes on Linux).

2. When a read error caused `handleMessage` to return, the `messageFifo` was not closed. Callers blocked on `ReadMessage()` / `Get()` (notably EdgeHub's `routeToEdge`) only learned about the disconnection when the next keepalive `Send` failed, delaying the reconnect by up to one heartbeat period.

The fix wires `WebSocketConfig.ReadDeadline` through `wsclient.Options.ReadDeadline` to `ConnectionOptions.ReadDeadlineInterval` and into `WSConnection`, then applies it on every `handleMessage` iteration, and closes the FIFO on a read error so the failure is observable immediately.

**Why the deadline is safe on idle connections (ping/pong keepalive)**: a read deadline alone would tear down idle-but-healthy connections, because nothing guarantees cloud-to-edge traffic within the interval (EdgeHub's application-level keepalive is one-way; CloudHub sends no response to it). To make deadline expiry mean "genuinely stalled" rather than "quiet", `pingLoop` sends WebSocket protocol pings every `readDeadlineInterval/2` and a pong handler extends the read deadline on every pong. The peer answers pings automatically (gorilla's default ping handler) as long as its read loop is running, so **no CloudHub change is needed** and the mechanism is backward-compatible with existing CloudHub deployments. A transient `WriteControl` timeout (write lock held by a large in-flight message on a slow link) does not stop the ping loop; only a genuine connection error does.

Timeout-related read errors are logged at `V(2)` consistently across the three layers that observe them (`WSConnection.handleMessage`, `WSLane.Read`, `packer.Reader.Read`); genuine transport errors still surface as `Errorf`.

`SetReadDeadline` / `SetWriteDeadline` now also propagate to the underlying connection (previously both returned `nil` without applying the deadline).

Part of #6965

_Not using `Fixes` on purpose: #6965 is addressed together with a companion enhancement PR; please keep the issue open until both are merged._

- This PR is intentionally scoped to the `viaduct` layer. A companion PR (#6967) addresses the EdgeHub-level reconnect backoff and reconnect-signal coalescing. The two PRs are independent (each builds and tests on its own); the maximum reduction in `NotReady` time is achieved when both are applied.
- QUIC has the same defect (`QuicConnection.SetReadDeadline` is also a no-op) and will be addressed in a separate PR; `Options.ReadDeadline` is documented as WebSocket-only for now.
- Added `pkg/viaduct/pkg/conn/ws_test.go` covering `SetReadDeadline` propagation, deadline-driven `handleMessage` exit on a stalled peer (TCP alive, no WebSocket-level processing), ping/pong keeping an idle connection alive, and the legacy zero-deadline behavior.
- Compared to TCP keepalive / `TCP_USER_TIMEOUT`: WebSocket-level ping/pong works identically across platforms, requires no socket options plumbing through the dialer, and verifies liveness end-to-end through the TLS session rather than only at the TCP layer.

Measured on a Raspberry Pi 5 edge node (Debian 12, WebSocket transport, `heartbeat: 120`, `readDeadline: 60`):
- Before: node `NotReady` for ~20 minutes after the network path broke, most of it spent in kernel TCP retransmission inside the read path.
- With this fix: after a wired-to-Wi-Fi interface switchover, the read deadline fired **59.7s** after the switchover (matching the configured bound; the default 15s bounds detection even tighter) and the connection was re-established ~66s after the switchover. An idle connection showed **no spurious reconnects over 7+ minutes** — the ping/pong keepalive kept the deadline extended.

```release-note
fix(viaduct): WSConnection now applies the configured read deadline to the underlying WebSocket connection and keeps idle connections alive with WebSocket protocol ping/pong. Edge nodes detect half-open TCP connections within `modules.edgeHub.websocket.readDeadline` (default 15s) instead of waiting for the kernel TCP retransmission timeout (~15 minutes on Linux); idle-but-healthy connections are unaffected. Read errors caused by a deadline expiry are logged at V(2), as they are the expected reconnect path.
```

### DIFF
--- a/pkg/viaduct/pkg/conn/ws.go
+++ b/pkg/viaduct/pkg/conn/ws.go
@@ -20,30 +20,32 @@ import (
 )
 
 type WSConnection struct {
-	WriteDeadline      time.Time
-	ReadDeadline       time.Time
-	handler            mux.Handler
-	wsConn             *websocket.Conn
-	state              *ConnectionState
-	syncKeeper         *keeper.SyncKeeper
-	connUse            api.UseType
-	consumer           io.Writer
-	autoRoute          bool
-	messageFifo        *fifo.MessageFifo
-	locker             sync.Mutex
-	OnReadTransportErr func(nodeID, projectID string)
+	WriteDeadline        time.Time
+	ReadDeadline         time.Time
+	handler              mux.Handler
+	wsConn               *websocket.Conn
+	state                *ConnectionState
+	syncKeeper           *keeper.SyncKeeper
+	connUse              api.UseType
+	consumer             io.Writer
+	autoRoute            bool
+	messageFifo          *fifo.MessageFifo
+	locker               sync.Mutex
+	OnReadTransportErr   func(nodeID, projectID string)
+	readDeadlineInterval time.Duration
 }
 
 func NewWSConn(options *ConnectionOptions) *WSConnection {
 	return &WSConnection{
-		wsConn:             options.Base.(*websocket.Conn),
-		handler:            options.Handler,
-		syncKeeper:         keeper.NewSyncKeeper(),
-		state:              options.State,
-		connUse:            options.ConnUse,
-		autoRoute:          options.AutoRoute,
-		messageFifo:        fifo.NewMessageFifo(),
-		OnReadTransportErr: options.OnReadTransportErr,
+		wsConn:               options.Base.(*websocket.Conn),
+		handler:              options.Handler,
+		syncKeeper:           keeper.NewSyncKeeper(),
+		state:                options.State,
+		connUse:              options.ConnUse,
+		autoRoute:            options.AutoRoute,
+		messageFifo:          fifo.NewMessageFifo(),
+		OnReadTransportErr:   options.OnReadTransportErr,
+		readDeadlineInterval: options.ReadDeadlineInterval,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


pkg/viaduct/pkg/conn/ws.go
```diff
@@ -99,16 +101,94 @@ func (conn *WSConnection) handleRawData() {
...
+					// The write lock was held past the deadline (e.g. a
+					// large message on a slow link); the connection may
+					// well be healthy. Keep pinging — liveness is judged
+					// by the read deadline, not here.
+					continue
+				}
+				// The connection is gone (closed or broken); the read side
+				// surfaces the failure via the read deadline.
+				klog.V(2).Infof("ping loop stopped: %v", err)
+				return
```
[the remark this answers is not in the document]
ryoheimorimoto: Done in a8cae66 — on a terminal write error the ping loop now closes the connection so the read loop unblocks immediately instead of waiting for the read deadline to expire. Good catch.

pkg/viaduct/pkg/conn/ws.go
```diff
@@ -99,16 +101,96 @@ func (conn *WSConnection) handleRawData() {
...
+					// large message on a slow link); the connection may
+					// well be healthy. Keep pinging — liveness is judged
+					// by the read deadline, not here.
+					continue
+				}
+				// The connection is gone (closed or broken); close it so
+				// the read loop unblocks immediately instead of waiting
+				// for the read deadline to expire.
+				klog.V(2).Infof("ping loop stopped: %v", err)
+				_ = conn.wsConn.Close()
```
fujitatomoya, line 149: i guess this is harmless but better to fix.

if the connection was already torn down by the read loop (handleMessage closes wsConn on a read error before stopPing is closed), this write error is a consequence of that shutdown, not a new failure? it exits quietly without logging a misleading "ping loop stopped" or double-closing.

fujitatomoya (approved):
i believe that this is a genuine bug, and the fix (arming the deadline on every read iteration) is right thing to do. minor comment but as i mentioned in the comment, that is harmless. so i will leave this maintainers, lgtm with green CI.

pkg/viaduct/pkg/conn/ws.go
ryoheimorimoto: Fixed in aad5880 — the ping loop now checks `stop` first and returns quietly when handleMessage already closed the connection, avoiding the misleading log and the double close.

ryoheimorimoto:
All review comments are addressed. The CI workflows on the latest commit are still awaiting maintainer approval to run, so could someone approve them? PTAL @fujitatomoya @chendave

fujitatomoya:
@kevin-wangzefeng @Shelley-BaoYue can you guys allocate some time for this?

ryoheimorimoto:
Squashed into a single commit and rebased onto the latest master, same as #6967.

ryoheimorimoto:
Same state as #6967: the CI workflows on the squashed commit are waiting for approval to run. @fujitatomoya the code is identical to what you approved, just squashed into one commit, so PTAL once CI is green. cc @chendave

ryoheimorimoto:
Rebased onto the latest master and resolved the conflict with #6900: ws_test.go now contains both the existing handleRawData regression test and the tests from this PR. Same remaining steps as #6967: approving the CI workflow runs, then re-lgtm and /approve.

/assign @DoisLONG

DoisLONG:
Thanks for the update.

I rechecked the latest version after the rebase and conflict resolution with #6900.

The PR is now squashed into a single commit, and the `ws_test.go` conflict looks handled properly: the existing `handleRawData` regression test is kept, and the read-deadline tests from this PR are still included.

The fix still looks good to me. Passing EdgeHub's configured WebSocket read deadline down into viaduct and applying it on each `handleMessage()` read iteration fixes the no-op read deadline issue. The ping/pong loop also makes the deadline safe for idle but healthy WebSocket connections, so quiet connections are not disconnected just because there is no cloud-to-edge application traffic.

Closing `messageFifo` on read error is also important, because it lets callers blocked on `ReadMessage()` observe the disconnect immediately instead of waiting for the next keepalive write failure.

/lgtm

ryoheimorimoto:
Thanks @DoisLONG for the lgtm. The remaining steps are approving the pending CI workflow runs and /approve. @WillardHu could you take a look when you have a moment?

/assign @WillardHu

ryoheimorimoto:
Added unit tests for the new code paths (constructor wiring, the ping loop's timeout, terminal-error and stop paths, SetWriteDeadline propagation, WSLane.Read error classification, and packer's silentReadErr with truncated streams). Locally this brings the patch coverage well above the 80% codecov target. Kept as a single commit as requested; the delta since the last push is test-only. The push also retriggers the QUIC v1.30 e2e job, which failed on the known pod-startup flake.

ryoheimorimoto:
@DoisLONG the delta since your lgtm is test-only (no production code changed), PTAL for re-lgtm when you have a moment.

ryoheimorimoto:
@DoisLONG thanks for approving the workflow runs, and sorry for asking you again and again on this PR. Two e2e jobs failed on a pod that stayed Pending for 4 minutes (E2E_MAPPER_1 on cri-o, E2E_POD_DEPLOYMENT_4 on QUIC v1.30). The WebSocket e2e jobs this PR touches passed on all three Kubernetes versions, and the cri-o failure is the same spec that failed on master in run 33755666342. Could you re-run the failed jobs when you have a moment? codecov/patch is now at 98%.

Shelley-BaoYue (commented):
Sorry for late reply. 
/approve

PR #6967, opened 2026-07-03
feat(edgehub): exponential backoff & coalesce reconnect signals

ryoheimorimoto:
**What type of PR is this?**

/kind feature

This PR replaces EdgeHub's fixed `Heartbeat * 2` reconnect wait time with an exponential backoff (initial 2s, factor 2, jitter 0.2, cap 30s, effectively unbounded steps). With the previous fixed wait, environments configuring a long heartbeat period (e.g. 120s to reduce cloudhub load on large clusters) had to wait up to 4 minutes before retrying after a transient outage, even when the network had already recovered.

**The backoff is reset only after the connection stayed up longer than the backoff cap** (`shouldResetBackoff`), not on every successful handshake. Resetting on handshake success would let a connect-then-die loop (e.g. an overloaded CloudHub or a broken LB) retry at the initial ~2s interval forever — a connection storm across a large fleet.

This PR also makes `reconnectChan` a buffered(1) channel and routes the transport-error senders (`routeToEdge` / `routeToCloud` / `keepalive`) through a non-blocking `triggerReconnect()` helper. With the previous unbuffered channel, multiple goroutines firing on the same outage could block on the send; one queued signal is sufficient to drive the reconnect loop, so coalescing is intentional. The labeled drain loop is replaced with a `drainReconnect()` helper (single non-blocking receive), and a stale signal queued during `chClient.Init()` is drained immediately after a successful connect so it does not trigger an unnecessary reconnect cycle.

**Certificate-rotation signals are deliberately NOT coalesced with transport reconnects**: they flow through a dedicated buffered(1) `rotateChan` and the reconnect wait `select`s on both channels. A rotation must always result in a connection that uses the new certificate (`chClient.Init()` reloads it from disk); routing the signal through the drainable `reconnectChan` could discard it when a rotation completes while a connect is in flight, leaving the old certificate in use until the next natural disconnect (`certManager.Done` fires only once per rotation). While connected, a rotation signal is never discarded; a stale signal is dropped only while disconnected, right before an `Init()` that reads the newest certificate from disk anyway (the certificate files are fully written before the signal is sent). Rotation-triggered reconnects are logged at Info instead of the broken-connection Warning, since the disconnect is intentional.

Part of #6965

_Not using `Fixes` on purpose: #6965 is addressed together with a companion bug-fix PR (#6966); please keep the issue open until both are merged._

- This PR is intentionally scoped to the EdgeHub layer. A companion PR (#6966) addresses the related `WSConnection` read-deadline defect in the `viaduct` transport. The two PRs are independent (each builds and tests on its own); the maximum reduction in `NotReady` time is achieved when both are applied.
- Tests cover the coalescing semantics of `triggerReconnect`, `drainReconnect` never consuming rotation signals, the `ifRotationDone` → `rotateChan` wiring, the `shouldResetBackoff` gating, and the value range of `reconnectBackoff` (positive, capped, jittered, resettable).
- The backoff constants (initial 2s, cap 30s) are chosen to recover quickly while staying well below the previous default wait (`heartbeat*2` = 30s with the default heartbeat). If maintainers prefer, exposing them through the EdgeHub config is a straightforward follow-up; kept out of this PR to avoid an API change.

Measured on a Raspberry Pi 5 edge node (Debian 12, WebSocket transport, `heartbeat: 120`, `readDeadline: 60`), with the companion read-deadline fix applied:
- `NotReady` time after an interface switchover dropped from ~20 minutes to **1-2 minutes**; the reconnect wait observed at recovery was **~2s** (vs. the previous fixed 4 minutes at `heartbeat: 120`).
- When the first reconnect attempt landed on a load-balancer instance still holding the stale session and was closed immediately, the backoff grew 2.0s → 4.3s and the second attempt succeeded — and after the connection stayed healthy well past the cap, the next outage started from the initial ~2s again (`shouldResetBackoff` behaving as designed).

```release-note
feat(edgehub): EdgeHub now uses exponential backoff with jitter (initial 2s, cap 30s) between reconnect attempts instead of a fixed `Heartbeat * 2` wait, so edge nodes recover from transient cloud outages within seconds even when `heartbeat` is configured to a long period. The backoff resets only after a connection stays healthy for longer than the cap, preventing connection storms from connect-then-die loops. Reconnect signals from concurrent senders are coalesced through a buffered, non-blocking channel; certificate-rotation signals use a dedicated non-coalesced channel so a rotation always results in a connection that uses the new certificate.
```

### DIFF
--- a/edge/pkg/edgehub/edgehub.go
+++ b/edge/pkg/edgehub/edgehub.go
@@ -1,9 +1,11 @@
 package edgehub
 
 import (
+	"math"
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/klog/v2"
 
@@ -18,14 +20,40 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/features"
 )
 
+// reconnectBackoff returns the backoff used between reconnect attempts.
+// It starts small and grows exponentially with jitter, capped at 30s, so
+// that a quickly-recovered network surfaces a NotReady node back to Ready
+// within seconds instead of waiting heartbeat*2 (up to 4 minutes when
+// heartbeat is configured to 120s).
+func reconnectBackoff() wait.Backoff {
+	return wait.Backoff{
+		Duration: 2 * time.Second,
+		Factor:   2.0,
+		Jitter:   0.2,
+		Cap:      30 * time.Second,
+		// Steps only bounds how many times Duration may grow — it never
+		// stops Step() from returning values: once Cap is reached, Step()
+		// keeps returning a jittered value in [Cap, Cap*(1+Jitter))
+		// indefinitely. MaxInt32 just ensures growth is never cut short.
+		Steps: math.MaxInt32,
+	}
+}
+
 // EdgeHub defines edgehub object structure
 type EdgeHub struct {
-	certManager   certificate.CertManager
-	chClient      clients.Adapter
+	certManager certificate.CertManager
+	chClient    clients.Adapter
+	// reconnectChan is buffered(1); the transport-error senders
+	// (routeToEdge / routeToCloud / keepalive) deliver through the
+	// non-blocking triggerReconnect, so signals for one outage coalesce.
 	reconnectChan chan struct{}
-	rateLimiter   flowcontrol.RateLimiter
-	keeperLock    sync.RWMutex
-	enable        bool
+	// rotateChan carries certificate-rotation signals separately from
+	// reconnectChan so they are never coalesced or drained away: a rotation
+	// must always be followed by a reconnect that reloads the certificate.
+	rotateChan  chan struct{}
+	rateLimiter flowcontrol.RateLimiter
+	keeperLock  sync.RWMutex
+	enable      bool
 }
 
 var _ core.Module = (*EdgeHub)(nil)
@@ -46,11 +74,42 @@ func newEdgeHub(enable bool) *EdgeHub {
 	NewCertSyncChannel()
 	return &EdgeHub{
 		enable:        enable,
-		reconnectChan: make(chan struct{}),
+		reconnectChan: make(chan struct{}, 1),
+		rotateChan:    make(chan struct{}, 1),
 		rateLimiter: flowcontrol.NewTokenBucketRateLimiter(
 			float32(config.Config.EdgeHub.MessageQPS),
 			int(config.Config.EdgeHub.MessageBurst)),
 	}
+}
+
+// triggerReconnect requests a reconnect without blocking. If a reconnect
+// signal is already pending, the call is a no-op since one signal is
+// sufficient to drive the reconnect loop.
+func (eh *EdgeHub) triggerReconnect() {
+	select {
+	case eh.reconnectChan <- struct{}{}:
+	default:
+	}
+}
+
+// drainReconnect discards a pending transport reconnect signal, if any.
+// It deliberately does not touch rotateChan: dropping a rotation signal
+// could leave a connection running on a stale certificate until the next
+// natural disconnect.
+func (eh *EdgeHub) drainReconnect() {
+	select {
+	case <-eh.reconnectChan:
+	default:
+	}
+}
+
+// shouldResetBackoff reports whether the connection stayed up long enough to
+// consider the previous outage over. Resetting on every successful handshake
+// would let a connect-then-die loop (e.g. an overloaded CloudHub or a broken
+// LB) retry at the initial ~2s interval forever — a connection storm across
+// a large fleet. Surviving longer than the backoff cap is taken as healthy.
+func shouldResetBackoff(connectedAt time.Time, b wait.Backoff) bool {
+	return time.Since(connectedAt) > b.Cap
 }
 
 // Register register edgehub
@@ -99,6 +158,7 @@ func (eh *EdgeHub) Start() {
 
 	go eh.ifRotationDone()
 
+	backoff := reconnectBackoff()
 	for {
 		select {
 		case <-beehiveContext.Done():
@@ -112,14 +172,37 @@ func (eh *EdgeHub) Start() {
 			return
 		}
 
-		waitTime := time.Duration(config.Config.Heartbeat) * time.Second * 2
-
+		// A rotation that completed while we were disconnected is already
+		// satisfied by the upcoming Init(), which reads the newest
+		// certificate from disk (certManager writes the files before
+		// signaling). Discard such a stale signal now — while still
+		// disconnected — so it does not trigger a redundant reconnect right
+		// after the connection is established. A rotation completing after
+		// this point sends a fresh signal that survives to the wait below.
+		select {
+		case <-eh.rotateChan:
+		default:
+		}
 		err = eh.chClient.Init()
 		if err != nil {
-			klog.Errorf("connection failed: %v, will reconnect after %s", err, waitTime.String())
-			time.Sleep(waitTime)
+			sleep := backoff.Step()
+			klog.Errorf("connection failed: %v, will reconnect after %s", err, sleep.String())
+			select {
+			case <-beehiveContext.Done():
+				klog.Warning("EdgeHub stop")
+				return
+			case <-time.After(sleep):
+			}
 			continue
 		}
+		// Drain any stale transport-error signal queued by the previous
+		// generation of goroutines while Init was in flight. Without this,
+		// the stale signal would be received immediately by the wait below,
+		// triggering an unnecessary reconnect cycle right after a successful
+		// connect. Certificate-rotation signals are deliberately not
+		// drained (see rotateChan).
+		eh.drainReconnect()
+		connectedAt := time.Now()
 		// execute hook func after connect
 		eh.pubConnectInfo(true)
 		go eh.routeToEdge()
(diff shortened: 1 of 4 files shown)


edge/pkg/edgehub/edgehub.go
```diff
@@ -112,14 +177,32 @@ func (eh *EdgeHub) Start() {
...
+		case <-eh.rotateChan:
+		default:
+		}
 		err = eh.chClient.Init()
 		if err != nil {
-			klog.Errorf("connection failed: %v, will reconnect after %s", err, waitTime.String())
-			time.Sleep(waitTime)
+			sleep := backoff.Step()
+			klog.Errorf("connection failed: %v, will reconnect after %s", err, sleep.String())
+			time.Sleep(sleep)
```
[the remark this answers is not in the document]
ryoheimorimoto: Done in 939b40c — the backoff sleep now selects on `beehiveContext.Done()`. (This wait not observing shutdown predates this PR, but since the PR rewrites these lines it makes sense to fix it here.)

edge/pkg/edgehub/edgehub.go
```diff
@@ -128,24 +211,39 @@ func (eh *EdgeHub) Start() {
...
-		<-eh.reconnectChan
+		rotated := false
+		select {
+		case <-eh.reconnectChan:
+		case <-eh.rotateChan:
+			// The certificate was rotated (possibly while the connect above
+			// was in flight). Re-establish the connection so chClient.Init()
+			// reloads the new certificate from disk.
+			rotated = true
+		}
```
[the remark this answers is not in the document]
ryoheimorimoto: Done in 939b40c — added a `beehiveContext.Done()` case to the reconnect wait. I kept the shutdown path minimal (log + return, mirroring the existing loop-top check) rather than also calling `UnInit()`/`pubConnectInfo(false)`: the process teardown closes the connection, and the disconnect fan-out isn't meaningful mid-shutdown. Happy to add explicit cleanup if maintainers prefer. (This wait blocking on shutdown also predates this PR.)

edge/pkg/edgehub/edgehub.go
```diff
@@ -128,24 +211,39 @@ func (eh *EdgeHub) Start() {
...
 		}
+		sleep := backoff.Step()
+		if rotated {
+			// Intentional disconnect: not a transport failure, so do not
+			// alarm operators with a broken-connection warning.
+			klog.Infof("certificate rotated, will reconnect after %s to reload it", sleep.String())
+		} else {
+			klog.Warningf("connection is broken, will reconnect after %s", sleep.String())
+		}
+		time.Sleep(sleep)
```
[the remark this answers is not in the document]
ryoheimorimoto: Done in 939b40c — same treatment as the other backoff sleep.

edge/pkg/edgehub/process.go
```diff
@@ -150,7 +150,19 @@ func (eh *EdgeHub) ifRotationDone() {
...
+			// dropped only while disconnected, right before an Init() that
+			// reads the newest certificate anyway. Buffered(1) +
+			// non-blocking keeps this loop from blocking when a signal is
+			// already pending — one pending rotation reconnect is enough,
+			// the newest certificate is always read from disk.
+			select {
+			case eh.rotateChan <- struct{}{}:
+			default:
+			}
 		}
```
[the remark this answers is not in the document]
ryoheimorimoto: Done in 5eefc79 — `ifRotationDone` now exits on `beehiveContext.Done()`. (Also pre-existing on master, but worth fixing together with the other shutdown-responsiveness changes in this PR.)

DoisLONG:
Thanks for the detailed explanation. The scope is clear to me: this PR focuses on the EdgeHub reconnect/backoff behavior, while the viaduct read-deadline fix is kept in #6966.

I noticed the previous inline comments around shutdown responsiveness have mostly been addressed by replacing the blocking sleeps/selects with `beehiveContext.Done()`-aware selects.

One thing I’d still like to confirm: in the connected state, when `beehiveContext.Done()` is received, the current code returns directly before calling `eh.chClient.UnInit()` and `eh.pubConnectInfo(false)`. For reconnect and certificate-rotation paths, those cleanup steps are still executed after the select.

Could you check whether the shutdown path should also perform the same cleanup before returning? If it is intentionally relying on process/module teardown to close the connection, it would be good to make that explicit; otherwise, it may be safer to call `UnInit()` and publish the disconnected state before returning.

Other than that, the reconnect signal coalescing and the dedicated `rotateChan` separation look reasonable to me.

edge/pkg/edgehub/edgehub.go
```diff
@@ -45,14 +67,56 @@ func NewCertSyncChannel() map[string]chan bool {
 func newEdgeHub(enable bool) *EdgeHub {
 	NewCertSyncChannel()
 	return &EdgeHub{
-		enable:        enable,
-		reconnectChan: make(chan struct{}),
+		enable: enable,
```
fujitatomoya: nitpick; i would avoid the unnecessary changes to show the difference clearly...

edge/pkg/edgehub/edgehub.go
```diff
@@ -128,24 +216,49 @@ func (eh *EdgeHub) Start() {
...
-				break clean
-			}
+		// Reset the backoff only after the connection proved healthy for a
+		// while, so a connect-then-die loop keeps backing off instead of
+		// hammering the cloud at the initial interval.
+		if shouldResetBackoff(connectedAt, backoff) {
+			backoff = reconnectBackoff()
+		}
+		sleep := backoff.Step()
+		if rotated {
```
fujitatomoya: non-blocking suggestion, but i was expecting that backoff sleep should be applied only when the else statement. if this statement is true, it should reconnect immediately without extra backoff sleep but curretly it only picks up the logging level.

the reason why i say this is non-blocking suggestion is that, cost is 2 seconds a few times a year per node...

fujitatomoya (approved):
lgtm with green CI.

ryoheimorimoto:
@DoisLONG good point — addressed in ee14ca2. On `beehiveContext.Done()` in the connected state it now calls `UnInit()` and publishes the disconnected state before returning, so the shutdown path cleans up the same way as the reconnect/rotation paths instead of relying on process teardown.

edge/pkg/edgehub/edgehub.go
ryoheimorimoto: Agreed — ee14ca2 now reconnects immediately on rotation (via `continue`) instead of taking a backoff step. The flag no longer only changes the log level.

edge/pkg/edgehub/edgehub.go
ryoheimorimoto: Done in ee14ca2 — moved the channel comments onto the struct fields so the constructor stays aligned with master and the diff is minimal.

ryoheimorimoto:
All review comments are addressed. The CI workflows on the latest commit are still awaiting maintainer approval to run, so could someone approve them? PTAL @fujitatomoya @DoisLONG

fujitatomoya:
@kevin-wangzefeng @Shelley-BaoYue same here, this is legitimate. can you take a look?

DoisLONG (commented):
Thanks for the detailed work here. I took another pass over the latest version and the direction looks good to me overall.

I also checked the tests. They cover the key semantics introduced by this PR: coalescing reconnect signals, not draining rotation signals, forwarding `certManager.Done` to `rotateChan`, the backoff reset condition, and the backoff range/reset behavior. These are meaningful tests rather than only coverage-oriented tests.

A few existing review threads are still unresolved, although some of them appear to be addressed by the latest code. Could you please resolve or follow up on the remaining threads so the discussion state matches the current patch?

After that, and with CI green, this looks reasonable to me.

ryoheimorimoto:
@DoisLONG thanks for the review. All review threads on this PR and #6966 are now resolved, and each has a reply pointing to the commit that addressed it. The CI workflows on the latest commits still need a maintainer to approve the run.

DoisLONG (commented):

The current direction looks reasonable to me now. The reconnect signal coalescing through a buffered `reconnectChan`, the dedicated `rotateChan` for certificate rotation, the backoff reset policy, and the shutdown cleanup path all look much clearer after the previous review rounds.

The existing review threads are resolved, and I do not see a remaining code blocker from my side.

One final cleanup before LGTM: this PR still has multiple commits. Could you please squash them into a single logical commit?

After that, and once CI is triggered/green, this should be good from my side.

DoisLONG (commented):

One final cleanup: this PR still has multiple commits. Could you please squash them into a single logical commit?

After that, and once CI is triggered/green, this should be good from my side.

ryoheimorimoto:
@DoisLONG done. Squashed into a single commit and rebased onto the latest master. I squashed #6966 the same way since the same convention should apply there. The CI workflows will need approval to run again.

DoisLONG (commented):
Thanks for the update.

/lgtm

ryoheimorimoto:
@DoisLONG thanks for the lgtm! The one remaining blocker is that the CI workflows on the squashed commit still need a maintainer to click "Approve and run" (the pending workflow runs on the Checks tab). #6966 is in the same state. Once CI is green, I believe only /approve remains. cc @kevin-wangzefeng

ryoheimorimoto:
@DoisLONG this has been quiet for a month, so let me ask you directly. Your conditions are met: the PR is a single squashed commit and all review threads are resolved. The remaining steps are approving the CI workflow runs and /approve. Could you take it from here when you have a moment?

/assign @DoisLONG

ryoheimorimoto:
Closed and reopened this PR to regenerate the CI workflow runs, which had expired after 30 days without approval (no jobs ever ran, so the failure marks were not real failures). The lgtm is in place, so the remaining steps are approving the new CI runs and /approve. @WillardHu could you take a look when you have a moment?

/assign @WillardHu

ryoheimorimoto:
Pushed an update to bring codecov/patch above the target: the reconnect loop in `Start` had no test coverage. Everything is still in a single commit. Changes on top of the previously reviewed version:

- `Start` now delegates to `run(done)`, and the client constructor is injectable (`newClient`, `clients.GetClient` in production), so the loop can be driven with the existing mock adapter. No behavior change by itself.
- Exercising the loop under the race detector surfaced two real issues, both fixed in this update:
  - With a backoff shorter than one heartbeat period, the previous connection's `keepalive` goroutine could survive into the next connection and keep pinging on it (one leaked goroutine per reconnect). The old fixed `Heartbeat*2` wait prevented this implicitly. Each connection now has a `stop` channel that its goroutines select on, and `keepalive` waits on it instead of `time.Sleep`.
  - `eh.chClient` was replaced by the loop while goroutines of the previous connection could still read it. It is now guarded by a mutex (`client()`).
- New tests: five loop tests in `run_test.go` and a keepalive teardown test. Locally, patch coverage is 95%; only the `klog.Exitf` branch is uncovered.

@DoisLONG sorry for the extra round, PTAL when you have time.

edge/pkg/edgehub/run_test.go
```diff
@@ -0,0 +1,279 @@
...
+func (h *runHarness) start(done <-chan struct{}) (exited chan struct{}) {
+	exited = make(chan struct{})
+	go func() {
+		h.eh.run(done)
+		close(exited)
+	}()
+	return exited
+}
+
+func waitFor(t *testing.T, ch <-chan struct{}, what string) {
```
Shelley-BaoYue, line 128: Maybe we can change this func name or add `waitFor` in codespell whitelist to fix code spell CI.  : )

Shelley-BaoYue (commented):
And rebase the master branch can fix the failed runtime CI.

edge/pkg/edgehub/run_test.go
ryoheimorimoto: Renamed the helper to awaitSignal, thanks. codespell --config codespell.cfg is clean locally.

ryoheimorimoto:
@Shelley-BaoYue thanks for the review and for running the CI. Rebased onto the latest master (picks up the cri-o mapper fix from #7263) and renamed the test helper that tripped codespell. Still a single commit; the production code is unchanged since your review. PTAL when you have a moment.

ryoheimorimoto:
Update on CI for the rebased head: everything is green except the cri-o runtime e2e, which still fails E2E_MAPPER_1 (the mapper pod stays Pending for 4 minutes). That job fails the same way on master itself in every push run since 2026-09-03, including the run for the #6966 merge commit, so it does not look related to this PR. One more thing for sequencing: #7295 touches the same code (per-connection stop of the routing goroutines and a buffered reconnectChan), which this PR already covers. I will leave the ordering to you.

Shelley-BaoYue (commented):
/approve 
cri-o e2e failed test will be fixed in another PR. 
PTAL @fujitatomoya @DoisLONG

PR #6970, opened 2026-07-04
edgehub: recover quickly from half-open cloud connections

### LINKED ISSUE #6965: edgecore stays NotReady for ~15-20 minutes after a network change (WiFi switchover / link flap)

Priyanshu6968:
**What type of PR is this?**
/kind bug

After a network change beneath an active edgecore->cloudcore websocket
connection (WiFi switchover, link flap, IP change), the connection goes
half open and the node stays NotReady for ~15-20 minutes before it
reconnects. Two independent issues cause the delay:

1. The read deadline is never applied to the underlying gorilla
   connection, so the read blocks until the kernel TCP timeout (~15 min).
2. EdgeHub waits a fixed Heartbeat*2 before reconnecting, adding more
   delay with longer heartbeat settings.

This wires the read deadline through to the connection and keeps it
refreshed with websocket ping/pong, so a stalled connection is detected
within the deadline instead of the kernel timeout. It also replaces the
fixed reconnect wait with capped exponential backoff and jitter. The
backoff is capped at the previous fixed wait, so the longest wait never
regresses while early attempts happen much sooner, and the jitter avoids
a reconnect storm when many nodes drop off one cloudcore at once.

The ping/pong is driven from the edge client only (gated on a non zero
read deadline); the cloud replies with pong automatically, so cloudcore,
QUIC, and the stream tunnel are unchanged.

Fixes #6965

```release-note
Fixed edgecore staying NotReady for ~15-20 minutes after a network change by detecting half open cloud connections via websocket ping/pong and using exponential backoff with jitter for reconnection.
```

### DIFF
--- a/edge/pkg/edgehub/clients/wsclient/websocket.go
+++ b/edge/pkg/edgehub/clients/wsclient/websocket.go
@@ -78,6 +78,7 @@ func (wsc *WebSocketClient) Init() error {
 		Addr:             wsc.config.URL,
 		AutoRoute:        false,
 		ConnUse:          api.UseTypeMessage,
+		ReadDeadline:     wsc.config.ReadDeadline,
 	}
 	exOpts := api.WSClientOption{Header: make(http.Header)}
 	exOpts.Header.Set("node_id", wsc.config.NodeID)
--- a/edge/pkg/edgehub/edgehub.go
+++ b/edge/pkg/edgehub/edgehub.go
@@ -40,17 +40,26 @@ const (
 	// reconnectJitter randomizes each wait so that many edge nodes losing the
 	// connection to the same cloudcore at once do not reconnect in lockstep.
 	reconnectJitter = 0.2
+	// defaultHeartbeat mirrors the config default and is used as a fallback when
+	// the configured heartbeat is non-positive, so the backoff cap stays valid.
+	defaultHeartbeat = 15
 )
 
 // newReconnectBackoff builds the exponential backoff used between reconnect
 // attempts. It is capped at the previous fixed wait (Heartbeat*2), so the
-// longest wait never regresses while early attempts happen much sooner.
+// longest wait never regresses while early attempts happen much sooner. A
+// non-positive heartbeat falls back to the default, otherwise the cap would be
+// zero and the backoff would grow without limit.
 func newReconnectBackoff() wait.Backoff {
+	heartbeat := config.Config.Heartbeat
+	if heartbeat <= 0 {
+		heartbeat = defaultHeartbeat
+	}
 	return wait.Backoff{
 		Duration: reconnectBaseDelay,
 		Factor:   reconnectFactor,
 		Jitter:   reconnectJitter,
-		Cap:      time.Duration(config.Config.Heartbeat) * time.Second * 2,
+		Cap:      time.Duration(heartbeat) * time.Second * 2,
 		Steps:    math.MaxInt32,
 	}
 }
--- a/edge/pkg/edgehub/edgehub.go
+++ b/edge/pkg/edgehub/edgehub.go
@@ -1,9 +1,11 @@
 package edgehub
 
 import (
+	"math"
 	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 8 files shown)


ryoheimorimoto:
Hi @Priyanshu6968, thanks for looking into #6965 (I'm the reporter)!

A quick heads-up in case they were easy to miss: two PRs implementing the proposed fix have been open since Jul 3 and are linked from the issue — #6966 (viaduct: read deadline + WebSocket ping/pong) and #6967 (edgehub: exponential backoff + reconnect-signal handling). They were split into a bug fix and an enhancement per the contributing guide's preference for smaller reviewable changes, and have been validated on real hardware (NotReady ~20 min → 1–2 min after an interface switchover).

Since this PR overlaps with both, it's probably best to let the maintainers decide which series to take forward so review effort isn't duplicated. A few details from the existing PRs that are worth carrying over whichever way that goes: closing the message FIFO on read error (so `routeToEdge` unblocks immediately instead of waiting for the next keepalive), tolerating transient `WriteControl` timeouts in the ping loop (a single timeout otherwise permanently disables the keepalive), and preserving certificate-rotation reconnect signals.

That said, deriving the backoff cap from `heartbeat*2` here is a nice touch for short-heartbeat setups — if maintainers prefer that behavior I'd be happy to fold it into #6967 with credit.

DoisLONG:
Thanks for working on this issue. The problem statement is valid, but this PR now overlaps heavily with #6966 and #6967.

#6966 focuses on the viaduct WebSocket read-deadline path, and #6967 focuses on the EdgeHub reconnect backoff / reconnect-signal handling. Those two PRs split the same issue into smaller, more focused changes, and they already cover several details that are also relevant here, such as FIFO closing, reconnect signal coalescing, shutdown responsiveness, and certificate-rotation handling.

To avoid duplicate review work and possible merge conflicts between overlapping fixes, I think it would be better to close this PR in favor of #6966 and #6967.

If you still see any gap that is not covered by those two PRs, it may be better to raise it directly in the corresponding PR or open a smaller follow-up PR.

Priyanshu6968:
Thanks for the review and the clarification. I wasn't aware that the work had already been split across #6966 and #6967. I'll close this PR in favor of those two PRs.

PR #6976, opened 2026-07-05
metaserver: close watch channel when hook registration fails

### LINKED ISSUE #6975: # metaserver: imitator.Watch returns nil channel causing permanent watch hang
**What happened**:

When `watchhook.NewWatchHook` fails in `imitator.Watch`, the function returns `nil` instead of a channel. The caller in `watcher.go` does `for wres := range wch` on that return value. In Go, ranging over a nil channel blocks forever, so the watch goroutine hangs and `kubectl watch` against the edge metaserver never receives events or an error.

If watch hook registration fails, the watch should end cleanly or return an error. It should not hang silently.

In `edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go`, `Watch` returns `nil` when `NewWatchHook` fails:

```go
wh, err := watchhook.NewWatchHook(key, rev, receiver)
if err != nil {
    klog.Errorf("add hook for %s failed, %v", key, err)
    return nil
}
```

The caller in `watcher.go` then blocks forever:

```go
wch := wc.watcher.client.Watch(wc.ctx, wc.key, uint64(wc.initialRev+1))
for wres := range wch {
```

`AddHook` fails when a hook with the same ID is already registered. Each hook uses a UUID so this is rare, but the error path is broken whenever it happens.

Suggested fix: close the channel and return it instead of returning `nil`.

* `imitator.Watch()` at `imitator.go` lines 204–210
* Called from `startWatching()` in `watcher.go` line 238
* `Client.Watch` returns `<-chan watch.Event` with no error return, so the fix belongs inside `imitator.Watch`
* Separate from the `watchhook.Trigger` map race fix

**Environment**:

* Kubernetes version (use `kubectl version`):
* KubeEdge version (e.g. `cloudcore --version` and `edgecore --version`):
* <details><summary>Cloud nodes Environment:</summary>

  * Hardware configuration (e.g. `lscpu`):
  * OS (e.g. `cat /etc/os-release`):
  * Kernel (e.g. `uname -a`):
  * Go version (e.g. `go version`):
  * Others:

  </details>
* <details><summary>Edge nodes Environment:</summary>

  * edgecore version (e.g. `edgecore --version`):
  * Hardware configuration (e.g. `lscpu`):
  * OS (e.g. `cat /etc/os-release`):
  * Kernel (e.g. `uname -a`):
  * Go version (e.g. `go version`):
  * Others:

  </details>

false200:
**What type of PR is this?**

/kind bug

When `watchhook.NewWatchHook` fails in `imitator.Watch`, the function returned `nil`. The caller in `watcher.go` ranges over that channel with `for wres := range wch`, which blocks forever on a nil channel in Go.

This PR closes the channel and returns it on registration failure, so the watch loop exits instead of hanging silently. The upper watcher then terminates with a deterministic clean `ResultChan` close.

Fixes #6975

Registration-failure returns a closed non-nil channel. `startWatching` signals only `watchClosedCh` when the source channel closes, so callers see a clean close rather than a race with `watch.Error`. Regression tests cover both the imitator failure path and the upper watcher closed-source path.

```release-note
Fix metaserver watch hang when hook registration fails by returning a closed channel and terminating the watch cleanly.
```

### DIFF
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go
@@ -201,13 +201,21 @@ func (s *imitator) SetRevision(version interface{}) {
 	}
 }
 
+// registerWatchHook registers the watch hook and can be replaced in tests
+// to exercise the registration-failure path.
+var registerWatchHook = watchhook.NewWatchHook
+
 func (s *imitator) Watch(ctx context.Context, key string, rev uint64) <-chan watch.Event {
 	wch := make(chan watch.Event)
 	receiver := watchhook.NewChanReceiver(wch)
-	wh, err := watchhook.NewWatchHook(key, rev, receiver)
+	wh, err := registerWatchHook(key, rev, receiver)
 	if err != nil {
+		// The detailed error is only logged because this interface cannot
+		// return it. Return a closed non-nil channel so the caller observes
+		// a clean watch termination instead of blocking forever.
 		klog.Errorf("add hook for %s failed, %v", key, err)
-		return nil
+		close(wch)
+		return wch
 	}
 
 	go func() {
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/watch_test.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/watch_test.go
@@ -1,0 +1,33 @@
+package imitator
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/watchhook"
+)
+
+func TestWatchReturnsClosedChannelOnRegistrationFailure(t *testing.T) {
+	original := registerWatchHook
+	registerWatchHook = func(key string, rev uint64, receiver watchhook.Receiver) (*watchhook.WatchHook, error) {
+		return nil, errors.New("hook registration failed")
+	}
+	defer func() { registerWatchHook = original }()
+
+	s := &imitator{}
+	ch := s.Watch(context.TODO(), "/core/v1/pods/default/foo", 1)
+	if ch == nil {
+		t.Fatal("Watch returned a nil channel on registration failure")
+	}
+
+	select {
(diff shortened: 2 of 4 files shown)


DoisLONG:
Thanks for fixing the nil-channel return. Returning a closed channel is safe in this branch because `NewWatchHook` failed before the hook was registered and before any producer or cleanup goroutine was started. It also correctly prevents the caller's `for range` from blocking forever.

There are two details to address before this behavior is complete.

### 1. Define and test the error semantics

`Client.Watch` cannot return an error, so the original hook-registration error is currently only logged.

When the closed channel reaches `startWatching`, the code sends a generic error to `errChan` and then closes `watchClosedCh`. `run` selects between both signals, so the client may either receive a generic `watch.Error` or observe `ResultChan` closing without an error.

Please decide which behavior is intended:

* if registration failure should appear as a clean watch termination, document that the detailed error is available only in the EdgeCore log;
* if the client should reliably receive a watch error, avoid signaling both the error and normal-close paths for the same failure.

Changing `Client.Watch` to return `(<-chan watch.Event, error)` would preserve the original error, although that can be a separate interface-level improvement.

### 2. Add regression coverage

Please add tests proving that:

* the registration-failure path never returns a nil channel;
* receiving from the returned channel completes immediately with `ok == false`;
* an upper-level watcher using a closed source channel terminates within a bounded time and does not leak goroutines;
* the expected `ResultChan` error or clean-close behavior is deterministic.

The existing fake client already exposes `WatchF`, so the upper-level watcher behavior can be tested by returning a pre-closed channel without relying on an actual UUID collision.

The two-line fix removes the permanent hang, but the final caller-visible behavior should be deterministic and covered by tests.

false200:
@DoisLONG, I've addressed your feedback on error semantics and regression coverage; PTAL when you have a moment. Thanks!

DoisLONG (commented):
Thanks for the update. This addresses my previous concern: the registration-failure path now returns a closed non-nil channel, and the upper-level watcher behavior is deterministic clean close instead of racing between `watch.Error` and normal close.

Two small cleanups before LGTM:

1. Please close the `release-note` block and use a short bugfix release note instead of `NONE`.
2. There are several fix-up commits in this PR. Please squash them into one clean commit before merge.

false200:
I've updated the requested changes. Thanks.

DoisLONG (commented):
/lgtm

PR #6969, opened 2026-07-03
test(common): achieve 100% statement coverage for common/message

Norway-02:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**:
Ensures complete test coverage of message parsing and construction utilities used across KubeEdge modules.

**Which issue(s) this PR fixes**:
NONE

**Special notes for your reviewer**:
None

**Does this PR introduce a user-facing change?**:
```release-note
Fix a production bug in ParseResourceEdge precedence.
```

PR #6981, opened 2026-07-06
test(util/config): add unit tests for certificate utility functions

Norway-02:
**What type of PR is this?**
/kind test

Adds tests for certificate config utilities in `pkg/util/config`.

NONE

None

```release-note
NONE
```

### DIFF
--- a/edge/pkg/common/util/config.go
+++ b/edge/pkg/common/util/config.go
@@ -26,6 +26,8 @@ import (
 	"os"
 	"time"
 
+	"path/filepath"
+
 	certutil "k8s.io/client-go/util/cert"
 	klog "k8s.io/klog/v2"
 )
@@ -70,7 +72,7 @@ func GenerateTestCertificate(path string, certFileName string, keyFileName strin
 	certFilePEM := pem.Block{
 		Type:  certutil.CertificateBlockType,
 		Bytes: cert}
-	err = createPEMfile(path+certFileName+".crt", certFilePEM)
+	err = createPEMfile(filepath.Join(path, certFileName+".crt"), certFilePEM)
 	if err != nil {
 		return err
 	}
@@ -79,7 +81,7 @@ func GenerateTestCertificate(path string, certFileName string, keyFileName strin
 		Bytes: pKey,
 	}
 
-	return createPEMfile(path+keyFileName+".key", keyFilePEM)
+	return createPEMfile(filepath.Join(path, keyFileName+".key"), keyFilePEM)
 }
 
 // createPEMfile creates an encoded file at the path given, with PEM Block specified
--- a/edge/pkg/common/util/config_test.go
+++ b/edge/pkg/common/util/config_test.go
@@ -1,0 +1,103 @@
+package util
+
+import (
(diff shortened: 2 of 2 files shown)


DoisLONG:
Thanks for adding coverage for the certificate utility functions. The current PR needs to be cleaned up and strengthened before it is ready.

### 1. Rebase this PR directly onto `master`

This branch contains all changes from the still-open #6969:

* `edge/pkg/common/message/message_test.go`
* `edge/pkg/common/message/nodetask_test.go`

and then adds `config_test.go` as a second commit.

Those message-package changes are unrelated to the stated scope of this PR and are already being reviewed separately in #6969. Please create a clean branch from the latest `master` and cherry-pick only the certificate-test commit.

This is important because the carried changes also remove an existing license header, weaken existing error-message assertions, and include unresolved behavioral issues from #6969.

### 2. Test the documented directory-path behavior

The success test currently calls:

```go id="x5rk4i"
GenerateTestCertificate(tmp+"/", "cert", "key")
```

This hides a defect in the implementation, which concatenates the directory and filename as strings.

According to the function documentation, callers should be able to pass a directory path without manually adding a trailing separator.

Please change the test to call:

```go id="qm4g9w"
GenerateTestCertificate(tmp, "cert", "key")
```

and verify the files at:

```go id="u03o0s"
filepath.Join(tmp, "cert.crt")
filepath.Join(tmp, "key.key")
```

The production implementation should use `filepath.Join` when constructing both paths.

### 3. Verify the generated certificate and key, not only file size

A non-empty file does not prove that the output is a valid or matching certificate/key pair.

Please extend the success test to:

* decode both PEM files;
* parse the certificate with `x509.ParseCertificate`;
* parse the private key with `x509.ParsePKCS1PrivateKey`;
* verify that the certificate public key matches the private key;
* load the pair with `tls.LoadX509KeyPair`;
* verify the expected certificate properties, such as `IsCA`, DNS names, usages, and validity period.

### 4. Verify the exact PEM block

`TestCreatePEMFileSuccess` should read the generated file, call `pem.Decode`, and assert:

* no trailing undecoded data;
* the expected block type;
* the exact original bytes.

Checking only that the file size is greater than zero would still pass for malformed or incorrect output.

### 5. Use an appropriate private-key file mode

`createPEMfile` currently uses `os.Create`, so the RSA private-key file may be created as `0644` depending on the process umask.

Please avoid making private keys world-readable. The key should be created with `0600`; certificate files may use a less restrictive mode if needed. Add permission assertions to the tests.

### 6. Repository cleanup

Please also:

* add the standard KubeEdge Apache 2.0 license header to the new test file;
* remove the `Fixes #<issue-number>` placeholder;
* consider a scope-specific title such as `test: add certificate utility coverage`;
* provide the focused test command and result after rebasing.

The `common/message` changes should continue only in #6969. For reference, its current `nodetask` tests call the functions without asserting the emitted messages, and `InitContext` alone does not register the EdgeHub module, so `beehiveContext.Send` can return without delivering anything.

Norway-02:
Hi @doislong and @WillardHu,

PR #6986, opened 2026-07-06
servicebus: serve traffic over TLS

### LINKED ISSUE #6985: ServiceBus server is exposed over plain HTTP with no TLS
## What happened

The ServiceBus server starts an `http.Server` with `ListenAndServe()` and does not support TLS. In `edge/pkg/servicebus/servicebus.go:234-252`, the listener is created without TLS configuration, and the code includes a TODO that TLS should be added later.

ServiceBus should support TLS-secured serving with configurable certificate and key paths, and it should fail validation when secure listener configuration is incomplete.

## How to reproduce it (as minimally and precisely as possible)

1. Open `edge/pkg/servicebus/servicebus.go`.
2. Review the server startup path around lines `234-252`.
3. Confirm the code calls `ListenAndServe()` and does not load TLS materials.
4. Review command: `nl -ba edge/pkg/servicebus/servicebus.go | sed -n '224,252p'`

## Anything else we need to know?

- Priority: P1
- Category: Security
- Why it matters: ServiceBus is an ingress path into edge-hosted services. Plain HTTP leaves traffic vulnerable to interception and tampering on shared or untrusted networks.
- Impact and likelihood: Impact high, likelihood medium to high in real edge deployments where local networks are often shared or physically exposed.
- Validation methods:
  - Static review confirming only `ListenAndServe()` is used and no TLS configuration exists.
  - Runtime probe in a non-production environment showing the endpoint negotiates plaintext HTTP rather than HTTPS.
- Suggested fix: Add TLS configuration, certificate and key loading, and validation for secure serving paths. Reuse the standard edge certificate defaults where possible.
- Estimated effort: M

## Environment

- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): Affected current repository state; exact release not yet bounded

  - Others: Not cloud-node specific.

  </details>

  - edgecore version (e.g. `edgecore --version`): Affected current repository state; exact release not yet bounded
  - OS (e.g. `cat /etc/os-release`): Any OS that runs ServiceBus
  - Others: Static code-path issue in the ServiceBus listener.

  </details>

rootp1:
/kind bug

Moves the edge ServiceBus listener off plaintext HTTP by serving with TLS and wiring certificate configuration through the edgecore ServiceBus config. The change also validates the configured cert and key paths in both edgecore config API versions so misconfigured secure listeners fail early.

Which issue(s) this PR fixes:
Fixes #6985

Verified with:
- `go test ./staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha1/validation ./staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/validation ./edge/pkg/servicebus`
- `make verify`

```release-note
The edge ServiceBus listener now serves HTTPS using configured TLS certificate and key files, and edgecore validation now rejects enabled ServiceBus configs with missing TLS listener credentials.
```

### DIFF
--- a/edge/pkg/servicebus/servicebus.go
+++ b/edge/pkg/servicebus/servicebus.go
@@ -2,11 +2,16 @@ package servicebus
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -24,11 +29,14 @@ import (
 	servicebusConfig "github.com/kubeedge/kubeedge/edge/pkg/servicebus/config"
 	"github.com/kubeedge/kubeedge/edge/pkg/servicebus/util"
 	"github.com/kubeedge/kubeedge/pkg/features"
+	utilvalidation "github.com/kubeedge/kubeedge/pkg/util/validation"
 )
 
 var (
 	inited int32
-	c      = make(chan struct{})
+
+	serverMu sync.Mutex
+	active   *http.Server
 )
 
 const (
@@ -36,10 +44,8 @@ const (
 	maxBodySize = 5 * 1e6
 )
 
-// servicebus struct
 type servicebus struct {
-	enable bool
-	// default 127.0.0.1
+	enable  bool
 	server  string
 	port    int
 	timeout int
@@ -71,7 +77,6 @@ func newServicebus(enable bool, server string, port, timeout int) *servicebus {
 	}
 }
 
-// Register register servicebus
 func Register(s *v1alpha2.ServiceBus) {
 	servicebusConfig.InitConfigure(s)
 	core.Register(newServicebus(s.Enable, s.Server, s.Port, s.Timeout))
@@ -100,19 +105,16 @@ func (sb *servicebus) RestartPolicy() *core.ModuleRestartPolicy {
 }
 
 func (sb *servicebus) Start() {
-	// no need to call TopicInit now, we have fixed topic
 	htc.Timeout = time.Second * 10
 	uc.Client = htc
 	if !sb.sbs.IsTableEmpty() {
-		if atomic.CompareAndSwapInt32(&inited, 0, 1) {
-			go server(c)
-		}
+		startServer()
 	}
-	//Get message from channel
 	for {
 		select {
 		case <-beehiveContext.Done():
 			klog.Warning("servicebus stop")
+			stopServer()
 			return
 		default:
 		}
@@ -122,7 +124,6 @@ func (sb *servicebus) Start() {
 			continue
 		}
 
-		// build new message with required field & send message to servicebus
 		klog.V(4).Info("servicebus receive msg")
 		go processMessage(&msg)
 	}
@@ -138,20 +139,16 @@ func processMessage(msg *beehiveModel.Message) {
 	switch msg.GetOperation() {
 	case message.OperationStart:
 		if err := dbc.InsertUrls(resource); err != nil {
-			// TODO: handle err
 			klog.Error(err)
 		}
-		if atomic.CompareAndSwapInt32(&inited, 0, 1) {
-			go server(c)
-		}
+		startServer()
 	case message.OperationStop:
 		if err := dbc.DeleteUrlsByKey(resource); err != nil {
-			// TODO: handle err
 			klog.Error(err)
 		}
 
 		if dbc.IsTableEmpty() {
-			c <- struct{}{}
+			stopServer()
 		}
 	default:
 		r := strings.Split(resource, ":")
@@ -186,7 +183,6 @@ func processMessage(msg *beehiveModel.Message) {
 			return
 		}
 
-		//send message with resource to the edge part
 		operation := httpRequest.Method
 		targetURL := "http://127.0.0.1:" + r[0] + r[1]
 		resp, err := uc.HTTPDo(operation, targetURL, httpRequest.Header, httpRequest.Body)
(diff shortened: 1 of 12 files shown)


rootp1:
Addressed the Gemini review point. The ServiceBus HTTPS listener now sets `TLSConfig.MinVersion` to `tls.VersionTLS12` and configures `ReadHeaderTimeout` to `10 * time.Second` to reduce exposure to weak protocol negotiation and Slowloris style header starvation. Verification completed with `go test ./edge/pkg/servicebus ./staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha1/validation ./staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/validation` and `make verify`.

DoisLONG:
Thanks for adding TLS 1.2, a header-read timeout, certificate configuration, and normal `ErrServerClosed` handling. Moving the ServiceBus listener away from plaintext HTTP is the right direction.

There are several blocking issues in the current certificate and listener lifecycle design.

### 1. The default certificate is not a valid ServiceBus server certificate

The new defaults reuse:

`/etc/kubeedge/certs/server.crt`
`/etc/kubeedge/certs/server.key`

These files are the EdgeHub node identity certificate and key.

The EdgeHub CSR uses `CN=system:node:<nodeName>` and does not request DNS or IP SANs. When no explicit usages are supplied, CloudCore signs the certificate with `ExtKeyUsageClientAuth` only.

The ServiceBus default address is `127.0.0.1`, so a normal TLS client verifying the server will reject the default certificate because:

* it is not valid for ServerAuth;
* it has no `127.0.0.1` IP SAN;
* its CommonName does not identify the ServiceBus endpoint.

Using `InsecureSkipVerify` would remove server authentication and would not address the interception and tampering threat described in #6985.

Please use a dedicated ServiceBus server certificate with ServerAuth and the appropriate DNS/IP SANs instead of reusing the EdgeHub client identity.

### 2. The validation order breaks first-time bootstrap

EdgeCore validates the full configuration before registering or starting modules. The default EdgeHub certificate is obtained later, when `CertManager.Start` runs inside the EdgeHub module.

If ServiceBus is enabled during the first bootstrap and uses the new default certificate paths, configuration validation fails before EdgeHub has an opportunity to create those files.

Please define a provisioning flow that works on the first start. For example, `keadm` could provision a dedicated ServiceBus certificate, or users could be required to provide an existing dedicated certificate explicitly.

The current defaults cannot rely on a certificate that is generated only after configuration validation.

### 3. Validate the key pair and certificate semantics, not only path existence

`FileIsExist` only calls `os.Stat`. The current “valid files” test creates two empty files and expects validation success.

Please validate at least:

* both paths are regular files;
* `tls.LoadX509KeyPair` succeeds;
* the key matches the certificate;
* the leaf certificate parses;
* the certificate is currently valid;
* the certificate supports ServerAuth;
* the certificate SAN policy is compatible with the configured endpoint;
* private-key permissions are appropriately restricted.

The current validation allows empty files, directories, mismatched pairs, expired certificates, and ClientAuth-only certificates to reach runtime startup.

### 4. Reset the server state on every startup failure

`inited` is set before the TLS listener goroutine starts. It is reset only by the shutdown goroutine after receiving `stopChan`.

If certificate loading, binding, or listener startup fails, `serveWithTLS` returns, but:

* `inited` remains `1`;
* the shutdown goroutine remains blocked;
* later `OperationStart` messages cannot retry the listener.

Please ensure every server exit path resets the initialization state and terminates its cleanup goroutine.

Ideally, load the certificate and bind the listener successfully before committing the initialized state.

Please add a regression test that:

1. starts with an invalid certificate configuration;
2. confirms startup failure;
3. replaces it with a valid certificate;
4. successfully starts the listener on the second attempt.

### 5. Handle certificate rotation

The default EdgeHub certificate files are overwritten during certificate rotation. EdgeHub responds by reconnecting, but ServiceBus has no `GetCertificate`, rotation notification, or controlled listener restart.

A running ServiceBus listener will therefore continue serving the previously loaded certificate after the files have rotated.

Please either:

* implement dynamic certificate reload;
* restart the listener safely when the certificate changes;
* or use a dedicated ServiceBus certificate and document the required restart and rotation procedure.

### 6. Clarify or add client authentication

The current TLS configuration provides server-side encryption only. It does not configure `ClientAuth` or `ClientCAs`, and the HTTP handler has no application-level authentication.

If ServiceBus is bound to a non-loopback address, any reachable client can establish a TLS connection and invoke an allowed ServiceBus route.

Please either:

* restrict the listener to loopback or Unix-socket use;
* add mTLS or another client-authentication mechanism;
* or explicitly narrow the security claim to transport encryption only.

### 7. Document the mandatory HTTPS migration

An enabled ServiceBus now always uses HTTPS, so existing applications calling `http://127.0.0.1:9060` will stop working.

Please document:

* certificate provisioning;
* client CA trust configuration;
* the required server name or IP;
* migration from HTTP;
* first-bootstrap behavior;
* certificate rotation;
* whether any compatibility mode is supported.

Please also update the API comments that still describe the listener as an HTTP server.

### 8. Add real TLS integration coverage

Please add a ServiceBus TLS test using a generated CA and a ServerAuth certificate with a matching SAN.

The8. Add real TLS integration coverage

Please add a ServiceBus TLS test using a generated CA and a ServerAuth certificate with a matching SAN.

The tests should verify:

* a trusted HTTPS client succeeds;
* plaintext HTTP does not work;
* a wrong CA or hostname fails;
* invalid and mismatched key pairs fail validation;
* startup failure does not wedge `inited`;
* shutdown and restart work;
* rotated certificates are served to new connections.

The transport change is valuable, but the default certificate must be verifiable as a server identity, and the listener lifecycle must recover safely from TLS startup failures.

rootp1:
@DoisLONG Thanks, addressed this round.

I reworked the ServiceBus TLS path around an explicit server-certificate contract instead of reusing the EdgeHub client identity:
- ServiceBus no longer defaults to the EdgeHub cert/key paths
- enabling ServiceBus now requires an explicit cert/key pair
- validation now checks a real key pair, leaf parsing, validity window, ServerAuth usage, SAN/hostname compatibility, regular-file constraints, and restricted key permissions
- the listener is restricted to loopback addresses while ServiceBus has no client-auth layer, which narrows the security scope to local transport protection

I also changed the listener lifecycle so startup failures do not wedge the module:
- the TLS config and listener are created before the server is marked initialized
- failed startup leaves `inited` clear so a later retry can succeed
- shutdown no longer depends on the blocked goroutine pattern from the previous `stopChan` flow
- the TLS config now reloads the certificate from disk on new handshakes, so rotated files are served to new connections without requiring stale in-memory cert reuse

Coverage was expanded with targeted tests for:
- invalid ServiceBus cert/key validation
- rejecting client-auth-only / SAN-mismatched certs
- retry after startup failure
- trusted HTTPS success
- plaintext requests not reaching the HTTP handler
- wrong CA / wrong hostname failure
- shutdown / restart
- certificate rotation serving the new leaf to subsequent connections

Verified locally with:
- `go test ./edge/pkg/servicebus`
- `go test ./staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha1/validation`
- `go test ./staging/src/github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/validation`
- `go test ./pkg/util/validation`
- `go test ./staging/src/github.com/kubeedge/api/apis/util/validation`

rootp1:
@DoisLONG can you please have a look whenever you have a chance to?

DoisLONG (commented):
Thanks for the updates. 

The earlier security feedback about TLS defaults has been addressed in the current version: the server now sets `MinVersion: tls.VersionTLS12`, and `ReadHeaderTimeout` is configured. The new lifecycle / rotation tests also give better coverage than the initial version.

The main concern is scope and compatibility. This PR overlaps heavily with #6922, which is also adding ServiceBus TLS support. The two PRs take different approaches: #6922 keeps TLS optional and preserves the existing plain-HTTP behavior by default, while this PR makes TLS credentials required whenever ServiceBus is enabled. That means existing deployments with `serviceBus.enable=true` but no TLS files would start failing validation after upgrade.

If the project wants to make ServiceBus TLS mandatory, that should be an explicit maintainer decision with migration documentation, release-note wording, and possibly a default certificate generation / configuration path. Otherwise, the safer compatibility-preserving approach is to make TLS optional when `tlsCertFile` and `tlsPrivateKeyFile` are empty.

There is also an implementation issue in the current code: `newTLSServer(cfg)` receives a `ServiceBus` config, but `loadCertificate()` derives the host from the global `servicebusConfig.Config.Server` instead of the passed config. That makes certificate validation depend on global state. Please pass the parsed host from `cfg.Server` into `loadTLSConfig()` / `loadCertificate()` instead.

Another design point to clarify is the new loopback-only restriction. I understand the reason: this PR adds server TLS but not caller authentication, so binding to non-loopback can expose ServiceBus to other machines. But this still changes existing ServiceBus binding semantics, so it should be called out explicitly or handled as part of a broader ServiceBus security model.

So from my side, the previous TLS-default review comments look addressed, but the larger compatibility and overlap questions are still unresolved. Before continuing this PR, I think we should first decide whether #6922 or this PR is the base direction for ServiceBus TLS.

rootp1:
@DoisLONG thanks for pointing that out. Fixed the global state issue, loadCertificate and loadTLSConfig now take the server host as a parameter and it is threaded through from the config passed into newTLSServer, so certificate validation no longer reaches into servicebusConfig.Config.Server directly. Pushed that change along with the existing test suite passing.

On the bigger question of overlap with #6922 and whether TLS should be mandatory or optional, that is a call for a maintainer to make since it affects existing deployments and the overall ServiceBus security direction. I do not want to guess at that and rework the PR twice, so I will wait for guidance on which approach the project wants to take as the base direction before making further scope changes here.

On the loopback only restriction, happy to document that explicitly as part of this change or fold it into a broader security writeup, whichever you and the maintainers prefer.

DoisLONG (commented):
I took another look at this PR together with the latest state of #6922.

This PR has improved a lot from the initial version. The loopback restriction, TLS file validation, TLS 1.2+, `ReadHeaderTimeout`, certificate reload through `GetCertificate`, and the server lifecycle cleanup are all useful pieces.

However, I think we should not continue carrying this as a separate ServiceBus TLS implementation while #6922 is already close to landing.

The main reason is still the compatibility difference. This PR makes TLS effectively mandatory whenever ServiceBus is enabled, because enabled ServiceBus configs must provide `tlsCertFile` and `tlsPrivateKeyFile`. #6922 takes the optional TLS direction instead: existing plain-HTTP ServiceBus deployments keep working by default, and users can opt in to HTTPS by configuring the cert/key files.

Given the compatibility impact, I think #6922 is the safer base direction unless maintainers explicitly decide that ServiceBus TLS should become mandatory as a breaking behavior change.

If there are still useful improvements from this PR, such as stricter validation or lifecycle cleanup, it may be better to rework them on top of #6922 after the base optional-TLS implementation is settled. Otherwise, continuing two overlapping TLS implementations will make the review and merge direction harder to reason about.

I suggest first aligning with #6922 or confirming from other maintainers that mandatory TLS is intentionally preferred.

rootp1:
@DoisLONG circling back here. The certificate handling, loopback restriction, lifecycle fixes and the global-state cleanup you asked for are all in from my side, and I squashed the history into one commit. The open question that's left is a product one, whether ServiceBus TLS should be mandatory (this PR's approach) or optional (the direction in #6922), so I'll hold off on further scope changes until that's decided. Everything else I can act on is done, happy to take another pass whenever you have time.

rootp1:
@DoisLONG I know this one still has the open design question, but wanted to mention that the code itself has been reviewed several times and is thoroughly tested on my end. If the direction gets settled, I would appreciate help getting this merged soon after.

rootp1:
@DoisLONG thanks for the detailed review. I agree the bigger question here is the project direction around ServiceBus TLS. I will keep this PR parked until maintainers decide whether the base path should follow the optional TLS approach or the mandatory TLS approach. If the decision is to build on the optional path, I can move the validation and lifecycle improvements over there instead of carrying two parallel implementations.

DoisLONG (commented):
I took another look at this PR together with the latest state of #6922.

This PR has improved a lot technically, and I can see that it now includes useful pieces such as TLS 1.2+, `ReadHeaderTimeout`, loopback binding, certificate validation, certificate reload, and server lifecycle cleanup.

However, I still do not think we should continue this as a separate ServiceBus TLS implementation while #6922 is already close to settling the optional-TLS direction.

The key difference is compatibility. This PR makes TLS effectively mandatory for enabled ServiceBus configs by rejecting configurations without TLS listener credentials. #6922 keeps TLS optional instead: existing plain-HTTP ServiceBus deployments continue to work by default, and users can opt in to HTTPS by configuring the cert/key files.

Given the compatibility impact, I think #6922 is the safer base direction unless maintainers explicitly decide that ServiceBus TLS should become mandatory as a breaking behavior change.

If there are still useful pieces from this PR, it may be better to rework them on top of #6922 after the optional-TLS base implementation is settled, rather than carrying two overlapping implementations.

I suggest aligning it with #6922, or getting maintainer confirmation that mandatory TLS is intentionally preferred.

rootp1:
@DoisLONG all the technical points you raised are addressed. The remaining open item is the mandatory versus optional TLS direction versus #6922, which needs a call from kadisi rather than another code change on my end. Following up in case that decision can move forward.

PR #6988, opened 2026-07-08
test(servicebus): add unit test for servicebus config

Norway-02:
**What type of PR is this?**
/kind test

Adds coverage for the config loading in servicebus.

NONE

None

```release-note
NONE
```

### DIFF
--- a/edge/pkg/servicebus/config/config_test.go
+++ b/edge/pkg/servicebus/config/config_test.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2025 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


DoisLONG:
Thanks for adding coverage for the servicebus config package.

One thing I think should be adjusted: `InitConfigure()` mutates package-level state through both `Config` and `once`. Since `once.Do()` only allows the initialization to run once, this test can affect later tests in the same package, and later tests may also affect this one depending on test order.

Could you reset and restore the package-level state in the test? For example, save the old `Config` and `once`, set `once = sync.Once{}` before calling `InitConfigure`, and restore both in a deferred cleanup.

The existing inline suggestion to compare the whole `ServiceBus` struct also looks reasonable, since it would make the test cover future fields automatically.

Also, please fix the PR description release-note block; it looks like the closing triple backticks are missing.

Norway-02:
@DoisLONG 

Thanks for the review! I have addressed all your feedback:

1. **State Reset & Restore**: Updated `TestInitConfigure` to save, reset, and restore the package-level variables `Config` and `once` using deferred cleanup to ensure test isolation.
2. **Struct Assertion**: Changed assertions to compare the entire `ServiceBus` struct directly to cover future fields automatically.
3. **PR Description**: Added the missing closing triple backticks to the release-note block in the PR description.

The updated, signed-off commit has been pushed. Please let me know if it looks good!

Norway-02:
Hi @doislong and @WillardHu,

PR #6991, opened 2026-07-09
fix(taskmanager): resolve shadowed error variables in task loop

Norway-02:
**What type of PR is this?**
/kind bug

Several error variables in the task manager loop were unintentionally shadowed by inner scope assignments. This caused genuine failures to be ignored, leading to incorrect task state reporting. This PR resolves the variable shadowing, ensuring errors are correctly captured and propagated.

NONE

None

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/taskmanager/downstream/config_update.go
+++ b/cloud/pkg/taskmanager/downstream/config_update.go
@@ -128,5 +128,5 @@ func (h *ConfigUpdateJobHandler) UpdateNodeTaskStatus(
 			}
 		},
 	}
-	status.GetConfigeUpdateJobStatusUpdater().UpdateStatus(opts)
+	status.GetConfigUpdateJobStatusUpdater().UpdateStatus(opts)
 }
--- a/cloud/pkg/taskmanager/status/init.go
+++ b/cloud/pkg/taskmanager/status/init.go
@@ -47,6 +47,6 @@ func GetNodeUpgradeJobStatusUpdater() *StatusUpdater {
 	return nodeUpgradeJobStatusUpdater
 }
 
-func GetConfigeUpdateJobStatusUpdater() *StatusUpdater {
+func GetConfigUpdateJobStatusUpdater() *StatusUpdater {
 	return configUpdateJobStatusUpdater
 }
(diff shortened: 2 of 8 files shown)


DoisLONG:
Thanks for submitting this fix. The issue itself looks valid, and the PR is pointing at the right area.

Before going deeper into the code review, I’d like to clarify the coordination around #6916 so we avoid duplicate work.

#6916 was originally opened by @g-k-s-03, and it is currently assigned to @Anshumancanrock. Since this PR is opened by another contributor, could you please confirm whether you have coordinated with them before working on this fix?

@g-k-s-03 @Anshumancanrock, could you also confirm whether you are still working on this issue, and whether you are okay with continuing the fix through this PR?

Once the ownership is clear, we can continue reviewing the implementation details.

Norway-02:

I noticed that the issue has been open for a while without an active pull request, so I went ahead and implemented this fix to resolve the bug, along with adding unit tests to prevent future regressions. 

@g-k-s-03 @Anshumancanrock, if either of you are currently working on a fix or have a different implementation in progress, please let me know. Otherwise, I would be glad to coordinate and proceed with this PR to get the issue resolved.

Norway-02:
Hi @doislong and @WillardHu,

DoisLONG (commented):
Thanks for working on this.

The issue itself is valid, and the direction of fixing the callback error shadowing is correct.

However, I noticed that #6938 was opened earlier for the same issue, and it is already linked to #6916. Since #6916 was originally opened by @g-k-s-03 and #6938 is from the same author, I think it would be better to continue the fix there to avoid duplicate PRs.

I left a review on #6938 asking for failure-path regression tests, which is also the useful part this PR is adding.

Could you please coordinate with the author of #6938 and the issue assignee? If #6938 continues, this PR can probably be closed, or the test ideas here can be moved into #6938.

g-k-s-03:
@DoisLONG I opened #6916 but haven't started implementing a fix — no PR of my own in progress.
No objection to either #69991 or #6938  continuing; happy to see whichever one the maintainers
prefer move forward.

PR #6992, opened 2026-07-09
Add resourceclaims/binding RBAC for DRA granular status authorization

shwetha-s-poojary:
**What type of PR is this?**

Adds `update` and `patch` permissions on `resource.k8s.io/resourceclaims/binding`
to the `system:kube-scheduler` ClusterRole in the vendored bootstrap RBAC policy.

In Kubernetes v1.36, the `DRAResourceClaimGranularStatusAuthorization` feature gate
is beta and on by default. It introduces a dedicated `resourceclaims/binding`
subresource to separate scheduler binding authorization from `resourceclaims/status`.
Without this permission, the scheduler cannot perform ResourceClaim binding operations
on v1.36+ clusters with DynamicResourceAllocation enabled.

The permission is inert on clusters older than v1.33 where the subresource does not
exist, so it is safe to add now in preparation.

 https://github.com/kubernetes/kubernetes/issues/138149https://github.com/kubernetes/kubernetes/issues/138149

The `DRAResourceClaimGranularStatusAuthorization` feature gate constant does not exist
in the vendored Kubernetes v1.32. The rule is therefore placed flat inside the existing
`DynamicResourceAllocation` block, which is the correct outer guard. When the vendored
Kubernetes is bumped to v1.33+, this can be moved under its own feature gate check.

```release-note
Add resourceclaims/binding RBAC for DRA granular status authorization to the
system:kube-scheduler ClusterRole. Required for Kubernetes v1.36 clusters with
DynamicResourceAllocation enabled.
```

### DIFF
--- a/vendor/k8s.io/kubernetes/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
+++ b/vendor/k8s.io/kubernetes/plugin/pkg/auth/authorizer/rbac/bootstrappolicy/policy.go
@@ -603,6 +603,10 @@ func ClusterRoles() []rbacv1.ClusterRole {
 			rbacv1helpers.NewRule(Read...).Groups(resourceGroup).Resources("deviceclasses").RuleOrDie(),
 			rbacv1helpers.NewRule(ReadUpdate...).Groups(resourceGroup).Resources("resourceclaims").RuleOrDie(),
 			rbacv1helpers.NewRule(ReadUpdate...).Groups(resourceGroup).Resources("resourceclaims/status").RuleOrDie(),
+			// Needed for DRAResourceClaimGranularStatusAuthorization (beta in k8s v1.36, on by default).
+			// The resourceclaims/binding subresource is inert on clusters older than v1.33 but safe to add in preparation.
+			// Ref: kubernetes/kubernetes#138149
+			rbacv1helpers.NewRule("update", "patch").Groups(resourceGroup).Resources("resourceclaims/binding").RuleOrDie(),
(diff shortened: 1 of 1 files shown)


DoisLONG:
Thanks for working on the Kubernetes DRA authorization migration. The `update` and `patch` permissions on `resourceclaims/binding` look consistent with the upstream Kubernetes scheduler bootstrap policy.

However, I do not think this patch has an effective runtime path in KubeEdge.

From what I can see, KubeEdge imports the vendored `bootstrappolicy` package for the CloudHub node authorizer, but the relevant call path uses:

```go
bootstrappolicy.NodeRules()
```

This PR modifies `bootstrappolicy.ClusterRoles()`. KubeEdge does not appear to call that function to create or reconcile the `system:kube-scheduler` ClusterRole, and I could not find a KubeEdge-owned scheduler or scheduler RBAC manifest that would consume this change.

So even if the RBAC rule itself is correct upstream, changing this vendored file would not update scheduler RBAC in a real KubeEdge deployment. It would also make the vendored Kubernetes policy diverge from its upstream source and could be overwritten by future vendor regeneration.

For these reasons, I do not think this patch is applicable to KubeEdge as-is, and I suggest closing this PR.

If there is a concrete KubeEdge component or deployment path that creates the `system:kube-scheduler` role or performs ResourceClaim binding through a custom scheduler, please point it out and we can reconsider the appropriate KubeEdge-owned RBAC manifest or code path to update.

shwetha-s-poojary:
@DoisLONG , closing this. The change has no runtime effect and wouldn't survive a vendor regeneration anyway. Appreciate you taking the time to trace the actual call path.

PR #6993, opened 2026-07-09
test(cloudcore): add comprehensive unit tests for cloudcore app and fix nil map panic

Norway-02:
**What type of PR is this?**
/kind test
/kind bug

1. Covers edge cases in the CloudCore app lifecycle.
2. Resolves a `nil` map initialization panic found during testing.

NONE

Added the `kind/bug` label due to the production fix included alongside the tests.

```release-note
Fix nil map panic and incomplete tunnel-port record handling in CloudCore.
```

### DIFF
--- a/cloud/cmd/cloudcore/app/server.go
+++ b/cloud/cmd/cloudcore/app/server.go
@@ -208,6 +208,13 @@ func NegotiateTunnelPort() (*int, error) {
 			return nil, err
 		}
 
+		if record.IPTunnelPort == nil {
+			record.IPTunnelPort = make(map[string]int)
+		}
+		if record.Port == nil {
+			record.Port = make(map[int]bool)
+		}
+
 		port, found := record.IPTunnelPort[localIP]
 		if found {
 			return &port, nil
@@ -292,10 +299,13 @@ func updateCloudCoreConfigMap(c *v1alpha1.CloudCoreConfig) {
 		return
 	}
 
+	if cloudCoreCM.Data == nil {
+		cloudCoreCM.Data = make(map[string]string)
+	}
 	cloudCoreCM.Data["cloudcore.yaml"] = string(configBytes)
 
 	_, err = kubeClient.CoreV1().ConfigMaps(constants.SystemNamespace).Update(context.TODO(), cloudCoreCM, metav1.UpdateOptions{})
 	if err != nil {
-		klog.Errorf("Failed to marshal cloudcore config: %v", err)
+		klog.Errorf("Failed to update CloudCore configMap: %v", err)
 	}
 }
--- a/cloud/cmd/cloudcore/app/server_test.go
+++ b/cloud/cmd/cloudcore/app/server_test.go
@@ -1,15 +1,38 @@
+/*
+Copyright 2019 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package app
 
 import (
+	"context"
+	"errors"
 	"reflect"
 	"testing"
 
 	"github.com/agiledragon/gomonkey/v2"
+	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	fakekube "k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+	"sigs.k8s.io/yaml"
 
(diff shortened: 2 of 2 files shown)


Norway-02:
@DoisLONG

DoisLONG:
Thanks for expanding the test coverage around CloudCore port negotiation and ConfigMap updates. The added fake-client reactors are useful, and the new edge-case subtests are a good direction.

I think there are still a few cases that should be covered:

1. `NegotiateTunnelPort()` should handle valid but incomplete tunnel-port records. For example, after unmarshalling `{}`, or a value missing either `ipTunnelPort` or `port`, one or both maps can remain nil. The current code then writes to `record.IPTunnelPort` and `record.Port`, which can panic with `assignment to entry in nil map`. Please add regression cases for these inputs and initialize both maps after unmarshalling when needed.

2. `updateCloudCoreConfigMap()` should handle a CloudCore ConfigMap with `Data == nil`. The current success test initializes `Data` to an empty map, so it does not cover the nil-map case. Please add this case and initialize `cloudCoreCM.Data` in the production code if necessary.

3. The ConfigMap get/update failure tests currently only assert that the function does not panic. Could you also assert the fake client actions, or use a flag in the reactor, to verify that the expected Get/Update path is actually reached? The Get failure case should also verify that no Update is attempted.

A couple of cleanup items as well:

* The success case only checks that the `cloudcore.yaml` key exists. It would be better to use a representative non-empty config and verify the serialized content.
* The original `TestNegotiateTunnelPort` still shares outer mutable state across subtests. The current test is sequential, so this is not an immediate race, but refactoring it so each subtest owns its fake client and patches would make it much safer.
* The update-failure path logs `Failed to marshal cloudcore config`, but the error is from updating the ConfigMap. Please correct the log message or track it separately.
* Please fix the invalid commit message / release-note formatting issue. The release note should be fenced as:

```release-note
NONE
```

Once the two nil-map cases and the behavior assertions are addressed, this test coverage will be much more robust.

Norway-02:
@DoisLONG 

Thanks for the detailed review! I have addressed all your feedback:

1. **Incomplete Tunnel-Port Records**: Initialized `record.IPTunnelPort` and `record.Port` maps to empty maps if they remain nil after unmarshalling. Also added three regression test cases to verify no panics on empty or partially missing records.
2. **Nil Data Map Initialization**: Added a safety check to initialize `cloudCoreCM.Data` if nil. Added the corresponding unit test verifying updates succeed under nil Data.
3. **Verify Expected Action Paths**: Asserted fake client actions list (`fakeClient.Actions()`) in all subtests to guarantee that updates are not attempted on get failure and that correct get/update path sequence is followed.
4. **YAML Content Serialization**: Configured a mock port in the success test case and verified that the ConfigMap's serialized string exactly matches the expected YAML marshalled content.
5. **Isolate Subtest State**: Refactored `TestNegotiateTunnelPort` to initialize patches, mock clients, and ConfigMaps independently within each subtest execution.
6. **Correct Update Path Log**: Changed log message to reflect Update failure instead of Marshal failure.
7. **PR Description Release-Note**: Fixed formatting on the PR description.

The updated, signed-off commit has been pushed. Please let me know if it looks good!

Norway-02:
Addressed Gemini Code Assist review feedback:

- Fixed gomonkey patching: replaced `ApplyFunc(client.GetKubeClient, func() ...)` with `ApplyFuncReturn(client.GetKubeClient, fakeClient)` — the original approach failed because the compiler inlines short functions, causing the patch to never fire and `kubeClient` to be nil at runtime (panic on line 189)
- Added `// go test -gcflags=all=-l` comment to document the required test invocation, consistent with other gomonkey tests in this repo (e.g. `cloud/pkg/cloudhub/servers/httpserver/node/checknode_test.go`)
- Fixed test isolation in `TestNegotiateTunnelPort`: fakeClient is now constructed before patching rather than inside the closure
- Moved `hostnameOverride`/`localIP` outside the sub-test loop
- Added `tt := tt` range-variable capture in both table-driven loops
- Consistent `gomonkey.NewPatches()` + `defer patches.Reset()` across all sub-tests

All 12 tests pass with `-gcflags=all=-l`.

/cc @wbc6080 @zc2638

Norway-02:
Hi @doislong and @WillardHu,

DoisLONG:
Thanks for the updates.

I went through the latest version again. The main concerns look addressed to me now.

`TestNegotiateTunnelPort` no longer shares the outer ConfigMap / patch state across subtests, the incomplete tunnel-port record maps are initialized after unmarshalling, and `updateCloudCoreConfigMap()` now handles a nil `Data` map. The update-failure log message is also clearer now.

The added tests also look more meaningful now: they cover the empty / partial tunnel-port record cases, ConfigMap create/update failures, and the CloudCore ConfigMap update path now checks the client actions and the serialized `cloudcore.yaml` content instead of only checking for no panic.

One small non-blocking cleanup: since this PR now includes production bug fixes, the release note probably does not need to be `NONE`. A short note about fixing the nil map panic / incomplete tunnel-port record handling would be more accurate.

The code changes look good to me.

/lgtm

Norway-02:
@DoisLONG Thanks for the review and the `/lgtm` tag! I have also updated the release-note block in the PR description as requested.

Norway-02:
/assign @willardhu 
Hi Willard, this PR has received the LGTM label and is fully ready for your final review and merge. Please take a look when you have a moment. Thank you for your time!

Norway-02:
Hi @DoisLONG, this PR has the `lgtm` label and is ready to merge. The commit is squashed and DCO-signed. Could you please `/approve` when you get a chance? Thank you!

DoisLONG (commented):
/lgtm

Norway-02:

PR #6997, opened 2026-07-10
imageparser: keep both tag and digest when parsing a reference

krishhna24:
**What type of PR is this?**

/kind bug
/kind test

`imageparser.Parse()` drops the digest when an image reference carries both a tag and a
digest (`nginx:1.19.1@sha256:...`). `reference.Parse` returns a value that satisfies both
`reference.Tagged` and `reference.Digested`, but `Parse()` matched them in a type switch,
which only ever runs the first matching case. `String()` then re-serialized the image
without the digest, so an EdgeApplication `ImageOverrider` on such an image silently
unpinned it — `nginx:1.19.1@sha256:...` came back out as `nginx:1.19.1`.

Fix:

- `Parse` asserts `reference.Tagged` and `reference.Digested` independently instead of
  type-switching on them.
- `String` emits `repository:tag@digest` when both are set, rather than tag-or-digest.
- `RemoveTagOrDigest` clears both fields. Its doc comment claimed tag and digest cannot
  co-exist, which is no longer true; clearing both keeps the `Tag`/`remove` override
  producing the same output it produces today (`nginx`), so the only user-visible change
  in this PR is that the digest is no longer lost.

`SetTagOrDigest` is left alone — replacing the version specifier should drop the stale
pin, which is what it already does.

The PR also carries the unit tests it started as: table-driven coverage for `Parse`,
`SplitHostname` and the `Components` accessors, including a regression case for the
tag+digest reference above. The package is at 100% statement coverage.

Refs #4327

This started as a test-only PR. Per review feedback the tag+digest characterization test
is back and the underlying parser bug is fixed here rather than deferred, so the test
asserts correct behavior instead of pinning the broken behavior.

`cloud/pkg/controllermanager/...` is green locally and `imageparser` is at 100%
statement coverage. Note that no GitHub Actions run has ever started on this PR — it is
still waiting on first-time-contributor workflow approval.

```release-note
Fixed EdgeApplication image overrides silently dropping the image digest when the image reference carries both a tag and a digest (for example nginx:1.19.1@sha256:...).
```

### DIFF
--- a/cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/imageparser.go
+++ b/cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/imageparser.go
@@ -71,15 +71,20 @@ func (c *Components) FullRepository() string {
 	return c.hostname + "/" + c.repository
 }
 
-// String returns the full name of the image, including repository and tag(or digest).
+// String returns the full name of the image, including repository, tag and digest.
+// A reference may carry both a tag and a digest (ie: nginx:1.19.1@sha256:...), in
+// which case both are emitted, the tag first, as in the reference itself.
 func (c Components) String() string {
-	if c.tag != "" {
+	switch {
+	case c.tag != "" && c.digest != "":
+		return c.FullRepository() + ":" + c.tag + "@" + c.digest
+	case c.tag != "":
 		return c.FullRepository() + ":" + c.tag
-	} else if c.digest != "" {
+	case c.digest != "":
 		return c.FullRepository() + "@" + c.digest
+	default:
+		return c.FullRepository()
 	}
-
-	return c.FullRepository()
 }
 
 // Tag returns the tag.
@@ -135,14 +140,19 @@ func (c *Components) SetTagOrDigest(input string) {
 	}
 }
 
-// RemoveTagOrDigest removes tag or digest.
-// Since tag and digest don't co-exist, so remove tag if tag not empty, otherwise remove digest.
+// RemoveTagOrDigest removes the tag and the digest.
+// A reference may carry either or both (ie: nginx:1.19.1@sha256:...), so clear each of them.
 func (c *Components) RemoveTagOrDigest() {
-	if c.tag != "" {
-		c.tag = ""
-	} else if c.digest != "" {
-		c.digest = ""
-	}
+	c.tag = ""
+	c.digest = ""
+}
+
+// taggedAndDigested is satisfied by a reference carrying both a tag and a digest,
+// ie: nginx:1.19.1@sha256:.... Such a reference also satisfies reference.Tagged and
+// reference.Digested on its own, so it has to be matched ahead of both of them.
+type taggedAndDigested interface {
+	reference.Tagged
+	reference.Digested
 }
 
 // Parse returns a Components of the given image.
@@ -158,6 +168,9 @@ func Parse(image string) (*Components, error) {
 	}
 
 	switch taggedOrDigested := ref.(type) {
+	case taggedAndDigested:
+		comp.tag = taggedOrDigested.Tag()
+		comp.digest = taggedOrDigested.Digest().String()
 	case reference.Tagged:
 		comp.tag = taggedOrDigested.Tag()
 	case reference.Digested:
--- a/cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/imageparser_test.go
+++ b/cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/imageparser_test.go
@@ -1,0 +1,274 @@
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package imageparser
+
+import "testing"
+
+const (
+	testRepo   = "nginx"
+	testDigest = "sha256:50d858e0985ecc7f60418aaf0cc5ab587f42c2570a884095a9e8ccacd0f6545c"
+)
+
+func TestParse(t *testing.T) {
+	tests := []struct {
+		name       string
+		image      string
+		wantErr    bool
+		hostname   string
+		repository string
+		tag        string
+		digest     string
+	}{
+		{
+			name:       "repository only",
+			image:      testRepo,
+			repository: testRepo,
+		},
+		{
+			name:       "repository and tag",
+			image:      testRepo + ":1.19.1",
+			repository: testRepo,
+			tag:        "1.19.1",
+		},
+		{
+			name:       "hostname repository and tag",
+			image:      "k8s.gcr.io/kube-apiserver:v1.0.0",
+			hostname:   "k8s.gcr.io",
+			repository: "kube-apiserver",
+			tag:        "v1.0.0",
+		},
+		{
+			name:       "hostname with port",
+			image:      "fictional.registry.example:10443/karmada/karmada-controller-manager:v1.0.0",
+			hostname:   "fictional.registry.example:10443",
+			repository: "karmada/karmada-controller-manager",
+			tag:        "v1.0.0",
+		},
+		{
+			name:       "digest reference",
+			image:      testRepo + "@" + testDigest,
+			repository: testRepo,
+			digest:     testDigest,
+		},
+		{
+			name:       "repository, tag and digest",
+			image:      testRepo + ":1.19.1@" + testDigest,
+			repository: testRepo,
+			tag:        "1.19.1",
+			digest:     testDigest,
+		},
+		{
+			name:    "invalid reference",
+			image:   "",
+			wantErr: true,
+		},
+	}
+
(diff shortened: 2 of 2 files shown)


krishhna24:
Addressed both in ab06bb4.

- `SetTagOrDigest`: added a case with input that matches neither regex, asserting the existing tag/digest are left untouched.
- tag+digest reference: added a case pinning current behavior. `reference.Parse` returns a value implementing both `Tagged` and `Digested`, and the type switch in `Parse` matches `Tagged` first, so the digest is dropped today. That does look like a real bug, but fixing it changes production behavior — I'd rather keep this PR test-only and send the `imageparser.go` fix in a separate PR so this one stays focused. The characterization test will force that follow-up to consciously flip the assertion.

Still at 100% statement coverage.

krishhna24:
@WillardHu @DoisLONG review comments are addressed. Could one of you approve the workflow runs? CI hasn't started since this is my first PR here.

DoisLONG:
Thanks for adding the imageparser tests. Most of the coverage looks useful to me, especially the `Parse`, `SplitHostname`, and accessor cases.

One thing: the tag+digest case currently pins the existing behavior where `Parse("nginx:tag@sha256:...")` keeps the tag but drops the digest.

That looks more like a bug in the production code than behavior we should lock in with a passing test. `Parse()` uses a type switch, so a reference that is both `reference.Tagged` and `reference.Digested` only hits the `Tagged` branch and silently loses the digest.

Could you either:

1. remove this case from this test-only PR and track the tag+digest behavior as a separate bug/follow-up, or
2. update this PR to fix the production code as well, so both tag and digest are preserved?

Also, the PR description currently has `Fixes #` without an issue number. Since this is part of the coverage effort, `Refs #4327` would be more appropriate unless this PR is intended to close a specific issue.

krishhna24:
Thanks for the review. Agreed the tag+digest behavior looks like a real bug in `Parse` rather than something to pin down, so I've dropped that case to keep this PR test-only. I'll track the digest-drop as a separate follow-up.

Also fixed the PR description: `Fixes #` → `Refs #4327`.

PTAL.

krishhna24:
@DoisLONG @WillardHu ping on this one.

The tag+digest case is gone as of a3b7413, so this is back to test-only — nothing pins production behavior anymore.

What's actually holding it up isn't review though: no GitHub Actions run has ever started on this PR. It's still waiting on first-time-contributor workflow approval, so there's no CI signal on a PR that is entirely tests. Could one of you kick the workflows off?

Locally:

```
$ go test -cover ./cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/...
ok  ...  coverage: 100.0% of statements
```

Happy to rebase if it's drifted.

DoisLONG (commented):
Thanks for the test coverage. The overall scope looks fine to me as a test-only PR for `imageparser`.

One important case is still missing before LGTM: please add coverage for an image reference that contains both a tag and a digest, for example `nginx:1.19.1@sha256:...`.

This is not just a coverage detail. It helps verify whether `Parse()` preserves both fields correctly for a reference that can be both tagged and digested. If the current production code drops the digest in that case, please fix the parser as part of this PR and keep the regression test.

The `SetTagOrDigest` invalid-input case already looks covered in the current patch, so that existing thread can probably be resolved after you confirm it.

krishhna24:
@DoisLONG done in 943dd830 — the tag+digest case is back and the parser is fixed here, per your last comment.

`Parse` now asserts `reference.Tagged` and `reference.Digested` separately instead of type-switching, so `nginx:1.19.1@sha256:...` keeps both, and `String()` emits both in canonical order. Without the `String()` half the parse fix wouldn't be observable, so both had to move.

One judgement call worth flagging: `RemoveTagOrDigest` documented that "tag and digest don't co-exist", which this PR makes false. I made it clear both fields. That keeps the `Tag`/`remove` override emitting `nginx` for a tag+digest image, exactly as it does today — the alternative (remove tag first, leave the digest) would have been a second behavior change on top of the bug fix. So the only user-visible change here is that the digest is no longer silently dropped. Happy to flip it if you'd rather `remove` leave the digest pin in place.

Retitled and the release note is no longer NONE, since this isn't test-only anymore. Rebased onto master (afc7d19).

/kind bug

Still no GitHub Actions run on this PR — it's never left "waiting for approval to run workflows" since I'm a first-time contributor here, so there's no CI signal at all. Could you kick the workflows off?

```
$ go test ./cloud/pkg/controllermanager/... 
ok      .../edgeapplication/overridemanager                     coverage: 74.6% of statements
ok      .../edgeapplication/overridemanager/imageparser         coverage: 100.0% of statements
```

DoisLONG (commented):
Thanks for the update. The current direction looks good to me.

This is no longer just a test-only change, and fixing the parser here makes sense. A reference can satisfy both `reference.Tagged` and `reference.Digested`, so replacing the type switch with independent assertions is the right fix. `String()` now preserves `repository:tag@digest`, and the regression coverage for tag+digest references is useful.

The invalid `SetTagOrDigest()` fallback case also looks covered now.

A couple of cleanup items before LGTM:

2. The existing review threads look addressed in the current diff, but they are still unresolved. Could you please resolve them if there is nothing pending there?

krishhna24:
Thanks! Both cleanup items are done.

- Squashed into a single commit (220c5e1b8). The tree is unchanged from the previous head, just the two commits collapsed with a combined message covering the `Parse` fix and the tests.
- Resolved the two open review threads — the tag+digest case and the invalid `SetTagOrDigest()` fallback are both covered in the current diff, nothing pending on either.

PTAL.

DoisLONG (commented):
Thanks for the cleanup.

/lgtm

krishhna24:
> Thanks for the cleanup.
> 
> /lgtm



Hey @DoisLONG What should I do next so that this PR gets `/approve` cause this is my first pr at kubeedge

cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/imageparser.go
```diff
@@ -71,15 +71,19 @@ func (c *Components) FullRepository() string {
...
+// String returns the full name of the image, including repository, tag and digest.
+// A reference may carry both a tag and a digest (ie: nginx:1.19.1@sha256:...), in
+// which case both are emitted in canonical order.
 func (c Components) String() string {
+	res := c.FullRepository()
 	if c.tag != "" {
-		return c.FullRepository() + ":" + c.tag
-	} else if c.digest != "" {
-		return c.FullRepository() + "@" + c.digest
+		res += ":" + c.tag
```
WillardHu: Changing `if..else if...` to `if ...; if...` is ambiguous. Originally, the priority of the returned value was c.tag > c.digest, but now c.digest has higher priority than c.tag.

cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/imageparser.go
```diff
@@ -157,11 +158,14 @@ func Parse(image string) (*Components, error) {
 		comp.hostname, comp.repository = SplitHostname(named.Name())
 	}
 
-	switch taggedOrDigested := ref.(type) {
-	case reference.Tagged:
-		comp.tag = taggedOrDigested.Tag()
-	case reference.Digested:
-		comp.digest = taggedOrDigested.Digest().String()
+	// A reference may satisfy both interfaces (ie: nginx:1.19.1@sha256:...), so check
```
WillardHu: The readability of a switch statement is better than that of an if statement.

cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/imageparser.go
krishhna24: Good catch, that was an unintended change in behaviour. Reworked it into a switch where each case returns directly, so the ordering is explicit instead of falling out of the accumulation:

```go
func (c Components) String() string {
	switch {
	case c.tag != "" && c.digest != "":
		return c.FullRepository() + ":" + c.tag + "@" + c.digest
	case c.tag != "":
		return c.FullRepository() + ":" + c.tag
	case c.digest != "":
		return c.FullRepository() + "@" + c.digest
	default:
		return c.FullRepository()
	}
}
```

Tag-only and digest-only now serialize exactly as they did before, and the both-set case emits `repo:tag@digest`, which is the form such a reference is parsed from.

cloud/pkg/controllermanager/edgeapplication/overridemanager/imageparser/imageparser.go
krishhna24: Agreed, moved back to a switch. The subtlety is that the type switch was the bug: `nginx:1.19.1@sha256:...` parses to a value satisfying both `reference.Tagged` and `reference.Digested`, and a type switch only ever runs the first matching case, so the digest was silently dropped.

So the switch stays, with the combined case named and matched ahead of the two single ones:

```go
// taggedAndDigested is satisfied by a reference carrying both a tag and a digest,
// ie: nginx:1.19.1@sha256:.... Such a reference also satisfies reference.Tagged and
// reference.Digested on its own, so it has to be matched ahead of both of them.
type taggedAndDigested interface {
	reference.Tagged
	reference.Digested
}

switch taggedOrDigested := ref.(type) {
case taggedAndDigested:
	comp.tag = taggedOrDigested.Tag()
	comp.digest = taggedOrDigested.Digest().String()
case reference.Tagged:
	comp.tag = taggedOrDigested.Tag()
case reference.Digested:
	comp.digest = taggedOrDigested.Digest().String()
}
```

Both comments are addressed in the amended commit. I also fixed the lint failure, which was `goconst` on the test file (`nginx` used in four comparisons), by hoisting it and the digest to file-level constants. PTAL.

krishhna24:
**Please add `/lgtm`**
@WillardHu @DoisLONG @fisherxu @Congrool

PR #7002, opened 2026-07-12
eventbus: fix data race on MQTT client during broker reconnect

### LINKED ISSUE #7000: eventbus: data race on MQTT PubCli/SubCli when the client is reassigned on broker reconnect
**What happened**:

The eventbus MQTT bridge stores its pub/sub client handles in a `Client` struct
with no synchronization:

    // edge/pkg/eventbus/mqtt/client.go:62-71
    type Client struct {
        ...
        PubCli MQTT.Client
        SubCli MQTT.Client
    }

`AutoReconnect` is disabled, so reconnection is driven entirely by the paho
connection-lost callbacks, which spawn a goroutine that rebuilds and reassigns
the handle.

At the same time, the eventbus message loop reads those same handles without any
lock:

    // eventbus.go:105  mqttBus.MQTTHub.PubCli.Publish(topic, 1, false, payload)
    // eventbus.go:204  mqttBus.MQTTHub.SubCli.Subscribe(topic, 1, mqttBus.OnSubMessageReceived)
    // eventbus.go:223  mqttBus.MQTTHub.SubCli.Unsubscribe(topic)

So when the broker connection drops (common on an edge node), the callback
goroutine reassigns `PubCli`/`SubCli` while the message loop may be reading them
to publish or (un)subscribe. `PubCli`/`SubCli` are interface values (two words),
so a concurrent read and write is a data race and can produce a torn read 
publishing/subscribing on a half constructed client or dereferencing a nil
method set. `Client` has no mutex.

Access to the shared MQTT client should be synchronized so that reconnecting
(which reassigns the client) cannot race with publish/subscribe. A `-race` build
exercising a reconnect while publishing should be clean.

1. Run edgecore with the external MQTT broker enabled (`mqttMode` set to Both or
   External).
2. While the eventbus is publishing/subscribing device traffic, drop or restart
   the broker so `OnConnectionLost` fires and `InitPubClient`/`InitSubClient`
   reassign the client.
3. Under the race detector, a driver that calls `pubMQTT` (eventbus.go:105) /
   `subscribe` (eventbus.go:204) concurrently with a simulated
   `onPubConnectionLost`/`onSubConnectionLost` reassignment reports a data race
   on `PubCli`/`SubCli`.

Code path: `onPubConnectionLost`/`onSubConnectionLost` (client.go:81-89) ->
`InitPubClient`/`InitSubClient` reassignment (client.go:141, 160), concurrent
with `pubMQTT` (eventbus.go:105) / `subscribe` (eventbus.go:204) / `unsubscribe`
(eventbus.go:223).

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

Priyanshu6968:
**What type of PR is this?**
/kind bug

The eventbus MQTT client handles (`PubCli`/`SubCli`) were read by the eventbus
loop (publish/subscribe/unsubscribe) while the paho connection lost callbacks
reassigned them from a separate goroutine, with no synchronization. Since the
handles are interface values, a broker disconnect (common on an edge node) could
race a concurrent read against the reassignment, producing a torn read and a
publish/subscribe on a half constructed client. `AutoReconnect` is disabled, so
this reassign on reconnect path is the normal one, not an edge case.

This unexports the two handles and guards them with a `RWMutex`, exposing them
through `PubCli()`/`SubCli()` accessors, so every read and the reassignment are
serialized. It also `Disconnect()`s the client being replaced, which otherwise
leaks its goroutines on every broker reconnect.

A related follow up (left out to keep this focused): `util.LoopConnect` does not
honor `beehiveContext.Done()`, so an in flight connect loop can't be reaped on
module shutdown. That change touches a shared util's signature and its tests, so
it is better as its own PR.

Fixes #7000

```release-note
Fixed a data race on the eventbus MQTT client when the connection to the broker is lost and re-established.
```

### DIFF
--- a/edge/pkg/eventbus/eventbus.go
+++ b/edge/pkg/eventbus/eventbus.go
@@ -102,7 +102,7 @@ func (eb *eventbus) Start() {
 }
 
 func pubMQTT(topic string, payload []byte) {
-	token := mqttBus.MQTTHub.PubCli().Publish(topic, 1, false, payload)
+	token := mqttBus.MQTTHub.Publish(topic, 1, false, payload)
 	if token.WaitTimeout(util.TokenWaitTime) && token.Error() != nil {
 		klog.Errorf("Error in pubMQTT with topic: %s, %v", topic, token.Error())
 	} else {
@@ -201,7 +201,7 @@ func (eb *eventbus) subscribe(topic string) {
 
 	if eventconfig.Config.MqttMode >= v1alpha2.MqttModeBoth {
 		// subscribe topic to external mqtt broker.
-		token := mqttBus.MQTTHub.SubCli().Subscribe(topic, 1, mqttBus.OnSubMessageReceived)
+		token := mqttBus.MQTTHub.Subscribe(topic, 1, mqttBus.OnSubMessageReceived)
 		if rs, err := util.CheckClientToken(token); !rs {
 			klog.Errorf("Edge-hub-cli subscribe topic: %s, %v", topic, err)
 			return
@@ -220,7 +220,7 @@ func (eb *eventbus) unsubscribe(topic string) {
 	}
 
 	if eventconfig.Config.MqttMode >= v1alpha2.MqttModeBoth {
-		token := mqttBus.MQTTHub.SubCli().Unsubscribe(topic)
+		token := mqttBus.MQTTHub.Unsubscribe(topic)
 		if rs, err := util.CheckClientToken(token); !rs {
 			klog.Errorf("Edge-hub-cli unsubscribe topic: %s, %v", topic, err)
 			return
--- a/edge/pkg/eventbus/eventbus.go
+++ b/edge/pkg/eventbus/eventbus.go
@@ -102,7 +102,7 @@ func (eb *eventbus) Start() {
 }
 
 func pubMQTT(topic string, payload []byte) {
-	token := mqttBus.MQTTHub.PubCli.Publish(topic, 1, false, payload)
+	token := mqttBus.MQTTHub.PubCli().Publish(topic, 1, false, payload)
 	if token.WaitTimeout(util.TokenWaitTime) && token.Error() != nil {
 		klog.Errorf("Error in pubMQTT with topic: %s, %v", topic, token.Error())
 	} else {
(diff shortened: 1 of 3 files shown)


DoisLONG:
Thanks for the fix. The main direction looks right to me.

Guarding the MQTT client handles with a lock and switching eventbus reads to accessors should address the direct race on the `PubCli` / `SubCli` interface fields. The added race test is also useful for this path.

One thing still looks worth fixing: `onPubConnectionLost` and `onSubConnectionLost` still ignore the client passed by the callback. If a superseded client fires a delayed connection-lost callback after it has already been replaced, it can still trigger another `InitPubClient()` / `InitSubClient()` and replace the current active client again.

Could you add a guard there so the reconnect only happens when the callback comes from the current active client? Something like checking `client == MQTTHub.PubCli()` / `client == MQTTHub.SubCli()` before starting the reconnect.

Priyanshu6968:
hi , thanks for the review, i will add a guard for the same

DoisLONG (commented):
Thanks for working on this. The original top-level race is real: the eventbus reads the `MQTT.Client` interface while the reconnect callback can replace it from another goroutine, and the `RWMutex` accessors address that specific interface read/write race.

I still see an issue with the concurrency model used by the new regression test. `TestClientReassignRace` starts two concurrent writers, but `InitPubClient()` and `InitSubClient()` only lock the field swap, not the complete reconnect lifecycle.

For example, one writer can install client B, a second writer can immediately replace it with C and call `Disconnect(B)`, and then the first writer can continue into `LoopConnect(B)`. That reconnects a superseded client that has already been disconnected. The test hides this because `LoopConnect` is patched to a no-op and the fake client's `Disconnect` only changes a boolean.

Could you either:

1. use a single writer in the race test and document that reconnect initialization is strictly single-writer, or
2. serialize the full create/connect/replace/cleanup lifecycle for each pub/sub client?

The regression proof also needs some follow-up. Main CI runs the normal `make test PROFILE=y` path, which does not enable `-race`, so the current successful CI run does not exercise the property this test is intended to verify. Please provide the before/after output from a focused race-detector run. It would also be good to convert the existing `TestOnPubConnectionLost` and `TestOnSubConnectionLost` boolean synchronization to a channel or atomic, since those tests themselves are not race-detector-safe.

One more compatibility point: this changes `PubCli` and `SubCli` from exported fields to methods, which is source-incompatible for downstream field access and struct literals. Please clarify whether this package is intentionally outside the compatibility surface, or mark/document the API change accordingly.

Finally, the accessors release the read lock before `Publish`/`Subscribe` is invoked, while the replacement path can immediately disconnect the returned old client. Please either confirm that concurrent use and `Disconnect` are supported by the Paho client version used here, or consider wrapper operations that hold the read lock for the actual client call.

Priyanshu6968:
thanks for your careful review, all addressed

PR #7004, opened 2026-07-13
dynamiccontroller: handle optional EndpointSlice port fields

### LINKED ISSUE #7003: dynamiccontroller: defaultMaster filter panics on EndpointSlice ports with nil fields
**What happened**:

The dynamic controller default-master filter can panic while rewriting the default Kubernetes EndpointSlice.

In `cloud/pkg/dynamiccontroller/filter/defaultmaster/defaultmaster.go`, `FilterResource` dereferences both `EndpointPort.Name` and `EndpointPort.Port` without checking whether either optional pointer is nil:

```go
for i := range eps.Ports {
    if *eps.Ports[i].Name != defaultEndpointSlicePortName {
        continue
    }
    *eps.Ports[i].Port = defaultMetaServerPort
}
```

The Kubernetes `discovery/v1.EndpointPort` API permits both fields to be omitted:

- `name` is optional and defaults to an empty string.
- `port` is optional; when omitted, ports are not restricted.

Therefore, a valid EndpointSlice containing an unnamed port causes a nil-pointer dereference at the name comparison. A port named `https` with no numeric port causes a nil-pointer dereference when assigning the metaserver port.

Because this filter runs in CloudCore dynamic-controller resource processing, the panic can terminate CloudCore.

The filter should handle valid EndpointSlice entries with omitted optional fields without panicking. It should only rewrite a matching `https` port when the required pointers are present, or safely set the missing port value.

1. Check out current master at `1853a181862ae637dbe5931a2ba508e123867add`.
2. In the `defaultmaster` package test, construct the default `kubernetes` EndpointSlice with at least one endpoint address and either of these port entries:

   ```go
   discovery.EndpointPort{Name: nil, Port: ptr.To[int32](443)}
   ```

   or:

   ```go
   discovery.EndpointPort{Name: ptr.To("https"), Port: nil}
   ```

3. Convert it to `unstructured.Unstructured` and call `FilterResource`.
4. Observe a nil-pointer panic at the corresponding dereference in `defaultmaster.go`.

The existing package tests pass because they only cover an EndpointPort where both pointers are non-nil.

Affected code:

- `cloud/pkg/dynamiccontroller/filter/defaultmaster/defaultmaster.go`
- The dereferences are currently around lines 80 and 83.

A focused regression test can cover missing `Name`, missing `Port`, and mixed port entries.

**Environment**:
- Kubernetes version (use `kubectl version`): Not required; source-level reproduction using the Kubernetes `discovery/v1` API type
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): master at `1853a181862ae637dbe5931a2ba508e123867add`

  - Hardware configuration (e.g. `lscpu`): Not required
  - OS (e.g. `cat /etc/os-release`): Not required
  - Kernel (e.g. `uname -a`): Not required
  - Go version (e.g. `go version`): Reproducible with the repository Go toolchain
  - Others: `go test ./cloud/pkg/dynamiccontroller/filter/defaultmaster -count=1` currently passes because the nil cases are not covered

  </details>

  - edgecore version (e.g. `edgecore --version`): Not applicable
  - Hardware configuration (e.g. `lscpu`): Not applicable
  - OS (e.g. `cat /etc/os-release`): Not applicable
  - Kernel (e.g. `uname -a`): Not applicable
  - Go version (e.g. `go version`): Not applicable
  - Others: Not applicable

  </details>

santhil-cyber:
**What type of PR is this?**

/kind bug

The dynamic controller's default-master EndpointSlice filter dereferenced the optional `EndpointPort.Name` and `EndpointPort.Port` fields without nil checks.

This change:

- ignores unnamed EndpointSlice ports instead of panicking;
- assigns the metaserver port pointer for a matching `https` entry, including when its original port is omitted;
- adds regression tests for nil name, nil port, and mixed port entries.

Both reported panic paths were reproduced independently before applying the fix:

- nil `Name` panicked at `defaultmaster.go:80`;
- nil `Port` panicked at `defaultmaster.go:83`.

Fixes #7003

Verified with:

```text
go test ./cloud/pkg/dynamiccontroller/filter/defaultmaster -count=1
go test ./cloud/pkg/dynamiccontroller/filter/... -count=1
go test -race ./cloud/pkg/dynamiccontroller/filter/defaultmaster -count=1
```

The focused package suite was also run twice after the initial fix.

```release-note
Fixed a CloudCore panic when the dynamic controller processes EndpointSlice ports with omitted optional fields.
```

### DIFF
--- a/cloud/pkg/dynamiccontroller/filter/defaultmaster/defaultmaster.go
+++ b/cloud/pkg/dynamiccontroller/filter/defaultmaster/defaultmaster.go
@@ -80,13 +80,14 @@ func (f *FilterImpl) FilterResource(_ string, obj runtime.Object) {
 			if eps.Ports[i].Name == nil || *eps.Ports[i].Name != defaultEndpointSlicePortName {
 				continue
 			}
-			eps.Ports[i].Port = &f.port
-			unstrRaw, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&eps)
-			if err != nil {
-				klog.Errorf("default endpointslice %v convert to unstructure error: %v", eps.Name, err)
-				return
-			}
-			unstruct.SetUnstructuredContent(unstrRaw)
+			port := f.port
+			eps.Ports[i].Port = &port
 		}
+		unstrRaw, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&eps)
+		if err != nil {
+			klog.Errorf("default endpointslice %v convert to unstructure error: %v", eps.Name, err)
+			return
+		}
+		unstruct.SetUnstructuredContent(unstrRaw)
 	}
 }
--- a/cloud/pkg/dynamiccontroller/filter/defaultmaster/defaultmaster.go
+++ b/cloud/pkg/dynamiccontroller/filter/defaultmaster/defaultmaster.go
@@ -77,10 +77,10 @@ func (f *FilterImpl) FilterResource(_ string, obj runtime.Object) {
 		eps.Endpoints[0].Addresses = eps.Endpoints[0].Addresses[:1]
 		eps.Endpoints[0].Addresses[0] = defaultMetaServerIP
 		for i := range eps.Ports {
-			if *(eps.Ports[i].Name) != defaultEndpointSlicePortName {
+			if eps.Ports[i].Name == nil || *eps.Ports[i].Name != defaultEndpointSlicePortName {
 				continue
 			}
-			*(eps.Ports[i].Port) = defaultMetaServerPort
+			eps.Ports[i].Port = &f.port
 			unstrRaw, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&eps)
 			if err != nil {
 				klog.Errorf("default endpointslice %v convert to unstructure error: %v", eps.Name, err)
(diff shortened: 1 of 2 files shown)


DoisLONG:
Thanks for the fix. The nil `EndpointPort.Name` / `EndpointPort.Port` cases are valid, and the added regression tests are useful.

One thing I think still needs cleanup: `eps.Ports[i].Port = &f.port` stores a pointer to the filter struct field in the EndpointSlice object. It would be safer to copy the value first and assign the pointer to that local value.

Also, `SetUnstructuredContent` is still inside the port loop. If no `https` port is matched, the endpoint address rewrite done before the loop will not be written back to the unstructured object. Could you move the conversion / `SetUnstructuredContent` after the loop, and add an assertion that the endpoint address is still rewritten even when an unnamed port is ignored?

santhil-cyber:
Thanks for the review. Both points make sense. I’ll update the implementation

santhil-cyber:
@DoisLONG can you verify?!

DoisLONG (commented):
Thanks for the update.

I re-checked the latest version carefully. The main issues look addressed now.

The nil EndpointPort.Name case is handled without dereferencing the pointer, and the https port rewrite now also works when the original EndpointPort.Port is nil. Using a local port := f.port before assigning eps.Ports[i].Port also avoids sharing a pointer to the filter struct field.

Moving ToUnstructured / SetUnstructuredContent outside the port loop looks right as well. This makes sure the endpoint address rewrite is written back even when no https port is matched.

The regression tests cover the nil name, nil port, and mixed-port cases. One small non-blocking improvement would be to also assert the number of ports and that port names are preserved, but I do not think that needs to block this fix.

santhil-cyber:
/assign @kevin-wangzefeng

@DoisLONG lgtm’d this on Jul 13 after the pointer-aliasing and `SetUnstructuredContent` points were addressed. DCO passes, there are no conflicts with master, and tide reports the `approved` label as the only thing outstanding. The workflow runs are still waiting on maintainer approval to start, so if a re-run is needed I am happy to push a rebase.

PR #7010, opened 2026-07-14
test(pkg/metaserver): migrate test setup assertions from assert to require

Norway-02:
**What type of PR is this?**
/kind cleanup

Migrates `assert.NoError` to `require.NoError` in the `pkg/metaserver` test suites. Using `assert` during test setup phases allows the test to continue executing when struct creation fails, which causes cascading and difficult-to-debug nil-pointer panics in CI logs. `require` immediately halts the test on failure.

NONE

NONE

```release-note
NONE
```

### DIFF
--- a/pkg/metaserver/application_test.go
+++ b/pkg/metaserver/application_test.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


Norway-02:
I have updated the loop assertions in `application_test.go` to use `if !assert.NoError(t, err) { continue }`. This ensures that test loops gracefully continue to evaluate subsequent test cases while still preventing cascading nil-pointer panics downstream!

PR #7011, opened 2026-07-14
test(cloud/dynamiccontroller): migrate test setup assertions from assert to require

Norway-02:
**What type of PR is this?**
/kind cleanup

Migrates `assert.NoError` to `require.NoError` in the `cloud/pkg/dynamiccontroller` test suites. Using `assert` during test setup phases allows the test to continue executing when struct creation fails, which causes cascading nil-pointer panics in CI logs. `require` immediately halts the test on failure.

NONE

NONE

```release-note
NONE
```


Norway-02:
I have removed all of the dead/commented-out test blocks in `endpointresource_test.go` entirely per Gemini's feedback! Thanks for the catch.

PR #7013, opened 2026-07-14
fix(cloud/dynamiccontroller): close data races by safely copying listener maps

Norway-02:
**What type of PR is this?**
/kind bug

This PR resolves potential data races and `concurrent map iteration and map write` panics in `cloud/pkg/dynamiccontroller/application/listener_manager.go`:

1. `GetListenersForNode` now returns a shallow copy of the node's listener map (`maps.Clone`) under `RLock`.
2. `GetListenersForGVR` now returns a newly allocated slice `[]*SelectorListener` under `RLock`. This avoids map allocation overhead during event dispatching and provides cache-friendly sequential iteration when broadcasting events to matching listeners.
3. Added comprehensive regression tests (`TestGetListenersReturnsCopy`) ensuring external mutations on the returned collections do not affect the internal state of `listenerManager`.

A shallow copy / slice copy is sufficient because callers only iterate and read the returned listeners to broadcast events, rather than mutating nested listener state.

NONE

- Exactly 1 squashed commit with DCO sign-off.
- Verified with `go test -race ./cloud/pkg/dynamiccontroller/...`.

```release-note
NONE
```


DoisLONG (commented):
Thanks for the fix.

The direction looks right to me. If `GetListenersForNode` / `GetListenersForGVR` return the internal listener maps directly, callers can end up iterating over a map that is concurrently modified by the manager, which can lead to data races or `concurrent map iteration and map write` panics.

Returning a copied map under the read lock makes sense here.

What I still want to see is a small regression test for this behavior. For example, it would be good to verify that the returned map is a copy rather than an alias of the internal map, so mutations on the returned value do not affect the manager state.

It would also help to clarify in the PR description whether a shallow map copy is sufficient here because callers only iterate/read the returned listeners, rather than mutating nested state.

Norway-02:
@doislong Added a regression test verifying that mutations on the returned map do not affect the manager's internal state. Also updated the PR description to clarify that a shallow copy is sufficient since callers only iterate over the maps. Let me know if everything looks good now!

DoisLONG (commented):
Thanks for the update.

Two small cleanup points:

2. The PR description still says both getters return shallow map copies, but `GetListenersForGVR()` now returns a `[]*SelectorListener`. Could you please update the description to match the current implementation?

Norway-02:
Hi @DoisLONG, thanks for the helpful suggestions!
I have:
1. Refactored `GetListenersForGVR` to return `[]*SelectorListener` for better performance during event broadcast iteration.
2. Added the regression test (`TestGetListenersReturnsCopy`) verifying that external mutations on returned collections do not mutate the manager's internal state.
3. Clarified in the PR description why shallow/slice copying is sufficient for broadcast iteration.
4. Squashed all changes into a single clean, DCO-signed commit.

All tests pass with `-race`. Please let me know if anything else is needed!

DoisLONG (commented):
Thanks for the update.

/lgtm

Norway-02:
Hi @gsssc, this PR has the `lgtm` label and all review feedback has been addressed. The change fixes a real data race in the DynamicController's listener map, is squashed into a single DCO-signed commit, and all tests pass with `-race`. Could you please `/approve` when you get a chance? Thank you!

Norway-02:

PR #7015, opened 2026-07-14
cloud/devicecontroller: fix detached pointer for newly created twin

### LINKED ISSUE #7014: cloud/devicecontroller: findOrCreateTwinByName returns a detached pointer, first twin update is lost
/kind bug

**What happened**:

In `cloud/pkg/devicecontroller/controller/upstream.go`, `findOrCreateTwinByName` builds a new `Twin` for a device property that isn't in the cached status yet, appends a **copy** of it to the slice, and then returns the **original local pointer** — not a pointer into the slice:

```go
func findOrCreateTwinByName(twinName string, properties []v1beta1.DeviceProperty, deviceStatus *DeviceStatus) *v1beta1.Twin {
	for i := range properties {
		if twinName == properties[i].Name {
			twin := findTwinByName(twinName, deviceStatus)
			if twin != nil {
				return twin
			}
			twin = &v1beta1.Twin{
				PropertyName: twinName,
			}
			deviceStatus.Status.Twins = append(deviceStatus.Status.Twins, *twin) // appends a COPY
			return twin // returns the ORIGINAL local, not the slice element
		}
	}
	return nil
}
```

The caller `updateDeviceStatus` mutates the returned pointer in place:

```go
deviceTwin := findOrCreateTwinByName(twinName, cacheDevice.Spec.Properties, deviceStatus)
if deviceTwin != nil {
	// ...
	deviceTwin.Reported = reported // mutates the detached local, not deviceStatus.Status.Twins
	// ...
	deviceTwin.ObservedDesired = observedDesired
}
```

Because `deviceTwin` points at a struct that was already copied by value into `deviceStatus.Status.Twins` *before* the mutation, the `Reported` / `ObservedDesired` writes land on a throwaway variable. The slice element that is actually marshalled and PATCHed to the `DeviceStatus` CRD keeps its zero-value `Reported` / `ObservedDesired`. This breaks only the *first* update for a given property — once the (empty) twin exists in the slice, `findTwinByName` on the next update correctly returns `&deviceStatus.Status.Twins[i]`, so subsequent updates work.

The very first reported/desired value for a newly-appeared device twin property should reach the `DeviceStatus` CRD, exactly like it does for twins that already exist in the status.

Register a device with a property that has never been reported before, then have the edge node send a `DeviceTwinUpdate` for that property with `Actual.Value` set. Inspect the resulting `DeviceStatus` CRD after the patch: `status.twins[].reported.value` for that property is empty even though the edge sent a real value.

Fix is to append the fully-built value and return the address of the slice element:

```go
deviceStatus.Status.Twins = append(deviceStatus.Status.Twins, v1beta1.Twin{
	PropertyName: twinName,
})
return &deviceStatus.Status.Twins[len(deviceStatus.Status.Twins)-1]
```

The sibling `findTwinByName` in the same file already returns a slice-backed pointer, which is the contract callers rely on.

**Environment**:

- KubeEdge version: master
- This is a code-level defect, independent of runtime environment.

Priyanshubhartistm:
**What type of PR is this?**

/kind bug

`findOrCreateTwinByName` in `cloud/pkg/devicecontroller/controller/upstream.go` is meant to return a pointer that the caller (`updateDeviceStatus`) mutates in place. For an existing twin it correctly returns `&deviceStatus.Status.Twins[i]` (via `findTwinByName`). For a *newly created* twin, however, it appended a copy of the twin to the slice (`append(..., *twin)`) and then returned the original local `twin` pointer — which is detached from the slice.

As a result, the `Reported` / `ObservedDesired` values that `updateDeviceStatus` writes through the returned pointer land on a throwaway local, while the slice element that actually gets marshalled into the `DeviceStatus` CRD patch keeps its zero value. The first status update for any newly reported device property is therefore silently lost. Subsequent updates work because the (empty) twin now exists in the slice and `findTwinByName` returns a proper slice-backed pointer.

The fix appends the twin and returns `&deviceStatus.Status.Twins[len-1]`, so the returned pointer references the element stored in the slice.

Fixes #7014 

- One-line behavioural fix; `findTwinByName` (same file) already demonstrates the intended pointer-into-slice pattern.
- Added `TestFindOrCreateTwinByNameNewTwinIsMutable`, which mutates the returned pointer and asserts the change is visible on the stored twin. The pre-existing `TestFindOrCreateTwinByName` passed even with the bug because it never mutated the returned pointer.
- Verified locally: `make verify`, `make lint`, `make test`, and `make integrationtest` all pass.

```release-note
Fix device twin reported/observedDesired values being dropped on the first device status update for a newly reported property.
```

### DIFF
--- a/cloud/pkg/devicecontroller/controller/upstream.go
+++ b/cloud/pkg/devicecontroller/controller/upstream.go
@@ -353,11 +353,10 @@ func findOrCreateTwinByName(twinName string, properties []v1beta1.DeviceProperty
 			if twin != nil {
 				return twin
 			}
-			twin = &v1beta1.Twin{
+			deviceStatus.Status.Twins = append(deviceStatus.Status.Twins, v1beta1.Twin{
 				PropertyName: twinName,
-			}
-			deviceStatus.Status.Twins = append(deviceStatus.Status.Twins, *twin)
-			return twin
+			})
+			return &deviceStatus.Status.Twins[len(deviceStatus.Status.Twins)-1]
 		}
 	}
 	return nil
--- a/cloud/pkg/devicecontroller/controller/upstream_test.go
+++ b/cloud/pkg/devicecontroller/controller/upstream_test.go
@@ -165,6 +165,31 @@ func TestFindOrCreateTwinByName(t *testing.T) {
 	}
 }
 
+func TestFindOrCreateTwinByNameNewTwinIsMutable(t *testing.T) {
+	assert := assert.New(t)
+
+	status := &DeviceStatus{
+		Status: v1beta1.DeviceStatusStatus{
+			Twins: []v1beta1.Twin{},
+		},
+	}
+	properties := []v1beta1.DeviceProperty{
+		{
+			Name: "humidity",
+		},
+	}
+
+	// The returned pointer must reference the twin stored in the slice, so that
+	// updates made through it (as updateDeviceStatus does) reach the DeviceStatus
+	// that is later marshalled into the CRD patch.
+	twin := findOrCreateTwinByName("humidity", properties, status)
+	assert.NotNil(twin)
+	twin.Reported = v1beta1.TwinProperty{Value: "60"}
+
+	assert.Len(status.Status.Twins, 1)
+	assert.Equal("60", status.Status.Twins[0].Reported.Value)
+}
+
 func TestFindTwinByName(t *testing.T) {
 	tests := []struct {
 		name         string


Priyanshubhartistm:
Hey @DoisLONG , Whenever you're free, could you please review this PR?
 I'd really appreciate your feedback. Thanks!

DoisLONG (commented):
Thanks for the fix. This looks good to me overall.

The bug is real: for a newly created twin, the old code appended a copy to `deviceStatus.Status.Twins` but returned the original local pointer. Since `updateDeviceStatus()` mutates the returned pointer, the first reported/observedDesired update could be written to the detached object instead of the slice element that is later patched back to the CRD.

Returning `&deviceStatus.Status.Twins[len(deviceStatus.Status.Twins)-1]` after append matches the contract already used by `findTwinByName`, where the caller gets a pointer to the stored twin.

The new regression test also checks the important behavior: mutating the returned pointer must update the stored twin in `DeviceStatus`.

Priyanshubhartistm:
Hi @DoisLONG, thank you for the kind words! Since the fix and the regression test look good to you, could you please put a /lgtm if everything is in order? That would really help move this forward. 
Thank you

DoisLONG:
Thanks for the fix.

I rechecked the latest version. The change is focused, and the regression test covers the important behavior: the pointer returned for a newly created twin must reference the stored slice element so later mutations are included in the DeviceStatus patch.

/lgtm

PR #7024, opened 2026-07-15
devicecontroller: tolerate DeviceStatus create races

fanny7d:
**What type of PR is this?**

/kind bug

This PR makes DeviceStatus creation idempotent when CloudCore runs with multiple replicas.

All devicecontroller replicas can receive the same Device add event. They may all observe that the DeviceStatus is absent, then race to create it. The first create succeeds and the remaining creates return `AlreadyExists`.

Previously, `deviceAdded()` treated that expected race as a fatal error and returned before sending the DeviceModel and Device messages to the edge node. The API server could therefore contain both Device and DeviceStatus objects while the edge node never received the corresponding Device ObjectSync.

The create path now handles `AlreadyExists` by fetching the DeviceStatus created by the winning replica and continuing the normal Device/DeviceModel downlink path.

Validation performed:

- Verified the regression test fails against the old logic with `devicestatuses.devices.kubeedge.io "lite-node" already exists`.
- `go test ./cloud/pkg/devicecontroller/... -count=1`
- `git diff --check upstream/master...HEAD`

```release-note
CloudCore devicecontroller now tolerates concurrent DeviceStatus create races so Device and DeviceModel downlink continues in HA deployments.
```

### DIFF
--- a/cloud/pkg/devicecontroller/controller/downstream.go
+++ b/cloud/pkg/devicecontroller/controller/downstream.go
@@ -210,7 +210,11 @@ func (dc *DownstreamController) getOrCreateDeviceStatusForDevice(device *v1beta1
 		return nil, err
 	}
 
-	return dc.crdClient.DevicesV1beta1().DeviceStatuses(device.Namespace).Create(context.Background(), deviceStatus, metav1.CreateOptions{})
+	deviceStatus, err = dc.crdClient.DevicesV1beta1().DeviceStatuses(device.Namespace).Create(context.Background(), deviceStatus, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
+		return dc.crdClient.DevicesV1beta1().DeviceStatuses(device.Namespace).Get(context.Background(), device.Name, metav1.GetOptions{})
+	}
+	return deviceStatus, err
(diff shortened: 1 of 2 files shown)


fanny7d:
Updated this branch onto the latest upstream `master` (including the v1.23.1 security fixes). The devicecontroller unit tests pass.

/assign @rohitsardesai83

DoisLONG (commented):
Thanks for the fix. This direction looks good to me.

The `AlreadyExists` handling makes sense here. In an HA CloudCore deployment, another replica may create the `DeviceStatus` after this replica observes `NotFound`, so treating `AlreadyExists` as a hard failure would incorrectly stop the normal device/model downlink path.

I also checked the caller side: `deviceAdded()` only needs `getOrCreateDeviceStatusForDevice()` to return successfully before continuing to send the membership, DeviceModel, and Device messages. So fetching the existing `DeviceStatus` after the create race should preserve the intended flow.

The regression test also captures the important sequence: first `Get` returns `NotFound`, `Create` returns `AlreadyExists`, and the second `Get` returns the existing object.

One small thing: could you also confirm the CI result once it is available? Other than that, this looks reasonable to me.

PR #7025, opened 2026-07-15
cloudhub: preserve ObjectSync specs for CRDs

fanny7d:
**What type of PR is this?**

/kind bug

This PR preserves ObjectSync and ClusterObjectSync GVK metadata for Kubernetes and KubeEdge runtime objects, including KubeEdge CRDs whose in-memory objects do not contain TypeMeta.

When a downstream `ServiceAccountAccess` object had an empty TypeMeta, `GetMessageAPIVersion` and `GetMessageResourceType` returned empty strings. CloudHub then created an ObjectSync with empty `spec.objectAPIVersion` and `spec.objectKind`, so synccontroller could not reconstruct and resend the object during compensation.

The change resolves GVK from the Kubernetes and KubeEdge runtime schemes when TypeMeta is absent. New ObjectSync records use that resolved spec, and existing records backfill only missing spec fields with conflict retries.

The backfill is guarded by the informer-cached spec. When the ObjectSync or ClusterObjectSync spec and status are already initialized, dispatch compares resource versions directly and performs no synchronous API server call. API GET/UPDATE calls only occur when a spec field is actually missing and a replacement value is available.

Validation performed:

- Added namespaced and cluster-scoped tests that assert the initialized dispatch hot path records zero client actions.
- Added namespaced and cluster-scoped tests that assert missing specs perform exactly GET and UPDATE before dispatch continues.
- Added KubeEdge CRD tests for API version and kind resolution without TypeMeta.
- `go test ./pkg/metaserver/util ./cloud/pkg/cloudhub/dispatcher ./cloud/pkg/cloudhub/session -count=1`
- `git diff --check upstream/master...HEAD`

The broader `go test ./cloud/pkg/cloudhub/...` command also reaches an unrelated existing panic in `cloud/pkg/cloudhub/servers/httpserver/node`; the same failure reproduces on the current upstream master without this change.

```release-note
CloudHub now preserves and conditionally backfills ObjectSync GVK metadata for Kubernetes and KubeEdge runtime objects without adding API server calls to the initialized dispatch hot path.
```

### DIFF
--- a/cloud/pkg/cloudhub/dispatcher/message_dispatcher.go
+++ b/cloud/pkg/cloudhub/dispatcher/message_dispatcher.go
@@ -24,6 +24,7 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 
 	"github.com/kubeedge/api/apis/reliablesyncs/v1alpha1"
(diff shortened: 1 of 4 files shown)


fanny7d:
Updated this branch onto the latest upstream `master` (including the v1.23.1 security fixes). The CloudHub dispatcher and metaserver util unit tests pass.

/assign @shelley-baoyue

DoisLONG (commented):
Thanks for working on this. The problem this PR is trying to solve is important: if `ObjectSync` / `ClusterObjectSync` is created with empty GVK fields, later compensation may not be able to reconstruct the object correctly.

I think the general direction makes sense, especially resolving GVK from the registered Kubernetes/KubeEdge schemes when TypeMeta is empty, and keeping the initialized dispatch hot path free from extra API calls.

I have a few concerns:

1. Please address the existing unresolved comments first. In particular, this repo currently uses Go 1.23, so `t.Context()` should be replaced with `context.TODO()` or `context.Background()` in the new tests.

2. The PR title/description says CRDs, but the implementation resolves kinds only from the Kubernetes scheme and the KubeEdge generated scheme. That is fine for registered Kubernetes/KubeEdge runtime objects, but it does not cover arbitrary third-party CRDs. Could you clarify this scope in the PR description/title to avoid overclaiming?

3. The backfill path now performs synchronous GET/UPDATE before deciding whether to enqueue the message. If the backfill fails, the message is not enqueued. Could you explain why this is the right failure behavior for reliable sync? My main concern is whether a temporary API conflict or backfill failure could block an otherwise valid newer message from being dispatched.

4. It would be useful to add one test for the failure path of `backfillObjectSyncSpec` / `backfillClusterObjectSyncSpec`, so the expected behavior is explicit. For example, when the ObjectSync exists in the lister with missing spec, but the API update fails, the test should verify that enqueue returns false and no message is dispatched.

PR #7037, opened 2026-07-16
keadm: assign log paths for resource copy containers

fanny7d:
/kind bug

`CopyResources` creates a temporary CRI sandbox and container during `keadm join`.
Those resources were created without an explicit pod log directory or container log
path, leaving runtime cleanup without a scoped log location.

This change:

- creates a unique pod log directory for each resource-copy sandbox;
- sets the temporary container log path relative to that directory;
- cleans up the temporary CRI resources and log directory with an independent,
  bounded context; and
- preserves a successful resource copy when only cleanup sees a transient CRI
  transport error.

None.

Validation performed:

- `go test -count=1 ./pkg/containers`
- `make verify`
- `make test WHAT='pkg keadm'`

The current Makefile does not define the `edge_test` target referenced by
CONTRIBUTING.md. The available integration script requires `sudo`, a Linux
EdgeCore process, and a local MQTT setup, so it was not run on macOS. The
focused unit tests cover the resource-copy lifecycle and cleanup behavior.

```release-note
Fixed keadm resource copying to use scoped CRI log paths and to tolerate
transient cleanup transport failures after a successful copy.
```

### DIFF
--- a/pkg/containers/container_runtime.go
+++ b/pkg/containers/container_runtime.go
@@ -18,7 +18,9 @@ package containers
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"time"
 
@@ -36,6 +38,13 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/image"
 )
 
+const (
+	defaultPodLogsDirectory     = "/var/log/pods"
+	copyResourcesCleanupTimeout = 45 * time.Second
+	copyResourcesStopTimeout    = int64(5)
+	copyResourcesContainerName  = "container"
+)
+
 type ContainerRuntime interface {
 	CopyResources(ctx context.Context, image string, files map[string]string) error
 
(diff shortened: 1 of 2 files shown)


DoisLONG (commented):
Thanks for working on this.

The general direction looks reasonable to me. `CopyResources()` creates a temporary CRI sandbox/container during `keadm join`, so setting a scoped pod log directory and a container log path makes sense. The cleanup refactor is also useful: using an independent bounded cleanup context avoids leaking the temporary CRI resources only because the caller context was cancelled.

The tests are meaningful as well. They check the sandbox `LogDirectory`, container `LogPath`, cleanup order, cleanup after a cancelled parent context, and the case where a cleanup error after a successful copy is treated as non-fatal.

I have two questions about the intended semantics:

1. This PR is marked as `/kind bug`, but `Which issue(s) this PR fixes` is `None`. Since this changes the `keadm join` resource-copy lifecycle and cleanup behavior, could you link the issue or describe the concrete failure this fixes more explicitly?

2. The PR description says cleanup transport errors after a successful copy should not make `keadm join` retry the whole resource copy. I understand that motivation, but the current implementation treats all cleanup errors as non-fatal after a successful copy, including `StopContainer`, `RemoveContainer`, `RemovePodSandbox`, and `os.RemoveAll(logDirectory)` failures. Could you clarify that this is the intended behavior for `keadm join`? In other words, if the resource copy succeeds but the temporary sandbox/container/log directory cleanup fails, should `CopyResources()` still return nil?

If that behavior is intentional, the code and tests look focused to me.

PR #7043, opened 2026-07-16
fix: handle dynamiccontroller delete tombstones

### LINKED ISSUE #7038: cloud/pkg/dynamiccontroller: delete events can be dropped on informer tombstones
**What happened**:

On latest `master`, delete events handled by `CommonResourceEventHandler` can be silently dropped when the informer sends `cache.DeletedFinalStateUnknown`.

In `cloud/pkg/dynamiccontroller/application/eventhandler.go`, `DeleteFunc` passes `obj` directly into `objToEvent`, but `objToEvent` only accepts `runtime.Object` and logs `Unknown type` otherwise.

`cache.DeletedFinalStateUnknown` is a valid informer delete payload. When that payload is delivered, the delete event is ignored instead of being forwarded.

Delete events should still be converted and forwarded when the informer provides `cache.DeletedFinalStateUnknown`.

1. Check out latest `master` at `555e57bb71fede21d3fc020e95f132132018986c`.
2. Observe current code path:

   - `cloud/pkg/dynamiccontroller/application/eventhandler.go`
     - `DeleteFunc` calls `handler.objToEvent(watch.Deleted, obj)`
     - `objToEvent` does `eventObj, ok := obj.(runtime.Object)` and returns if `ok == false`

3. Run a focused repro such as passing a tombstone payload into the same handler path:

   ```go
   tombstone := cache.DeletedFinalStateUnknown{Obj: &unstructured.Unstructured{}}
   _, ok := interface{}(tombstone).(runtime.Object)
   fmt.Println(ok)
   ```

4. Actual result:

   ```text
   false
   ```

   So the current code logs unknown type and drops the delete event.

5. Expected result:

   ```text
   tombstone should be unwrapped to tombstone.Obj and delete event should be forwarded
   ```

Current code on latest `master`:

- `cloud/pkg/dynamiccontroller/application/eventhandler.go`
  - `DeleteFunc` forwards raw informer delete payload
  - `objToEvent` only accepts `runtime.Object`
- `cache.DeletedFinalStateUnknown`
  - is standard client-go delete payload when final state is not directly available
- other KubeEdge delete handlers already unwrap tombstones before continuing:
  - `cloud/pkg/devicecontroller/manager/common.go`
  - `cloud/pkg/edgecontroller/manager/common.go`

This makes the bug small, isolated, and consistent with an existing repo pattern for the fix.

**Environment**:
  - Not required for this source-level repro
  - latest `master` at `555e57bb71fede21d3fc020e95f132132018986c`

    - Not required for this repro
    - Not required for this repro
  - Kernel (e.g. `uname -a`):
    - Not required for this repro
    - `go version go1.26.4 linux/amd64`
  - Others:
    - Repro is source-level and based on valid informer tombstone payload handling

  </details>

    - latest `master` at `555e57bb71fede21d3fc020e95f132132018986c`
    - Not required for this repro
    - Not required for this repro
  - Kernel (e.g. `uname -a`):
    - Not required for this repro
    - `go version go1.26.4 linux/amd64`
  - Others:
    - Repro is source-level and based on valid informer tombstone payload handling

  </details>

RedZapdos123:
## Description:

Issue #7038 reported that dynamiccontroller delete events could be dropped when the informer delivered `cache.DeletedFinalStateUnknown`.

This PR updates `cloud/pkg/dynamiccontroller/application/eventhandler.go` to unwrap tombstone delete payloads before converting them to `runtime.Object`, so valid delete events are still forwarded.

It also adds regression coverage in `cloud/pkg/dynamiccontroller/application/eventhandler_test.go` for both valid and invalid tombstone delete payloads.

The fix is minimal and limited to the affected dynamiccontroller delete event conversion path.

## Validation:

- `gofmt -w cloud/pkg/dynamiccontroller/application/eventhandler.go cloud/pkg/dynamiccontroller/application/eventhandler_test.go`
- `go test ./cloud/pkg/dynamiccontroller/application -count=1`
- `go test -v ./cloud/pkg/dynamiccontroller/application -run TestEventHandlerObjectConversion -count=1`
- `golangci-lint run ./cloud/pkg/dynamiccontroller/application --timeout=5m`

Closes #7038.

## Checklist:

- [x] I have added regression test cases for valid and invalid tombstone delete payloads.
- [x] I have run validation using the focused `go test` and `golangci-lint` commands above.

## Release note:

```release-note
DynamicController now correctly forwards delete events when informer delete payloads arrive as tombstones.
```

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/eventhandler.go
+++ b/cloud/pkg/dynamiccontroller/application/eventhandler.go
@@ -144,6 +144,9 @@ func NewCommonResourceEventHandler(
 }
 
 func (c *CommonResourceEventHandler) objToEvent(t watch.EventType, obj interface{}) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+	}
 	eventObj, ok := obj.(runtime.Object)
 	if !ok {
 		klog.Warningf("Unknown type: %T, ignore", obj)
--- a/cloud/pkg/dynamiccontroller/application/eventhandler_test.go
+++ b/cloud/pkg/dynamiccontroller/application/eventhandler_test.go
@@ -711,7 +711,7 @@ func TestEventHandlerObjectConversion(t *testing.T) {
 		select {
 		case <-freshHandler.events:
 			t.Fatal("Unexpected event received for invalid tombstone object")
-		case <-time.After(100 * time.Millisecond):
+		default:
 		}
 	})
 
--- a/cloud/pkg/dynamiccontroller/application/eventhandler_test.go
+++ b/cloud/pkg/dynamiccontroller/application/eventhandler_test.go
@@ -674,6 +674,47 @@ func TestEventHandlerObjectConversion(t *testing.T) {
 		}
 	})
 
+	t.Run("Delete tombstone object", func(t *testing.T) {
+		freshHandler := &CommonResourceEventHandler{
+			events: make(chan watch.Event, 10),
+		}
+
+		freshHandler.objToEvent(watch.Deleted, cache.DeletedFinalStateUnknown{
+			Key: "default/test-deployment",
+			Obj: testObj,
+		})
+
+		select {
+		case event := <-freshHandler.events:
+			assert.Equal(t, watch.Deleted, event.Type)
+			actualObj, ok := event.Object.(*unstructured.Unstructured)
+			if assert.True(t, ok, "Expected *unstructured.Unstructured, got %T", event.Object) {
+				assert.Equal(t, testObj.GetName(), actualObj.GetName())
+				assert.Equal(t, testObj.GetNamespace(), actualObj.GetNamespace())
+				assert.Equal(t, testObj.GetLabels(), actualObj.GetLabels())
+			}
+		case <-time.After(100 * time.Millisecond):
+			t.Fatal("No event received for tombstone object")
+		}
+	})
+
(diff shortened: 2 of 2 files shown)


RedZapdos123:
Hey @DoisLONG , Could you take a look at this PR, whenever you have some time? 

I would appreciate your review and any feedback or suggestions you may have.

Thank You.

DoisLONG (commented):
Thanks for the fix. 

I have one small test cleanup suggestion : in the invalid tombstone test, we do not need to wait for `time.After(100 * time.Millisecond)` to prove no event was sent. `objToEvent()` sends synchronously into a buffered channel, so after the call returns, a non-blocking select should be enough and avoids slowing down the test suite.

For example:

```go
select {
case <-freshHandler.events:
    t.Fatal("Unexpected event received for invalid tombstone object")
default:
}
```

Also, since this is a bug fix, please add the normal PR metadata, such as `/kind bug` and a release-note block.

RedZapdos123:
Thanks for the review. I updated the invalid tombstone test to use a non blocking select since `objToEvent()` sends synchronously into the buffered channel so the timeout was not needed.

I added the PR metadata: the kind/bug label and a release note.

DoisLONG (commented):
Thanks for the update.

The invalid tombstone test now uses a non-blocking select, and the release note has been added. The production change is focused, and the regression tests cover both the valid tombstone forwarding path and the invalid tombstone ignore path.

/lgtm

PR #7044, opened 2026-07-16
fix: handle CloudCore pod delete tombstones in iptables

### LINKED ISSUE #7039: cloud/pkg/cloudstream/iptables: pod delete tombstones skip cleanup enqueue
**What happened**:

On latest `master`, pod delete events watched by cloudstream iptables manager can be ignored when the informer sends `cache.DeletedFinalStateUnknown`.

In `cloud/pkg/cloudstream/iptables/iptables.go`, `DeleteFunc` does a direct `*v1.Pod` type assertion and returns on any other type.

`cache.DeletedFinalStateUnknown` is a valid informer delete payload. When that payload is delivered, the deleted pod is skipped and tunnel-port cleanup is never enqueued.

Pod delete handling should unwrap `cache.DeletedFinalStateUnknown` and continue cleanup for the underlying pod object.

1. Check out latest `master` at `555e57bb71fede21d3fc020e95f132132018986c`.
2. Observe current code path in `cloud/pkg/cloudstream/iptables/iptables.go`:

   - `DeleteFunc` does `pod, ok := obj.(*v1.Pod)`
   - if `ok == false`, it logs warning and returns

3. Run a focused repro such as:

   ```go
   tombstone := cache.DeletedFinalStateUnknown{Obj: &v1.Pod{}}
   _, ok := interface{}(tombstone).(*v1.Pod)
   fmt.Println(ok)
   ```

4. Actual result:

   ```text
   false
   ```

   So the current delete handler returns early and never enqueues cleanup.

5. Expected result:

   ```text
   tombstone should be unwrapped to tombstone.Obj and pod cleanup should still be enqueued
   ```

Current code on latest `master`:

- `cloud/pkg/cloudstream/iptables/iptables.go`
  - `DeleteFunc` only accepts `*v1.Pod`
  - any tombstone payload is ignored
- `cache.DeletedFinalStateUnknown`
  - is standard client-go behavior for some delete deliveries

This issue is small and isolated. It affects tunnel-port cleanup when pod delete events arrive as informer tombstones.

**Environment**:
  - Not required for this source-level repro
  - latest `master` at `555e57bb71fede21d3fc020e95f132132018986c`

    - Not required for this repro
    - Not required for this repro
  - Kernel (e.g. `uname -a`):
    - Not required for this repro
    - `go version go1.26.4 linux/amd64`
  - Others:
    - Repro is source-level and based on valid informer tombstone payload handling

  </details>

    - latest `master` at `555e57bb71fede21d3fc020e95f132132018986c`
    - Not required for this repro
    - Not required for this repro
  - Kernel (e.g. `uname -a`):
    - Not required for this repro
    - `go version go1.26.4 linux/amd64`
  - Others:
    - Repro is source-level and based on valid informer tombstone payload handling

  </details>

RedZapdos123:
## Description:

Issue #7039 reported that the cloudstream iptables manager could ignore pod delete events when the informer delivered `cache.DeletedFinalStateUnknown`.

This PR updates `cloud/pkg/cloudstream/iptables/iptables.go` to unwrap tombstone delete payloads before converting them to `*v1.Pod` in the CloudCore pod delete path, so deleted CloudCore pods are still enqueued for tunnel-port cleanup.

It also adds regression coverage in `cloud/pkg/cloudstream/iptables/iptables_test.go` for direct pod deletes, tombstone pod deletes, and invalid tombstone payloads.

The fix is minimal and limited to the affected CloudCore pod delete handling path in the iptables manager.

## Validation:

- `gofmt -w cloud/pkg/cloudstream/iptables/iptables.go cloud/pkg/cloudstream/iptables/iptables_test.go`
- `go test ./cloud/pkg/cloudstream/iptables -count=1`
- `go test -v ./cloud/pkg/cloudstream/iptables -run TestHandleCloudCorePodDelete -count=1`
- `golangci-lint run ./cloud/pkg/cloudstream/iptables --timeout=5m`

Closes #7039.

## Checklist:

- [x] I have added regression test cases for direct, tombstone, and invalid delete payloads.
- [x] I have run validation using the focused `go test` and `golangci-lint` commands above.

## Release note:

```release-note
CloudStream iptables cleanup now handles CloudCore pod delete tombstones correctly.
```

### DIFF
--- a/cloud/pkg/cloudstream/iptables/iptables.go
+++ b/cloud/pkg/cloudstream/iptables/iptables.go
@@ -273,22 +273,7 @@ func (im *Manager) listenCloudCore(ctx context.Context) {
 	_, err := podInformer.Informer().AddEventHandler(
 		cache.ResourceEventHandlerFuncs{
 			DeleteFunc: func(obj interface{}) {
-				// listen pod delete event
-				pod, ok := obj.(*v1.Pod)
-				if !ok {
-					klog.Warningf("object type: %T unsupported when listen CloudCore pod delete", obj)
-					return
-				}
-				value, ok := pod.Labels[constants.SystemName]
-				if ok && value == constants.CloudConfigMapName {
-					// only handle coudcore pod delete
-					podIP := pod.Status.PodIP
-					if len(podIP) == 0 {
-						return
-					}
-					// find deleted cloudcore pod, put it in queue
-					im.enqueuePod(pod)
-				}
+				im.handleCloudCorePodDelete(obj)
 			},
 		},
 	)
@@ -296,6 +281,29 @@ func (im *Manager) listenCloudCore(ctx context.Context) {
 		klog.Fatalf("new podInformer failed, add event handler err: %v", err)
 	}
 	podInformer.Informer().Run(ctx.Done())
+}
+
+func (im *Manager) handleCloudCorePodDelete(obj interface{}) {
+	if tombstone, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = tombstone.Obj
+	}
+
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		klog.Warningf("object type: %T unsupported when listen CloudCore pod delete", obj)
+		return
+	}
+
+	value, ok := pod.Labels[constants.SystemName]
+	if ok && value == constants.CloudConfigMapName {
+		// only handle coudcore pod delete
+		podIP := pod.Status.PodIP
+		if len(podIP) == 0 {
+			return
+		}
+		// find deleted cloudcore pod, put it in queue
+		im.enqueuePod(pod)
+	}
 }
 
 func (im *Manager) CleanCloudCoreIPPort(ctx context.Context, pod *v1.Pod) error {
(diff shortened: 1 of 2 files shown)


DoisLONG (commented):
Thanks for the fix. I took a closer look and this looks reasonable to me.

The issue is real: informer delete handlers may receive `cache.DeletedFinalStateUnknown`, and the previous code only accepted `*v1.Pod`. In that case, a CloudCore pod delete event could be ignored and the tunnel-port cleanup would not be enqueued.

The production change is focused. Moving the delete handling into `handleCloudCorePodDelete()` keeps the existing behavior for normal pod delete events, while adding the missing tombstone unwrap before the `*v1.Pod` assertion.

The tests also cover the right behavior. I like that they mock `enqueuePod` and verify the actual effect, rather than only checking that the function does not panic. The cases for direct CloudCore pod delete, tombstone CloudCore pod delete, invalid tombstone payload, non-CloudCore pod, and CloudCore pod without PodIP are all relevant.

One small thing before LGTM: could you please add the normal PR metadata, such as `/kind bug` and a release-note block? The release note can simply mention that CloudStream iptables cleanup now handles CloudCore pod delete tombstones correctly.

RedZapdos123:
Thanks for the review. I updated the PR metadata.

The kind/bug label is now there. I added the release note block.

DoisLONG:
Thanks for the update.

The PR metadata and release note have been added. The fix is focused, and the tests cover the direct delete, tombstone delete, invalid tombstone, non-CloudCore pod, and missing PodIP paths.

/lgtm

PR #7052, opened 2026-07-17
Fix goroutine leak in EdgeHub upon reconnection

### LINKED ISSUE #7051: Bug: Goroutine leak and spurious reconnects in EdgeHub on connection drop
**What happened**:
In `EdgeHub`, the `routeToCloud` and `keepalive` goroutines can permanently leak upon reconnects. Furthermore, because `reconnectChan` is unbuffered, delayed errors from these leaked goroutines can cause spurious reconnects by sending a signal that tears down a newly established, healthy connection.

The old `routeToCloud` goroutine never exits because it is blocked on `beehiveContext.Receive()`. When it finally receives a message, it uses the shared `eh.chClient` (which has been updated to the new connection) and successfully sends the message, continuing its infinite loop. After `N` reconnects, there are `N+1` `routeToCloud` goroutines leaked and competing to read from the channel.
Additionally, if the old goroutine encounters a send error *before* the new client is ready, it sends to the unbuffered `eh.reconnectChan`. This blocks the goroutine and will later trigger a spurious disconnect for the next healthy connection.

Previous `routeToCloud` and `keepalive` goroutines should exit when the connection drops, ensuring only one active instance of each goroutine per active connection.

1. Start KubeEdge (`edgecore` and `cloudcore`).
2. Disconnect the network between edge and cloud (e.g., stop `cloudcore` or block port).
3. The `routeToEdge` goroutine detects the connection failure, sends to `eh.reconnectChan`, and exits.
4. `EdgeHub.Start()` loop handles the reconnect, calls `UnInit()`, and sleeps.
5. Meanwhile, the `routeToCloud` goroutine is blocked on `beehiveContext.Receive(modules.EdgeHubModuleName)` and does not exit.
6. Once the network is restored, `Start()` creates a new connection, assigns it to `eh.chClient`, and spawns a NEW set of `routeToCloud`, `routeToEdge`, and `keepalive` goroutines.
7. Repeat the disconnect/reconnect process multiple times.

**Root cause analysis:**
1. `EdgeHub.Start()` spawns `routeToCloud` on every reconnect but provides no cancellation context to stop them.
2. `routeToCloud` only exits if `sendToCloud` returns an error. However, if it is blocked on `beehiveContext.Receive()` during the disconnect, it will not hit the error path.
3. `eh.chClient` is shared and updated on reconnect, so when the leaked `routeToCloud` eventually receives a message, it uses the new connection, succeeds, and stays alive.
4. `eh.reconnectChan <- struct{}{}` blocks if no receiver is ready. A delayed error will satisfy the `<-eh.reconnectChan` read of a *subsequent* healthy connection, causing a spurious reconnect cycle.

**Possible fix direction:**
- Pass a `context.Context` to `routeToCloud`, `routeToEdge`, and `keepalive` that gets cancelled when `reconnectChan` is triggered or when `UnInit` is called.
- Make `reconnectChan` buffered, or use a non-blocking send (`select { case eh.reconnectChan <- struct{}{}: default: }`) to prevent blocking leaked goroutines.
- Ensure old goroutines exit by checking the context instead of relying solely on network write errors.

**Environment**:
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): Latest master branch

  - Hardware configuration (e.g. `lscpu`): Any
  - OS (e.g. `cat /etc/os-release`): Any
  - Kernel (e.g. `uname -a`): Any
  - Go version (e.g. `go version`): Go 1.22+
  - Others:

  </details>

  - edgecore version (e.g. `edgecore --version`): Latest master branch
  - Hardware configuration (e.g. `lscpu`): Any
  - OS (e.g. `cat /etc/os-release`): Any
  - Kernel (e.g. `uname -a`): Any
  - Go version (e.g. `go version`): Go 1.22+
  - Others:

  </details>

vishalmore90:
### What this PR does / why we need it
This PR resolves #7051 by improving goroutine lifecycle management, context cancellation, and reconnection stability in `EdgeHub`.

#### Summary of Changes:
- **Context Lifecycle**: Passed `ctx` to `routeToEdge`, `routeToCloud`, and `keepalive` goroutines in [`edge/pkg/edgehub/process.go`](file:///c:/Users/Vishal/Code_File/Repo%20Folder/kubeedge/edge/pkg/edgehub/process.go), replacing `time.Sleep` with a selectable `time.NewTimer` listening on `ctx.Done()`.
- **Stale Stop Message Filtering**: Discarded `OperationStop` messages in `routeToCloud` to avoid forwarding stale unblocking control messages to CloudHub.
- **Non-blocking Reconnect**: Buffered `reconnectChan` to size 1 in [`edge/pkg/edgehub/edgehub.go`](file:///c:/Users/Vishal/Code_File/Repo%20Folder/kubeedge/edge/pkg/edgehub/edgehub.go) and used non-blocking `select-default` channel sends to prevent deadlocks.
- **Unit Tests**: Added 4 unit tests in [`edge/pkg/edgehub/process_test.go`](file:///c:/Users/Vishal/Code_File/Repo%20Folder/kubeedge/edge/pkg/edgehub/process_test.go) for context cancellation, message filtering, and channel buffering. Removed accidental `umask` files.

### Which issue(s) this PR fixes
Fixes #7051

### Special notes for your reviewer
- All unit tests in `edge/pkg/edgehub` pass cleanly (`go test -v ./edge/pkg/edgehub/...`).

### Does this PR introduce a user-facing change?
```release-note
NONE

### DIFF
--- a/edge/pkg/edgehub/edgehub.go
+++ b/edge/pkg/edgehub/edgehub.go
@@ -1,6 +1,7 @@
 package edgehub
 
 import (
+	"context"
 	"sync"
 	"time"
 
@@ -10,6 +11,7 @@ import (
 	"github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2"
 	"github.com/kubeedge/beehive/pkg/core"
 	beehiveContext "github.com/kubeedge/beehive/pkg/core/context"
+	messagepkg "github.com/kubeedge/kubeedge/edge/pkg/common/message"
 	"github.com/kubeedge/kubeedge/edge/pkg/common/modules"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/certificate"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub/clients"
@@ -46,7 +48,7 @@ func newEdgeHub(enable bool) *EdgeHub {
 	NewCertSyncChannel()
 	return &EdgeHub{
 		enable:        enable,
-		reconnectChan: make(chan struct{}),
+		reconnectChan: make(chan struct{}, 1),
 		rateLimiter: flowcontrol.NewTokenBucketRateLimiter(
 			float32(config.Config.EdgeHub.MessageQPS),
 			int(config.Config.EdgeHub.MessageBurst)),
@@ -122,13 +124,20 @@ func (eh *EdgeHub) Start() {
 		}
 		// execute hook func after connect
 		eh.pubConnectInfo(true)
-		go eh.routeToEdge()
-		go eh.routeToCloud()
-		go eh.keepalive()
+		ctx, cancel := context.WithCancel(context.Background())
+		go eh.routeToEdge(ctx)
+		go eh.routeToCloud(ctx)
+		go eh.keepalive(ctx)
 
 		// wait the stop signal
 		// stop authinfo manager/websocket connection
 		<-eh.reconnectChan
+		cancel()
+
+		// Send a stop message to unblock routeToCloud if it's waiting on Receive()
+		stopMsg := messagepkg.BuildMsg(modules.HubGroup, "", modules.EdgeHubModuleName, "", messagepkg.OperationStop, nil)
+		beehiveContext.Send(modules.EdgeHubModuleName, *stopMsg)
+
 		eh.chClient.UnInit()
 
 		// execute hook fun after disconnect
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -39,18 +39,24 @@ func (eh *EdgeHub) dispatch(message model.Message) error {
 	return msghandler.ProcessHandler(message, eh.chClient)
 }
 
-func (eh *EdgeHub) routeToEdge() {
+func (eh *EdgeHub) routeToEdge(ctx context.Context) {
 	for {
 		select {
 		case <-beehiveContext.Done():
 			klog.Warning("EdgeHub RouteToEdge stop")
+			return
+		case <-ctx.Done():
+			klog.Warning("EdgeHub RouteToEdge reconnect stop")
 			return
 		default:
 		}
 		message, err := eh.chClient.Receive()
 		if err != nil {
 			klog.Errorf("websocket read error: %v", err)
-			eh.reconnectChan <- struct{}{}
+			select {
+			case eh.reconnectChan <- struct{}{}:
+			default:
+			}
 			return
 		}
 		klog.V(4).Infof("[edgehub/routeToEdge] receive msg from cloud, msg: %+v", message)
@@ -72,11 +78,14 @@ func (eh *EdgeHub) sendToCloud(message model.Message) error {
 	return nil
 }
 
-func (eh *EdgeHub) routeToCloud() {
+func (eh *EdgeHub) routeToCloud(ctx context.Context) {
 	for {
 		select {
 		case <-beehiveContext.Done():
 			klog.Warning("EdgeHub RouteToCloud stop")
+			return
+		case <-ctx.Done():
+			klog.Warning("EdgeHub RouteToCloud reconnect stop")
 			return
 		default:
 		}
@@ -84,6 +93,18 @@ func (eh *EdgeHub) routeToCloud() {
 		if err != nil {
 			klog.Errorf("failed to receive message from edge: %v", err)
 			time.Sleep(time.Second)
+			continue
+		}
+
+		select {
+		case <-ctx.Done():
+			klog.Warning("EdgeHub RouteToCloud reconnect stop after receive")
+			return
+		default:
+		}
+
+		if message.GetOperation() == messagepkg.OperationStop {
+			klog.Warning("EdgeHub RouteToCloud received stale stop message, discarding")
 			continue
 		}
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by DoisLONG — COMMENTED** [review]
Thanks for working on this, and welcome to the project.

I do not think the current implementation fully fixes the leak described in #7051 yet.

The main leak path in the issue is that an old `routeToCloud` goroutine may be blocked on `beehiveContext.Receive()` when the connection is torn down. Passing a `context.Context` into `routeToCloud()` does not interrupt that blocking receive. In the current patch, `ctx.Done()` is checked only after a message has already been received and sent to cloud. So if the goroutine is blocked in `Receive()`, cancelling the context will not make it exit.

There is also a reconnect signal race in the current version. The PR changes reconnect sends to non-blocking sends, but `reconnectChan` is still unbuffered. If a route goroutine hits an error before `Start()` is waiting on `<-eh.reconnectChan`, the reconnect signal can be dropped and the main loop may block indefinitely. The existing Gemini comment points this out as well.

Another important point is that this PR overlaps heavily with #6967, which already changes the EdgeHub reconnect path by making `reconnectChan` buffered(1), adding non-blocking reconnect signal coalescing, draining stale reconnect signals, and changing the reconnect loop behavior. We should avoid having two independent PRs change the same reconnect semantics in different ways.

Before moving this forward, I think this PR needs to either:

1. rebase on top of #6967 and focus only on the remaining goroutine lifecycle issue, or
2. clearly explain why this should be handled separately from #6967.

It also needs regression tests that prove the intended behavior, not only signature updates. In particular, please add tests showing that old per-connection goroutines exit after reconnect/cancel, and that reconnect signals are not lost.

For now, I would not LGTM this implementation, but the problem itself is valid and worth continuing.

**Comment by vishalmore90** [comment]
**Context:** 
In `EdgeHub`, the `routeToCloud` and `keepalive` goroutines could permanently leak upon network disconnects. Additionally, the `reconnectChan` was unbuffered, causing potential race conditions that dropped reconnect signals and indefinitely blocked the main loop. 

**Changes:**
1. **Buffered `reconnectChan`**: Initialized `reconnectChan` with a buffer size of 1 to ensure reconnect signals are not dropped.
2. **Graceful Goroutine Exit (`routeToCloud`)**: Passed a dummy `stop` message to `modules.EdgeHubModuleName` when the context is cancelled to safely wake up and terminate any blocked `routeToCloud` goroutine waiting on `beehiveContext.Receive()`.
3. **Timer Leak Fix (`keepalive`)**: Replaced `time.After` with `time.NewTimer` in the `keepalive` loop to correctly stop the timer when the context is cancelled, preventing resource leaks.

**Verification:**
- Added `TestRouteToCloudGracefulExit` to confirm that `routeToCloud` terminates gracefully on `stop` message and context cancellation.
- Added `TestReconnectChanBuffered` to guarantee `reconnectChan` does not block or drop signals when the receiver is not ready.
- All `edgehub` package tests pass successfully.

Fixes #7051

**Note for your reviewer**:

This PR incorporates fixes for both the goroutine leakage and the reconnect signal race conditions. The dummy `stop` message pattern reliably resolves the blocking `beehiveContext.Receive()` call during reconnects without altering the `beehive` core framework. If PR #6967 is merged before this, I will be happy to rebase the channel buffering changes on top of it.

**Review by DoisLONG — COMMENTED** [review]
Thanks for the update.

This is improved compared with the earlier version: reconnectChan is now buffered, and keepalive no longer uses a plain time.Sleep() that ignores cancellation.

However, I still do not think this implementation is safe enough yet.

The main concern is the synthetic stop message used to unblock routeToCloud(). It is sent through the normal EdgeHub beehive queue, so it is not scoped to a specific routeToCloud goroutine or connection generation. If a stale stop message is consumed by a later healthy routeToCloud() with a non-cancelled context, the current code does not return; it can continue and send that internal stop message toward cloud. Also, if there are multiple old routeToCloud goroutines blocked on the same module queue, one stop message does not prove that all stale goroutines exit.

This PR also overlaps with the reconnect-signal work in #6967, which already changes reconnectChan buffering, reconnect signal coalescing, stale signal draining, and reconnect-loop behavior. We should avoid having two independent PRs change EdgeHub reconnect semantics in different ways.

Could you please either rebase/align this work with #6967, or narrow this PR to a clearly separate goroutine-lifecycle fix with tests proving that stale routeToCloud goroutines exit and that no internal stop message can be forwarded to cloud?

**Review by DoisLONG — COMMENTED** [review]
Thanks for the update. This addresses several of the earlier concerns: `reconnectChan` is now buffered, `keepalive` can be interrupted by the reconnect context, and the stale internal stop-message case is handled.

One lifecycle gap still remains in `routeToCloud()`.

After `beehiveContext.Receive()` returns, the code only checks the cancelled context early for `OperationStop`. If the old cancelled `routeToCloud()` goroutine is woken up by a normal business message, it can still go through `tryThrottle()` and `sendToCloud()` before noticing the context cancellation.

Could you please add a context check immediately after `Receive()` returns, before handling either stop messages or normal messages, and add a regression test covering the cancelled-context + normal-message case?

**Comment by vishalmore90** [comment]
Thanks for reviewing my PR again and clear explanation,  I've pushed an update that adds the context check immediately after Receive() returns, and I've also added a new regression test (TestRouteToCloudCancelledContextWithNormalMessage) to cover this specific scenario.

**Review by DoisLONG — COMMENTED** [review]
Thanks for the update.
please update the PR description to match the current implementation. It now also sends an internal stop message to unblock routeToCloud() and explicitly drops stale stop messages, which is not mentioned in the current “Changes” list.

**Review by DoisLONG — COMMENTED** [review]
Thanks for the update.

The reconnect lifecycle looks good to me now. `reconnectChan` is buffered, `routeToCloud()` checks the reconnect context immediately after `Receive()` returns, stale internal stop messages are discarded, and the PR description now matches the current implementation.

One final cleanup before LGTM: could you please squash the commits into a single logical commit for this PR?

I do not see any remaining code blocker from my side after the commit cleanup.

**Review by DoisLONG — COMMENTED** [review]
Thanks for the cleanup.

The commits are squashed now, the reconnect lifecycle looks good to me, and the PR description matches the current implementation. The existing review threads are also resolved.

/lgtm

**Comment by WillardHu** [comment]
Please resolve the lint issue.
```
Error: edge/pkg/edgehub/edgehub.go:136:1: File is not properly formatted (gofmt)
````

PR #7054, opened 2026-07-17
cloud/router: check GetContentData error instead of stale ok

### LINKED ISSUE #7045: cloud/router: Forward checks a stale ok variable instead of the GetContentData error (servicebus + eventbus)
**What happened**:

Both router providers check the wrong variable after `GetContentData()`, so its error is never examined.

In `cloud/pkg/router/provider/servicebus/servicebus.go`, `Forward`:

```go
message, ok := data.(*model.Message)
if !ok {                                       // servicebus.go:86
    klog.Errorf("message type %T error", data)
    return nil, fmt.Errorf("message type %T error", data)
}
res := make(map[string]interface{})
content, err := message.GetContentData()
if !ok {                                       // servicebus.go:92  <-- BUG
    klog.Errorf("get message %s content err: %v", message.GetID(), err)
    return nil, fmt.Errorf("get message %s content err: %v", message.GetID(), err)
}
```

Line 92 re-checks `ok` from the type assertion on line 85, which is necessarily `true` at that point the function returned at line 86 otherwise. The `err` returned by `GetContentData()` is therefore never checked, and the error branch is unreachable dead code. `err` is then silently overwritten by the `resp, err := target.GoToTarget(...)` assignment on line 97.

The identical block appears in `cloud/pkg/router/provider/eventbus/eventbus.go:105-106`  the same `content, err := message.GetContentData()` followed by `if !ok`, with `ok` likewise fixed to `true` by the guard on line 100. The two are evidently copy-paste siblings and should be fixed together.

`if err != nil` in both files, so a content-extraction failure is reported rather than passing an empty `content` to `GoToTarget`.

Code inspection. Note this is a latent defect rather than an observable runtime failure — see below.

Scope, stated honestly: this is not currently reachable in production. `GetContentData()` (`beehive/pkg/core/model/message.go:183`) returns early for `[]byte` and `string` content, and only fails if `json.Marshal` fails - which requires a channel, func, NaN/Inf float, or cyclic value in `Content`. Messages reaching `Forward` arrive via `beehiveContext.Receive()` in `cloud/pkg/router/listener/message.go:64`, i.e. deserialized from the wire, so their content is always re-marshalable.

The fix is still worth making: the guard is inert as written, so it provides no protection if content handling ever changes, and the dead branch is misleading to readers. `errcheck` does not flag it because the error *is* assigned — just never tested.

One-character fix in each file (`!ok` → `err != nil`).

/kind bug

Priyanshubhartistm:
**What type of PR is this?**

/kind bug

`Forward` in the servicebus and eventbus router providers calls `message.GetContentData()`, which returns `([]byte, error)`, but then checks the stale `ok` variable from the earlier `data.(*model.Message)` type assertion instead of the returned `err`:

```go
message, ok := data.(*model.Message)
if !ok {
    return nil, fmt.Errorf("message type %T error", data)
}
res := make(map[string]interface{})
content, err := message.GetContentData()
if !ok {   // always false here — err is never checked
    ...
}
```

At this point `ok` is always `true` (the type assertion already succeeded), so `if !ok` is always false / unreachable dead code and the `GetContentData` error is silently discarded before the following `resp, err := target.GoToTarget(...)` call. This PR checks `err != nil` in both providers so a content-extraction error is returned instead of silently passing empty content on to `GoToTarget`. The two blocks are identical copy-paste siblings, so both are fixed together.

Note this is a latent defect: `GetContentData` only fails on malformed message content, which does not occur for the wire-deserialized messages that currently reach `Forward`. The guard protects the code path if content handling changes.

Fixes #7045

Two-character change (`!ok` → `err != nil`) in each file. `ok` remains used afterwards in both functions (servicebus reuses it for the `*http.Response` assertion, eventbus for the `*model.Message` guard), so nothing becomes an unused variable. `go build`, `go vet`, and `golangci-lint` all pass on both packages.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/router/provider/eventbus/eventbus.go
+++ b/cloud/pkg/router/provider/eventbus/eventbus.go
@@ -103,7 +103,7 @@ func (*EventBus) Forward(target provider.Target, data interface{}) (response int
 	}
 	res := make(map[string]interface{})
 	content, err := message.GetContentData()
-	if !ok {
+	if err != nil {
 		klog.Errorf("get message %s content err: %v", message.GetID(), err)
 		return nil, fmt.Errorf("get message %s content err: %v", message.GetID(), err)
 	}
--- a/cloud/pkg/router/provider/eventbus/eventbus_test.go
+++ b/cloud/pkg/router/provider/eventbus/eventbus_test.go
@@ -12,6 +12,18 @@ import (
 	"github.com/kubeedge/kubeedge/cloud/pkg/router/constants"
 )
 
+// spyTarget is a test double for provider.Target that records whether
+// GoToTarget was invoked.  It never calls into any real infrastructure.
+type spyTarget struct {
+	called bool
+}
+
+func (s *spyTarget) Name() string { return "spy" }
+func (s *spyTarget) GoToTarget(_ map[string]interface{}, _ chan struct{}) (interface{}, error) {
+	s.called = true
+	return nil, fmt.Errorf("spy: GoToTarget should not have been called")
+}
+
 func TestPathJoin(t *testing.T) {
 	var s1, s2 string
 	s1 = fmt.Sprintf("%s/node/%s/%s/%s", "bus", "nodeName", "namespace", "subTopic")
@@ -190,6 +202,25 @@ func TestForward(t *testing.T) {
 			}
 		})
 	}
+}
+
+// TestForward_GetContentDataError is the regression test requested by PR #7054
+// review: when GetContentData() fails, Forward must return an error and must
+// never invoke target.GoToTarget().
+func TestForward_GetContentDataError(t *testing.T) {
+	eb := &EventBus{}
+	spy := &spyTarget{}
+
+	// GetContentData returns (data, nil) for string and []byte content;
+	// it calls json.Marshal for anything else.  A channel cannot be marshalled
+	// so it reliably triggers the error branch.
+	msg := model.NewMessage("")
+	msg.Content = make(chan struct{})
+
+	_, err := eb.Forward(spy, msg)
+
+	assert.Error(t, err, "Forward must return an error when GetContentData fails")
+	assert.False(t, spy.called, "GoToTarget must NOT be called when GetContentData fails")
 }
 
 func TestGoToTarget(t *testing.T) {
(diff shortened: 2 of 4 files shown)


DoisLONG (commented):
Thanks for the fix. The code change itself looks correct: after `message.GetContentData()` returns `content, err`, the code should check `err` instead of the stale `ok` value from the earlier type assertion.

However, the current PR has a CI/test coverage problem: this bugfix changes error-handling behavior, but there is no regression test covering the `GetContentData()` error path.

Please add test cases for both servicebus and eventbus `Forward()` to verify that when `GetContentData()` returns an error:

1. `Forward()` returns an error;
2. `target.GoToTarget()` is not called.

After the missing test coverage is added and CI passes, this should be straightforward to review.

Priyanshubhartistm:
@DoisLONG , thank you for the suggestion! I've incorporated the change in the latest commit
Please let me know if you have any feedback

DoisLONG (commented):
Thanks for the update. The added regression tests now cover the `GetContentData()` error path for both servicebus and eventbus, and verify that `target.GoToTarget()` is not called on failure.

One small cleanup before LGTM: the PR description still has a few broken/truncated sentences. Could you please clean it up so the issue is easier to read?

Priyanshubhartistm:
@DoisLONG thanks for the suggestion , i updated PR description
Please let me know if you have any other feedback

DoisLONG (commented):
Thanks for updating the tests and cleaning up the PR description. The code change and the regression tests look good to me.

One tiny wording nit in the PR description: the comment says `if !ok` is "always true here", but since the earlier type assertion has already succeeded, `ok` is true and `if !ok` is actually always false / unreachable at that point.

Could you please adjust that wording, for example:

```go
if !ok { // always false here — err is never checked
    ...
}

Priyanshubhartistm:
@DoisLONG ,Done. Fixed the wording in the PR description changed
could you please take another look

DoisLONG (commented):
Thanks for the update.

This looks good to me now. The PR description is clear, both servicebus and eventbus now check the GetContentData() error instead of the stale ok value, and the regression tests cover that GoToTarget() is not called when content extraction fails.

/lgtm

Priyanshubhartistm:
Thank you for the review, @DoisLONG , @WillardHu

PR #7061, opened 2026-07-19
 Fix: Prevent edgecore crash (DoS) on unsupported edgestream MessageType

### LINKED ISSUE #7060: Bug: edgecore crashes (DoS) when receiving unsupported MessageType in edgestream tunnel
**What happened**:
In `edgecore`, the `edgestream` component processes incoming messages from the tunnel using a new goroutine `go s.ServeConnection(mess)`. If a message with an unsupported `MessageType` (specifically `MessageType >= 7`) is received, `ServeConnection` falls into the `default` switch case and explicitly calls `panic(fmt.Sprintf("Wrong message type %v", m.MessageType))`. 
Because this `panic` occurs inside a newly spawned goroutine without any `defer recover()` block, it crashes the entire `edgecore` process. This introduces a Denial of Service (DoS) vulnerability that can be triggered by a malformed message, a compromised component, or simply by a version mismatch if a newer CloudCore sends a newly implemented stream message type to an older EdgeCore.

`edgecore` should gracefully log an error and ignore the unknown message type (or safely close the stream connection), allowing the daemon to continue running normally without crashing.

1. Start `edgecore` with `edgestream` enabled and establish a stream tunnel connection from `cloudcore`.
2. Send a `stream.Message` across the tunnel with a `MessageType` of `7` (or any value `>= stream.MessageTypeAttachConnect` excluding `6` which is handled).
3. The `tunnel.go` loop reads the message. The condition `(mess.MessageType >= stream.MessageTypeAttachConnect)` evaluates to true.
4. `edgecore` spawns a goroutine: `go s.ServeConnection(mess)`.
5. Inside `ServeConnection`, the `switch` statement hits the `default` case and calls `panic()`.
6. The entire `edgecore` daemon crashes with `panic: Wrong message type 7`.

- **Root cause:** In `edge/pkg/edgestream/tunnel.go` line 194, the condition `(mess.MessageType >= stream.MessageTypeAttachConnect)` acts as a catch-all for any message type greater than `6`. When `ServeConnection` handles these values, it panics without a recovery block, taking down the entire `edgecore` application.
- **Possible fix:** 
  1. Replace the `panic()` in `ServeConnection`'s `default` case with `klog.Errorf` and return.
  2. Tighten the condition before calling `go s.ServeConnection(mess)` to only allow known valid message types rather than all values `>= stream.MessageTypeAttachConnect`.

**Environment**:
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): Latest `master` branch

  - Hardware configuration (e.g. `lscpu`): Any
  - OS (e.g. `cat /etc/os-release`): Any
  - Kernel (e.g. `uname -a`): Any
  - Go version (e.g. `go version`): Any Go version
  - Others: N/A

  </details>

  - edgecore version (e.g. `edgecore --version`): Latest `master` branch
  - Hardware configuration (e.g. `lscpu`): Any
  - OS (e.g. `cat /etc/os-release`): Any
  - Kernel (e.g. `uname -a`): Any
  - Go version (e.g. `go version`): Any Go version
  - Others: N/A

  </details>

vishalmore90:
This PR resolves a Denial of Service (DoS) vulnerability in edgecore. 

Previously, if the edgestream tunnel received a stream message with an unsupported or unknown message type, the system would panic in `ServeConnection()`, causing edgecore to crash.

With these changes:
1. We replaced the `panic` with a `klog.Errorf` log and an early `return` in `ServeConnection()`, allowing edgecore to gracefully handle unknown message types without crashing.
2. We refactored the `Serve()` message dispatching logic to use a switch statement. It now explicitly handles valid message types and gracefully ignores unknown or unsupported ones, ensuring safety and better code clarity.

Fixes #7060

**Note**:
I verified that swapping `panic()` for a log statement does not leak any resources or break the stream loop, as the goroutine simply exits after the switch statement. `go build` and `go test` pass cleanly.

```release-note
Fixed a Denial of Service vulnerability in edgecore where an unsupported edgestream message type could crash the entire daemon.
```

### DIFF
--- a/edge/pkg/edgestream/tunnel.go
+++ b/edge/pkg/edgestream/tunnel.go
@@ -124,7 +124,8 @@ func (s *TunnelSession) ServeConnection(m *stream.Message) {
 			klog.Errorf("Serve Attach connection error %s", m.String())
 		}
 	default:
-		panic(fmt.Sprintf("Wrong message type %v", m.MessageType))
+		klog.Errorf("Wrong message type %v", m.MessageType)
+		return
 	}
 
 	s.DeleteLocalConnection(m.ConnectID)
@@ -191,10 +192,18 @@ func (s *TunnelSession) Serve() error {
 			return fmt.Errorf("close tunnel stream connection, error:%s", string(mess.Data))
 		}
 
-		if (mess.MessageType < stream.MessageTypeData) || (mess.MessageType >= stream.MessageTypeAttachConnect) {
+		switch mess.MessageType {
+		case stream.MessageTypeLogsConnect,
+			stream.MessageTypeExecConnect,
+			stream.MessageTypeMetricConnect,
+			stream.MessageTypeAttachConnect:
 			go s.ServeConnection(mess)
+		case stream.MessageTypeData,
+			stream.MessageTypeRemoveConnect:
+			s.WriteToLocalConnection(mess)
+		default:
+			klog.Errorf("Wrong message type %v", mess.MessageType)
 		}
-		s.WriteToLocalConnection(mess)
 	}
 }
 
--- a/edge/pkg/edgestream/tunnel_test.go
+++ b/edge/pkg/edgestream/tunnel_test.go
@@ -1,0 +1,153 @@
+package edgestream
+
+import (
+	"bytes"
+	"io"
+	"testing"
+	"time"
+
+	"github.com/gorilla/websocket"
+
+	"github.com/kubeedge/kubeedge/pkg/stream"
+)
+
+type mockEdgedConnection struct {
+	cleanCalled bool
+	closeCalled bool
+	cachedMsgs  []*stream.Message
+}
+
+func (m *mockEdgedConnection) CreateConnectMessage() (*stream.Message, error) {
+	return nil, nil
+}
+func (m *mockEdgedConnection) Serve(tunnel stream.SafeWriteTunneler) error {
+	return nil
+}
+func (m *mockEdgedConnection) CacheTunnelMessage(msg *stream.Message) {
+	m.cachedMsgs = append(m.cachedMsgs, msg)
+}
+func (m *mockEdgedConnection) GetMessageID() uint64 {
+	return 1
+}
+func (m *mockEdgedConnection) CloseReadChannel() {
+	m.closeCalled = true
+}
+func (m *mockEdgedConnection) CleanChannel() {
+	m.cleanCalled = true
+}
+func (m *mockEdgedConnection) String() string {
+	return "mockEdgedConnection"
+}
+
+type mockTunneler struct {
+	messages []*stream.Message
+	index    int
+}
+
+func (m *mockTunneler) WriteMessage(message *stream.Message) error {
+	return nil
+}
+func (m *mockTunneler) WriteControl(messageType int, data []byte, deadline time.Time) error {
+	return nil
+}
+func (m *mockTunneler) Close() error {
+	return nil
+}
+func (m *mockTunneler) NextReader() (int, io.Reader, error) {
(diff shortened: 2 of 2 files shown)


DoisLONG (commented):
Thanks for working on this.

Replacing the panic is the right direction, but the current version should not continue past the default case. If an unsupported message type carries a ConnectID that matches an existing local connection, the code will still fall through and call s.DeleteLocalConnection(m.ConnectID), which can close an unrelated active connection.

Please return immediately from the default case after logging the unsupported message type.

It would also be good to add a small regression test covering this behavior: an unsupported message type should not panic, and it should not delete an existing local connection with the same ConnectID.

vishalmore90:
> Thanks for working on this.
> 
> Replacing the panic is the right direction, but the current version should not continue past the default case. If an unsupported message type carries a ConnectID that matches an existing local connection, the code will still fall through and call s.DeleteLocalConnection(m.ConnectID), which can close an unrelated active connection.
> 
> Please return immediately from the default case after logging the unsupported message type.
> 
> It would also be good to add a small regression test covering this behavior: an unsupported message type should not panic, and it should not delete an existing local connection with the same ConnectID.

Now, I have updated the code and added regression tests

DoisLONG (commented):
Thanks for the update. Returning from the default case in ServeConnection fixes the panic path and also avoids deleting an existing local connection with the same ConnectID.

One remaining concern is the dispatch logic in Serve().

Currently, Serve() still treats any message type greater than or equal to MessageTypeAttachConnect as a connection message. That means an unknown message type no longer panics after this PR, but it can still continue to WriteToLocalConnection(mess).

If the malformed message uses a ConnectID that matches an active local connection, the unsupported message may still be cached into that connection.

Could you also tighten the dispatch logic so that only known connection message types are passed to ServeConnection, and unknown message types are logged and ignored before WriteToLocalConnection?

The exact structure is up to you, but the important behavior should be:

- known connection message types should go to ServeConnection;
- data/remove messages should go to WriteToLocalConnection;
- unknown message types should not panic;
- unknown message types should not delete an existing connection;
- unknown message types should not be forwarded into an active local connection.

It would also be helpful to update the regression test to cover the real Serve() dispatch path if possible, or at least add a test that verifies unknown message types are ignored before reaching WriteToLocalConnection.

vishalmore90:
Hi @DoisLONG , Thanks for the guidance,

I have updated the dispatch logic in `Serve()` using an explicit `switch`:
- **Known connection types** (`Logs`, `Exec`, `Metric`, `Attach`) -> `go s.ServeConnection(mess)`
- **Data & Remove types** (`Data`, `RemoveConnect`) -> `s.WriteToLocalConnection(mess)`
- **Unknown types** -> logged and ignored before reaching `WriteToLocalConnection` or `ServeConnection`

I've also added unit tests in `tunnel_test.go` covering the full `Serve()` loop to verify that unknown message types are safely ignored without being cached.

The PR has been updated.

DoisLONG (commented):
Thanks for the update.

Two small cleanups:
1. Please resolve the old review thread since the requested `return` has been added.
2. Please update the PR note. It still says the goroutine “simply exits after the switch statement”, but the current PR also changes the `Serve()` dispatch behavior, so the description should match the final implementation.

vishalmore90:
> Thanks for the update.
> 
> Two small cleanups:
> 
> 1. Please resolve the old review thread since the requested `return` has been added.
> 2. Please update the PR note. It still says the goroutine “simply exits after the switch statement”, but the current PR also changes the `Serve()` dispatch behavior, so the description should match the final implementation.

Done.

DoisLONG (commented):
/lgtm

PR #7063, opened 2026-07-19
Fix silent pod status unmarshal errors in upstream controller

### LINKED ISSUE #7062: Bug: CloudCore silently swallows json.Unmarshal errors in unmarshalPodStatusMessage
**What happened**:
In `cloud/pkg/edgecontroller/controller/upstream.go`, the `unmarshalPodStatusMessage` function receives and parses pod status messages from edge nodes. However, it completely swallows `json.Unmarshal` errors without logging them.

```go
	if name, _ := messagelayer.GetResourceName(msg); name == "" {
		// multi pod status in one message
		_ = json.Unmarshal(data, &podStatuses) // <-- Error swallowed here
		return
	}

	// one pod status per message
	var status edgeapi.PodStatusRequest
	if err := json.Unmarshal(data, &status); err != nil {
		return // <-- Error swallowed here without any logs
	}
```

If an edge node sends a malformed pod status payload, or if there is a version mismatch resulting in an unmarshaling failure, CloudCore silently drops the pod status update. This missing visibility makes debugging edge-to-cloud synchronization issues nearly impossible.

CloudCore should log a warning or error message indicating that it failed to unmarshal the pod status message (e.g., `klog.Warningf("message: %s process failure, unmarshal content data failed with error: %s", msg.GetID(), err)`).

1. Start CloudCore and EdgeCore.
2. Intercept or mock an EdgeCore pod status message sent to CloudCore, injecting invalid JSON into the message content (e.g., `{"UID": "xxx", "Name": ...` truncated).
3. Observe the CloudCore logs. No log output is produced for the failure, and the pod status simply isn't synchronized.

**Environment**:
- Kubernetes version (use `kubectl version`): (Any)
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): Latest master (Commit 1910b0f8e)

  - Hardware configuration (e.g. `lscpu`): (Any)
  - OS (e.g. `cat /etc/os-release`): (Any)
  - Kernel (e.g. `uname -a`): (Any)
  - Go version (e.g. `go version`): Go 1.22+
  - Others:

  </details>

  - edgecore version (e.g. `edgecore --version`): (Any)
  - Hardware configuration (e.g. `lscpu`): (Any)
  - OS (e.g. `cat /etc/os-release`): (Any)
  - Kernel (e.g. `uname -a`): (Any)
  - Go version (e.g. `go version`): (Any)
  - Others:

  </details>

vishalmore90:
This PR fixes an issue in CloudCore where `json.Unmarshal` errors are silently swallowed during the parsing of pod status messages (`unmarshalPodStatusMessage`). 

Previously, if CloudCore received a malformed pod status message, the error was ignored, which made troubleshooting edge-to-cloud synchronization failures difficult. Furthermore, in the multi-pod status path, failing to handle these errors properly could result in partially decoded or corrupted data being passed along for further processing.

This fix:
1. Logs a warning when `json.Unmarshal` fails for both single and multi-pod status messages, making issues traceable.
2. Explicitly drops and sets `podStatuses = nil` on unmarshal failure in the multi-pod status path, ensuring that CloudCore avoids returning or processing partially decoded, incomplete, or corrupted data.
3. Adds regression tests to cover valid and invalid inputs for both single and multi-pod status messages.

Fixes #7062

N/A

```release-note
Fix an issue in CloudCore where pod status unmarshal errors were silently ignored, and prevent partially decoded multi-pod statuses from being processed upon JSON decode failure.
```

### DIFF
--- a/cloud/pkg/edgecontroller/controller/upstream.go
+++ b/cloud/pkg/edgecontroller/controller/upstream.go
@@ -1474,13 +1474,18 @@ func (uc *UpstreamController) unmarshalPodStatusMessage(msg model.Message) (ns s
 
 	if name, _ := messagelayer.GetResourceName(msg); name == "" {
 		// multi pod status in one message
-		_ = json.Unmarshal(data, &podStatuses)
+		err = json.Unmarshal(data, &podStatuses)
+		if err != nil {
+			klog.Warningf("message: %s process failure, unmarshal content data with error: %s", msg.GetID(), err)
+			podStatuses = nil
+		}
 		return
 	}
 
 	// one pod status per message
 	var status edgeapi.PodStatusRequest
 	if err := json.Unmarshal(data, &status); err != nil {
+		klog.Warningf("message: %s process failure, unmarshal content data with error: %s", msg.GetID(), err)
 		return
 	}
 	podStatuses = append(podStatuses, status)
--- a/cloud/pkg/edgecontroller/controller/upstream_test.go
+++ b/cloud/pkg/edgecontroller/controller/upstream_test.go
@@ -1173,3 +1173,86 @@ func TestUpdatePodStatus(t *testing.T) {
 		t.Errorf("Pod phase mismatch, expected %s, got %s", corev1.PodRunning, updatedPod.Status.Phase)
 	}
 }
+
+func TestUnmarshalPodStatusMessage(t *testing.T) {
+	uc := &UpstreamController{}
+
+	// Case 1: Multi-pod status valid unmarshal
+	multiStatuses := []edgeapi.PodStatusRequest{
+		{Name: "pod1", UID: types.UID("uid-1")},
+		{Name: "pod2", UID: types.UID("uid-2")},
+	}
+	multiData, err := json.Marshal(multiStatuses)
+	if err != nil {
+		t.Fatalf("Failed to marshal multi-pod statuses: %v", err)
+	}
+
+	msgMulti := model.Message{
+		Header: model.MessageHeader{ID: "msg-1"},
+		Router: model.MessageRoute{
+			Resource: "node/node1/default/podstatus",
+		},
+		Content: string(multiData),
+	}
+
+	ns, res := uc.unmarshalPodStatusMessage(msgMulti)
+	if ns != "default" {
+		t.Errorf("expected namespace 'default', got '%s'", ns)
+	}
+	if len(res) != 2 {
+		t.Errorf("expected 2 pod statuses, got %d", len(res))
+	}
+
+	// Case 2: Multi-pod status invalid JSON (verify podStatuses is set to nil)
+	msgMultiInvalid := model.Message{
+		Header: model.MessageHeader{ID: "msg-2"},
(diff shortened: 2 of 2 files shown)


DoisLONG (commented):
Thanks for the fix.

Logging the unmarshal error is useful, but for the multi-pod status path we should also avoid returning partially decoded data if json.Unmarshal fails. Otherwise CloudCore may still process incomplete or corrupted pod statuses.

Please set podStatuses = nil when unmarshalling the multi-pod payload fails.

It would also be helpful to add a small regression test for malformed multi-pod status content, verifying that the function does not return any partially decoded pod statuses when unmarshal fails.

vishalmore90:
> Thanks for the fix.
> 
> Logging the unmarshal error is useful, but for the multi-pod status path we should also avoid returning partially decoded data if json.Unmarshal fails. Otherwise CloudCore may still process incomplete or corrupted pod statuses.
> 
> Please set podStatuses = nil when unmarshalling the multi-pod payload fails.
> 
> It would also be helpful to add a small regression test for malformed multi-pod status content, verifying that the function does not return any partially decoded pod statuses when unmarshal fails.

Thanks for review, I've updated the code to explicitly set podStatuses = nil in the error block .

DoisLONG (commented):
Thanks for the update. Setting `podStatuses = nil` addresses the concern about returning partially decoded data in the multi-pod status path.

Could you also add a small regression test before we move forward? The test can use malformed multi-pod status content and verify that `unmarshalPodStatusMessage` does not return any pod statuses when `json.Unmarshal` fails.

This would help prevent the same silent/partial-unmarshal behavior from regressing later.

vishalmore90:
Thank you for the review — if `json.Unmarshal` fails midway through parsing a multi-pod status message, returning a partially populated slice could lead to invalid status processing downstream. 

I have updated the code to explicitly reset `podStatuses = nil` whenever unmarshaling fails for multi-pod status messages, and added unit test coverage in `upstream_test.go` to verify this behavior.

DoisLONG (commented):
Thanks, this looks good overall.

The unmarshal failure path now avoids returning partially decoded multi-pod status data, and the added tests cover both valid and invalid single/multi pod status messages.

Before LGTM, could you please resolve the existing review thread and update the PR description to mention that the fix also drops partially decoded multi-pod status data on unmarshal failure?

vishalmore90:
> Thanks, this looks good overall.
> 
> The unmarshal failure path now avoids returning partially decoded multi-pod status data, and the added tests cover both valid and invalid single/multi pod status messages.
> 
> Before LGTM, could you please resolve the existing review thread and update the PR description to mention that the fix also drops partially decoded multi-pod status data on unmarshal failure?

Done.

DoisLONG (commented):
/lgtm

WillardHu:
Please resove the lint issues. https://github.com/kubeedge/kubeedge/actions/runs/29758389679/job/91884522073?pr=7063

DoisLONG (commented):
/lgtm

PR #7067, opened 2026-07-20
test: add unit tests for pure functions across components

Norway-02:
**What type of PR is this?**
/kind test

**What this PR does / why we need it**:
This PR groups several small, pure-function unit tests across different components to reduce review overhead (addressing maintainer feedback to group coverage-only changes).

It adds 100% statement coverage for the following models and utility functions:
- `edge/pkg/metamanager/dao/models/` (`eventbus.go`, `servicebus.go`, `meta.go`, `device.go`): Tests pure `TableName()` receiver functions.
- `edge/pkg/metamanager/metaserver/kubernetes/fakers/convertor.go`: Tests the pure `fakeObjectConvertor` implementation.
- `cloud/pkg/common/messagelayer/nodetask.go`: Tests the pure `BuildNodeTaskRouter` function.

All tests are clean, pure-function tests that do not mutate global state and do not require any production code modifications.

**Which issue(s) this PR fixes**:
NONE

**Special notes for your reviewer**:
This replaces multiple individual PRs to consolidate review efforts as requested.

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #7069, opened 2026-07-20
test: add unit tests for version info generation

Norway-02:
**What type of PR is this?**
/kind test

This PR adds a 100% statement coverage unit test for the pure function `Get()` in `pkg/version/version.go`, verifying that all variables from `base.go` and the `runtime` package are perfectly mapped to the returned `apimachineryversion.Info` struct.

NONE

This is a small, pure-function coverage PR that explicitly does not touch any production code.

```release-note
NONE
```

### DIFF
--- a/pkg/version/version_test.go
+++ b/pkg/version/version_test.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


Norway-02:
Hi @DoisLONG, here is the second of the two small pure-function unit test PRs. This explicitly adds simple, pure-function coverage to the version package without modifying any production logic!
/assign @DoisLONG @WillardHu

DoisLONG (commented):
Thanks for adding the version package test. The test itself is straightforward and I do not see any code issue here.

One scope/process question before moving forward: since this is another very small pure-function coverage-only change, could we fold it into the consolidated pure-function test PR (#7067) instead of keeping it as a separate PR?

The goal of #7067 was to reduce review overhead by grouping these small coverage-only tests. This version `Get()` test seems to fit that same category: it does not change production logic, does not mutate global state, and only verifies a simple field mapping.

If there is a reason to keep this separate, please mention it in the PR. Otherwise, I would prefer consolidating it with #7067.

Norway-02:
Done! I've folded this into #7067 as requested to reduce review overhead. Closing this PR. Thanks @DoisLONG!

PR #7074, opened 2026-07-21
fix(cloud/common): fix a nil panic in fakeManager.GetLister() & add 100% test coverage

Norway-02:
**What type of PR is this?**
/kind test
/kind bug

This PR accomplishes two things:
1. **Fixes a nil panic** in `fakeManager.GetLister()` where an unhandled `nil` `informerPair` could cause a panic if the GVR was not found. We now safely return `nil` with an error for unsupported GVRs.
2. **Adds 100% statement coverage** for the pure helper functions in `cloud/pkg/common/context` and `cloud/pkg/common/informers`. 

It groups these similar pure-function coverage tests into a single PR to reduce review overhead.

NONE

This follows the pattern of grouping similar test packages into a single PR and provides a crucial nil-guard bug fix to prevent panics in the fake informer manager.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/common/context/context_test.go
+++ b/cloud/pkg/common/context/context_test.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package context
+
+import (
+	"context"
+	"testing"
+
+	authenticationv1 "k8s.io/api/authentication/v1"
+	"k8s.io/kubernetes/cmd/kubeadm/app/constants"
+
+	"github.com/kubeedge/beehive/pkg/core/model"
+)
+
+func TestWithEdgeNode(t *testing.T) {
+	nodeID := "test-node"
+	ctx := context.Background()
+
+	newCtx := WithEdgeNode(ctx, nodeID)
+
+	userVal := newCtx.Value(authenticationv1.ImpersonateUserHeader)
+	if userVal != constants.NodesUserPrefix+nodeID {
+		t.Errorf("WithEdgeNode() user = %v, want %v", userVal, constants.NodesUserPrefix+nodeID)
+	}
+
+	groupVal := newCtx.Value(authenticationv1.ImpersonateGroupHeader)
+	if groupVal != constants.NodesGroup {
+		t.Errorf("WithEdgeNode() group = %v, want %v", groupVal, constants.NodesGroup)
+	}
+}
+
+func TestFromMessage(t *testing.T) {
+	tests := []struct {
+		name     string
+		resource string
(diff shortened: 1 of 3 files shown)


Norway-02:
/assign @DoisLONG @WillardHu
Hi @DoisLONG, here is PR 1, fixing a nil panic in GetLister and adding full statement coverage for the cloud common helpers!

DoisLONG (commented):
Thanks for grouping these small cloud/common tests together.

The `GetLister()` nil guard looks reasonable to me. `GetInformerPair()` can already return `nil, nil` for unsupported GVRs, so `GetLister()` should not dereference the returned pair without checking it first.

A few cleanups:

1. Please address the existing review threads in the new tests. In particular, keeping the fake manager isolated per subtest and using `t.Fatalf` / early return on unexpected errors will make the tests easier to reason about.

2. Please avoid calling this a “panic vulnerability” in the PR description. Since this is the fake informer manager, “fix a nil panic in fakeManager.GetLister()” is more accurate and less over-stated.

3. Since this is mostly test-helper behavior and does not affect user-facing runtime behavior, the `release-note` as `NONE` is fine to me.

Norway-02:
Thanks @DoisLONG! I've updated the tests to properly isolate the stateful manager inside each `t.Run` block and used `t.Fatalf` for unexpected errors. I also updated the PR title and description to drop the 'vulnerability' phrasing since, as you noted, it's just a test helper panic fix.

DoisLONG (commented):
Thanks for the update. The PR description is clearer now, and the `GetLister()` nil guard looks reasonable to me.

Only the remaining test cleanup threads need to be addressed before LGTM. In particular, please use `t.Fatalf` or return early after unexpected errors / expected nil cases, so the later assertions do not run on invalid test state.

DoisLONG (commented):
Thanks for the update.

The test cleanup looks good to me now. `TestGetInformerPair` and `TestGetLister` both create a fresh fake manager inside each subtest, and the expected error paths now return early instead of continuing into later assertions.

The code change also looks reasonable. Returning an error for unsupported GVRs avoids the nil informer-pair path that could previously lead to a panic in `GetLister()`.

A few final cleanups:

2. The old review threads are still unresolved, although they look outdated now. Please resolve them if the latest code has addressed them.
3. The PR description still says the code now safely returns `nil, nil`, but the current implementation returns `nil` with an error for unsupported GVRs. Could you please update the description to match the current behavior?

After these cleanups, this should be good from my side.

Norway-02:
I have squashed the commits into a single logical commit and updated the PR description to correctly state that it returns `nil` with an error. The test cleanups (early returns and proper isolation) were addressed in the previous commit and are now squashed into this final version. Please let me know if anything else is needed!

DoisLONG (commented):
Thanks for the update.

This looks good to me now.

The PR is now squashed into a single commit, and the description is also aligned with the current behavior: unsupported GVRs now return `nil` with an error, rather than `nil, nil`.

The code change is small and reasonable. Returning an error from `GetInformerPair()` for unsupported GVRs avoids the nil informer-pair path that could previously lead to a panic in `GetLister()`.

The test cleanup also looks good now. `TestGetInformerPair` and `TestGetLister` both create a fresh fake manager inside each subtest, and the expected error paths return early instead of continuing into later assertions.

/lgtm

Norway-02:
Hi @DoisLONG, this PR has the `lgtm` label and all your feedback (squashed commits, early-return test cleanup, accurate PR description) has been fully addressed. Could you please `/approve` when you get a chance? Thank you!

DoisLONG (commented):
/lgtm

Norway-02:

PR #7081, opened 2026-07-22
upgrade: restrict upgrade report file permissions

### LINKED ISSUE #7080: pkg/upgrade/edge writes upgrade_report.json with overly broad permissions
**What happened**:

`pkg/upgrade/edge/upgrade_reporter.go` writes `upgrade_report.json` with `os.ModePerm`, so the file is created with mode `0777` subject to umask. `edge/pkg/taskmanager/hub_connected_hooker.go` later trusts that file to report upgrade success or trigger rollback. That leaves the status artifact unnecessarily writable by other local users on hosts where the containing path is accessible.

## Summary
Upgrade status is persisted to a predictable path with overly broad file permissions, which makes the status report vulnerable to local tampering.

## Affected Area
- `pkg/upgrade/edge/upgrade_reporter.go`
- `edge/pkg/taskmanager/hub_connected_hooker.go`

`upgrade_report.json` should be created with restrictive owner-only permissions so upgrade and rollback status cannot be modified by unrelated local users.

## Expected Behaviour
Upgrade status artifacts should be owner-readable and owner-writable only, for example `0600`.

1. Run an upgrade-related path that calls `JSONFileReporter.Report` and creates `/etc/kubeedge/upgrade_report.json`.
2. Check the resulting file mode, for example with `stat -c %a /etc/kubeedge/upgrade_report.json`.
3. If the containing directory is traversable by another local user, edit the file contents as that user.
4. Restart the edge-side status reporting path.
5. Observe that KubeEdge accepts the modified payload.

## Steps to Reproduce
1. Trigger `JSONFileReporter.Report` through an upgrade or rollback flow.
2. Inspect the mode on the generated `upgrade_report.json` file.
3. Modify the file from another local account when directory permissions allow traversal.
4. Restart the edge-side upgrade status reporting path.

## Actual Behaviour
The reporter currently calls `os.WriteFile(upgradeReportJSONFile, bff, os.ModePerm)`, which maps to `0777` before umask is applied.

## Technical Evidence
- `pkg/upgrade/edge/upgrade_reporter.go` writes the report file with `os.ModePerm`.
- `edge/pkg/taskmanager/hub_connected_hooker.go` reads the same file on startup and uses its contents to report status or trigger rollback.
- A focused unit test can assert that the report file mode is no broader than `0600`.

## Suspected Root Cause
`JSONFileReporter.Report` uses `os.ModePerm` instead of a restrictive mode for a status file that should only be writable by its owner.

## Proposed Direction
Create `upgrade_report.json` with restrictive permissions such as `0600`. A follow-up hardening step could also validate ownership and mode before reading the file back.

## Acceptance Criteria
- `upgrade_report.json` is created with restrictive permissions such as `0600`.
- A unit test verifies the created file mode.
- Upgrade status reporting and rollback handling continue to work after the permission change.

**Environment**:
- Kubernetes version (use `kubectl version`): Not applicable for this code-level bug report.
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): Observed in repository source as of July 22, 2026.

  - Hardware configuration (e.g. `lscpu`): Not applicable.
  - OS (e.g. `cat /etc/os-release`): Not applicable.
  - Kernel (e.g. `uname -a`): Not applicable.
  - Go version (e.g. `go version`): Not applicable.
  - Others: Not applicable.

  </details>

  - edgecore version (e.g. `edgecore --version`): Not applicable.
  - Hardware configuration (e.g. `lscpu`): Not applicable.
  - OS (e.g. `cat /etc/os-release`): Not applicable.
  - Kernel (e.g. `uname -a`): Not applicable.
  - Go version (e.g. `go version`): Not applicable.
  - Others: Source inspection and unit-test reproduction are sufficient for this bug.

  </details>

rootp1:
**What type of PR is this?**

/kind bug

Restricts `pkg/upgrade/edge` report-file permissions so `upgrade_report.json`
is written as owner-only and rewritten back to the same restrictive mode when an
existing file is already present.

Fixes #7080

## Summary
- tighten `upgrade_report.json` writes from `os.ModePerm` to `0600`
- apply `os.Chmod` after each write so pre-existing report files are also reset
  to restrictive permissions
- add regression coverage for new and pre-existing report files

## What Changed
- update `JSONFileReporter.Report` to write and enforce owner-only permissions
- extend `upgrade_reporter_test.go` to assert the final file mode directly

## Verification
- `go test ./pkg/upgrade/edge/...` — passed
- `make verify` — failed: `verify-vendor` rewrote unrelated module files (`go.mod` and `staging/src/github.com/kubeedge/mapper-framework/go.mod`) on this macOS Go 1.26.5 host and exited with `FAILED: Vendor Verify failed`
- `make test WHAT=edge` — failed: repo-wide edge tests hit existing host/toolchain issues outside this change, including workspace resolution for `github.com/lucas-clemente/quic-go-certificates`, a `SIGBUS` in `edge/pkg/metamanager/client`, and a `gomonkey` crash in `edge/pkg/taskmanager`
- `make integrationtest` — failed: integration scripts require Linux-style local tooling and elevated access; this macOS host lacks `pidof` and `sudo` is non-interactive in the current session

## Scope
- only `pkg/upgrade/edge/upgrade_reporter.go` and `pkg/upgrade/edge/upgrade_reporter_test.go` were changed

```release-note
NONE
```

### DIFF
--- a/pkg/upgrade/edge/upgrade_reporter.go
+++ b/pkg/upgrade/edge/upgrade_reporter.go
@@ -37,6 +37,7 @@ const (
 	EventTypeUpgrade      = "Upgrade"
 	EventTypeRollback     = "Rollback"
 	EventTypeConfigUpdate = "ConfigUpdate"
+	upgradeReportFileMode = 0o600
 )
 
 var upgradeReportJSONFile = filepath.Join(constants.KubeEdgePath, "upgrade_report.json")
@@ -85,7 +86,16 @@ func (r JSONFileReporter) Report(inerr error) error {
 	if err != nil {
 		return fmt.Errorf("failed to marshal upgrade result, err: %v", err)
 	}
-	if err := os.WriteFile(upgradeReportJSONFile, bff, os.ModePerm); err != nil {
+	if _, statErr := os.Stat(upgradeReportJSONFile); statErr == nil {
+		if err := os.Chmod(upgradeReportJSONFile, upgradeReportFileMode); err != nil {
+			return fmt.Errorf("failed to tighten permissions on existing upgrade result file %s, err: %v",
+				upgradeReportJSONFile, err)
+		}
+	} else if !os.IsNotExist(statErr) {
+		return fmt.Errorf("failed to stat upgrade result file %s, err: %v",
+			upgradeReportJSONFile, statErr)
+	}
+	if err := os.WriteFile(upgradeReportJSONFile, bff, upgradeReportFileMode); err != nil {
 		return fmt.Errorf("failed to write upgrade result to file %s, err: %v",
 			upgradeReportJSONFile, err)
 	}
--- a/pkg/upgrade/edge/upgrade_reporter_test.go
+++ b/pkg/upgrade/edge/upgrade_reporter_test.go
@@ -18,6 +18,8 @@ package edge
 
 import (
 	"errors"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


DoisLONG (commented):
Thanks for the PR. The direction looks right to me: `upgrade_report.json` should not be written with world-readable permissions, and the regression tests for new and pre-existing files are useful.

One issue: the current implementation writes the report first and then calls `Chmod(0600)`. For a pre-existing overly permissive file, that means the new report content may still be visible with the old file mode during the small window before `Chmod` runs.

Could you please tighten the file mode before writing/truncating an existing report file, or use an `OpenFile`-based flow that ensures the file is owner-only before the new content is written?

Also, the bot reported an invalid commit message, so please fix that as well.

rootp1:
@DoisLONG thanks for catching that window. I moved the permission tightening ahead of the write now. The Report function stats the file first, and if it already exists it chmods it to 0600 right away, before any new content is written or truncated in. That closes the gap where a pre existing loose permission file could briefly show the new report content under the old mode.

I also fixed the commit message the bot flagged and removed the issue closing keyword from it, referencing the issue as related instead.

Ran go test on the package again and it passes with the existing regression tests still covering both the new file and pre existing file cases. Please take another look when you get a chance.

DoisLONG (commented):
Thanks for the update.

The permission window I was concerned about is addressed now. For a pre-existing report file, the code tightens the mode to `0600` before writing/truncating the new report content, and new files are also written with the restrictive mode. The regression coverage for both new and pre-existing files looks good to me.

CI still looks like it needs to be triggered/approved, but I do not see any remaining code blocker from my side.

/lgtm

rootp1:
@DoisLONG thanks again for the lgtm earlier. I had to amend the commit just now to add a Signed-off-by line so the DCO check would pass (the original commit was missing it), which changed the commit SHA and caused the bot to drop the lgtm label automatically. The code itself is exactly what you reviewed, nothing changed besides the trailer. Would you mind reapplying /lgtm when you have a moment?

rootp1:
@DoisLONG quick nudge on this one. The fix has cleared your review multiple times now and is fully tested, so if everything looks good from your side, I would really appreciate it if this could be approved and merged.

DoisLONG (commented):
/lgtm

DoisLONG (commented):
/lgtm

rootp1:
@DoisLONG the lgtm got cleared because I amended the commit to drop the body text down to a subject-only message per convention, the code itself is unchanged from what you reviewed. Could you take another look and re-add lgtm?

DoisLONG (commented):
/lgtm

PR #7082, opened 2026-07-22
metamanager: propagate local meta cache write failures instead of acking the cloud OK

### LINKED ISSUE #7070: metamanager: local meta cache write failures are swallowed while the cloud is acked OK, causing silent metaserver divergence
**What happened**:

When CloudCore sends an object add/update/delete to an edge node, metamanager writes it to the local `meta_v2` cache (the etcd-imitator that backs the edge metaserver's list/watch) via `imitator.DefaultV2Client.Inject(...)`, and then acks the cloud with `OK`. If the local cache write fails, the failure is swallowed and the cloud is still acked `OK`, so the object is never resent. The edge metaserver then serves a stale object (and emits no watch event) for that key indefinitely.

A failed local cache write should not be acked to the cloud as success. Either the ack should be withheld / an error response returned so the cloud resends, or the write should be retried locally, so the edge metaserver cache does not silently and permanently diverge from the cloud.

1. On an edge node, make the `meta_v2` write fail for one cloud originated update so `RetryInsertOrReplaceMetaV2` exhausts its retries for that message.
2. Observe that the cloud receives an `OK` response for the update.
3. `GET`/`LIST`/`watch` the object through the edge metaserver (port 10550): the object is stale and no watch event was delivered, and it stays stale until the next successful write to the same key.

Related, in the same area: `processVolume` (`process.go:421-432`) logs the `SendSync` timeout to edged and then still sends a response to the cloud with empty content, masking a failed CSI volume operation as success.

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

Priyanshu6968:
**What type of PR is this?**
/kind bug

When CloudCore sends an object add/update/delete to an edge node, metamanager writes it to the local `meta_v2` cache (the etcd imitator backing the edge metaserver's list/watch) via `imitator.Inject`, then acks the cloud with `OK`. The result of `Inject` was ignored, so when the local cache write failed (e.g. SQLite `database is locked` / disk full, after `RetryInsertOrReplaceMetaV2` exhausts its retries) the object was never resent: the metaserver kept serving the stale object with no watch event, and the cloud considered it synced. `Inject` also dropped the error from `ObjectResourceVersion`, storing an `RV=0` row and stalling the imitator revision. Separately, `processVolume` logged a `SendSync` timeout but still acked the cloud with empty content, masking a failed CSI volume operation as success.

This PR:
- makes `imitator.Inject` return an aggregated error and only trigger the watch hook on a successful write;
- checks the previously ignored `ObjectResourceVersion` error in `InsertOrUpdateObj`;
- has `processInsert` / `processUpdate` / `processDelete` report the failure via `feedbackError` instead of acking `OK`, so the cloud does not record the object as synced (edge-sourced paths are left unchanged, so local caching stays best effort for edge->cloud traffic);
- has `processVolume` report the `SendSync` timeout instead of forwarding an empty success.

Fixes #7070 

- The `Inject` error is only acted upon on the paths that previously sent a false `OK` to the cloud; edgesourced early returns (`processRemote`, edge pod delete forwarding) are untouched, so there is no behavior change for edge->cloud messages.
- The `Inject` signature change ripples to the `imitator.Client` interface and its fake.
- Added unit tests: `imitator` (Inject returns/aggregates the write error, succeeds when the write succeeds, and `InsertOrUpdateObj` rejects an unparseable resourceVersion) and `metamanager` (each of processInsert/Update/Delete feeds back the error and does not ack the cloud; processVolume reports the timeout).

```release-note
Fixed a bug where the edge metamanager acknowledged cloud object updates as successful even when the local meta cache write failed, causing the edge metaserver's list/watch to silently and permanently diverge from the cloud. Such failures are now reported so the update can be retried.

```

### DIFF
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/client.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/client.go
@@ -24,8 +24,9 @@ var Versioner = storage.APIObjectVersioner{}
 
 type Client interface {
 	// This set of functions is for metamanager
-	// Inject the msg to the backend storage
-	Inject(msg model.Message)
+	// Inject the msg to the backend storage. It returns an error if the message
+	// fails to be saved to the local cache.
+	Inject(msg model.Message) error
 	InsertOrUpdateObj(ctx context.Context, obj runtime.Object) error
 	DeleteObj(ctx context.Context, obj runtime.Object) error
 	InsertOrUpdatePassThroughObj(ctx context.Context, obj []byte, key string) error
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/fake/fake.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/fake/fake.go
@@ -12,7 +12,7 @@ import (
 
 // Client fake
 type Client struct {
-	InjectF                       func(msg model.Message)
+	InjectF                       func(msg model.Message) error
 	InsertOrUpdateObjF            func(ctx context.Context, obj runtime.Object) error
 	DeleteObjF                    func(ctx context.Context, obj runtime.Object) error
 	InsertOrUpdatePassThroughObjF func(ctx context.Context, obj []byte, key string) error
@@ -25,8 +25,8 @@ type Client struct {
 }
 
 // Inject fake
-func (c Client) Inject(msg model.Message) {
-	c.InjectF(msg)
+func (c Client) Inject(msg model.Message) error {
+	return c.InjectF(msg)
 }
 
 // InsertOrUpdateObj fake
--- a/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go
+++ b/edge/pkg/metamanager/metaserver/kubernetes/storage/sqlite/imitator/imitator.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -37,8 +38,11 @@ type imitator struct {
 }
 
 // Inject transform the message to watch.event, save internal obj/objs to table meta_v2
-// and trigger the corresponding hook to serve watch
-func (s *imitator) Inject(msg model.Message) {
+// and trigger the corresponding hook to serve watch. It returns an aggregated error
+// if any event fails to be saved, so the caller can avoid acknowledging the cloud as
+// if the local cache had been updated.
+func (s *imitator) Inject(msg model.Message) error {
+	var errs []error
 	for _, e := range s.Event(&msg) {
 		// save to meta_v2
 		var err error
@@ -50,12 +54,14 @@ func (s *imitator) Inject(msg model.Message) {
 		}
 		if err != nil {
 			key := metaserver.KeyFunc(e.Object)
-			klog.Errorf("failed to serve event {type:%v,key:%v}", e.Type, key)
+			klog.Errorf("failed to serve event {type:%v,key:%v}, err: %v", e.Type, key, err)
+			errs = append(errs, fmt.Errorf("serve event {type:%v,key:%v}: %w", e.Type, key, err))
 			continue
 		}
 		// TODO: move Trigger inside InsertOrUpdateObj and DeleteObj
 		watchhook.Trigger(e)
 	}
+	return errors.Join(errs...)
 }
 
 // TODO: filter out insert or update req that the obj's rev is smaller than the stored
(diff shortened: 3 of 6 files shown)


DoisLONG (commented):
Thanks for the PR. The problem looks real to me: if the local `meta_v2` cache write fails but metamanager still ACKs `OK` to cloud, the edge metaserver can silently keep serving stale state. Propagating the `Inject` error instead of acknowledging success is the right direction.

The main flow also looks reasonable: `Inject` now returns an error, hooks are triggered only after a successful local write, `ObjectResourceVersion` errors are no longer ignored, and the cloud-sourced insert/update/delete paths now call `feedbackError` on cache write failure.

A couple of points:

1. The PR description says edge-sourced paths are left unchanged. For insert/update this looks clear because the Edged-source paths still return before acting on `injectErr`. For delete, only the Edged pod-delete forwarding path is special-cased before the new `injectErr` check. Could you please confirm whether there are any other edge-sourced delete paths that should also remain best-effort?

2. `Inject` can now partially apply events: successful events trigger watch hooks, failed events are joined into the returned error. Please confirm that this partial-success + `feedbackError` behavior is safe with the expected cloud retry path, or add a short comment/test if there is a specific retry assumption here.

Priyanshu6968:
Thanks for reviewing.

1. Good catch. There's one more edge sourced delete: VolumeAttachment delete from Edged (client/volumeattachment.go). It doesn't hit the pod special case, so it was falling through to the injectErr check and getting failed on a cache write error, which it shouldn't.

Instead of listing resource types, I just gated the feedback on the source:


if injectErr != nil && message.GetSource() != modules.EdgedModuleName {
So all edge sourced deletes stay best effort and only cloud sourced ones propagate the failure. Reasoning: Inject writes the meta_v2 cache the metaserver serves locally for cloud state that's what should not go stale, but edge -> cloud messages are just reporting up, so a failed secondary write shouldn't block them. Did the same in processUpdate for consistency (its edge updates are status/lease, which Event already skips, so injectErr is nil there anyway). Added a test.

2. Confirmed safe. Partial apply only happens for list payloads single objects (which is what the cloud sync path sends) are all-or-nothing. On a nack, synccontroller resends the whole object (objectsync.go sendEvents), and the writes are idempotent: upsert for insert/update, delete-by-key for delete (deleting a missing key isn't an error). So replaying already applied events just refires the watch hook, which consumers already tolerate.

Priyanshu6968:
@DoisLONG please have a look

DoisLONG (commented):
Thanks for the update.

I rechecked the latest version. The previous concerns look addressed to me.

The cloud-sourced insert/update/delete paths now report the `Inject` failure instead of acknowledging success, while edge-sourced paths remain best-effort. In particular, the delete path now gates the feedback on the message source, so edge-sourced deletes such as VolumeAttachment are not blocked by a local meta_v2 cache write failure.

The partial-apply behavior in `Inject` is also clearer now. Successful events are persisted and trigger the watch hook, failed events are returned as an aggregated error, and the added test covers the case where one object in a list succeeds while another fails. The comment also explains why replaying the whole object is safe with idempotent writes.

Main CI, codespell, CIFuzz, and FOSSA are green. ARM64 is cancelled, so an approver may still want to rerun or check it before merge.

One final cleanup before LGTM: this PR currently has multiple commits. Could you please squash them into a single logical commit?

Priyanshu6968:
@DoisLONG there is only a single commit now and all the required checks are passing and 1 'not required' check is failing. Let me know if that works or i need to handle that.

DoisLONG (commented):
Thanks for the update.

The PR is now squashed into a single commit, and the code side still looks good to me.

The previous concerns are addressed: cloud-sourced insert/update/delete paths now report the `Inject` failure instead of acking success, while edge-sourced paths remain best-effort. The partial-apply behavior in `Inject` is also clearer now, and the added test covers the case where one object in a list succeeds while another fails.

One remaining thing before LGTM: I still see Main CI failing on the current head. Could you please check that failure and either fix it or confirm why it is unrelated / safe to ignore? ARM64 also still looks queued from what I can see.

Priyanshu6968:
@DoisLONG The cri-o e2e failure is not from this PR. It fails at the mapper image build,
before any test runs:

  make: *** [Makefile:296: e2e] Error 1

tests/scripts/generate_mapper.sh:63 tags a fixed name, localhost/v1.0.0:latest,
after podman load. podman no longer loads it under that name, so the tag fails and
the image is never created. Only the cri-o branch tags, so docker and isulad are
fail-fast cancellations. #7263 fixes it and is labelled kind/failing-test. This job stays red on every PR until that merges.

imitator.go also changed in this push. Inject returned an error for any failed event,
including ones where the meta_v2 key could not be built. A list payload becomes one
event per item, and those items carry no apiVersion or kind, so KeyFuncObj fails on
all of them and the message never reached edged. Inject now returns only cache write
errors. Key, encode and resourceVersion errors are logged and skipped as before,
since a resend cannot fix them. Covered by TestInjectSkipsUnrepresentableListItems.

process.go, client.go and fake.go are unchanged from your review.
RetryInsertOrReplaceMetaV2 and processVolume still return errors.

DoisLONG (commented):
Thanks for the update and the CI explanation.

/lgtm

PR #7084, opened 2026-07-22
edge/devicetwin: skip bad rows when restoring DMI cache from DB

### LINKED ISSUE #7071: edge/devicetwin: initDeviceInfoFromDB / initDeviceModelInfoFromDB / initDeviceMapperInfoFromDB stop loading on first unmarshal error
/kind bug
**What happened**:

In `edge/pkg/devicetwin/dtmanager/dmiworker.go`, all three DB-restore loops (devices, device models, device mappers) `return` on the first row that fails to unmarshal:

```go
for _, meta := range *metas {
    device := v1beta1.Device{}
    if err := json.Unmarshal([]byte(meta), &device); err != nil {
        klog.Errorf("fail to unmarshal device info from db with err: %v", err)
        return                                   // dmiworker.go:229 — aborts the whole loop
    }
    dw.dmiCache.PutDevice(&device)
}
```

Same pattern in `initDeviceModelInfoFromDB` (`return` at `dmiworker.go:211`) and in
`initDeviceMapperInfoFromDB` (`return` at `dmiworker.go:247`). One corrupt or schema-migrated row in the metadata DB silently prevents **every subsequent** device/model/mapper from being restored into the DMI cache after an edgecore restart - mappers then can't resolve those devices while offline.

Note `initDeviceMapperInfoFromDB` already gets a *later* error path right: the `CreateDMIClient` failure in the same loop (`dmiworker.go:250-253`) correctly uses `continue`. Its own unmarshal failure a few lines above still uses the broken `return`.

Skip the bad row and keep loading the rest, exactly as the `CreateDMIClient` error branch in the same function already does.

Insert two device rows into the meta DB, corrupt the JSON of the first, restart edgecore - the second (valid) device is never loaded into the DMI cache. Same reproduction for device models and device mappers.

Fix is `return` → `continue` in all three unmarshal-error branches (`dmiworker.go:211`, `:229`, `:247`). Also worth noting: the trailing `klog.Infoln("success to init ...")` in each function currently prints even when the load aborted halfway.

Priyanshubhartistm:
**What type of PR is this?**

/kind bug

The three DB-restore loops in `edge/pkg/devicetwin/dtmanager/dmiworker.go`  `initDeviceModelInfoFromDB`, `initDeviceInfoFromDB` and `initDeviceMapperInfoFromDB`  `return` on the first row that fails to `json.Unmarshal`. As a result, a single corrupt or schema-migrated row in the metadata DB silently prevents **every subsequent** device, device model or device mapper from being restored into the DMI cache after an edgecore restart, leaving mappers unable to resolve those devices
while offline.

This PR changes the unmarshal-error branch in each of the three loops from `return` to `continue`, so a bad row is logged and skipped while the remaining rows are still loaded. This matches the `CreateDMIClient` error path already present in `initDeviceMapperInfoFromDB`, which correctly uses `continue`.

Fixes #7071

Minimal three-line change (`return` → `continue`) in the unmarshal-error branch of each of the three functions. No test is added: `MetaService` wraps an unexported `*gorm.DB` obtained from `dao.GetDB()` with no injection point, so exercising these loops would require adding production code (a DB setter) or a full sqlite `InitDBManager` harness  both beyond the minimal scope of this fix. Verified locally with `gofmt`, `go vet`, `golangci-lint`, `make verify` and
`go test ./edge/pkg/devicetwin/dtmanager/...` (all pass).

```release-note
edge/devicetwin: fix DMI cache restore aborting on the first unreadable metadata DB row, which prevented later devices, device models and device mappers from being loaded after an edgecore restart.
```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/dmiworker.go
+++ b/edge/pkg/devicetwin/dtmanager/dmiworker.go
@@ -209,7 +209,7 @@ func (dw *DMIWorker) initDeviceModelInfoFromDB() {
 		deviceModel := v1beta1.DeviceModel{}
 		if err := json.Unmarshal([]byte(meta), &deviceModel); err != nil {
 			klog.Errorf("fail to unmarshal device model info from db with err: %v", err)
-			return
+			continue
 		}
 		dw.dmiCache.PutDeviceModel(&deviceModel)
 	}
@@ -227,7 +227,7 @@ func (dw *DMIWorker) initDeviceInfoFromDB() {
 		device := v1beta1.Device{}
 		if err := json.Unmarshal([]byte(meta), &device); err != nil {
 			klog.Errorf("fail to unmarshal device info from db with err: %v", err)
-			return
+			continue
 		}
 		dw.dmiCache.PutDevice(&device)
 	}
@@ -245,7 +245,7 @@ func (dw *DMIWorker) initDeviceMapperInfoFromDB() {
 		deviceMapper := pb.MapperInfo{}
 		if err := json.Unmarshal([]byte(meta), &deviceMapper); err != nil {
 			klog.Errorf("fail to unmarshal device mapper info from db with err: %v", err)
-			return
+			continue
 		}
 
 		err := dmiclient.DMIClientsImp.CreateDMIClient(deviceMapper.Protocol, string(deviceMapper.Address), true)


Priyanshubhartistm:
@DoisLONG could you please take a look
thanks

DoisLONG (commented):
Thanks, this looks good to me.

The change is small and focused. Skipping only the malformed DB row and continuing to restore the remaining device models, devices, and device mappers is safer than aborting the whole DMI cache restore on the first unmarshal error. It is also consistent with the existing `CreateDMIClient` error path in the mapper restore flow.

I agree that adding a direct unit test for this path would require more setup than the actual fix, so I do not think it should block this PR.

/lgtm

PR #7085, opened 2026-07-22
edge/devicetwin: guard message.Content type assertion against panic

### LINKED ISSUE #7076: edge/devicetwin: unchecked message.Content.([]byte) assertions in dtmanager can panic edgecore
**What happened**:

Four message handlers in `edge/pkg/devicetwin/dtmanager` assert `message.Content.([]byte)` in single-value form, which panics if the content is anything else:

- `device.go:83` — `dealDeviceStateUpdate`
- `device.go:161` — `dealDeviceAttrUpdate`
- `dmiworker.go:121` and `dmiworker.go:164` — `dealMetaDeviceOperation`

```go
updatedDevice, err := dttype.UnmarshalDeviceUpdate(message.Content.([]byte))
```

Beehive `model.Message.Content` is `interface{}`; modules on the same process frequently fill it with structs or `string` rather than raw bytes. Every handler already guards the outer `msg.(*model.Message)` assertion with the two-value form and returns a proper error but then trusts `Content` blindly. A single malformed message on these hot paths takes down the whole edgecore process instead of being rejected with an error.

Two-value assertion with an error return, consistent with how the same functions already handle the `*model.Message` assertion:

```go
content, ok := message.Content.([]byte)
if !ok {
    return errors.New("message content is not []byte")
}
```

Unit test — call `dealDeviceStateUpdate(ctx, "dev1", &model.Message{Content: "not-bytes"})`; panics with an interface-conversion runtime error. Same for `dealDeviceAttrUpdate` and `dealMetaDeviceOperation`.

**Distinct from existing work** (checked all open items touching these files):

- **#6761** (PR) validates `message.Content` type — but only for `dealLifeCycle` in  `communicate.go`, a different function. The four sites above are untouched.
- **#6904** / **#6951** (issue + PR) handle "silently swallowed errors" (`// TODO: handle error`)  inside `dealDeviceStateUpdate` / `dealDeviceAttrUpdate`. That is a different root cause (a swallowed error return, not a panic); its diff modifies the *later* error-handling lines and does **not** change the `message.Content.([]byte)` assertion at line 83 / 161.
- **#6957** (issue) asks for unit tests for `device.go` — a test-addition, not a fix.
- **#7001** (PR) only changes `resources[3]` → `resources[1]` in  `dmiworker.go`; it does not touch the `Content.([]byte)` assertions at lines 121 / 164.

None of these guard the four assertions reported here — this is a distinct panic-on-crash defect, not a swallowed-error or test gap.

Same fix pattern in all four sites (two-value assertion + error return). Other dtmanager callbacks (e.g. `dealTwinUpdate`) already use safe `GetContentData()`-style paths; these four predate that convention.

/kind bug

Priyanshubhartistm:
**What type of PR is this?**

/kind bug

Four message handlers in `edge/pkg/devicetwin/dtmanager` converted `message.Content` to `[]byte` using the single-value type assertion (`message.Content.([]byte)`), which panics with an interface-conversion runtime error if the content is anything else:

- `device.go:83` — `dealDeviceStateUpdate`
- `device.go:161` — `dealDeviceAttrUpdate`
- `dmiworker.go:121` and `dmiworker.go:164` — `dealMetaDeviceOperation`

`model.Message.Content` is `interface{}`, and modules on the same process frequently fill it with structs or strings rather than raw bytes. Every handler already guards the outer `msg.(*model.Message)` assertion with the two-value form and returns a proper error, but then trusted `Content` blindly. A single malformed message on these hot paths could take down the whole edgecore process instead of being rejected with an error.

This PR switches all four sites to the two-value assertion form, returning an error instead of panicking matching the pattern already used for the `*model.Message` assertion in the same functions, and by
`dealTwinUpdate`/`dealMembershipDetail` elsewhere in the package.

Fixes #7076

Same fix pattern applied at all four sites: two-value assertion + `errors.New("invalid message content")` return, consistent with the existing convention in `twin.go`/`membership.go`. In `dmiworker.go`, the two `Content.([]byte)` sites within `dealMetaDeviceOperation` share a single assertion placed before the `switch` on resource type, since both branches need identical content-type handling.

Unit tests were added to `device_test.go` and a new `dmiworker_test.go` covering the previously-panicking case (non-`[]byte` `Content`), which now returns an error instead of crashing.

```release-note
edgecore: fix panic in devicetwin dtmanager when a message's Content field is not []byte
```

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/device.go
+++ b/edge/pkg/devicetwin/dtmanager/device.go
@@ -80,7 +80,12 @@ func dealDeviceStateUpdate(context *dtcontext.DTContext, resource string, msg in
 		return errors.New("msg not Message type")
 	}
 
-	updatedDevice, err := dttype.UnmarshalDeviceUpdate(message.Content.([]byte))
+	content, ok := message.Content.([]byte)
+	if !ok {
+		return errors.New("invalid message content")
+	}
+
+	updatedDevice, err := dttype.UnmarshalDeviceUpdate(content)
 	if err != nil {
 		klog.Errorf("Unmarshal device info failed, err: %#v", err)
 		return err
@@ -158,7 +163,12 @@ func dealDeviceAttrUpdate(context *dtcontext.DTContext, resource string, msg int
 		return errors.New("msg not Message type")
 	}
 
-	updatedDevice, err := dttype.UnmarshalDeviceUpdate(message.Content.([]byte))
+	content, ok := message.Content.([]byte)
+	if !ok {
+		return errors.New("invalid message content")
+	}
+
+	updatedDevice, err := dttype.UnmarshalDeviceUpdate(content)
 	if err != nil {
 		klog.Errorf("Unmarshal device info failed, err: %#v", err)
 		return err
--- a/edge/pkg/devicetwin/dtmanager/device_test.go
+++ b/edge/pkg/devicetwin/dtmanager/device_test.go
@@ -149,6 +149,16 @@ func TestDealDeviceStateUpdate(t *testing.T) {
 			wantErr: true,
 		},
 		{
+			name:     "dealDeviceStateUpdateTest-InvalidContentType",
+			context:  dtContexts,
+			resource: "DeviceD",
+			msg:      &model.Message{Content: "not-bytes"},
+			setupMock: func(m *mocks.MockDeviceService) {
+				// No database call expected
+			},
+			wantErr: true,
+		},
+		{
 			name:     "dealDeviceStateUpdateTest-DeviceDoesNotExist",
 			context:  dtContexts,
 			resource: "DeviceB",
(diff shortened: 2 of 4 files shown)


Priyanshubhartistm:
@DoisLONG could you please take a look
thanks

DoisLONG (commented):
Thanks, the code change looks good to me.

Replacing the single-value `message.Content.([]byte)` assertions with checked assertions is the right direction here. A malformed or unexpected message content type should return an error instead of panicking edgecore, and the added tests cover the invalid-content cases for both the device update handlers and `dealMetaDeviceOperation()`.

One final cleanup: could you please squash the commits into a single logical commit for this PR?

DoisLONG (commented):
Thanks for the update.

This looks good to me now. The commits are squashed, and the fix is small and focused. Replacing the single-value `message.Content.([]byte)` assertions with checked assertions prevents malformed message content from panicking edgecore, and the added tests cover the invalid-content cases in both the device update handlers and `dealMetaDeviceOperation()`.

Most CI is already green, with ARM64 still queued, but I do not see any remaining code blocker from my side.

/lgtm

PR #7087, opened 2026-07-23
keadm: fix stale keys leaking between debug get pod/node records

### LINKED ISSUE #7086: keadm debug get: getPodsFromDatabase / getNodeFromDatabase reuse one map across json.Unmarshal in a loop, leaking stale keys between records
**What happened**:

`keadm/cmd/keadm/app/cmd/debug/get.go` declares a `map[string]interface{}` **once, outside** the loop and then `json.Unmarshal`s each record into that **same** map every iteration:

```go
func (g *GetOptions) getPodsFromDatabase(resNS string, resNames []string) ([]models.Meta, error) {
    var results []models.Meta
    podJSON := make(map[string]interface{})        // get.go:299 — declared once
    podStatusJSON := make(map[string]interface{})  // get.go:300
    ...
    for _, v := range ... {
        if err := json.Unmarshal([]byte(v.Value), &podJSON); err != nil {          // get.go:325 — reused
            return nil, err
        }
        if err := json.Unmarshal([]byte((*podStatusRecords)[0]), &podStatusJSON); err != nil {  // get.go:328
            return nil, err
        }
        podJSON["status"] = podStatusJSON["Status"]
        data, err := json.Marshal(podJSON)
        ...
    }
}
```

`json.Unmarshal` into a non-empty map does **not** clear keys absent from the new JSON  it only overwrites keys present in the incoming document. So keys from record N that record N+1 doesn't contain persist and get marshalled into record N+1's output. Same defect in `getNodeFromDatabase` `nodeJSON`/`nodeStatusJSON` at get.go:345-346, reused at get.go:371 / :374.

Each record unmarshalled into a fresh map so no state leaks between iterations.

Store two pod records where the second pod's JSON omits a top-level key the first has (or omits `status`), then run `keadm debug get pods` the second pod's output carries the first pod's stale key/status.

Fix: move `make(map[string]interface{})` **inside*t the top of each iteration on Go 1.21+), in both`getPodsFromDatabase` and `getNodeFromDatabase`.

Verified empirically  `json.Unmarshal` into a reused map does not clear stale keys (confirmed with a standalone repro, not just reading the Go docs).

Checked open PR #6777 ("Clean up keadm debug get fs same file its diff only extracts thenamespace/name filtering into a `filterMetaRecords` helper and does not touch the `podJSON`/`podStatusJSON`/`nodeJSON`/`nodeStatusJSmarshal` calls, so this bug is untouched by it.

**Environment**: code-level defect , master

/kind bug

pujitha24:
Motivation:
keadm/cmd/keadm/app/cmd/debug/get.go declared the podJSON/podStatusJSON
(and nodeJSON/nodeStatusJSON) maps once outside the per-record loop in
getPodsFromDatabase and getNodeFromDatabase, then reused the same map
across json.Unmarshal calls for every record. json.Unmarshal into a
non-empty map only overwrites keys present in the new document; it does
not clear keys absent from it. So a key present in record N (e.g. an
extra top-level field, or "status") could persist into record N+1's
marshalled output if record N+1's JSON didn't redefine that key. This
is a data-correctness bug in `keadm debug get pods`/`get nodes` output,
not a crash: nothing panics or exits, but the emitted JSON/YAML for a
later record can silently contain stale fields carried over from an
earlier, unrelated record.

Approach:
Move the map allocations inside the loop body, right before the
json.Unmarshal calls, so every record gets a fresh map. Applied to both
getPodsFromDatabase and getNodeFromDatabase, matching the fix
suggested in the issue.

Validation:
Added TestGetPodsFromDatabaseNoStaleKeys and
TestGetNodeFromDatabaseNoStaleKeys, which spin up a real in-memory
sqlite DB via dao.Init, insert two records where the second omits a
key the first has, and assert the second record's output does not
contain the first record's stale key. Confirmed both tests fail
against the pre-fix code and pass against the fix.

  go build ./keadm/...
  go test ./keadm/cmd/keadm/app/cmd/debug/... -run 'TestGetPodsFromDatabaseNoStaleKeys|TestGetNodeFromDatabaseNoStaleKeys' -v

PASS (both tests), go vet ./keadm/cmd/keadm/app/cmd/debug/... clean.

/kind bug

**What type of PR is this?**
/kind bug

Fixes a data-correctness bug where `keadm debug get pods` and
`keadm debug get nodes` could emit stale fields from a previous
record's JSON in a later, unrelated record's output, because the
unmarshal target maps were reused across loop iterations instead of
being allocated fresh per record.

Fixes #7086

None.

```release-note
Fixed a bug in `keadm debug get pods`/`keadm debug get nodes` where stale fields from a previously processed record could leak into a later record's JSON output due to a reused json.Unmarshal target map.
```

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/keadm/cmd/keadm/app/cmd/debug/get.go
+++ b/keadm/cmd/keadm/app/cmd/debug/get.go
@@ -296,8 +296,6 @@ func (g *GetOptions) queryDataFromDatabase(resType string, resNames []string) ([
 
 func (g *GetOptions) getPodsFromDatabase(resNS string, resNames []string) ([]models.Meta, error) {
 	var results []models.Meta
-	podJSON := make(map[string]interface{})
-	podStatusJSON := make(map[string]interface{})
 
 	podRecords, err := ms.QueryAllMeta("type", model.ResourceTypePod)
 	if err != nil {
@@ -322,6 +320,8 @@ func (g *GetOptions) getPodsFromDatabase(resNS string, resNames []string) ([]mod
 			results = append(results, v)
 			continue
 		}
+		podJSON := make(map[string]interface{})
+		podStatusJSON := make(map[string]interface{})
 		if err := json.Unmarshal([]byte(v.Value), &podJSON); err != nil {
 			return nil, err
 		}
@@ -342,8 +342,6 @@ func (g *GetOptions) getPodsFromDatabase(resNS string, resNames []string) ([]mod
 
 func (g *GetOptions) getNodeFromDatabase(resNS string, resNames []string) ([]models.Meta, error) {
 	var results []models.Meta
-	nodeJSON := make(map[string]interface{})
-	nodeStatusJSON := make(map[string]interface{})
 
 	nodeRecords, err := ms.QueryAllMeta("type", model.ResourceTypeNode)
 	if err != nil {
@@ -368,6 +366,8 @@ func (g *GetOptions) getNodeFromDatabase(resNS string, resNames []string) ([]mod
 			results = append(results, v)
 			continue
 		}
+		nodeJSON := make(map[string]interface{})
+		nodeStatusJSON := make(map[string]interface{})
 		if err := json.Unmarshal([]byte(v.Value), &nodeJSON); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


pujitha24:
/assign @zc2638

This has been rebased and green for a while — assigning for review whenever you have a chance.

PR #7094, opened 2026-07-23
 keadm/debug: fix stale field leakage across get records

### LINKED ISSUE #7086: keadm debug get: getPodsFromDatabase / getNodeFromDatabase reuse one map across json.Unmarshal in a loop, leaking stale keys between records

Priyanshubhartistm:
**What type of PR is this?**

/kind bug

`getPodsFromDatabase` and `getNodeFromDatabase` in `keadm/cmd/keadm/app/cmd/debug/get.go` each declared their merge maps (`podJSON`/`podStatusJSON`, `nodeJSON`/`nodeStatusJSON`) once outside the loop over database records, then reused the same map instance via `json.Unmarshal` on every iteration. `json.Unmarshal` into a non-empty map only overwrites keys present in the incoming JSON document — it does not clear keys that are absent from it. As a result, keys (and the merged `status` field) from one record persisted into the next record's output whenever the next record's JSON didn't redefine them, producing corrupted output for `keadm debug get pods` / `keadm debug get nodes`.

The fix moves the map declarations inside the loop body so each record is unmarshalled into a fresh map.

Fixes #7086

Added `get_test.go` with regression tests for both `getPodsFromDatabase` and `getNodeFromDatabase` that store two records where the second omits a field the first has, and assert the second record's output doesn't carry the first's stale field. Confirmed these tests fail against the pre-fix code and pass against the fix.

```release-note
Fixed `keadm debug get pods`/`keadm debug get nodes` returning stale fields from a previous record when a later record's JSON omits a key the previous one had.
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/debug/get.go
+++ b/keadm/cmd/keadm/app/cmd/debug/get.go
@@ -296,8 +296,6 @@ func (g *GetOptions) queryDataFromDatabase(resType string, resNames []string) ([
 
 func (g *GetOptions) getPodsFromDatabase(resNS string, resNames []string) ([]models.Meta, error) {
 	var results []models.Meta
-	podJSON := make(map[string]interface{})
-	podStatusJSON := make(map[string]interface{})
 
 	podRecords, err := ms.QueryAllMeta("type", model.ResourceTypePod)
 	if err != nil {
@@ -322,6 +320,8 @@ func (g *GetOptions) getPodsFromDatabase(resNS string, resNames []string) ([]mod
 			results = append(results, v)
 			continue
 		}
+		podJSON := make(map[string]interface{})
+		podStatusJSON := make(map[string]interface{})
 		if err := json.Unmarshal([]byte(v.Value), &podJSON); err != nil {
 			return nil, err
 		}
@@ -342,8 +342,6 @@ func (g *GetOptions) getPodsFromDatabase(resNS string, resNames []string) ([]mod
 
 func (g *GetOptions) getNodeFromDatabase(resNS string, resNames []string) ([]models.Meta, error) {
 	var results []models.Meta
-	nodeJSON := make(map[string]interface{})
-	nodeStatusJSON := make(map[string]interface{})
 
 	nodeRecords, err := ms.QueryAllMeta("type", model.ResourceTypeNode)
 	if err != nil {
@@ -368,6 +366,8 @@ func (g *GetOptions) getNodeFromDatabase(resNS string, resNames []string) ([]mod
 			results = append(results, v)
 			continue
 		}
+		nodeJSON := make(map[string]interface{})
+		nodeStatusJSON := make(map[string]interface{})
 		if err := json.Unmarshal([]byte(v.Value), &nodeJSON); err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


Priyanshubhartistm:
@DoisLONG, could you please take a look and let me know if any changes are needed?

DoisLONG (commented):
Thanks, the fix looks reasonable to me.

Moving the merge maps inside the per-record loop is the right fix here. Reusing the same map across `json.Unmarshal` calls can keep keys that are absent from the next JSON document, so this could indeed leak stale fields from one pod/node record into the next `keadm debug get` output. The added regression tests cover both pod and node records and verify that fields from the first record are not carried into the second one.

A couple of cleanup items before LGTM:

2. Main CI is currently failing. Could you please check whether the failure is related to this PR, or retrigger/fix it if needed?

DoisLONG (commented):
Thanks for the cleanup.

This looks good to me now. The PR is squashed into a single commit, and the fix correctly avoids reusing the same merge maps across database records. Creating fresh maps per record prevents fields from a previous pod/node record from leaking into the next `keadm debug get` output.

Main CI, codespell, FOSSA, and CIFuzz are green, with ARM64 still queued. I do not see any remaining code blocker from my side.

/lgtm

Priyanshubhartistm:
@Shelley-BaoYue could you please take a look.

PR #7098, opened 2026-07-24
taskmanager: secure node upgrade task logs

### LINKED ISSUE #7097: Upgrade and rollback logs use unsafe /tmp path
**What happened**:

Privileged node upgrade and rollback paths write Keadm output to a fixed `/tmp/keadm.log` file. Because `/tmp` is world-writable, using a predictable shared filename exposes local symlink risk and lets concurrent or near-concurrent upgrade tasks corrupt each other's logs.

Upgrade and rollback logs should be written to a KubeEdge-owned log directory with restrictive file permissions. Log file creation should avoid following symlink targets, and separate upgrade or rollback executions should not share a single global `/tmp/keadm.log` path.

1. On a disposable Linux host, create a symlink such as `ln -sf /root/test-target /tmp/keadm.log`.
2. Trigger a node upgrade or rollback path that opens or redirects to `/tmp/keadm.log`.
3. Inspect the symlink target after the command starts.
4. Observe that privileged task output can be written through the fixed `/tmp/keadm.log` path, or that two task executions share and overwrite or append to the same log file.

A safer static/code-level reproduction is to inspect the code locations below and confirm that all use the same fixed path under `/tmp`.

## Summary
Privileged upgrade and rollback command output is written to a predictable shared path, `/tmp/keadm.log`, instead of a private KubeEdge-owned log path.

## Affected Area
- `edge/pkg/taskmanager/actions`
- `edge/pkg/taskmanager/v1alpha1/taskexecutor`

## Steps to Reproduce
1. Inspect the node upgrade and rollback code paths listed under Technical Evidence.
2. Confirm that they open or redirect command output to `/tmp/keadm.log`.
3. On a disposable Linux host, create a symlink such as `ln -sf /root/test-target /tmp/keadm.log`.
4. Trigger an affected upgrade or rollback path and inspect whether output is written through the fixed `/tmp/keadm.log` path.

## Expected Behaviour
Upgrade and rollback logs should be written under a KubeEdge-owned directory with restrictive permissions and isolated filenames. Symlink targets should be rejected.

## Actual Behaviour
Multiple upgrade and rollback code paths use a shared fixed `/tmp/keadm.log` path, either by opening it directly or by shell redirection.

## Technical Evidence
- `edge/pkg/taskmanager/actions/nodeupgradejob.go` opens `/tmp/keadm.log` for upgrade and redirects rollback output to `/tmp/keadm.log`.
- `edge/pkg/taskmanager/v1alpha1/taskexecutor/node_upgrade.go` opens `/tmp/keadm.log` for v1alpha1 upgrade logs.
- `edge/pkg/taskmanager/v1alpha1/taskexecutor/node_rollback.go` redirects rollback output to `/tmp/keadm.log`.

## Suspected Root Cause
The task runners hardcode a shared `/tmp/keadm.log` path instead of creating log files in a KubeEdge-owned directory with symlink-safe file creation semantics.

## Proposed Direction
Move the logs to a private KubeEdge log directory, use restrictive file modes, create log files in a way that rejects symlink targets, and give upgrade/rollback paths isolated log filenames. Where possible, pass an open log file to `exec.Command` instead of relying on shell redirection.

## Acceptance Criteria
- Upgrade and rollback actions no longer write to a shared `/tmp/keadm.log` path.
- Concurrent jobs use isolated log files.
- Regression tests verify the new path handling and symlink rejection.

**Environment**:
- Kubernetes version (use `kubectl version`): Not applicable; this is visible in current repository code paths.
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): Affected current `master`; exact released versions not yet bounded.

  - Hardware configuration (e.g. `lscpu`): Not applicable.
  - OS (e.g. `cat /etc/os-release`): Not applicable.
  - Kernel (e.g. `uname -a`): Not applicable.
  - Go version (e.g. `go version`): Not applicable.
  - Others: Issue is in edge-side upgrade and rollback log handling.

  </details>

  - edgecore version (e.g. `edgecore --version`): Affected current repository state; exact release range not yet bounded.
  - Hardware configuration (e.g. `lscpu`): Not applicable.
  - OS (e.g. `cat /etc/os-release`): Linux edge nodes using node upgrade or rollback tasks.
  - Kernel (e.g. `uname -a`): Not applicable.
  - Go version (e.g. `go version`): Not applicable.
  - Others: Local exploitability depends on local access to the edge node and privileges of the task process.

  </details>

rootp1:
## Summary
- Move keadm upgrade and rollback logs from the shared `/tmp/keadm.log` path to private KubeEdge log files.
- Add symlink rejection and restrictive log file permissions for node task logs.

## Linked Issue
Fixes #7097

## What Changed
- Added `pkg/nodetask/tasklog` for safe Keadm task log creation under `constants.KubeEdgeLogPath`.
- Updated v1alpha1 and v1alpha2 node upgrade/rollback paths to use per-task log filenames.
- Replaced rollback shell redirection to `/tmp/keadm.log` with explicit command argv and attached stdout/stderr.
- Added focused regression tests for private log creation, symlink rejection, log naming, and changed upgrade/rollback paths.

## Verification
- `go test ./pkg/nodetask/tasklog` — passed
- `go test ./edge/pkg/taskmanager/actions -run '^(TestNodeUpgradeJobUpgrade|TestNodeUpgradeJobRollback|TestNodeUpgradeJobLogName)$'` — passed\n- `go test ./edge/pkg/taskmanager/v1alpha1/taskexecutor` — passed\n- `go test ./pkg/nodetask/tasklog ./edge/pkg/taskmanager/v1alpha1/taskexecutor ./edge/pkg/taskmanager/actions -run '^(TestOpenKeadmLogAt|TestSafeName|TestBuildKeadmUpgradeArgsDoesNotUseShell|TestKeadmUpgradeReturnsErrorWhenCommandMissing|TestV1alpha1KeadmLogName|TestKeadmUpgradeWritesPrivateTaskLog|TestNodeUpgradeJobUpgrade|TestNodeUpgradeJobRollback|TestNodeUpgradeJobLogName)$'` — passed\n- `make verify` — passed\n- `make lint` — not completed: installed required `golangci-lint` v1.64.5 and passed whitespace checking after staging intended files, but the root `golangci-lint` process hung after reporting 12m execution time and was interrupted.\n- `PATH="$HOME/go/bin:$PATH" golangci-lint run --timeout=3m ./pkg/nodetask/tasklog ./edge/pkg/taskmanager/actions ./edge/pkg/taskmanager/v1alpha1/taskexecutor` — not completed: focused lint also hung past the requested timeout and was interrupted.\n- `make edge_test` — not run: target is not defined in this Makefile.\n- `make test WHAT=edge` — failed: unrelated existing/local failures in `edge/pkg/metamanager/client` runtime crash and `edge/pkg/taskmanager/actions` imageprepull/pre-check tests; changed `taskexecutor` package passed in this run.\n- `make edge_integration_test` — not run: target is not defined in this Makefile.\n- `make integrationtest WHAT=edge` — failed: local environment lacks passwordless `sudo` and `pidof`; `edgecore` build step completed before script failure.\n\n## Scope\n- Only node upgrade task log handling and focused tests were changed; unrelated files, audit files, and issue-list files were not modified.\n\n**What type of PR is this?**\n/kind bug\n\n**What this PR does / why we need it**:\nKeadm upgrade and rollback task output should not use a predictable shared log file in `/tmp`. This change writes task logs under the KubeEdge log directory with restrictive permissions and isolated task-specific filenames, and rejects symlinked log paths.\n\n**Which issue(s) this PR fixes**:\nFixes #7097\n\n**Special notes for your reviewer**:\nThe broader edge test and integration targets were attempted locally. Failures were outside the changed code path or due to local macOS environment prerequisites, as listed in Verification.\n\n**Does this PR introduce a user-facing change?**:\n```release-note\nKubeEdge now writes node upgrade and rollback keadm task logs to private per-task files under the KubeEdge log directory instead of a shared `/tmp/keadm.log` path.\n```

### DIFF
--- a/edge/pkg/taskmanager/actions/nodeupgradejob.go
+++ b/edge/pkg/taskmanager/actions/nodeupgradejob.go
@@ -37,6 +37,7 @@ import (
 	"github.com/kubeedge/kubeedge/pkg/containers"
 	"github.com/kubeedge/kubeedge/pkg/nodetask/actionflow"
 	taskmsg "github.com/kubeedge/kubeedge/pkg/nodetask/message"
+	"github.com/kubeedge/kubeedge/pkg/nodetask/tasklog"
 	upgradeedge "github.com/kubeedge/kubeedge/pkg/upgrade/edge"
 	"github.com/kubeedge/kubeedge/pkg/util/execs"
 	"github.com/kubeedge/kubeedge/pkg/util/validation"
@@ -81,6 +82,8 @@ type nodeUpgradeJobActionHandler struct {
 	backupFiles []string
 	logger      logr.Logger
 }
+
+var openNodeUpgradeJobLog = tasklog.OpenKeadmLog
 
 func (nodeUpgradeJobActionHandler) preRun(
 	_ctx context.Context,
@@ -242,7 +245,7 @@ func (h *nodeUpgradeJobActionHandler) backup(
 
 func (h *nodeUpgradeJobActionHandler) upgrade(
 	_ctx context.Context,
-	_jobname, _nodename string,
+	jobname, nodename string,
 	specser SpecSerializer,
 ) ActionResponse {
 	resp := new(nodeUpgradeJobActionResponse)
@@ -263,9 +266,9 @@ func (h *nodeUpgradeJobActionHandler) upgrade(
 		return resp
 	}
 
-	logFile, err := os.OpenFile("/tmp/keadm.log", os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0600)
+	logFile, err := openNodeUpgradeJobLog(nodeUpgradeJobLogName("upgrade", jobname, nodename), os.O_APPEND)
 	if err != nil {
-		resp.err = fmt.Errorf("failed to open keadm log file: %w", err)
+		resp.err = err
 		resp.interrupt = true // No upgrade yet, no need to roll back.
 		return resp
 	}
@@ -291,16 +294,38 @@ func buildNodeUpgradeJobCommandArgs(spec *operationsv1alpha2.NodeUpgradeJobSpec)
 
 func (h *nodeUpgradeJobActionHandler) rollback(
 	_ctx context.Context,
-	_jobname, _nodename string,
+	jobname, nodename string,
 	specser SpecSerializer,
 ) ActionResponse {
 	resp := new(nodeUpgradeJobActionResponse)
-	// Roll back to the previous version
-	cmdline := "keadm rollback edge >> /tmp/keadm.log 2>&1"
-	cmd := execs.NewCommand(cmdline)
-	h.logger.V(2).Info("run rollback cmd", "cmd", cmdline)
-	resp.err = cmd.Exec()
+	logFile, err := openNodeUpgradeJobLog(nodeUpgradeJobLogName("rollback", jobname, nodename), os.O_APPEND)
+	if err != nil {
+		resp.err = err
+		return resp
+	}
+	defer logFile.Close()
+
+	args := []string{"rollback", "edge"}
+	cmd := exec.Command("keadm", args...)
+	cmd.Stdout = logFile
+	cmd.Stderr = logFile
+	h.logger.V(2).Info("run rollback cmd", "cmd", "keadm", "args", args)
+	resp.err = cmd.Run()
 	return resp
+}
+
+func nodeUpgradeJobLogName(action, jobname, nodename string) string {
+	return fmt.Sprintf("keadm-%s-%s-%s.log",
+		tasklog.SafeName(action),
+		tasklog.SafeName(defaultLogNamePart(jobname)),
+		tasklog.SafeName(defaultLogNamePart(nodename)))
+}
+
+func defaultLogNamePart(part string) string {
+	if part == "" {
+		return "unknown"
+	}
+	return part
 }
 
 func (nodeUpgradeJobActionHandler) getSpecSerializer(specData []byte) (SpecSerializer, error) {
--- a/edge/pkg/taskmanager/actions/nodeupgradejob_test.go
+++ b/edge/pkg/taskmanager/actions/nodeupgradejob_test.go
@@ -19,6 +19,8 @@ package actions
 import (
 	"context"
 	"errors"
+	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 
@@ -35,6 +37,7 @@ import (
 	"github.com/kubeedge/kubeedge/edge/pkg/metamanager/dao/dbclient"
 	"github.com/kubeedge/kubeedge/pkg/containers"
 	taskmsg "github.com/kubeedge/kubeedge/pkg/nodetask/message"
+	"github.com/kubeedge/kubeedge/pkg/nodetask/tasklog"
 	upgradeedge "github.com/kubeedge/kubeedge/pkg/upgrade/edge"
 	"github.com/kubeedge/kubeedge/pkg/util/execs"
 )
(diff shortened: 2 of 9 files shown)


rootp1:
@chendave the bot flagged an issue closing keyword in the commit message, so I reworded it to reference the issue instead of closing it directly, and pushed the update. No open review comments otherwise, please take a look when you have a chance.

rootp1:
@DoisLONG this one is all clear from my end now. No open threads that I can see beyond the usual approval gate, so flagging it for a look whenever you get a chance.

rootp1:
@DoisLONG following up once more here. This change has now gone through several review rounds and every point raised has been addressed and verified with tests. Whenever you get the chance, it would be great if this could move toward being merged.

DoisLONG (commented):
Thanks for the PR. The overall direction looks good to me.

Moving upgrade and rollback logs away from the shared `/tmp/keadm.log` path is the right fix. The per-task log names under the KubeEdge log directory are better scoped, and replacing rollback shell redirection with explicit argv plus attached stdout/stderr is also a good cleanup. I also like that the helper sanitizes log names, rejects symlinks, and uses `O_NOFOLLOW` on Unix.

One issue: `OpenKeadmLogAt()` passes `0600` to `os.OpenFile`, but that mode only applies when the file is newly created. If the target log file already exists as a regular file with overly permissive permissions, opening it with `O_CREATE|O_WRONLY|...` will keep the old mode and then write the new upgrade/rollback output into that file.

Could you please tighten the permissions for an existing regular log file before writing/truncating/appending to it, similar to the upgrade report fix? It would also be useful to add a regression test where the log file already exists with permissive permissions and is reset to `0600`.

DoisLONG (commented):
I rechecked the latest diff, and my previous concern still applies.

`OpenKeadmLogAt()` still relies on the `0600` mode passed to `os.OpenFile`, but that only applies when the file is newly created. If the log file already exists as a regular file with permissive permissions, the old mode is still kept and the new upgrade/rollback output can be written into that file.

Could you please address the existing-file case and add the regression test mentioned above?

While updating this helper, it would also be good to consider the existing-directory case: `os.MkdirAll(logDir, 0750)` does not tighten permissions if the log directory already exists with a permissive mode.

One more thing I noticed in the current diff: the v1alpha1 rollback path still uses `cmd.Process.Release()` after `cmd.Start()`. Since this PR rewrites that launcher, could you also avoid leaving this background child process path without a `Wait()`? A background `Wait()` goroutine would be safer.

rootp1:
@DoisLONG I tightened permissions on existing log dirs and files before writing, same as the cert fix, and added a regression test for it. I also swapped the rollback and upgrade Process.Release() calls for a background Wait so the detached keadm process gets reaped.

DoisLONG (commented):
Thanks for the update. The previous concerns look addressed now.

The existing-directory and existing-file permission cases are now handled, and the regression test for permissive existing paths looks good. The v1alpha1 upgrade/rollback paths also no longer use `Process.Release()` and now reap the background keadm process with a `Wait()` goroutine, which is safer.

One remaining hardening point: `OpenKeadmLogAt()` currently rejects symlinks, but for an existing log path it does not verify that the path is a regular file before chmodding and opening it. If the target path already exists as a directory, FIFO, device file, or another non-regular file, the helper should reject it instead of chmodding it and then trying to open it as a log file.

Could you please add an `info.Mode().IsRegular()` check for the existing-file case and a small regression test for a non-regular existing path?

rootp1:
@DoisLONG I added the IsRegular check in OpenKeadmLogAt for the existing path case and a regression test covering a directory in place of the log file. Please take another look when you have a chance.

DoisLONG (commented):
Thanks for the update.

This addresses my last hardening concern. `OpenKeadmLogAt()` now rejects an existing non-regular log path before chmodding or opening it, and the added regression test covers the directory case.

Together with the earlier fixes, the log helper now handles the important cases I was concerned about: private file mode for new logs, permission tightening for existing directories/files, symlink rejection, non-regular path rejection, and safe per-task log names. The v1alpha1 upgrade/rollback paths also now reap the background keadm process with a `Wait()` goroutine instead of using `Process.Release()`.


/lgtm

PR #7108, opened 2026-07-29
edgecore, keadm: hot reload heartbeat and remoteQueryTimeout without a restart

### LINKED ISSUE #7107: Non-disruptive Hot Reload and Automatic Rollback of EdgeCore Module Configurations

**Why is this needed**:
KubeEdge already supports distributing EdgeCore configurations from the cloud to edge nodes via CRDs. However, the edge node must restart the EdgeCore process for new configurations to take effect. This restart disrupts the cloud-edge connection, delays device data reporting, and can cause loss of unpersisted messages, compromising edge business continuity. This project aims to introduce hot-reload capability into EdgeCore, enabling configuration changes for critical modules to take effect immediately without restarting EdgeCore. It will also include an automatic rollback mechanism based on health checks to ensure the safety and reliability of configuration changes.

bhuvan-somisetty:
**What type of PR is this?**

/kind feature

Today, any configuration change delivered to an edge node through a `ConfigUpdateJob` restarts `edgecore.service`, even when the changed field never actually required a restart. On a real deployment this drops the cloud-edge connection, delays device data reporting, and can lose messages that had not been persisted yet, which hurts edge business continuity for what is often a trivial change like a heartbeat interval.

Looking at the modules, `edgehub`'s heartbeat interval and `metamanager`'s remote query timeout are already read fresh from the in-memory config on every loop iteration / request rather than being cached once at startup, so they can safely change while edgecore keeps running. This PR takes advantage of that:

- `keadm config-update` now checks the fields in `--set` against a small, explicit allow list of fields that have been verified to be re-read live by their module. If every field being changed is on that list, keadm signals edgecore to reload instead of restarting the service.
- edgecore listens for that signal, re-reads only the config file, and applies just the allow-listed fields through the existing per-module config packages. Reads/writes of those fields are now mutex-guarded since they can be updated from a signal-handler goroutine while other goroutines read them.
- After signaling, keadm watches `edgecore.service` for a short window to confirm it stayed healthy. If it did not, keadm restores the previous config file and falls back to the normal restart, so a bad value is never left in place silently.
- Any field that is not on the allow list keeps going through the existing restart path, unchanged.

This is intentionally scoped to the two fields that were verified to be safe today (`modules.edgeHub.heartbeat`, `modules.metaManager.remoteQueryTimeout`); other fields (ports, TLS material, database paths, etc.) are read once at module registration and still need a restart. The allow list is a single map, so more fields can be added later as their modules are verified to support it.

Fixes #7107

- The hot reload path only ever touches the two allow-listed fields; everything else is untouched and behaves exactly as before.
- I moved away from `execs.NewCommand` (which shells out via `bash -c`) for the two new commands (`systemctl kill -s HUP` / `systemctl is-active`) and used `exec.Command` with an argument list instead, in the same spirit as the existing hardening in this package.
- Added unit tests for the allow-list matching logic (`isHotReloadable`) and for the edgecore-side reload function, including the missing-file and missing-`modules`-section edge cases.

```release-note
EdgeCore now hot reloads `modules.edgeHub.heartbeat` and `modules.metaManager.remoteQueryTimeout` config updates delivered via `keadm config-update` / ConfigUpdateJob instead of restarting edgecore, with an automatic rollback to the previous config if edgecore is not healthy afterwards. All other fields continue to restart edgecore as before.
```

### DIFF
--- a/edge/cmd/edgecore/app/server.go
+++ b/edge/cmd/edgecore/app/server.go
@@ -20,6 +20,7 @@ import (
 	"github.com/kubeedge/api/apis/componentconfig/edgecore/v1alpha2/validation"
 	"github.com/kubeedge/beehive/pkg/core"
 	"github.com/kubeedge/kubeedge/edge/cmd/edgecore/app/options"
+	"github.com/kubeedge/kubeedge/edge/pkg/common/hotreload"
 	"github.com/kubeedge/kubeedge/edge/pkg/devicetwin"
 	"github.com/kubeedge/kubeedge/edge/pkg/edged"
 	"github.com/kubeedge/kubeedge/edge/pkg/edgehub"
@@ -133,6 +134,10 @@ offering HTTP client capabilities to components of cloud to reach HTTP servers r
 			}
 
 			registerModules(config)
+
+			// Watch for a hot reload signal so a subset of the configuration
+			// can be updated without restarting edgecore.
+			hotreload.Watch(opts.ConfigFile)
 
 			// start all modules
 			core.Run()
--- a/edge/pkg/common/hotreload/hotreload.go
+++ b/edge/pkg/common/hotreload/hotreload.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2026 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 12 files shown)


luomengY:
Sorry, this issue has been taken over by another internship arrangement. Please disregard. https://github.com/kubeedge/kubeedge/issues/7107

PR #7114, opened 2026-07-29
cloudhub: fix token refresh goroutine spinning after ctx cancel

### LINKED ISSUE #7101: cloud/cloudhub: token refresh goroutine spins forever after context cancel (break inside select)
**What happened**:

In `cloud/pkg/cloudhub/servers/httpserver/pre_server.go`, `GenerateAndRefreshToken` starts a goroutine to refresh the node-join token periodically:

```go
t := time.NewTicker(time.Hour * hubconfig.Config.CloudHub.TokenRefreshDuration)  // pre_server.go:174
go func() {
    for {
        select {
        case <-t.C:
            ...
        case <-ctx.Done():
            break        // pre_server.go:185
        }
    }
}()
```

The `break` on line 185 only exits the `select`, not the `for` loop. Once `ctx` is cancelled, `ctx.Done()` is a closed channel, so the `select` completes immediately on every iteration and the goroutine spins in a tight loop, consuming a full CPU core for the remaining lifetime of the process. The ticker `t` is also never stopped, so its resources are never released.

**What you expected to happen**: on context cancellation the goroutine should stop the ticker and return.

**How to reproduce**: cancel the context passed to `GenerateAndRefreshToken` (e.g. a cloudcore shutdown path that cancels but doesn't immediately exit the process) and observe CPU usage, or step through the loop in a unit test with a cancelled context.

**Fix**:

```go
case <-ctx.Done():
    t.Stop()
    return
```

(or `defer t.Stop()` at the top of the goroutine). Classic Go `break`-in-`select` pitfall; `go vet` does not catch it.

**Verified**: present at origin/master

Priyanshubhartistm:
**What type of PR is this?**

/kind bug

`GenerateAndRefreshToken` in `cloud/pkg/cloudhub/servers/httpserver/pre_server.go` starts a goroutine that periodically refreshes the node-join token using a `select` over the refresh ticker and `ctx.Done()`. In the `ctx.Done()` case, the code used a `break` statement, which in Go only exits the enclosing `select`, not the surrounding `for` loop. Once the context was cancelled, `ctx.Done()` returns an already-closed channel, so the `select` resolves
immediately on every iteration and the goroutine spins in a tight busy loop for the remaining lifetime of the process, pinning a full CPU core. The ticker was also never stopped, leaking its resources.

This PR changes the `ctx.Done()` case to `return` (exiting the goroutine) and adds `defer t.Stop()` to release the ticker, so cancellation actually stops the refresh goroutine as intended.

Fixes #7101

Added `TestGenerateAndRefreshToken_StopsOnContextCancel`, which fails against the original code (the goroutine never terminates after cancellation, so `runtime.NumGoroutine()` never drops back to baseline) and passes against the fix. Uses gomonkey to patch `client.GetKubeClient` with a fake clientset,

following the existing pattern in
`cloud/pkg/cloudhub/servers/httpserver/node/checknode_test.go`. Run with
`go test -gcflags=all=-l ./cloud/pkg/cloudhub/servers/httpserver/...`.

```release-note
Fix a bug in CloudHub where cancelling the token-refresh context caused the
refresh goroutine to spin in a tight loop and consume a full CPU core
instead of exiting, and where the refresh ticker was never stopped.
```

### DIFF
--- a/cloud/pkg/cloudhub/servers/httpserver/pre_server_test.go
+++ b/cloud/pkg/cloudhub/servers/httpserver/pre_server_test.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2024 The KubeEdge Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Note: Tests in this file use gomonkey for function patching.
+// Run tests with inlining disabled:
+// go test -gcflags=all=-l ./cloud/pkg/cloudhub/servers/httpserver/...
+
+package httpserver
+
+import (
+	"context"
+	"runtime"
+	"testing"
+	"time"
+
+	"github.com/agiledragon/gomonkey/v2"
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+
+	hubconfig "github.com/kubeedge/kubeedge/cloud/pkg/cloudhub/config"
+	"github.com/kubeedge/kubeedge/cloud/pkg/common/client"
+	"github.com/kubeedge/kubeedge/pkg/security/certs"
+)
+
+// TestGenerateAndRefreshToken_StopsOnContextCancel guards against a
+// break-in-select bug where cancelling ctx only exited the select
+// statement instead of the refresh goroutine, leaving it spinning in a
+// tight loop forever. On a fixed implementation the goroutine (and its
+// ticker) must terminate shortly after ctx is cancelled.
+func TestGenerateAndRefreshToken_StopsOnContextCancel(t *testing.T) {
+	fakeClient := fake.NewSimpleClientset()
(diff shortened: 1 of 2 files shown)


Priyanshubhartistm:
Hii @DoisLONG could you please take a look at this PR? I would really appreciate your review and any feedback or suggestions you may have.

Thank you

DoisLONG (commented):
Thanks for the PR.

This looks good to me. The root cause is clear: `break` in the `ctx.Done()` branch only exits the `select`, not the surrounding `for` loop, so after the context is cancelled the goroutine can keep spinning on the already-closed `ctx.Done()` channel. Changing it to `return` correctly exits the refresh goroutine, and stopping the ticker on exit avoids leaking the ticker resource.

The added regression test also covers the intended behavior: the refresh goroutine starts after `GenerateAndRefreshToken()` and exits after the context is cancelled.

/lgtm

Priyanshubhartistm:
@WillardHu , Could I get a review from you?

cloud/pkg/cloudhub/servers/httpserver/pre_server_test.go
```diff
@@ -0,0 +1,91 @@
+/*
+Copyright 2024 The KubeEdge Authors.
```
Shelley-BaoYue, line 2:
```suggestion
Copyright 2026 The KubeEdge Authors.
```

Shelley-BaoYue (commented):
Looks good for me. : )
Please update the date of new testing file.

cloud/pkg/cloudhub/servers/httpserver/pre_server_test.go
Priyanshubhartistm: @Shelley-BaoYue , Done.

PR #7121, opened 2026-07-29
certs: restrict cert file and directory permissions

### LINKED ISSUE #7120: Certificate bootstrap writes PEM files and directories with overly broad permissions
**What happened**:
`WriteDERToPEMFile` creates certificate directories with mode `0722` and PEM files with mode `0622`. This makes private-key and certificate artifacts writable by non-owner users on the host.

Certificate directories and files should be created with owner-only or at least restrictive permissions, for example `0700` for directories and `0600` for private-key material.

1. On a disposable host, trigger certificate bootstrap or rotation so `pkg/security/certs.WriteDERToPEMFile` writes `ca.crt`, `edge.crt`, or `edge.key`.
2. Run `stat -c '%a %n' <dir> <file>` on the created directory and PEM files.
3. Observe `0722` on the directory and `0622` on the file.
4. As a non-owner local user, verify that the files are writable when filesystem permissions otherwise allow access.

- Evidence in current upstream code:
  - `pkg/security/certs/io.go` uses `os.MkdirAll(dir, 0722)` and `os.WriteFile(file, bff, 0622)`.
  - `edge/pkg/edgehub/certificate/certmanager.go` uses this helper for CA, cert, and key writes.
- Searched upstream issues and PRs for related terms (`WriteDERToPEMFile`, "certificate file permissions", "insecure file permissions", "0722"); no matching open issue or PR found.

**Environment**:
- Kubernetes version (use `kubectl version`): not Kubernetes-version specific
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): `upstream/master` at `afc7d19423f775d90009f770d01866c73cc2cd2b`

  - Hardware configuration (e.g. `lscpu`): not hardware-specific
  - OS (e.g. `cat /etc/os-release`): any Linux distribution where EdgeCore writes certs to disk
  - Kernel (e.g. `uname -a`): not kernel-specific
  - Go version (e.g. `go version`): source audit against current repo
  - Others: host-local filesystem permissions relevant

  </details>

  - edgecore version (e.g. `edgecore --version`): current `master`
  - Hardware configuration (e.g. `lscpu`): not hardware-specific
  - OS (e.g. `cat /etc/os-release`): Linux
  - Kernel (e.g. `uname -a`): not kernel-specific
  - Go version (e.g. `go version`): source audit against current repo
  - Others: issue affects certificate bootstrap and rotation paths

  </details>

rootp1:
**What type of PR is this?**

/kind bug

`WriteDERToPEMFile` in `pkg/security/certs/io.go` created certificate directories with mode `0722` and PEM files with mode `0622`. These modes are group/world writable, which allows non-owner users on the host to modify or overwrite private key and certificate material written during certificate bootstrap and rotation (used by `edge/pkg/edgehub/certificate/certmanager.go`). This PR restricts directory creation to `0700` and file creation to `0600` so only the owner can write these artifacts.

Fixes #7120

Added `TestWriteDERToPEMFilePermissions` in `pkg/security/certs/io_test.go` to assert the created directory and file have the expected restrictive permissions. Existing tests in the package continue to pass.

```release-note
Restrict permissions on certificate files and directories written by WriteDERToPEMFile to 0600/0700 to prevent non-owner write access to private key material.
```

### DIFF
--- a/pkg/security/certs/io.go
+++ b/pkg/security/certs/io.go
@@ -36,13 +36,23 @@ func ReadPEMFile(file string) (*pem.Block, error) {
 
 func WriteDERToPEMFile(file, t string, der []byte) (*pem.Block, error) {
 	dir := filepath.Dir(file)
-	if err := os.MkdirAll(dir, 0722); err != nil {
+	if err := os.MkdirAll(dir, 0700); err != nil {
 		return nil, fmt.Errorf("failed to create dir %s, err: %v", dir, err)
+	}
+	// MkdirAll does not change the mode of a directory that already exists,
+	// so an existing dir created with permissive bits would otherwise stay that way.
+	if err := os.Chmod(dir, 0700); err != nil {
+		return nil, fmt.Errorf("failed to chmod dir %s, err: %v", dir, err)
 	}
 	block := &pem.Block{Type: t, Bytes: der}
 	bff := pem.EncodeToMemory(block)
-	if err := os.WriteFile(file, bff, 0622); err != nil {
+	if err := os.WriteFile(file, bff, 0600); err != nil {
 		return nil, fmt.Errorf("failed to write file %s, err: %v", file, err)
+	}
+	// WriteFile only applies the mode when creating a new file, so an existing
+	// file with permissive bits needs an explicit chmod to be corrected.
+	if err := os.Chmod(file, 0600); err != nil {
+		return nil, fmt.Errorf("failed to chmod file %s, err: %v", file, err)
 	}
 	return block, nil
 }
--- a/pkg/security/certs/io_test.go
+++ b/pkg/security/certs/io_test.go
@@ -24,6 +24,62 @@ func TestReadWrite(t *testing.T) {
 	}
 }
 
+func TestWriteDERToPEMFilePermissions(t *testing.T) {
+	dir := filepath.Join(t.TempDir(), "certs")
+	file := filepath.Join(dir, "edge.key")
+	if _, err := WriteDERToPEMFile(file, "test data", []byte("test")); err != nil {
+		t.Fatal(err)
+	}
+
+	dirInfo, err := os.Stat(dir)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if perm := dirInfo.Mode().Perm(); perm != 0700 {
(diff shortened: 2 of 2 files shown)


rootp1:
@DoisLONG can you please have a look whenever you have a chance to

DoisLONG (commented):
Thanks for the PR. The direction is right, but I do not think this fully fixes the permission issue yet.

Changing the `MkdirAll` mode to `0700` and the `WriteFile` mode to `0600` fixes the permissions for newly created directories and files. However, these mode arguments do not tighten permissions when the directory or PEM file already exists.

That matters here because `WriteDERToPEMFile()` is also used for certificate rotation. If an existing cert/key file was previously created with permissive permissions such as `0622`, `os.WriteFile(..., 0600)` will truncate and rewrite the file but keep the old file mode. Similarly, an existing cert directory with permissive permissions would remain unchanged.

Could you please also tighten permissions for existing paths? For example, after `MkdirAll`, explicitly `Chmod` the cert directory to `0700`, and ensure an existing regular PEM file is reset to `0600` before or after writing. Please also add a regression test where the directory and file already exist with permissive modes and are corrected by `WriteDERToPEMFile()`.

CI currently looks like it needs workflow approval, but this existing-path permission case is a code blocker from my side.

rootp1:
@DoisLONG thanks for catching that. You're right, MkdirAll and WriteFile only apply the mode on creation, so an existing permissive dir or file would have survived rotation. I added explicit Chmod calls for both the directory and the file after creation, and a regression test that pre creates the dir at 0755 and the file at 0644 and verifies WriteDERToPEMFile corrects both to 0700 and 0600. Can you please have a look whenever you have a chance to

DoisLONG (commented):
Thanks for the update.

This addresses my previous concern. `WriteDERToPEMFile()` now explicitly tightens the existing certificate directory and PEM file permissions instead of relying only on the modes passed to `MkdirAll` and `WriteFile`.

The added regression test also covers the existing-path case by starting with a permissive directory and file, then verifying they are corrected to `0700` and `0600`.

CI currently still needs workflow approval, but I do not see any remaining code blocker from my side.

/lgtm

PR #7133, opened 2026-07-30
charts: derive admission webhook namespace and service name from release

### LINKED ISSUE #7132: Admission manifests hardcode webhook namespace and service name, ignoring release namespace
**What happened**:

The admission Deployment template in the cloudcore Helm chart always passes `--webhook-namespace=kubeedge` and `--webhook-service-name=kubeedge-admission-service` as hardcoded literal argv values, regardless of the actual Helm release namespace or configured service name. This means Helm installs into a non-default namespace cannot correctly self-register the webhook against the real release namespace/service.

The admission container args should derive the webhook namespace from `.Release.Namespace` (the pattern already used elsewhere in this chart, e.g. `manifests/charts/cloudcore/templates/deployment_controllermanager.yaml`) and the webhook service name from a chart value, instead of hardcoded constants.

1. Fetch the latest upstream default branch: `git fetch upstream master`.
2. Inspect the Deployment template arguments: `git show upstream/master:manifests/charts/cloudcore/templates/deployment_admission.yaml | sed -n '40,47p'`.
3. Render the chart for a non-default namespace: `helm template demo manifests/charts/cloudcore -n custom-edge --set admission.enable=true | grep -- '--webhook-'`.
4. Observe that the args still point to `kubeedge` and `kubeedge-admission-service` even though the release namespace is `custom-edge`.

Verified against upstream `master` at commit `afc7d1942`.

**Environment**:
- Kubernetes version (use `kubectl version`): Helm render / static inspection
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): upstream `master` at `afc7d1942`

  - Others: Helm 3 is sufficient for reproduction

  </details>

  - edgecore version (e.g. `edgecore --version`): N/A
  - Others: N/A

  </details>

rootp1:
## Summary
- Fix the cloudcore admission Helm chart so the webhook namespace and service name are no longer hardcoded

## Linked Issue
Fixes #7132

## What Changed
- `manifests/charts/cloudcore/templates/deployment_admission.yaml`: derive `--webhook-namespace` from `.Release.Namespace` and `--webhook-service-name` from a new `admission.serviceName` value, instead of hardcoded `kubeedge` / `kubeedge-admission-service`
- `manifests/charts/cloudcore/templates/service_admission.yaml`: use the same `admission.serviceName` value for the Service name so it stays consistent with the Deployment args
- `manifests/charts/cloudcore/values.yaml`: add `admission.serviceName` default value `kubeedge-admission-service`

## Verification
- `make verify` — passed
- `helm lint manifests/charts/cloudcore --set admission.enable=true` — passed
- `helm template demo manifests/charts/cloudcore -n custom-edge --set admission.enable=true | grep -- '--webhook-'` — passed, confirms `--webhook-namespace=custom-edge` is now rendered instead of the previous hardcoded `kubeedge`

## Scope
- Only the three files listed above were changed; no unrelated files were modified

### DIFF
--- a/manifests/charts/cloudcore/templates/deployment_admission.yaml
+++ b/manifests/charts/cloudcore/templates/deployment_admission.yaml
@@ -41,8 +41,8 @@ spec:
             - --tls-private-key-file=/admission.local.config/certificates/tls.key
             - --ca-cert-file=/admission.local.config/certificates/ca.crt
             {{- end }}
-            - --webhook-namespace=kubeedge
-            - --webhook-service-name=kubeedge-admission-service
+            - --webhook-namespace={{ .Release.Namespace }}
+            - --webhook-service-name={{ .Values.admission.serviceName | default "kubeedge-admission-service" }}
             - --port=443
             - -v=4
             - 2>&1
--- a/manifests/charts/cloudcore/templates/service_admission.yaml
+++ b/manifests/charts/cloudcore/templates/service_admission.yaml
@@ -5,7 +5,7 @@ metadata:
   {{- with .Values.admission.labels }}
   labels: {{- toYaml . | nindent 4 }}
   {{- end }}
-  name: kubeedge-admission-service
+  name: {{ .Values.admission.serviceName | default "kubeedge-admission-service" }}
 spec:
   ports:
     - port: 443
--- a/manifests/charts/cloudcore/values.yaml
+++ b/manifests/charts/cloudcore/values.yaml
@@ -147,6 +147,7 @@ admission:
   enable: false
   replicaCount: 1
   certsSecretName: ""
+  serviceName: "kubeedge-admission-service"
   image:
     repository: "kubeedge/admission"
     tag: "v1.23.0"


rootp1:
@DoisLONG could you take a look when you get a moment?

DoisLONG (commented):
Thanks for the PR.

The Helm chart direction looks reasonable to me. Deriving `--webhook-namespace` from `.Release.Namespace` and using the same `admission.serviceName` value for both the admission Deployment argument and the Service name should fix the hardcoded namespace/service-name issue while preserving the existing default.

There are two cleanup items:

1. This PR currently includes unrelated cert permission changes in `pkg/security/certs/io.go` and `pkg/security/certs/io_test.go`. Those changes do not belong to this chart fix and also make the PR body inaccurate, since the Scope section says only the three chart files were changed. Could you please rebase/drop the unrelated cert changes so this PR only contains the admission Helm chart fix?

2. This PR currently has multiple commits. After the unrelated changes are removed, could you please squash it into a single logical commit?

rootp1:
@DoisLONG I dropped the unrelated cert permissions commit and squashed this down to a single commit scoped to the namespace and service name change. Please have another look when convenient.

PR #7135, opened 2026-07-30
test: pin and cache integration bootstrap tools without sudo

### LINKED ISSUE #7134: Integration test bootstrap installs unpinned tools with sudo into /usr/bin

Refactor `cloud/test/integration/scripts/execute.sh` so that it installs pinned `setup-envtest` and `ginkgo` versions into a repo-local tool cache (e.g. `.cache/tools`) instead of installing `@latest` and copying the binaries into `/usr/bin` with `sudo`.

**Why is this needed**:

The current bootstrap in `cloud/test/integration/scripts/execute.sh` runs:

```
go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
sudo cp $GOPATH/bin/setup-envtest /usr/bin/
...
go install github.com/onsi/ginkgo/v2/ginkgo@latest
sudo cp $GOPATH/bin/ginkgo /usr/bin/
```

This has two problems:
1. It installs `@latest` for both tools, which is nondeterministic — a new upstream release of either tool can silently change local/CI test behavior without any change to this repository.
2. It requires `sudo` to copy binaries into `/usr/bin`, which is unnecessary privilege escalation for a test bootstrap step and complicates running the integration suite in restricted/CI environments.

Pinning both tools to versions consistent with `go.mod` (`sigs.k8s.io/controller-runtime v0.19.7`, `github.com/onsi/ginkgo/v2 v2.21.0`) and installing them into a repo-local directory via `GOBIN` removes both the nondeterminism and the privilege requirement, and lets the bootstrap reuse already-installed tools on repeat runs instead of reinstalling every time.

## Affected Area
- `cloud/test/integration/scripts/execute.sh`

## Acceptance Criteria
- The integration bootstrap installs pinned tool versions into a repo-local or user-local directory without `sudo`.
- Re-running the bootstrap reuses cached tools when versions already match.
- The integration suite (`make integrationtest` / `cloud/test/integration/scripts/execute.sh`) continues to pass with the pinned tool versions.

rootp1:
**What type of PR is this?**

/kind test

Refactors `cloud/test/integration/scripts/execute.sh` so the integration test bootstrap installs pinned `setup-envtest` (release-0.19) and `ginkgo` (v2.21.0, matching go.mod) into a repo-local `.cache/tools` directory via `GOBIN`, instead of installing `@latest` and copying the binaries into `/usr/bin` with `sudo`. This removes nondeterminism from unpinned tool versions and removes the unnecessary privilege escalation, while allowing re-runs to reuse already-installed tools.

Fixes #7134

`.cache/` was added to `.gitignore` so the local tool cache is never committed.

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ build/edge/tmp
 
 coverage.out
 go.work.sum
+
+.cache/
--- a/cloud/test/integration/scripts/execute.sh
+++ b/cloud/test/integration/scripts/execute.sh
@@ -17,21 +17,33 @@
 ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd -P)"
 TEST_DIR=$(dirname $(dirname "${BASH_SOURCE[0]}"))
 APPS_CRD_DIR="${ROOT_DIR}/build/crds/apps"
-ENVTEST_DOWNLOAD_DIR="/tmp/envtest/bin"
+TOOL_CACHE_DIR="${ROOT_DIR}/.cache/tools"
 ENVTEST_BIN_DIR=""
 
+SETUP_ENVTEST_VERSION="release-0.19"
+GINKGO_VERSION="v2.21.0"
+
+# Bin dirs are keyed by tool version so bumping SETUP_ENVTEST_VERSION or
+# GINKGO_VERSION always triggers a fresh install instead of silently
+# reusing a stale cached binary from a previous version.
+SETUP_ENVTEST_BIN_DIR="${TOOL_CACHE_DIR}/bin/setup-envtest-${SETUP_ENVTEST_VERSION}"
+GINKGO_BIN_DIR="${TOOL_CACHE_DIR}/bin/ginkgo-${GINKGO_VERSION}"
+ENVTEST_DOWNLOAD_DIR="${TOOL_CACHE_DIR}/envtest/${SETUP_ENVTEST_VERSION}/bin"
+
 function do_preparation() {
-    which setup-envtest &> /dev/null || {
-        go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
-        sudo cp $GOPATH/bin/setup-envtest /usr/bin/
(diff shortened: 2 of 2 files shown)


rootp1:
@DoisLONG whenever you have a chance, could you have a look at this one?

DoisLONG (commented):
Thanks for the PR.

The direction looks good to me. Installing the integration-test bootstrap tools into a repo-local cache through `GOBIN` is better than installing `@latest` and copying binaries into `/usr/bin` with `sudo`. Pinning `setup-envtest` and `ginkgo` also makes the test bootstrap more reproducible.

One thing I think should be fixed before LGTM: the cache is currently not versioned. The script only checks whether `${TOOL_CACHE_DIR}/bin/setup-envtest` and `${TOOL_CACHE_DIR}/bin/ginkgo` are executable. If those binaries already exist from a previous local run, branch, or older version of this script, they will be reused even if they do not match `SETUP_ENVTEST_VERSION` or `GINKGO_VERSION`.

That weakens the purpose of pinning the tool versions.

Could you make the cache path versioned, or otherwise verify the cached binary versions before reusing them? For example, the tool cache directory could include the pinned setup-envtest and ginkgo versions so a version change automatically installs a fresh copy.

CI currently looks like it only needs workflow approval. After the cache-versioning issue is addressed, I can take another look.

rootp1:
@DoisLONG I made the tool cache paths version specific for both setup-envtest and ginkgo, so bumping either version forces a fresh install instead of reusing a stale binary. Ran the integration script end to end to confirm it installs correctly and the tests pass. Could you take another look?

DoisLONG (commented):
Thanks for the update.

This addresses my previous concern. The cached tool paths are now version-specific for both `setup-envtest` and `ginkgo`, so bumping either pinned version will install into a new cache directory instead of silently reusing an older binary.

The current approach also keeps the original goal of this PR: avoid installing `@latest`, avoid copying binaries into `/usr/bin` with `sudo`, and keep the integration-test bootstrap reproducible and local to the repository workspace.

/lgtm

PR #7137, opened 2026-07-31
[keadm] Fix misleading argument error and YAML unmarshaling in keadm ctl edit device

### LINKED ISSUE #7136: [keadm] Fix misleading error message and YAML unmarshaling failure in keadm ctl edit device
**What happened**:
In `keadm ctl edit device`, two issues exist in argument validation and editor content parsing:

1. **Misleading Error Message for Missing Arguments**:
   In `editDevice(args []string)` (`keadm/cmd/keadm/app/cmd/ctl/edit/device.go`), the argument count check uses `if len(args) == 1 ... else return fmt.Errorf("too many args, edit one device at once")`.
   When a user executes `keadm ctl edit device` without specifying any device name (`len(args) == 0`), the CLI outputs:
   ```text
   Error: too many args, edit one device at once
   ```
   This error message is confusing and incorrect when zero arguments are passed.

2. **JSON Unmarshaling Failure on Edited YAML Content**:
   When editing a device, `keadm ctl edit device` launches the text editor with a temporary YAML representation of the device. After the user edits and saves the file, `edit()` attempts to unmarshal the saved YAML content using `json.Unmarshal` (`encoding/json`) instead of `yaml.Unmarshal` (`sigs.k8s.io/yaml`). Standard Go `json.Unmarshal` fails to parse YAML formatted content, resulting in syntax errors (`invalid character ... looking for beginning of value`).

3. **Missing Unit Tests**: `keadm/cmd/keadm/app/cmd/ctl/edit` currently lacks unit test coverage.

- Running `keadm ctl edit device` with no arguments should output `Error: device name is required` (or `Error: required argument <device-name> missing`).
- Running `keadm ctl edit device <name>` should correctly parse the saved YAML file using `yaml.Unmarshal` and update the device resource without JSON parsing errors.

1. Run `keadm ctl edit device` (without specifying a device name).
2. Observe error output: `Error: too many args, edit one device at once`.
3. Run `keadm ctl edit device <device-name>` in an edge environment and save valid YAML edits in the opened editor.
4. Observe error: `invalid character ... looking for beginning of value` caused by `json.Unmarshal`.

Code location: `keadm/cmd/keadm/app/cmd/ctl/edit/device.go` inside `editDevice()` and `edit()`.

**Environment**:
- KubeEdge version: `v1.17.0+` / `main` branch

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

This PR addresses argument validation and YAML unmarshaling issues in `keadm ctl edit device`:
1. Refactors argument count validation in `editDevice` so running `keadm ctl edit device` with 0 arguments returns `Error: device name is required` instead of `Error: too many args, edit one device at once`.
2. Replaces `json.Unmarshal` with `yaml.Unmarshal` (`sigs.k8s.io/yaml`) when parsing saved temporary YAML device files in `edit()` to avoid JSON syntax parsing failures on YAML content.
3. Adds unit test coverage for `keadm ctl edit device` in `device_test.go`.

Fixes #7136

NONE

```release-note
Fix misleading argument error message and YAML unmarshaling failure in `keadm ctl edit device`.
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/ctl/edit/device.go
+++ b/keadm/cmd/keadm/app/cmd/ctl/edit/device.go
@@ -85,6 +85,13 @@ func NewEditDeviceOpts() *DeviceEditOptions {
 }
 
 func (o *DeviceEditOptions) editDevice(args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("device name is required")
+	}
+	if len(args) > 1 {
+		return fmt.Errorf("too many args, edit one device at once")
+	}
+
 	config, err := util.ParseEdgecoreConfig(constants.EdgecoreConfigPath)
 	if err != nil {
 		return fmt.Errorf("get edge config failed with err:%v", err)
@@ -93,27 +100,23 @@ func (o *DeviceEditOptions) editDevice(args []string) error {
 
 	ctx := context.Background()
 
-	if len(args) == 1 {
-		deviceRequest := &client.DeviceRequest{
-			Namespace:  o.Namespace,
-			DeviceName: args[0],
-		}
+	deviceRequest := &client.DeviceRequest{
+		Namespace:  o.Namespace,
+		DeviceName: args[0],
+	}
 
-		device, err := deviceRequest.GetDevice(ctx)
-		if err != nil {
+	device, err := deviceRequest.GetDevice(ctx)
+	if err != nil {
+		return err
+	}
+
+	if device.Spec.NodeName == nodeName {
+		if err = o.edit(device); err != nil {
 			return err
 		}
-
-		if device.Spec.NodeName == nodeName {
-			if err = o.edit(device); err != nil {
-				return err
-			}
-			klog.Infof("Send update message to DeviceTwin")
-		} else {
-			klog.Errorf("Can't query device: \"%s\" for node: \"%s\"", device.Name, device.Spec.NodeName)
-		}
+		klog.Infof("Send update message to DeviceTwin")
 	} else {
-		return fmt.Errorf("too many args, edit one device at once")
+		klog.Errorf("Can't query device: \"%s\" for node: \"%s\"", device.Name, device.Spec.NodeName)
 	}
 
 	return nil
@@ -160,7 +163,7 @@ func (o *DeviceEditOptions) edit(dl *v1beta1.Device) error {
 	jsonEdited := cmdutil.StripComments(edited)
 
 	var editedDevice v1beta1.Device
-	err = json.Unmarshal(jsonEdited, &editedDevice)
+	err = yaml.Unmarshal(jsonEdited, &editedDevice)
 	if err != nil {
 		return preservedFile(err, file)
 	}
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
CC @fisherxu @DoisLONG @tangming1996 @subpathdev @chendave @Shelley-BaoYue @WillardHu - Requesting a review for this fix when you have a moment. Thanks!

DoisLONG (commented):
Thanks for the PR.

The YAML parsing fix looks right to me. Since the editor writes the device as YAML, using `yaml.Unmarshal` here is the correct behavior and should avoid the JSON syntax error when users save normal YAML content.

I also like that this PR adds coverage for the edit-device path, including the argument validation, successful YAML edit, cancelled edit, and invalid YAML cases.

There are two things to address:

1. The argument count validation is still done after `ParseEdgecoreConfig()`. That means `keadm ctl edit device` with no device name can still fail on the edgecore config first if the config is missing or unreadable, instead of returning the intended `device name is required` error. Could you move the `len(args)` validation before parsing the edgecore config? It would also be good to add a regression test for the zero-argument case that does not depend on a successfully parsed edgecore config.

2. Main CI is currently failing in the lint job due to `goconst`:

```text
keadm/cmd/keadm/app/cmd/ctl/edit/device_test.go:184:25: string `temp-file` has 3 occurrences, make it a constant (goconst)
```

Could you please fix the lint issue as well, for example by extracting the repeated `"temp-file"` test value into a constant?

bhuvan-somisetty:
@DoisLONG thanks for the review!

1. Moved the `len(args)` validation to run before `ParseEdgecoreConfig()`, so `keadm ctl edit device` with no device name now returns `device name is required` right away instead of possibly failing on the edgecore config first. Added a regression test (`TestEditDeviceZeroArgsDoesNotParseConfig`) that mocks `ParseEdgecoreConfig` to fail and asserts it's never called in the zero-arg case.
2. Extracted the repeated `"temp-file"` literal into a `testTempFile` constant to fix the `goconst` lint failure.

Pushed in fdafe0234. PTAL, thanks again!

DoisLONG (commented):
Thanks for the update.

The previous review comments look addressed now. The argument validation is now done before parsing the edgecore config, and the added zero-argument regression test makes sure `ParseEdgecoreConfig()` is not called before returning `device name is required`.

The previous lint issue is also fixed by extracting the repeated temp file name into a test constant.

There are two remaining items:

2. Main CI is currently failing. The earlier lint failure is gone, but the latest failed job is `E2e test (WebSocket, v1.30.0)`. Could you please check whether this is unrelated/flaky and retrigger it if needed?

bhuvan-somisetty:
Thanks for the thorough review @DoisLONG! Both are sorted now:

1. Squashed the two commits into one logical commit.
2. That WebSocket v1.30.0 failure was indeed unrelated/flaky pod scheduling just timed out waiting to go Running, while the same suite passes on v1.31.0/v1.32.0 and QUIC v1.30.0. Pushing the squashed commit retriggered CI and it passed clean this time.

All checks are green now. Let me know if anything else needs a look.

Issue #7106, opened 2026-07-27
Add RuntimeClass support for confidential workloads on KubeEdge


DoisLONG:
**What would you like to be added/modified**:

KubeEdge currently does not provide complete RuntimeClass support for workloads running on edge nodes.

This enhancement aims to enable end-to-end RuntimeClass support in KubeEdge and validate it with confidential container runtimes.

The proposed scope includes:

- Investigate the Kubernetes RuntimeClass workflow and identify missing components in KubeEdge.
- Support RuntimeClass resource synchronization from cloud side to edge nodes.
- Enable edge workloads to use `spec.runtimeClassName`.
- Ensure Edged can correctly select the configured container runtime handler.
- Validate RuntimeClass with alternative runtimes such as:
  - Kata Containers
  - Confidential Containers ecosystem runtimes

The implementation should consider:

- CloudCore
- EdgeCore
- MetaManager
- MetaServer
- Edged
- Container Runtime Interface integration

Optional validation may include confidential computing platforms such as Intel TDX when suitable infrastructure is available.

**Why is this needed**:

Confidential computing and workload isolation are becoming increasingly important for edge scenarios.

Kubernetes already provides RuntimeClass as the standard mechanism for selecting different runtime environments, but KubeEdge edge workloads currently cannot fully leverage this capability.

Supporting RuntimeClass will provide a foundation for secure and isolated workloads at the edge, enabling scenarios such as confidential containers, stronger workload isolation, and hardware-assisted security.

**Expected outcome / acceptance criteria**:

- RuntimeClass resource synchronization works between Kubernetes control plane and KubeEdge edge nodes.
- Edge workloads can specify and use RuntimeClass through `spec.runtimeClassName`.
- Alternative runtime handlers can be selected correctly on edge nodes.
- Missing or invalid RuntimeClass configurations are handled gracefully.
- Unit tests and end-to-end tests are added.
- Documentation is provided describing how to deploy RuntimeClass-based workloads on KubeEdge.
- Confidential container runtime validation is demonstrated when available.

**References**:

- #6334 v1.20 kubeedge do not support the runtimeclass
- #4416 RuntimeClass support discussion

Moemenmohamed24:
Hi @DoisLONG , I am planning to apply for the LFX Mentorship (Term 3) with this project.

[...]

PR #7141, opened 2026-08-01
feat(edgecontroller,metamanager): add RuntimeClass cloud-to-edge sync and MetaManager client

Norway-02:
Related to #7106

While debugging why pods fail with `runtimeclass.node.k8s.io "kata" not found` even though the RuntimeClass exists in the cluster, I found that the problem starts in CloudCore.

CloudCore doesn't watch RuntimeClass resources, so they never get sent to the edge. Because of that, MetaManager never stores them, MetaServer can't serve them, and when edged tries to resolve `spec.runtimeClassName`, it can't find anything.

This PR fixes the first part of that problem by syncing RuntimeClass resources from CloudCore to the edge.

### Example

```bash
kubectl get runtimeclasses

NAME    HANDLER   AGE
kata    kata      2h

# Pod uses spec.runtimeClassName: kata

# edgecore log
failed to "CreatePodSandbox" for pod: runtimeclass.node.k8s.io "kata" not found
```

The RuntimeClass exists in the cluster. It just never reaches the edge node.

### Changes

On the cloud side:

* Added RuntimeClass to CloudHub's authorization map with the correct GVR (`node.k8s.io/v1`, `runtimeclasses`).
* Added a `RuntimeClassManager` that watches RuntimeClass resources using a `SharedIndexInformer`.
* Added `syncRuntimeClass()` to the downstream controller to send RuntimeClass updates to edge nodes.
* Since RuntimeClass is cluster-scoped, updates are broadcast to every connected edge node.
* Added `GetAllNodes()` to `LocationCache` to get the current list of connected nodes for broadcasting.

On the edge side:

* Added a RuntimeClass client under `metamanager/client`, following the same pattern used for PersistentVolumes.
* It first looks in the local SQLite store. If the object isn't there yet, it sends a sync request to MetaManager.
* Added `RuntimeClassesGetter` to `CoreInterface` and exposed `RuntimeClasses()` from `metaClient`.

There are no behavior changes unless a workload uses `spec.runtimeClassName`.

### Why only this part?

I wanted to verify that RuntimeClass objects can actually reach the edge and be stored in MetaManager before working on the lookup path.

The next step is to let MetaServer serve RuntimeClass objects from the local store so edged's `runtimeClassManager` can resolve them during pod creation.

### Files changed

* `staging/.../model/message.go` - added `ResourceTypeRuntimeClass`
* `cloud/pkg/cloudhub/authorization/resource_attributes.go` - added RuntimeClass authorization
* `cloud/pkg/edgecontroller/manager/runtimeclass.go` - added `RuntimeClassManager`
* `cloud/pkg/edgecontroller/manager/location.go` - added `GetAllNodes()`
* `cloud/pkg/edgecontroller/controller/downstream.go` - added `syncRuntimeClass()`
* `edge/pkg/metamanager/client/runtimeclass.go` - added RuntimeClass client
* `edge/pkg/metamanager/client/metaclient.go` - added `RuntimeClassesGetter` and `RuntimeClasses()`
* Added unit tests

### Testing

* `go build ./...` clean.
* `go test ./cloud/pkg/edgecontroller/... ./cloud/pkg/cloudhub/authorization/... ./edge/pkg/metamanager/client/... -v -count=1` clean.
* `go vet` clean.
* `go test -race ./cloud/pkg/cloudhub/authorization/... ./edge/pkg/metamanager/client/...` clean.

### DIFF
--- a/cloud/pkg/cloudhub/authorization/resource_attributes.go
+++ b/cloud/pkg/cloudhub/authorization/resource_attributes.go
@@ -23,6 +23,7 @@ import (
 	certificatesv1 "k8s.io/api/certificates/v1"
 	coordinationv1 "k8s.io/api/coordination/v1"
 	v1 "k8s.io/api/core/v1"
+	nodev1 "k8s.io/api/node/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
@@ -199,4 +200,5 @@ var resourceTypeToKubeResources = map[string]kubeResource{
 	beehivemodel.ResourceTypePodPatch:                 {resource: "pods", subresource: "status", groupVersion: v1.SchemeGroupVersion, namespaced: true},
 	beehivemodel.ResourceTypeLease:                    {resource: "leases", groupVersion: coordinationv1.SchemeGroupVersion, namespaced: true},
 	beehivemodel.ResourceTypeCSR:                      {resource: "certificatesigningrequests", groupVersion: certificatesv1.SchemeGroupVersion, namespaced: false},
+	beehivemodel.ResourceTypeRuntimeClass:             {resource: "runtimeclasses", groupVersion: nodev1.SchemeGroupVersion, namespaced: false},
 }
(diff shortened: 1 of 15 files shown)


DoisLONG (commented):
Thanks for the investigation and the implementation work in this PR.

This change covers a substantial part of the core scope of the upcoming LFX Mentorship project tracked in #7106. While contributions to KubeEdge are always welcome, this work was started before we had aligned on the overall architecture, scope boundaries, compatibility requirements, and end-to-end validation plan for the project.

To avoid fragmented implementation and duplicated work, please pause further development of the remaining RuntimeClass scope and do not submit additional follow-up PRs for now.

Please also convert this PR to draft while the maintainers and mentors review the proposed approach. Before proceeding, we need to agree on several key points, including:

- the complete cloud-to-edge RuntimeClass workflow;
- synchronization for newly joined and reconnected edge nodes;
- offline storage and deletion behavior;
- MetaServer and Edged integration;
- compatibility and upgrade considerations;
- the unit and end-to-end testing strategy.

Please note that this LFX project is not first-come-first-served. Submitting an early or large implementation PR does not provide an advantage in the mentee selection process. Candidate selection will be evaluated separately based on technical understanding, proposal quality, communication, collaboration, and overall project fit.

After the design and scope are reviewed, we will decide whether this PR should continue as-is, be re-scoped, or be deferred and incorporated into the mentorship implementation plan.

PR #7151, opened 2026-08-02
keadm batch: return an error when node processing fails

### LINKED ISSUE #7150: keadm batch returns exit code 0 when node processing fails
**What happened**:

`keadm batch` returns exit code 0 even when processing one or more nodes fails. The failure is printed to the log and written to `batch_process_log.txt`, but it is not propagated to the command result.

`keadm batch` should continue processing the remaining nodes, but return a non-zero exit code after all nodes finish if any node failed.

1. Build `keadm` from the current `master` branch.
2. Create a configuration file:

```yaml
nodes:
  - nodeName: failing-node
    ssh:
      auth:
        type: invalid
```

3. Run:

```bash
keadm batch -c config.yaml
echo $?
```

4. Observe the output:

```text
Failed to process node failing-node: failed to connect to failing-node: unsupported authentication type: invalid
0
```

The command returns exit code `0` even though the node processing failed.

This reproduction does not require a real SSH server. The unsupported authentication type is rejected before an SSH connection is attempted.

The relevant execution path is:

```text
keadm batch
  -> RunE
  -> processBatchProcess
  -> batchProcessNodes
  -> processNode
```

In `keadm/cmd/keadm/app/cmd/edge/batch.go`, errors returned by `processNode` are converted into log messages inside goroutines. The `sync.WaitGroup` only waits for all goroutines to finish, and `batchProcessNodes` unconditionally returns `nil`.

Since `keadm/cmd/keadm/keadm.go` exits with status `1` only when `app.Run()` returns an error, node processing failures do not affect the final process exit code.

This makes it difficult for Jenkins, Ansible, shell scripts, and other automation to detect failed or partially completed batch operations. A possible fix would be to aggregate node processing errors after all nodes have finished, while preserving the existing best-effort behavior of continuing to process the remaining nodes.

**Environment**:

- Kubernetes version: N/A; standalone `keadm batch` reproduction
- KubeEdge version: current `master`, commit `afc7d19423f775d90009f770d01866c73cc2cd2b`
- OS: macOS 26.5, arm64
- Go version: `go1.26.5`
- Remote SSH nodes: Not required

DadaVinqi:
**What type of PR is this?**

/kind bug

`keadm batch` currently logs errors returned by `processNode`, but `batchProcessNodes` always returns `nil` after all node-processing goroutines complete. As a result, SSH failures, unreachable nodes, file-transfer failures, and remote command failures are not propagated to the CLI exit code.

This PR:

1. Collects the result of each node through a buffered channel.
2. Serializes writes to the shared batch log writer.
3. Aggregates node failures after all nodes have finished processing.
4. Returns an error when one or more nodes fail, causing `keadm batch` to exit with a non-zero status.
5. Preserves the existing best-effort behavior: remaining nodes continue to be processed after an individual node fails.

No fail-fast behavior, global cancellation, or automatic rollback is introduced.

Fixes #7150

- The result channel is sized for all configured nodes so workers cannot block while the caller waits for completion.
- Regression tests mock `processNode` and do not require SSH connectivity or user configuration.
- The tests cover:
  - all nodes succeeding;
  - a single failed node;
  - mixed successful and failed nodes;
  - processing all nodes before returning the aggregated error;
  - one log line per processed node.

Validation:

- Targeted batch tests pass with the repository's `-gcflags "all=-N -l"` test configuration.
- The invalid SSH authentication reproduction now exits with code `1`.
- `git diff --check` passes.
- `make test WHAT=keadm` runs the new batch tests successfully, but the overall command is affected by unrelated platform-dependent tests involving cloud, debug, edge, Helm, and system utilities.

Yes. `keadm batch` now returns a non-zero exit code when one or more nodes fail, while continuing to process the remaining nodes.

```release-note
keadm batch now returns a non-zero exit code when one or more nodes fail, while continuing to process the remaining nodes.
```

### DIFF
--- a/keadm/cmd/keadm/app/cmd/edge/batch.go
+++ b/keadm/cmd/keadm/app/cmd/edge/batch.go
@@ -235,8 +235,15 @@ func batchProcessNodes(cfg *common.Config, logWriter *bufio.Writer) error {
 	if cfg.MaxRunNum == 0 {
 		cfg.MaxRunNum = defaultMaxRunNum
 	}
+	type batchNodeResult struct {
+		nodeName string
+		message  string
+		err      error
+	}
+
 	var wg sync.WaitGroup
 	sem := make(chan struct{}, cfg.MaxRunNum)
+	results := make(chan batchNodeResult, len(cfg.Nodes))
 	for _, node := range cfg.Nodes {
 		wg.Add(1)
 		go func(node common.Node) {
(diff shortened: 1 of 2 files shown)


DadaVinqi:
Rebased onto the latest `master` and updated the signed commit message to satisfy the repository's commit-message policy.

Validation:
- `go test -gcflags=\"all=-N -l\" ./keadm/cmd/keadm/app/cmd/edge -run 'TestBatchProcessNodes' -count=1 -v`
- `git diff --check origin/master...HEAD`

The invalid-commit-message label is now cleared and the PR is ready for review.

/cc @kevin-wangzefeng @zhu733756

PR #7160, opened 2026-08-04
cloudhub: cleanup ackMessageCache when sendMessageWithRetry exits

### LINKED ISSUE #6175: Is there something i didn't found out about the channel named ackChan make in node_session.go of cloudhub module in method sendMessageWithRetry
I am reading the code of kubeedge, when i read the sendMessageWithRetry method of cloudhub module file name node_session.go, i found there is only one close call to ackChan, but i think there should have a timeout call to close the channel.

The channel named ackChan make in file node_session.go of cloudhub module in method sendMessageWithRetry, i didn't found the close method called after timeout，i am wondering what would happed if there is a lot of ack message not responsed, will there be a memory leak?

**Environment**:
git hub dev version

vee1e:
**What type of PR is this?**

/kind bug

CloudHub's `sendMessageWithRetry` function tracks each message it sends to an edge node and waits for an acknowledgement. When the edge node confirms a message, CloudHub removes the tracking entry. This cleanup happens only on acknowledgement.

Two situations skip that cleanup:

1. Sending the message fails.
2. The edge node never acknowledges the message, and the retry limit runs out.

In both cases, the tracking entry stays in memory forever. A busy edge node with many failed or unacknowledged messages slowly builds up memory that is never freed.

This PR makes sure the tracking entry is removed no matter how the send attempt ends. It also stops the retry timer when the attempt ends.

Messages that time out still get retried later, exactly as before. The fix only stops the leftover memory growth.

Fixes #6175

The change is small and contained to the message-sending path. A late acknowledgement after cleanup does not cause a problem. The system simply ignores it.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/cloudhub/session/node_session.go
+++ b/cloud/pkg/cloudhub/session/node_session.go
@@ -338,10 +338,14 @@ func (ns *NodeSession) syncAckMessage() (bool, error) {
 func (ns *NodeSession) sendMessageWithRetry(copyMsg, msg *beehivemodel.Message) error {
 	ackChan := make(chan struct{})
 	ns.ackMessageCache.Store(copyMsg.GetID(), ackChan)
+	// Delete the ack channel from the cache on every return path.
+	// This prevents a leak when the edge node never acknowledges the message.
+	defer ns.ackMessageCache.Delete(copyMsg.GetID())
 
 	// initialize retry count and timer for sending message
 	retryCount := 0
 	ticker := time.NewTimer(sendRetryInterval)
+	defer ticker.Stop()
 
 	err := ns.connection.WriteMessageAsync(copyMsg)
 	if err != nil {


vee1e:
hi @DoisLONG could i get a review on this if/when you have time?

DoisLONG (commented):
Thanks for the PR.

The fix direction looks right to me. `ackMessageCache` currently stores the ack channel before sending the message, but the entry is only removed on the ack path. Adding a deferred `Delete` makes the cleanup happen on all return paths, including initial send failure and retry exhaustion. Stopping the retry timer on exit also makes sense.

One thing I would like to see is a small regression test for the cleanup behavior. Since this PR is specifically fixing a leak in the unacknowledged / failed-send paths, could you please add focused coverage showing that the `ackMessageCache` entry is removed when `sendMessageWithRetry` exits without receiving an ack?

The two most useful cases would be:

1. `WriteMessageAsync` fails and `sendMessageWithRetry` returns early.
2. the message is sent but no ack is received until the retry limit is reached.

vee1e:
thanks for the review @DoisLONG!

added a focused regression test in https://github.com/kubeedge/kubeedge/pull/7160/commits/5fd7d0224567f1f8471e012c004d839093a7da4c covering the cleanup behavior. The test verifies that the entry is removed in three cases:

- `WriteMessageAsync` fails and `sendMessageWithRetry` returns early.
- The message is sent but no ack arrives until the retry limit is reached.
- The edge node acknowledges the message normally.

The test fails without the fix and passes with it.

DoisLONG (commented):
Thanks for the update.

The test coverage looks good to me now. It covers the cleanup behavior for the main paths I was worried about: normal ack, `WriteMessageAsync` failure, and retry timeout without ack.

The code change is also still small and focused. The deferred `ackMessageCache.Delete()` makes the cleanup happen on every return path, and stopping the retry timer on exit makes sense.

One final cleanup before LGTM: this PR currently has multiple commits. Could you please squash them into a single logical commit?

vee1e:
@doislong done, squashed into one commit

DoisLONG (commented):
Thanks for the update.

/lgtm

PR #7163, opened 2026-08-05
metaserver: fix application concurrency races, state transition in Reset(), and cleanup in CloseApplication

### LINKED ISSUE #7162: [metaserver] Fix concurrency data race, improper state transition in Reset(), and unhandled watch cleanup in ApplicationAgent
**What happened**:

While analyzing `pkg/metaserver` and `edge/pkg/metamanager/metaserver/agent`, several concurrent state-management and synchronization defects were identified in the `Application` and `ApplicationAgent` subsystems:

1. **Data Race on Application Fields**: `app.Status`, `app.Reason`, `app.Error`, `app.RespBody`, `app.ctx`, and `app.cancel` are read and modified across concurrent goroutines (`Agent.doApply`, `Agent.Apply`, `Application.Reset`, `Application.Close`) without mutex protection. The current `countLock` only protects `app.count` and `app.Timestamp`. Running EdgeCore under high concurrency or under the Go race detector (`go test -race ./pkg/metaserver/...`) flags data races between concurrent `GetStatus()`, `Reset()`, `RespBodyTo()`, and `doApply()` writes.

2. **Missing State Transition to `InApplying` / `PreApplying` in `Reset()`**: When an application reaches `Completed` status and is re-used by a subsequent request, `Agent.Apply` calls `app.Reset()` and spawns `go a.doApply(app)`. However, `app.Reset()` does not update `app.Status` to `InApplying` (or `PreApplying`). As a result, `app.Status` remains `Completed` while `doApply()` is running in the background. If a second concurrent request calls `Apply()` on the same application during this window, it sees `app.Status == Completed`, invokes `app.Reset()` again, and launches a duplicate, redundant `doApply()` goroutine to CloudCore.

3. **Unsynchronized `Wait()` and Context Re-initialization**: `Application.Wait()` reads `a.ctx` without synchronization, while `Reset()` concurrently re-assigns `a.ctx` (`a.ctx, a.cancel = context.WithCancel(...)`). If `Wait()` evaluates `a.ctx` while `Reset()` is replacing it, goroutines can block on a stale or canceled context channel.

4. **Inconsistent Cleanup in `CloseApplication`**: `CloseApplication` removes the application from `a.Applications` map without calling `app.Cancel()` or cleaning up associated background resources, leading to potential resource leaks during rapid list/watch re-subscriptions.

1. All reads and writes to `Application` fields (`Status`, `Reason`, `Error`, `RespBody`, `ctx`, `cancel`) should be thread-safe under concurrent access.
2. `app.Reset()` (or `Agent.Apply`) should atomically transition `app.Status` to `InApplying` before launching `doApply()`, preventing duplicate background requests to CloudCore.
3. `Application.Wait()` and `Reset()` should safely synchronize access to context channels.
4. `CloseApplication` should properly cancel and clean up application contexts upon removal.

1. Run unit tests in `pkg/metaserver` with the Go race detector enabled:
   ```bash
   go test -race -v ./pkg/metaserver/...
   ```
2. Inspect `edge/pkg/metamanager/metaserver/agent/agent.go`:
   - In `Apply()` (line 101):
     ```go
     case metaserver.Completed:
         app.Reset()
         go a.doApply(app)
     ```
   - In `Reset()` (`pkg/metaserver/application.go`, line 177):
     ```go
     func (a *Application) Reset() {
         if a.ctx != nil && a.cancel != nil {
             a.cancel()
         }
         a.ctx, a.cancel = context.WithCancel(beehiveContext.GetContext())
         a.Reason = ""
         a.RespBody = []byte{}
     }
     ```
     `app.Status` is not updated here, leaving it as `Completed` during asynchronous `doApply()` execution.

### Proposed Fix Plan:

1. Extend `Application` mutex protection to cover all mutable fields (`Status`, `Reason`, `Error`, `RespBody`, `ctx`, `cancel`).
2. Update `Reset()` to reset `Status` to `PreApplying` (or atomically transition to `InApplying` in `Agent.Apply`).
3. Ensure `Application.Wait()` safely reads the active context under lock.
4. Update `CloseApplication` to properly cancel contexts when removing applications.
5. Add concurrent unit test cases in `pkg/metaserver/application_test.go` and `edge/pkg/metamanager/metaserver/agent/agent_test.go`.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.30.0
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): v1.23.0-alpha

  - Hardware configuration (e.g. `lscpu`): x86_64
  - OS (e.g. `cat /etc/os-release`): Ubuntu 22.04 LTS
  - Kernel (e.g. `uname -a`): Linux 5.15.0-88-generic
  - Others: N/A

  </details>

  - edgecore version (e.g. `edgecore --version`): v1.23.0-alpha
  - Hardware configuration (e.g. `lscpu`): x86_64
  - OS (e.g. `cat /etc/os-release`): Ubuntu 22.04 LTS
  - Kernel (e.g. `uname -a`): Linux 5.15.0-88-generic
  - Others: N/A

  </details>

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

This PR fixes several systemic concurrency defects in the `Application` and `Agent` subsystems in `pkg/metaserver` and `edge/pkg/metamanager/metaserver/agent`, identified in issue #7162.

1. **Data Race Protection**: Adds a pointer `RWMutex` (`mu *sync.RWMutex`) to `Application` to protect concurrent access to `Status`, `Reason`, `Error`, `RespBody`, `ctx`, and `cancel`. Introduces atomic accessors: `SetStatus`, `SetReason`, `GetReason`, `GetError`, `UpdateFromResponse`, and lazy lock helpers (`lock`, `unlock`, `rLock`, `rUnlock`).
2. **State Transition in `Reset()`**: `Application.Reset()` now transitions `Status` back to `PreApplying`. Previously, after `Close()` set `Status = Completed` and a second request called `Apply()`, the `Completed` state triggered `app.Reset()` followed immediately by `go a.doApply(app)`. Since `Reset()` did not update `Status`, a third concurrent `Apply()` call could also see `Completed` and launch a duplicate `doApply()` goroutine, causing redundant requests to CloudCore.
3. **Safe `Wait()` Context Read**: `Application.Wait()` now snapshots `a.ctx` under the read lock before blocking on the channel, preventing a race with `Reset()` which reassigns `a.ctx`.
4. **Resource Cleanup in `CloseApplication()`**: `Agent.CloseApplication()` now calls `app.Cancel()` before removing the application from the `sync.Map`, ensuring goroutines blocked in `Wait()` are released.
5. **Unit Tests**: Adds unit tests for `SetStatus`, `SetReason`, `GetError`, `UpdateFromResponse`, the `Reset` status transition, and a concurrent stress test `TestConcurrentStatusAccess` that exercises all protected fields under 50 concurrent goroutines.

Fixes #7162

- Using `*sync.RWMutex` (pointer) instead of embedded `sync.RWMutex` avoids `go vet` `copylocks` warnings since `Application` is copied by value in map operations and test struct literals.
- The `mu` and `countLock` fields are lazily initialized in helper methods to support `Application{}` struct literals used in existing tests without panicking.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/dynamiccontroller/application/application_test.go
+++ b/cloud/pkg/dynamiccontroller/application/application_test.go
@@ -97,13 +97,13 @@ func TestOptionTo(t *testing.T) {
 }
 
 func TestApplicationString(t *testing.T) {
-	app := &metaserver.Application{
+	app := metaserver.NewApplicationForTest(metaserver.Application{
 		ID:       "test-id",
 		Nodename: "test-node",
 		Key:      "apps/v1/deployments/default/nginx",
 		Verb:     metaserver.Get,
 		Status:   metaserver.Approved,
-	}
+	})
 
 	str := app.String()
 
@@ -388,8 +388,8 @@ func TestGetWatchDiff(t *testing.T) {
 	}
 
 	allWatchAppInEdge := map[string]metaserver.Application{
-		"listener1": {ID: "listener1", Nodename: "node1"},
-		"listener3": {ID: "listener3", Nodename: "node1"},
+		"listener1": *metaserver.NewApplicationForTest(metaserver.Application{ID: "listener1", Nodename: "node1"}),
+		"listener3": *metaserver.NewApplicationForTest(metaserver.Application{ID: "listener3", Nodename: "node1"}),
 	}
 
 	added, removed := center.getWatchDiff(allWatchAppInEdge, "node1")
--- a/edge/pkg/metamanager/metaserver/agent/agent.go
+++ b/edge/pkg/metamanager/metaserver/agent/agent.go
@@ -102,9 +102,10 @@ func (a *Agent) Apply(app *metaserver.Application) error {
 		app.Reset()
 		go a.doApply(app)
 	case metaserver.Rejected:
-		return &app.Error
+		appErr := app.GetError()
+		return &appErr
 	case metaserver.Failed:
-		return errors.New(app.Reason)
+		return errors.New(app.GetReason())
 	case metaserver.Approved:
 		return nil
 	case metaserver.InApplying:
@@ -112,10 +113,11 @@ func (a *Agent) Apply(app *metaserver.Application) error {
 	}
 	app.Wait()
 	if app.GetStatus() == metaserver.Rejected {
-		return &app.Error
+		appErr := app.GetError()
+		return &appErr
 	}
 	if app.GetStatus() != metaserver.Approved {
-		return errors.New(app.Reason)
+		return errors.New(app.GetReason())
 	}
 	return nil
 }
@@ -123,32 +125,33 @@ func (a *Agent) Apply(app *metaserver.Application) error {
 func (a *Agent) doApply(app *metaserver.Application) {
 	defer app.Cancel()
 	// encapsulate as a message
-	app.Status = metaserver.InApplying
+	app.SetStatus(metaserver.InApplying)
 	msg := model.NewMessage("").SetRoute(metaserver.MetaServerSource, modules.DynamicControllerModuleGroup).FillBody(app)
 	msg.SetResourceOperation("null", "null")
 	resp, err := beehiveContext.SendSync(edgemodule.EdgeHubModuleName, *msg, 10*time.Second)
 	if err != nil {
-		app.Status = metaserver.Failed
-		app.Reason = fmt.Sprintf("failed to access cloud Application center: %v", err)
+		app.SetStatus(metaserver.Failed)
+		app.SetReason(fmt.Sprintf("failed to access cloud Application center: %v", err))
 		return
 	}
 
 	retApp, err := metaserver.MsgToApplication(resp)
 	if err != nil {
-		app.Status = metaserver.Failed
-		app.Reason = fmt.Sprintf("failed to get Application from resp msg: %v", err)
+		app.SetStatus(metaserver.Failed)
+		app.SetReason(fmt.Sprintf("failed to get Application from resp msg: %v", err))
 		return
 	}
 
-	//merge returned application to local application
-	app.Status = retApp.Status
-	app.Reason = retApp.Reason
-	app.Error = retApp.Error
-	app.RespBody = retApp.RespBody
+	// merge returned application to local application under a single lock
+	app.UpdateFromResponse(retApp)
 }
 
 func (a *Agent) CloseApplication(appID string) {
-	a.Applications.Delete(appID)
+	if v, ok := a.Applications.LoadAndDelete(appID); ok {
+		// Cancel the application context to release any goroutines blocked in Wait()
+		app := v.(*metaserver.Application)
+		app.Cancel()
+	}
 	a.SyncWatchAppOnConnected()
 }
 
@@ -207,7 +210,9 @@ func (a *Agent) syncWatchApplications() error {
 		return err
 	}
 
-	klog.Errorf("failed to process watch apps: %+v", failedWatchApps)
+	if len(failedWatchApps) > 0 {
+		klog.Warningf("failed to process watch apps: %+v", failedWatchApps)
+	}
 
 	return nil
 }
--- a/edge/pkg/metamanager/metaserver/agent/agent.go
+++ b/edge/pkg/metamanager/metaserver/agent/agent.go
@@ -97,14 +97,18 @@ func (a *Agent) Apply(app *metaserver.Application) error {
 	app = store.(*metaserver.Application)
 	switch app.GetStatus() {
 	case metaserver.PreApplying:
-		go a.doApply(app)
+		if app.TryStartApplying() {
+			go a.doApply(app)
+		}
 	case metaserver.Completed:
-		app.Reset()
-		go a.doApply(app)
+		if app.TryResetAndStartApplying() {
+			go a.doApply(app)
+		}
 	case metaserver.Rejected:
-		return &app.Error
+		appErr := app.GetError()
+		return &appErr
 	case metaserver.Failed:
-		return errors.New(app.Reason)
+		return errors.New(app.GetReason())
 	case metaserver.Approved:
 		return nil
 	case metaserver.InApplying:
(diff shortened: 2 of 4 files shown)


bhuvan-somisetty:
Hi @chendave @DoisLONG @WillardHu @fisherxu @tangming1996 @subpathdev, please help review this PR when you get a chance. Thanks!

DoisLONG (commented):
Thanks for the PR.

The direction looks right to me. Using accessors for the mutable Application fields, updating the response fields under one lock, and cancelling the app from `CloseApplication()` are all reasonable.

Two things:

1. The lazy initialization of `mu` is not safe. `lock()` / `rLock()` currently do `if a.mu == nil { a.mu = &sync.RWMutex{} }` without any guard. If a zero-value `Application` is accessed by multiple goroutines, the first lock initialization can race, and different goroutines may even end up locking different mutexes. Please either make the initialization path safe, or make sure all `Application` instances are created with initialized locks and remove the unsafe lazy init.

2. Please double-check the duplicate `doApply()` fix. `Reset()` now sets the status back to `PreApplying`, but `InApplying` is still set later inside `doApply()`. There is still a window where another `Apply()` may observe `PreApplying` before the goroutine runs. I think the transition to `InApplying` should happen synchronously under the same lock before starting the goroutine, or be handled by a single atomic helper.

Main CI, codespell, CIFuzz and FOSSA look good, with ARM64 cancelled. But for this PR, I would also like to see the relevant package pass with `go test -race`

bhuvan-somisetty:
Thanks for the detailed review @DoisLONG! I've updated the PR to address both points:

1. **Thread-safe Lock Init**: \ensureInit()\ now uses atomic CAS (\atomic.CompareAndSwapPointer\) to safely initialize \mu\ and \countLock\ without any data race even if an uninitialized \Application\ struct is accessed concurrently.
2. **Atomic \InApplying\ State Transition**: Introduced \TryStartApplying()\ and \TryResetAndStartApplying()\ to atomically transition \Status\ to \InApplying\ under lock *before* spawning \go a.doApply(app)\, closing the duplicate \doApply()\ race window.

Also added unit tests covering the new atomic state transitions in \application_test.go\. Please take another look when you get a chance!

DoisLONG (commented):
Thanks for the update.

The `TryStartApplying()` / `TryResetAndStartApplying()` change is a good improvement. Moving the transition to `InApplying` before starting `doApply()` addresses the duplicate `doApply()` window I was worried about.

One thing still does not look right to me: the new lazy lock initialization uses `unsafe` and `atomic.CompareAndSwapPointer` for `mu` and `countLock`.

Since this PR is fixing concurrency races, I would prefer not to introduce a new unsafe lock-initialization path here. The current code still does a normal `if a.mu == nil` / `if a.countLock == nil` check before the atomic CAS, which makes the initialization logic harder to reason about.

Could we avoid lazy initialization here? For production paths, it looks like `Application` can be initialized explicitly in `NewApplication()`, `MsgToApplication()`, and `MsgToApplications()`. For tests using `Application{}` literals, I think it would be cleaner to initialize the locks in the test setup/helper instead of adding `unsafe` to the main implementation.

Please also check the current CI issue and fix it together with this cleanup. Since this PR is about concurrency safety, I would also like to see the relevant package pass with `go test -race`.

After the lock initialization is simplified and the CI issue is fixed, I can take another look.

DoisLONG (commented):
Thanks for the update.

The `TryStartApplying()` / `TryResetAndStartApplying()` change is good. Moving the transition to `InApplying` before starting `doApply()` addresses the duplicate `doApply()` window I was worried about.

But the main concern from my previous review is still there. The implementation still uses `unsafe` and `atomic.CompareAndSwapPointer` to lazily initialize `mu` and `countLock`.

Since this PR is specifically fixing concurrency races, I do not think we should add a new unsafe lock-initialization path in the main implementation. The current `ensureInit()` also still does a normal nil check before the atomic CAS, which makes the initialization path harder to reason about.

Could you please avoid the lazy initialization here? For the production paths, `Application` can be initialized explicitly in `NewApplication()`, `MsgToApplication()`, and `MsgToApplications()`. For tests that use `Application{}` literals, I think it is cleaner to initialize the locks in the test setup/helper instead of adding `unsafe` to the production code.

Also, Main CI is currently failing on the latest head. Please check and fix the CI failure together with this cleanup. Since this is a concurrency-safety fix, I would also like to see the relevant package pass with `go test -race`.

After the unsafe lock initialization is removed and the CI failure is fixed, I can take another look.

bhuvan-somisetty:
Thanks for the detailed review, @DoisLONG.

Removed the `unsafe.Pointer`/`atomic.CompareAndSwapPointer` lazy lock initialization entirely. `Application.mu` and `Application.countLock` are now always initialized explicitly at construction time — in `NewApplication`, `MsgToApplication`, and `MsgToApplications` (the last one was missing this before and has been fixed) — and `lock`/`unlock`/`rLock`/`rUnlock` plus the count helpers now assume the locks are already set rather than lazily creating them.

For tests that build an `Application` via a struct literal outside this package, I added `NewApplicationForTest`, which returns an `*Application` with both locks initialized. Same-package tests just set `mu`/`countLock` directly on the literal.

Locally verified `go build`, `go vet`, and `go test` are clean for `pkg/metaserver`, `edge/pkg/metamanager/metaserver/agent`, and `cloud/pkg/dynamiccontroller/application` (couldn't run `-race` locally — this box has no cgo toolchain — but it should be exercised by CI here).

Ready for another look whenever you have time.

DoisLONG (commented):
Thanks for the update.

This version looks much better to me.

The unsafe / atomic lazy lock initialization has been removed, and `mu` / `countLock` are now initialized explicitly in `NewApplication()`, `MsgToApplication()`, and `MsgToApplications()`. That addresses the main concern I had about adding an unsafe lock-initialization path in a concurrency-fix PR.

The `TryStartApplying()` / `TryResetAndStartApplying()` path also still looks good. Moving the transition to `InApplying` before starting `doApply()` closes the duplicate `doApply()` window.

A few final cleanups:

2. The PR description still mentions lazy lock helpers / lazy initialization, but the latest implementation no longer does lazy initialization. Please update the description to match the current code.
3. CI is still running on the latest head from what I can see. Main CI and CIFuzz are still in progress, and ARM64 is queued. Since this PR is about concurrency safety, I would still like to see the relevant checks finish, especially the race-related validation if covered by CI.

After those are handled, this should be good from my side.

PR #7164, opened 2026-08-05
edgehub: requeue message when sending to cloud fails

### LINKED ISSUE #6615: Reliability of custom message routing
I was testing the reliability of sending messages from Edge to Cloud using an EventBus to Rest Rule. Specifically, I was testing what happens when the connection between Edge and Cloud goes down.
I consistently ran into the following issue: The first message sent after the connection goes down is lost and is not received on the Cloud side.

Setup EventBus to Rest Rule.
Setup Rest listener.
Continually send messages to MQTT broker.
Cut connection.
Wait for some messages to be sent on Edge side.
Restore connection.
Verify which messages arrive on Cloud side after the connection is restored.

The problem is in the following part of the code:
https://github.com/kubeedge/kubeedge/blob/a4926f0c9a55cc8bfccb552bc6eb303db45d94f2/edge/pkg/edgehub/process.go#L96-L102

If the EdgeHub is unable to send a message to the Cloud, it will drop the connection and try to reconnect. The problem is that the message that it failed to send is not requeued. I was able to fix the issue by adding the following line to the `if err != nil` block:
`beehiveContext.Send(modules.EdgeHubModuleName, message)`

My question is the following: How reliable is the custom message routing of KubeEdge? Is the issue I encountered an oversight or by design?

**Environment**:
- Kubernetes version (use `kubectl version`): 1.34.0
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): 1.22.0

vee1e:
**What type of PR is this?**

/kind bug

EdgeHub sends messages from the edge to the cloud over one connection. When the connection breaks, the message that is being sent at that moment is lost. EdgeHub then closes the connection and starts reconnecting. The failed message is not put back into the queue.

This PR puts the failed message back into EdgeHub's queue before reconnecting. After the new connection is established, EdgeHub sends the message again. No message is lost when the connection breaks.

Messages published while the connection is down are already buffered. They are sent after reconnection. This change covers the one message that is in flight when the connection breaks.

Fixes #6615

The change is one line in the message-routing loop, plus a short comment. It reuses the existing queue, so no new state is added.

I validated the change on a live single-machine KubeEdge setup:

- The full path works end to end: MQTT topic to EventBus to EdgeHub to CloudCore router to a REST sink.
- I cut the connection and confirmed that EdgeHub detects the break and enters the reconnect loop.
- Messages published during the outage wait in the queue and are not dropped.

```release-note
NONE
```

### DIFF
--- a/edge/pkg/edgehub/process.go
+++ b/edge/pkg/edgehub/process.go
@@ -97,6 +97,9 @@ func (eh *EdgeHub) routeToCloud() {
 		err = eh.sendToCloud(message)
 		if err != nil {
 			klog.Errorf("failed to send message to cloud: %v", err)
+			// Requeue the message so it is sent again after reconnection.
+			// Otherwise the message is lost when the connection drops.
+			beehiveContext.Send(modules.EdgeHubModuleName, message)
 			eh.reconnectChan <- struct{}{}
 			return
 		}


vee1e:
hi @DoisLONG could i get a review on this if/when you have time?

DoisLONG (commented):
Thanks for the PR.

Requeueing the in-flight message before triggering reconnect should prevent the specific case where the connection breaks while EdgeHub is sending a message and that message is otherwise lost.

I would like to see two small follow-ups.

First, could you please add a focused regression test for this path? Since this PR changes the delivery behavior in the EdgeHub reconnect path, it would be useful to cover the case where `sendToCloud(message)` returns an error and verify that the message is put back into the EdgeHub queue before reconnect is triggered.

Second, please clarify the duplicate-delivery semantics. If `sendToCloud` returns an error after the message has already been partially written or accepted by the peer, requeueing may cause the same message to be delivered again after reconnect. That may be acceptable as at-least-once delivery, but it would be good to document that assumption in the PR description or a short code comment.

vee1e:
@DoisLONG thanks for the review, and for the clear summary of the delivery behavior.

1. added a focused unit test that makes sendToCloud return an error and verifies the message is requeued into the EdgeHub queue before the reconnect signal is sent. The test fails without the fix and passes with it.

2. comment added stating that the cloud may already have received the message, so at-least-once delivery means it can arrive twice.

DoisLONG (commented):
Thanks for the update.

The two points from my previous review look addressed now.

The added test covers the send failure path and verifies that the message is put back into the EdgeHub queue before reconnect is triggered. The code comment also documents the at-least-once delivery semantics, including the possibility that the cloud may already have received the message and the retry may deliver it again.

The code change is still small and focused.

vee1e:
@doislong done, squashed into one commit

DoisLONG (commented):
Thanks for the update.

The PR is now squashed into a single commit, and the code still looks good to me.

The send-failure path now requeues the in-flight message before triggering reconnect, which addresses the lost-message case when the connection breaks during `sendToCloud(message)`. The code comment also documents the at-least-once delivery behavior, including the possibility that the cloud may already have received the message and the retry may deliver it again.

The added regression test covers the path I was looking for: `sendToCloud` fails, the message is put back into the EdgeHub queue, and reconnect is triggered.

/lgtm

PR #7168, opened 2026-08-06
Fix race condition in dtmanager callback initialization

### LINKED ISSUE #7167: bug: missing sync.Once in initDeviceActionCallBack and initMemActionCallBack causing potential map race condition

Use `sync.Once` for one-time initialization of `deviceActionCallBack` in `edge/pkg/devicetwin/dtmanager/device.go` and `memActionCallBack` in `edge/pkg/devicetwin/dtmanager/membership.go`, matching the pattern already used in `edge/pkg/devicetwin/dtmanager/twin.go`.

Proposed changes:
1. Add `initDeviceActionCallBackOnce sync.Once` in `device.go` and wrap `initDeviceActionCallBack()` in `initDeviceActionCallBackOnce.Do(...)`.
2. Add `initMemActionCallBackOnce sync.Once` in `membership.go` and wrap `initMemActionCallBack()` in `initMemActionCallBackOnce.Do(...)`.

**Why is this needed**:

Currently, `initTwinActionCallBack()` in `twin.go` uses `sync.Once` to prevent re-initialization. However, `initDeviceActionCallBack()` and `initMemActionCallBack()` execute `make(map[string]CallBack)` directly every time `DeviceWorker.Start()` or `MemWorker.Start()` is called (such as during module restarts). 

Re-creating these global maps without synchronization can lead to race conditions and potential map read/write panics when workers start or restart concurrently. Adding `sync.Once` ensures thread-safe, single initialization across the `dtmanager` package.

rahulkr182:
**What type of PR is this?**

/kind bug

This PR fixes a potential map race condition in the `dtmanager` package. Currently, `initTwinActionCallBack()` in `twin.go` uses `sync.Once` to prevent re-initialization. However, `initDeviceActionCallBack()` and `initMemActionCallBack()` execute `make(map[string]CallBack)` directly every time `DeviceWorker.Start()` or `MemWorker.Start()` is called (e.g., during module restarts). 

This PR adds `sync.Once` for the initialization of `deviceActionCallBack` and `memActionCallBack` to ensure thread-safe, single initialization across the `dtmanager` package, matching the existing pattern in `twin.go`.

Fixes #7167

### DIFF
--- a/edge/pkg/devicetwin/dtmanager/device.go
+++ b/edge/pkg/devicetwin/dtmanager/device.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"k8s.io/klog/v2"
@@ -21,7 +22,8 @@ import (
 
 var (
 	//deviceActionCallBack map for action to callback
-	deviceActionCallBack map[string]CallBack
+	deviceActionCallBack         map[string]CallBack
+	initDeviceActionCallBackOnce sync.Once
 
 	// DeviceServiceFactory is a function variable that can be mocked in tests
 	DeviceServiceFactory = func() interface {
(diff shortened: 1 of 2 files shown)


rahulkr182:
Hi @DoisLONG , could you please review this pr .

DoisLONG (commented):
Thanks for the PR.

This looks good to me. The change is small and keeps the device and membership callback initialization consistent with the existing `sync.Once` pattern used by the twin callback initialization.

Using `sync.Once` here avoids recreating `deviceActionCallBack` and `memActionCallBack` when the workers are started more than once, and prevents concurrent/repeated initialization from replacing the callback maps while they may already be in use.

/lgtm

rahulkr182:
Hi @DoisLONG , How can this PR be merged?

PR #7178, opened 2026-08-06
admissioncontroller: validate Rule and RuleEndpoint updates

### LINKED ISSUE #7173: admissioncontroller: Rule and RuleEndpoint webhooks deny every UPDATE request
**What happened**:

In cloud/pkg/admissioncontroller, the rules and ruleendpoints validating
webhooks are registered for CREATE, UPDATE and DELETE (admission.go:227-230 and
:255-258, both FailurePolicy: Fail at :247 and :275), but the handlers implement
only CREATE and DELETE/CONNECT and reject everything else:

    admit_rule.go:25,39,43
    admit_ruleendpoint.go:17,31,35

The default branch returns toAdmissionResponse (common.go:115), which sets only
Result and leaves Allowed false, so UPDATE is denied.

admit_device.go:20 and admit_devicemodel.go:20 use
`case admissionv1.Create, admissionv1.Update:`. These two are the only handlers
that omit Update.

UPDATE validated the same way as CREATE, as the device handlers do.

Calling admitRuleEndpoint with Operation: admissionv1.Update returns
Allowed=false, Result.Message="unsupported webhook operation UPDATE". The same
call against admitDeviceModel returns Allowed=true.

On a cluster this should reject a repeat `kubectl apply`, a `kubectl edit`, or
adding a label to an existing Rule or RuleEndpoint.

Fix is `case admissionv1.Create, admissionv1.Update:` in both handlers, plus
tests for the UPDATE path.

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

Priyanshu6968:
**What type of PR is this?**
/kind bug

The `rules` and `ruleendpoints` validating webhooks are registered for CREATE,
UPDATE and DELETE with `FailurePolicy: Fail` (`admission.go:226-231`, `:254-259`),
but `admitRule` and `admitRuleEndpoint` only handled CREATE and DELETE/CONNECT.
Every UPDATE fell through to the default branch, which returns
`toAdmissionResponse` (`common.go:115`) and leaves `Allowed` false, so the API
server rejected the request. `admitDevice` and `admitDeviceModel` already use
`case admissionv1.Create, admissionv1.Update:`; these two handlers were the only
ones that omitted Update.

This blocks a repeat `kubectl apply`, a `kubectl edit`, and any label or
annotation change on an existing Rule or RuleEndpoint.

It also blocks the router's own status writes. The Rule CRD has no `status`
subresource, so `updateRuleStatus` patches the `rules` resource itself
(`cloud/pkg/edgecontroller/controller/upstream.go:392`), which the API server
presents to the webhook as an UPDATE. The webhook denies it and the error is
only logged at `:394`, so `rule.status.successMessages` / `failMessages` /
`errors` never advance on a cluster running the admission webhook.

Note that adding UPDATE to the switch is not sufficient on its own for rules.
`validateSourceRuleEndpoint` scans the existing rules of the namespace for a
source-property conflict, and on an update that list contains the stored copy of
the rule being updated, which always matches its own `path` (or `topic` +
`node_name`). Updates would still be denied, with a misleading "source properties
exist in Rule" message. This PR skips the rule being validated so only other
rules can conflict with it.

`AdmissionController.CrdClient` is now held as `versioned.Interface` rather than
`*versioned.Clientset`, matching how `cloud/pkg/taskmanager/v1alpha1/util/controller`
already holds it, so the handlers can be tested against the generated fake
clientset.

Fixes #7173 

```release-note
Fixed the Rule and RuleEndpoint validating webhooks rejecting every update, which blocked `kubectl apply`/`kubectl edit` on existing objects and prevented the router from recording rule delivery status.
```

### DIFF
--- a/cloud/pkg/admissioncontroller/admission.go
+++ b/cloud/pkg/admissioncontroller/admission.go
@@ -65,7 +65,7 @@ func addToScheme(scheme *runtime.Scheme) {
 // AdmissionController implements the admission webhook for validation of configuration.
 type AdmissionController struct {
 	Client    *kubernetes.Clientset
-	CrdClient *versioned.Clientset
+	CrdClient versioned.Interface
 }
 
 func strPtr(s string) *string { return &s }
--- a/cloud/pkg/admissioncontroller/admit_rule.go
+++ b/cloud/pkg/admissioncontroller/admit_rule.go
@@ -22,7 +22,7 @@ func admitRule(review admissionv1.AdmissionReview) *admissionv1.AdmissionRespons
 	reviewResponse := admissionv1.AdmissionResponse{}
 
 	switch review.Request.Operation {
-	case admissionv1.Create:
+	case admissionv1.Create, admissionv1.Update:
 		raw := review.Request.Object.Raw
 		rule := rulesv1.Rule{}
 		deserializer := codecs.UniversalDeserializer()
@@ -55,7 +55,7 @@ func validateRule(rule *rulesv1.Rule) error {
 	} else if sourceEndpoint == nil {
 		return fmt.Errorf("source ruleEndpoint %s has not been created", sourceKey)
 	}
-	if err = validateSourceRuleEndpoint(sourceEndpoint, rule.Spec.SourceResource); err != nil {
+	if err = validateSourceRuleEndpoint(sourceEndpoint, rule); err != nil {
 		return err
 	}
 	targetKey := fmt.Sprintf("%s/%s", rule.Namespace, rule.Spec.Target)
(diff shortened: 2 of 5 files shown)


Priyanshu6968:
@DoisLONG  please if you can have a look at this one

DoisLONG (commented):
Thanks for the PR.

This looks good to me. The root cause is clear: the Rule and RuleEndpoint validating webhooks are registered for update operations, but the handlers only accepted create, delete, and connect paths. As a result, normal updates such as `kubectl apply`, `kubectl edit`, metadata changes, and Rule status updates could be rejected by the webhook.

The implementation looks properly scoped. `admitRule` and `admitRuleEndpoint` now handle `Create` and `Update` consistently, and the Rule validation also skips the rule being updated when checking existing source-property conflicts. That avoids the self-conflict case where the stored copy of the same Rule is returned by the list call.

The added tests cover the important cases I would expect here: Rule create/update success, update conflict with another Rule, invalid update validation, EventBus source update, and RuleEndpoint update validation.


/lgtm

PR #7195, opened 2026-08-08
dynamiccontroller: scope defaultMaster EndpointSlice filtering

### LINKED ISSUE #7194: dynamiccontroller: defaultMaster rewrites unrelated EndpointSlices
**What happened**:

The DynamicController `defaultMaster` filter can rewrite every EndpointSlice in a list response, including EndpointSlices that belong to application Services.

For a single object, `FilterImpl.NeedFilter` requires the `default/kubernetes` EndpointSlice. For an `UnstructuredList`, it only checks that the first item has kind `EndpointSlice`. `filter.MessageFilter` then passes every list item to `FilterResource`, whose current implementation does not recheck name or namespace before truncating endpoints and rewriting the first address and `https` port to `127.0.0.1:10550`.

As a result, an unrelated EndpointSlice returned beside the Kubernetes API Service EndpointSlice can be mutated before the response is sent to the edge node.

Only the `default/kubernetes` EndpointSlice should be rewritten to the local MetaServer endpoint. EndpointSlices for every other Service must remain unchanged in both single-object and list responses.

1. Check out `master` at `16cd8548eedb81af293b94f27844144c080eaa89`.
2. Construct an `UnstructuredList` containing an application EndpointSlice followed by `default/kubernetes`.
3. Register the default-master filter and call `filter.MessageFilter` on the list.
4. Observe that both items are truncated and rewritten, although only `default/kubernetes` should change.

The existing tests exercise a single target object but do not cover a mixed EndpointSlice list.

- Affected code: `cloud/pkg/dynamiccontroller/filter/defaultmaster/defaultmaster.go`
- Dispatch path: `cloud/pkg/dynamiccontroller/filter/filter.go`
- Proposed regression coverage: put an unrelated EndpointSlice first, keep `default/kubernetes` second, and assert that only the second object changes.
- This is separate from the optional EndpointPort fix in #7003/#7004. The old PR #4519 is a stale GVR-level filter refactor, not a focused fix for this mutation boundary.

**Environment**:
- Kubernetes version: Not required; reproducible with Kubernetes `discovery/v1` API objects in a focused unit test.
- KubeEdge version: `master` at `16cd8548eedb81af293b94f27844144c080eaa89`.
- Cloud node hardware/OS/kernel: Not required for the source-level reproduction.
- Go version: repository toolchain (`go 1.23.12`).
- Edge node environment: Not applicable; the mutation occurs in CloudCore before the response is sent to the edge.

areycruzer:
**What type of PR is this?**

/kind bug

The default-master filter accepted an EndpointSlice list based only on its first item's kind and then rewrote every item in that list. This could make application EndpointSlices point to the local MetaServer instead of their real backends.

This change:

- centralizes recognition of the `default/kubernetes` EndpointSlice;
- applies that identity check both before selecting a list and before mutating each item; and
- adds a mixed-list regression test proving an unrelated EndpointSlice remains unchanged while the target is still rewritten to `127.0.0.1:10550`.

Fixes #7194

Verified with:

```text
go test ./cloud/pkg/dynamiccontroller/filter/... -count=1
PASS

go test -race ./cloud/pkg/dynamiccontroller/filter/defaultmaster -count=1
PASS

make verify
PASS (clean full checkout; yq v4.44.3 supplied on PATH)
```

The patch changes only the default-master filter and its tests; it does not alter EndpointSlice synchronization, public APIs, configuration, dependencies, or CRDs.

Additional broad-test note: `TMPDIR=/tmp make test WHAT=cloud` passed the modified DynamicController packages, but the overall target remains blocked by an unrelated `cloud/pkg/taskmanager/upstream` gomonkey test failure that reproduces at the base commit `16cd8548e`. The default macOS temporary path also exceeds the Unix-socket limit in existing CSI tests; using `TMPDIR=/tmp` clears that separate failure.

```release-note
Fixed the DynamicController default-master filter so it no longer rewrites EndpointSlices belonging to application Services in list responses.
```


cloud/pkg/dynamiccontroller/filter/defaultmaster/defaultmaster.go
```diff
@@ -40,25 +40,30 @@ func (f *FilterImpl) Name() string {
 	return filterName
 }
 
+func isDefaultMasterEndpointSlice(obj *unstructured.Unstructured) bool {
+	return obj.GetObjectKind().GroupVersionKind().Kind == resourceName &&
+		obj.GetName() == defaultEndpointSliceName &&
```
brice-ruppen, line 45: Thanks for working on this. I may be misunderstanding the expected EndpointSlice shape.

On OVH Managed Kubernetes, the default/kubernetes EndpointSlice looks like this:
```
metadata:
  name: kubernetes-8gbjh
  generateName: kubernetes-
  namespace: default
  labels:
    kubernetes.io/service-name: kubernetes
```

The test isDefaultMasterEndpointSlice appear to expect the EndpointSlice name to be exactly _kubernetes_.

Is my provider producing an invalid or unusual EndpointSlice here, or should KubeEdge also support generated names (and potentially multiple slices) for the kubernetes Service?

I can provide a fuller reproduction if useful.

PR #7215, opened 2026-08-10
cloudhub: eliminate synchronous K8s API calls in DispatchDownstream

### LINKED ISSUE #6876: [Bug] CloudHub DispatchDownstream single-goroutine bottleneck causes node NotReady when new node joins
# CloudHub DispatchDownstream single-goroutine bottleneck causes node NotReady when new node joins

## Bug Description

When a new edge node joins the cluster, existing edge nodes intermittently become NotReady and then self-recover. The root cause is that `DispatchDownstream` in `cloud/pkg/cloudhub/dispatcher/message_dispatcher.go` is a single goroutine consuming messages from a single beehivecontext channel (capacity 1024). All downstream messages for ALL nodes (lease response, node status, pod creation, initial sync, etc.) share this one FIFO channel with no priority or per-node isolation.

## Reproduction Steps

1. Deploy KubeEdge with 2+ edge nodes (e.g., edge-node-01, edge-node-02)
2. Install edgemesh on edge nodes (adds many DestinationRule/ServiceEntry resources)
3. Create a new edge node (edge-node-03) and register it to the cluster
4. Observe existing edge nodes intermittently becoming NotReady

## Expected Behavior

Existing edge nodes should not be affected when a new node joins. Lease responses and other critical messages should be delivered within the kube-controller-manager 40s heartbeat threshold.

## Actual Behavior

Existing nodes become NotReady. Analysis from production logs shows:

### Timeline (MessageID: eed7da35-26e4-4c48-853f-00275c09a9df)

| Time | Event |
|------|-------|
| 15:22:10.447 | CloudCore receives lease message from edge-node-01 |
| 15:22:10.637 | Lease updated in API server, response sent to beehivecontext channel |
| **15:22:44.917** | **DispatchDownstream finally reads the response from channel (34s delay)** |
| 15:22:44.918 | Response sent to edge-node-01 |

### Message traffic during the 34s delay window (15:22:10~15:22:44)

| Node | Messages dispatched | Percentage |
|------|-------------------|------------|
| edge-node-03 (new node) | **1034** | 97.6% |
| edge-node-01 | 13 | 1.2% |
| edge-node-02 | 12 | 1.1% |
| **Total** | **1059** | 100% |

The new node's initial sync messages (namespace, destinationrule, serviceentry from DynamicController) overwhelmed the shared channel, causing edge-node-01's lease response to be queued for 34 seconds. The next lease report exceeded the 40s threshold, triggering kube-controller-manager to mark the node NotReady.

## Root Cause Analysis

```
EdgeController ──Send()──→ CloudHub channel (1024) ──→ DispatchDownstream ──→ per-node Queue
DynamicController ──Send()──→        ↑ single goroutine                    ↑ parallel
DeviceController ──Send()──→        ↑                                      per-node
All Controllers ──Response()──→      ↑
```

1. **Single consumption point**: `DispatchDownstream` is a single goroutine calling `beehivecontext.Receive()` in a loop (`message_dispatcher.go:126-161`)
2. **No node isolation**: Node A's 1000+ messages block Node B's 1 critical message in the shared channel
3. **No priority**: Lease/nodeStatus and initial sync messages have equal priority
4. **No flow control**: `sendAllObjects` in `dynamiccontroller/application/listener.go:34-42` sends all initial sync messages in an unbounded loop
5. **Sync API calls block main loop**: `enqueueNonNamespacedResource`/`enqueueNamespacedResource` create objectSync via K8s API when not found, further blocking the single goroutine

### Per-node dispatch channel (recommended, minimal change)

Add a per-node buffered channel between `DispatchDownstream` and `enqueue`. The main loop only does lightweight routing (parse nodeID → write to node channel), while each node has its own goroutine for the heavier enqueue logic:

```
beehivecontext CloudHub channel (1024)
        ↓
  DispatchDownstream (single goroutine, only routing, no enqueue)
        ↓ parse nodeID
   ┌────┼────┐
   ↓    ↓    ↓
 node-01  node-02  node-03    ← per-node buffered channel (1024)
 channel  channel  channel
   ↓       ↓       ↓
 node-01  node-02  node-03    ← per-node dispatch goroutine
 worker   worker   worker       (enqueueAck/enqueueNoAck)
   ↓       ↓       ↓
 Queue    Queue    Queue
```

**Key changes** (only `message_dispatcher.go`):

- Add `nodeDispatchChannels sync.Map` field
- `DispatchDownstream` calls `dispatchToNodeChannel(nodeID, &msg)` instead of direct enqueue
- `startNodeDispatchLoop(nodeID)` starts a per-node goroutine that reads from the node's channel and calls enqueue
- `AddNodeMessagePool` starts the node dispatch loop; `DeleteNodeMessagePool` stops it

This ensures that a new node's 1000+ initial sync messages only occupy its own channel, without affecting other nodes' lease responses or business messages.

### Additional optimizations (can be done separately)

1. **Lazy objectSync creation**: In `enqueueNonNamespacedResource`/`enqueueNamespacedResource`, when objectSync is not found, directly return `true` to enqueue (skip the sync Create+UpdateStatus API calls), and let `saveSuccessPoint` create the objectSync after the edge node acknowledges.
2. **Rate-limited initial sync**: Add batching/rate-limiting to `sendAllObjects` in `listener.go` to avoid flooding the channel.

## Environment

- KubeEdge version: v1.23 (affects all versions since the cloudhub refactor in v1.12+)
- Kubernetes version: v1.27+
- Edge nodes: 3+ (more nodes = more likely to reproduce)
- Edgemesh installed (increases resource count significantly)

rizzler13:
**What type of PR is this?**

/kind bug

This PR resolves the `DispatchDownstream` single-goroutine channel bottleneck by:
1. Removing synchronous `Create` and `UpdateStatus` Kubernetes API server HTTP calls on `ObjectSync` / `ClusterObjectSync` inside `enqueueNamespacedResource` and `enqueueNonNamespacedResource` in `cloud/pkg/cloudhub/dispatcher/message_dispatcher.go`.
2. Deferring `ObjectSync` creation to be handled lazily when the edge node sends back an acknowledgment (`saveSuccessPoint` in `node_session.go`).
3. Reducing `DispatchDownstream()` per-message processing latency from ~30ms down to < 1µs, preventing channel backup and node `NotReady` status flips when new edge nodes join or sync initial resources.

Fixes #6876

- All unit tests in `cloud/pkg/cloudhub/dispatcher/...` pass cleanly.

```release-note
cloudhub: eliminated synchronous ObjectSync K8s API calls in DispatchDownstream to prevent channel bottlenecks
```


rizzler13:
hi @fisherxu @kevin-wangzefeng @Iceber @DoisLONG , could you please approve ci workflow when you have a moment ? Thanks!

DoisLONG (commented):
Thanks for the PR.

The direction makes sense to me. Removing the synchronous ObjectSync / ClusterObjectSync API calls from DispatchDownstream should help with the bottleneck here.

But I think this needs a bit more validation before LGTM, because this is changing the reliable-sync flow, not just moving some code around.

A few things:

1. The release-note block looks unfinished. It starts with ```release-note but I do not see the closing ```.

2. The new test covers the enqueue side, which is good, but I would also like to see the ACK side covered. Since the PR says ObjectSync creation is now deferred to `saveSuccessPoint()`, please add or point me to tests showing that ObjectSync / ClusterObjectSync are still created correctly after the edge ACK, including the expected spec and resourceVersion.

3. Please also clarify the failure case before ACK. With the old flow, the sync CR existed before sending. With this change, if the message is enqueued/sent but CloudHub restarts before the edge ACK comes back, do we still have a path that resends or recreates the missing sync state?

After those are clarified, this should be easier to review.

rizzler13:
Thanks for the review @DoisLONG 
addressed all three points:

1. fixed the release-note block 

2. added two integration tests covering the ACK side:
 TestSaveSuccessPoint_CreatesObjectSyncOnACK — namespaced (Pod)
 TestSaveSuccessPoint_CreatesClusterObjectSyncOnACK — cluster-scoped (Node)
 Both verify that zero K8s API calls happen during enqueue, and that the ObjectSync/ClusterObjectSync is created with    the correct spec + resourceVersion only after edge ACK triggers saveSuccessPoint().

3. on the failure case, if CloudHub restarts before the edge ACK, we're safe. The synccontroller reconciliation loop already handles this: it periodically checks for resources missing a sync CR and re-sends them. When the node reconnects, the message flows through the normal path again and saveSuccessPoint() creates the CR on the next successful ACK. This recovery path is unchanged from before, we just moved when the CR gets created (enqueue → ACK), not whether it gets created.

please review whenever you can find time, thanks !

DoisLONG (commented):
Thanks for the update.

The main concerns from my previous review are mostly addressed now.

The release-note block is fixed, and the new tests cover the ACK-side lazy creation path for both namespaced resources and cluster-scoped resources. The production change also stays focused: `DispatchDownstream` no longer blocks on synchronous ObjectSync / ClusterObjectSync create/update calls, and the sync CR creation is deferred to the ACK / `saveSuccessPoint()` path.

The explanation for the CloudHub restart-before-ACK case also makes sense to me: if the ACK is lost before the sync CR is created, the existing synccontroller reconciliation should resend the resource and the next successful ACK can create the sync CR.

One small test cleanup before LGTM:

The new ACK-side tests currently use `time.Sleep(2 * time.Second)` to wait for ACK and `saveSuccessPoint()`. Could you please replace the fixed sleep with `require.Eventually` / `assert.Eventually` and check until the expected ObjectSync / ClusterObjectSync appears in the reactor?

That would make the tests faster and less timing-sensitive in CI. Please also stop the keepalive ticker in the helper goroutine before returning.

rizzler13:
Hi @DoisLONG, thank you for reviewing !

I've updated the ACK-side lazy creation unit tests (`TestSaveSuccessPoint_CreatesObjectSyncOnACK` and `TestSaveSuccessPoint_CreatesClusterObjectSyncOnACK`):

1. Tests now poll the reactor every 10ms (up to 2s) and finish immediately once the expected `ObjectSync` / `ClusterObjectSync` CR is created.
2. Added `defer ticker.Stop()` in the keepalive helper goroutine.
3. Deferred `stopCh` closure, `nodeSession.Terminating()`, and `wg.Wait()` to guarantee background goroutines and tickers clean up safely under all execution paths.

All unit tests in `cloudhub/dispatcher` pass cleanly. Ready for another look!

DoisLONG (commented):
Thanks for the update.

rizzler13:
Done, 
squashed into a single commit. 
Thanks for the review, @DoisLONG! Let me know if there's anything else, otherwise this should be ready to approve workflow.

DoisLONG:
Thanks for the update.

/lgtm

rizzler13:
thanks @DoisLONG , the CI checks hasn't been triggered on this one yet, could you approve the workflows ?

rizzler13:
Hi @DoisLONG,
addressed the gofmt lint error in the test file. (the CRI-O failure seems to be the existing issue on master from #7262/#7263)
could you please reapprove the CI run? Thanks!

rizzler13:
Hi @DoisLONG 
All other 22 checks passed (Docker, CRI-O, and WebSocket/QUIC). The single isulad failure was just a runner timeout,
could you please re run the failed isulad job? 
Thanks!

PR #7218, opened 2026-08-10
edge/metamanager: fix serviceaccount token local cache key mismatch

### LINKED ISSUE #5958: 当网络断开时，重启边缘设备，设备无网，无法拉起本地服务
**What happened**:

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

pujitha24:
**What type of PR is this?**

/kind bug

`edge/pkg/metamanager/client.KeyFunc` builds the local sqlite cache key for a
service account `TokenRequest` from `name`, `namespace`, `Spec.Audiences`,
`Spec.ExpirationSeconds`, and `Spec.BoundObjectRef`. This same function is
used both when edgecore persists a token it received from the apiserver
(`edge/pkg/metamanager/process.go:parseResource`, keyed from the apiserver's
*response*, whose `Spec.Audiences` is always resolved/populated) and when it
later looks the token up locally (`getTokenLocally`, keyed from the
kubelet-issued *request*, whose `Spec.Audiences` is `nil` unless the pod's
projected-volume source explicitly sets an `audience` — see
`vendor/k8s.io/kubernetes/pkg/volume/projected/projected.go`, which is the
common/default case).

Because the two keys differ, `getTokenLocally` always misses for the common
case (logged as `query meta ... length error`), and `GetServiceAccountToken`
always falls back to a remote fetch. When the edge node has no network (e.g.
edgecore restarts while offline), that remote fetch fails and any pod that
mounts a projected service account token can never start.

Fixes #

- The fix drops `Spec.Audiences` from `KeyFunc`, so the key is now
  `name/namespace/ExpirationSeconds/BoundObjectRef`. This makes the store-time
  and query-time keys identical for the common case, so the local cache
  actually hits and edgecore no longer needs the network to serve an
  already-fetched token.
- Trade-off: a pod that mounts two projected service-account-token volumes
  for the *same* service account and expiration but with two different
  explicit `audience` values would now collide on one cache entry (whichever
  was stored last local-hits for both). This is a narrow, uncommon
  configuration; today it is already broken (permanent cache miss) for the
  much more common case of no explicit per-volume audience, which this PR
  fixes.
- Validation: added `TestKeyFuncRequestResponseRoundTrip` in
  `edge/pkg/metamanager/client/serviceaccount_test.go`, which reproduces the
  exact key mismatch from the issue's log line (same `BoundObjectRef`, one
  `TokenRequest` with nil `Audiences` simulating the request, one with a
  populated `Audiences` simulating the apiserver response) and asserts
  `KeyFunc` returns the same key for both. Confirmed the test fails on the
  pre-fix code and passes on the post-fix code:
  `go test ./edge/pkg/metamanager/client/... -run TestKeyFuncRequestResponseRoundTrip -v -count=1`
  Also ran `go build ./...` (passes) and `go vet ./edge/pkg/metamanager/client/...`
  (clean). A broader `make test WHAT=edge` run has a handful of unrelated,
  pre-existing failures/panics on unmodified master in other packages
  (`edge/pkg/metamanager/client.TestPersistentVolumeClaimsGet`,
  `edge/pkg/metamanager/metaserver/handlerfactory.TestConfirmUpgrade`,
  `edge/pkg/taskmanager` tests) that were confirmed to reproduce without this
  change; this PR does not touch those code paths. No live edge/cloud cluster
  was available to reproduce the full offline-restart scenario end-to-end;
  the unit test above proves the underlying key-mismatch defect directly.

```release-note
Fixed a bug where edgecore's local service account token cache never hit due to a key mismatch between the stored token and the lookup, forcing every token fetch to go through the cloud and breaking pods that mount a projected service account token when the edge node restarts without network connectivity.
```

Fixes #5958


pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #7226, opened 2026-08-11
fix: parse edited device YAML correctly

### LINKED ISSUE #7196: keadm ctl edit device fails with invalid character when parsing edited YAML
## What happened

`keadm ctl edit device` fails when parsing the edited Device YAML.

The `edit device` command prints the Device object as YAML using `yaml.JSONToYAML()`. However, after the user edits and saves the YAML, the command attempts to parse the edited content using `json.Unmarshal()`:

```go
jsonEdited := cmdutil.StripComments(edited)

var editedDevice v1beta1.Device
err = json.Unmarshal(jsonEdited, &editedDevice)
```

Since the edited content is YAML rather than JSON, `json.Unmarshal()` fails with an error similar to:

```text
invalid character 'a' looking for beginning of value
```

This prevents `keadm ctl edit device` from successfully applying an edited Device object.

The edited Device YAML should be successfully parsed back into a `v1beta1.Device` object and passed to `UpdateDevice()`.

For example, after editing:

```yaml
apiVersion: devices.kubeedge.io/v1beta1
kind: Device
metadata:
  name: device1
spec:
  nodeName: edge-node
```

the command should deserialize the YAML successfully and update the Device.

## How to reproduce it

1. Run the `keadm ctl edit device` command for an existing Device.
2. The command opens the Device definition in YAML format.
3. Make a valid change to the YAML and save it.
4. The command attempts to deserialize the edited YAML.
5. The operation fails because `json.Unmarshal()` is used on YAML data.

The same behavior can be reproduced with the following minimal test:

```go
originalJSON := []byte(`{
    "apiVersion": "devices.kubeedge.io/v1beta1",
    "kind": "Device",
    "metadata": {
        "name": "device1"
    },
    "spec": {
        "nodeName": "edge-node"
    }
}`)

yamlData, err := yaml.JSONToYAML(originalJSON)
if err != nil {
    panic(err)
}

var device map[string]interface{}
err = json.Unmarshal(yamlData, &device)
```

This produces:

```text
invalid character 'a' looking for beginning of value
```

The repository test `TestPrintObjOutputCanBeUnmarshaledAsJSON` also reproduces the same failure.

## Proposed fix

Use YAML-aware deserialization for the edited content instead of `json.Unmarshal()`.

The parsing step should accept the YAML generated by `PrintObj()` and deserialize it into `v1beta1.Device` before calling `UpdateDevice()`.

A regression test should also verify that the YAML produced by `PrintObj()` can be successfully parsed by the same deserialization path used by `edit()`.

## Additional context

The issue is located in:

`keadm/cmd/keadm/app/cmd/ctl/edit/device.go`

Relevant functions:

* `editPrinterOptions.PrintObj()`
* `DeviceEditOptions.edit()`

The problem is specifically the mismatch between the YAML output produced by `PrintObj()` and the JSON-only parser used afterward.

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

Moemenmohamed24:
What type of PR is this?

/kind bug

Fixes an issue where the edited Device YAML output was parsed using json.Unmarshal, even though the editor provides YAML content.

This change parses the edited Device manifest using yaml.Unmarshal, allowing edited Device resources to be correctly converted back into the Kubernetes Device object.

A unit test is also added to verify that the YAML output generated by PrintObj can be successfully unmarshaled back into a Device object.

Which issue(s) this PR fixes:

Fixes https://github.com/kubeedge/kubeedge/issues/7196

The change is limited to the Device edit flow and its corresponding unit test.

Yes. keadm ctl edit device can now correctly parse edited Device YAML and apply the changes.

`keadm ctl edit device` now correctly parses edited Device YAML.


Moemenmohamed24:
Hi @FengyunPan2  , @zhu733756 , could you please approve the FOSSA workflow run for my PR? It requires approval before it can run. Thanks!

PR #5863, opened 2024-09-18
OTA(Over-The-Air) Upgrades For Edge Node

HT0403:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
In order to make the edge node more convenient and rapid upgrade, we introduce a remote upgrade scheme OTA (Over-The-Air) into KubeEdge. In the main process of OTA(i.e. make the bundle, download the bundle, verify the bundle and firmware upgrade), we have realized most steps. Our release will generate a new image version called installation-package, then we use the NodeUpgradeJob CRD to obtain the installation tool keadm in the image and run the command to upgrade the edge node. During this process, if the hacker masquerades the image in the edge node, this will result in the untrusted binary keadm. We need to verify the digest of the image before the keadm executes the upgrade, which is the third step of OTA to verify the bundle. And in some business scenarios (Internet of vehicles, Internet of Things), we also need to provide an option to make the node wait for confirmation from a person with permission before upgrading the edge node.
**Which issue(s) this PR fixes**:

Fixes #5589 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #7227, opened 2026-08-12
cloud/admission: add strategy and allowDowngrade validation for NodeUpgradeJob

### LINKED ISSUE #6278: Add CRD validation + defaulting webhook for OTAUpgrade API
Provide a validation & defaulting admission webhook (or CRD structural schema) for the `OTAUpgrade` API introduced in PR #5863.  
Key validations:  
* `spec.image` must be a valid OCI‑compatible reference.  
* `spec.strategy.type` must be one of `AtOnce`, `Rolling`, `Canary`.  
* Reject updates that downgrade version unintentionally (e.g., 1.17.0 → 1.16.3) unless `allowDowngrade=true`.  
Also supply sane defaults (e.g., `strategy.type=Rolling`, `maxUnavailable=1`).
**Why is this needed**:
Without server‑side validation, an invalid `OTAUpgrade` could brick thousands of edge nodes in production.  
A webhook prevents typos at submit‑time and makes CI easier because we can rely on API guarantees instead of ad‑hoc checks in controllers.

om7057:
**What type of PR is this?**

/kind feature
/kind api-change

Adds admission validation and defaulting for a rollout strategy on the NodeUpgradeJob API, and adds protection against unintentional EdgeCore downgrades, building on the existing NodeUpgradeJob admission webhook introduced in #5863.

This PR is scoped to the `operations.kubeedge.io/v1alpha1` API only, matching the scope of the existing admission webhook (which is only registered for v1alpha1). It does not touch the v1alpha2 Go type or served schema.

Changes:
- Adds `spec.strategy` (`type`, `maxUnavailable`) and `spec.allowDowngrade` to `NodeUpgradeJobSpec` in the v1alpha1 API, with a `+kubebuilder:validation:Enum=AtOnce;Rolling;Canary` marker on `strategy.type` and a `+kubebuilder:validation:Minimum=1` marker on `strategy.maxUnavailable`.
- Validating webhook rejects a `strategy.type` outside that enum, and rejects `strategy.maxUnavailable` below 1 (matching the CRD marker).
- Validating webhook rejects a version update that downgrades EdgeCore (e.g. 1.17.0 to 1.16.3) unless `allowDowngrade: true` is set on the job.
- `version` and `allowDowngrade` are the only spec fields an update is allowed to change once a NodeUpgradeJob is created (previously the whole spec was immutable after create). `version` may only be changed while `status.State` is still empty, i.e. before the job has started dispatching any per-node action, so a stalled/not-yet-started job can be retried with a different version without risking some nodes ending up on the old version and others on a newly-set one mid-rollout. `allowDowngrade` can be changed at any time since it doesn't by itself affect already-dispatched actions.
- Mutating webhook defaults `strategy.type` to `Rolling` and `strategy.maxUnavailable` to `1` when not specified.
- `strategy` is validated and defaulted by this PR, but **the upgrade execution path does not yet change behavior based on `strategy.type`** (AtOnce/Rolling/Canary rollout semantics are not implemented here). This is documented directly on the `Strategy` field. Wiring actual rollout behavior into the task execution controller is left as follow-up work, since it touches a different, larger part of the codebase (`cloud/pkg/taskmanager`) and deserves its own review.
- Regenerates the v1alpha1 deepcopy code and the v1alpha1 schema block of the multi-version `operations_v1alpha2_nodeupgradejob.yaml` CRD manifest for the new fields (that file also contains an unrelated, unmodified v1alpha2 schema block, hence the filename), and updates the sample manifest under `build/crd-samples`.

Fixes #6278

- OCI image reference validation continues to use the existing `validation.ValidateImageRepo` regex rather than pulling in a new dependency such as go-containerregistry, since the existing check already covers registry/repository/tag/digest forms.

```release-note
Add `spec.strategy` and `spec.allowDowngrade` fields to the NodeUpgradeJob v1alpha1 API. The admission webhook validates `spec.strategy.type` (AtOnce, Rolling, Canary) and `spec.strategy.maxUnavailable` (must be >= 1), and defaults `spec.strategy.type` to Rolling and `spec.strategy.maxUnavailable` to 1. On update, any `spec.version` change is rejected once the job has started executing, regardless of `spec.allowDowngrade`; before execution starts, a `spec.version` change that downgrades EdgeCore additionally requires `spec.allowDowngrade` to be set to true. This PR only adds API validation/defaulting; it does not change NodeUpgradeJob rollout execution behavior.
```


om7057:
Hi @WillardHu @DoisLONG @kevin-wangzefeng @Shelley-BaoYue, could any of you take a look at this PR?

DoisLONG (commented):
Thanks for the PR.

The general direction makes sense to me: preventing accidental EdgeCore downgrades is useful, and adding an explicit rollout strategy to NodeUpgradeJob could be valuable. But this PR is an API change, so I think we need to be more careful about version consistency and actual behavior.

A few concerns:

1. v1alpha2 consistency

The PR updates the v1alpha2 CRD schema with `spec.strategy` and `spec.allowDowngrade`, but the PR description says the v1alpha2 Go type does not carry these fields yet and the webhook is only scoped to v1alpha1.

I do not think we should expose new fields in the served v1alpha2 schema without also aligning the v1alpha2 Go type and admission webhook behavior. Otherwise v1alpha2 users may see fields that are accepted by the schema but not consistently defaulted, validated, or preserved through typed clients.

Could we either:
- include the v1alpha2 API type and webhook support in this PR; or
- keep this PR clearly scoped to v1alpha1 and avoid changing the v1alpha2 served schema until the follow-up is ready?

2. Strategy behavior

This PR defaults and validates `spec.strategy`, but I do not see controller-side logic that actually consumes `AtOnce`, `Rolling`, or `Canary` during NodeUpgradeJob execution.

If the rollout behavior is not implemented in this PR, users may see `strategy.type: Rolling` defaulted in the object but the upgrade execution behavior may not actually change. That would be confusing for a new API field.

Please either wire the strategy into the upgrade execution path, or make the PR scope / release note clear that this only adds the API and admission groundwork, without changing rollout behavior yet.

3. maxUnavailable validation

`strategy.maxUnavailable` should probably be validated. Right now it is defaulted to `1` when missing, but a user can still set `0` or a negative value. For a rolling strategy, that does not look like a valid value.

Could you add validation, and ideally the CRD marker as well, to require `maxUnavailable >= 1`? It would also help to clarify whether `maxUnavailable` is only meaningful for `Rolling`, and what happens when it is set with `AtOnce` or `Canary`.

After the API versioning and validation points are clarified, I can take another look.

om7057:
Thanks for the detailed review, @DoisLONG. 

Pushed an update addressing all three points:

**v1alpha2 consistency**: this was a miscommunication on my part in the original PR description, not an actual schema change. `build/crds/operations/operations_v1alpha2_nodeupgradejob.yaml` is a single file containing two version blocks (`v1alpha1` and `v1alpha2`) because that's how this CRD is already set up on master. My diff only ever touched the `v1alpha1` block (lines 17-260ish); the `v1alpha2` block (starting further down) has no `strategy` or `allowDowngrade` fields and was never modified, since I only changed the v1alpha1 Go type. I've reworded the commit message to be explicit about that instead of the confusing "regenerates the v1alpha2 CRD manifest" phrasing that caused the misread. So this PR is already scoped to v1alpha1 only, both in Go types and served schema, matching your second option.

**Strategy execution behavior**: added a doc comment directly on the `Strategy` field clarifying that it's currently validated and defaulted by the admission webhook only, and that the upgrade execution path does not yet change behavior based on `Strategy.Type` (that wiring is real, non-trivial work in the task execution controller that I think deserves its own PR/design discussion rather than folding it in here). Went with your second suggested option rather than attempting the execution wiring in this PR.

**maxUnavailable validation**: added both a `+kubebuilder:validation:Minimum=1` marker (so the API server itself rejects 0/negative values structurally) and a matching check in the validating webhook with a clear error message. Also documented that `maxUnavailable` only applies to the `Rolling` strategy and has no effect for `AtOnce`/`Canary`.

Also fixed the DCO.

DoisLONG (commented):
Thanks for the update.

The three points from my previous review are mostly addressed now.

- `maxUnavailable` now has both CRD-side `Minimum=1` validation and admission-side validation.
- The strategy field now documents that it is only defaulted/validated for now, and the execution path does not yet change based on `Strategy.Type`.
- The v1alpha2 concern sounds like it was mostly a wording issue around the generated CRD file.

A couple of things still need cleanup:

1. The PR description is still confusing

The PR body still says the v1alpha2 CRD schema picks up the new fields automatically and that the v1alpha2 Go type does not carry them yet. That was exactly the confusing part from the previous review.

Since this PR is intended to be scoped to v1alpha1 only, please update the PR description to make that clear. It should not say that v1alpha2 schema gets the new fields if that is not actually the intended behavior.

It would also be good to update the release note to clarify that this PR only adds API/admission defaulting and validation, and does not change the actual rollout execution behavior yet.

2. Version update semantics still need clarification

This PR changes the previous immutability rule and allows `spec.version` and `spec.allowDowngrade` to be updated after the NodeUpgradeJob is created.

I think this needs a bit more clarification. If a NodeUpgradeJob has already started, changing `spec.version` may result in some nodes using the old version and later nodes using the new version. If the intent is to support retrying a stalled job, we should be explicit about when version changes are allowed.

Could you clarify the expected behavior here? For example, is changing `spec.version` allowed only before execution starts, or can it also happen while the job is already in progress? If there is status/phase information available, we may need validation and tests for that case.

After the PR description/release note are corrected and the version update semantics are clarified, I can take another look.

om7057:
Thanks @DoisLONG, pushed an update for both points:

**PR description/release note**: rewrote the PR description to remove the confusing "v1alpha2 CRD schema picks up the new fields automatically" line and state plainly that this PR is scoped to v1alpha1 only (Go type, admission webhook, and the v1alpha1 block of the CRD manifest). Also reworded the release note to say explicitly that this only adds API validation/defaulting and does not change rollout execution behavior yet.

**Version update semantics**: you're right that this needed tightening. Added a check in the validating webhook: a version change is now only allowed while `status.State` is still empty, i.e. before the job has started dispatching any per-node action (checking/backing-up/upgrading/rolling-back). Once execution has started, changing `spec.version` is rejected with `version cannot be changed once the upgrade has started executing (current state: <state>)`. `allowDowngrade` on its own can still be changed at any time since it doesn't retroactively affect actions already dispatched. Documented the same rule on the `Version` field's godoc, and added test cases for both the rejected mid-execution change and the still-allowed allowDowngrade-only change.

I traced this against `cloud/pkg/taskmanager/v1alpha1/nodeupgradecontroller`, which is the controller that actually processes v1alpha1 NodeUpgradeJobs and drives `status.State` through Checking -> BackingUp -> Upgrading -> RollingBack via the FSM in `pkg/util/fsm`, so gating on `status.State == ""` should reliably mean "no per-node action has been dispatched yet."

DoisLONG (commented):
Thanks for the update.

This looks much better now. The PR description is now clear that this PR is scoped to the v1alpha1 API only, and the release note also makes it clear that this PR only adds API/admission validation and defaulting, without changing the actual rollout execution behavior.

The version update semantics also look reasonable to me now. Allowing `spec.version` to change only while `status.State` is still empty avoids the risky case where a job is already dispatching per-node actions and some nodes could end up using the old version while later nodes use the new version. The added tests for rejecting version changes after execution starts and still allowing an `allowDowngrade`-only change are useful.

Two final cleanups before LGTM:

2. The release note wording is still a little ambiguous around `allowDowngrade`. This part:

`rejects a version downgrade or a version change after the job has started executing, unless spec.allowDowngrade is set to true`

could be read as if `allowDowngrade=true` also allows changing the version after the job has started. Could you please split that into two clearer statements? For example: downgrades require `spec.allowDowngrade=true`, and any `spec.version` change is rejected once the job has started executing.

After those two cleanups, this should be good from my side.

om7057:
Thanks @DoisLONG, both cleanups done:


> On update, any \`spec.version\` change is rejected once the job has started executing, regardless of \`spec.allowDowngrade\`; before execution starts, a \`spec.version\` change that downgrades EdgeCore additionally requires \`spec.allowDowngrade\` to be set to true.

That splits it into the two separate statements you suggested: the execution-state gate applies unconditionally, and \`allowDowngrade\` is a separate, narrower requirement that only matters for downgrades attempted before execution starts.

Would appreciate a review once you get time!

DoisLONG (commented):
Thanks for the update.

/lgtm

om7057:
> Thanks for the update.
> 
> /lgtm

Sure, would appreciate an approval @DoisLONG!

PR #7230, opened 2026-08-12
admissioncontroller: follow certificate rotation in the webhook config

### LINKED ISSUE #7220: admissioncontroller: webhook self-registration and CA handling block consolidation onto the controller-runtime Manager
**What happened**:

Admission registers its own ValidatingWebhookConfiguration and
MutatingWebhookConfiguration at process start: Run reads the CA from
--ca-cert-file (admission.go:93) and calls registerWebhooks (:99, defined at
:157), which builds the configuration objects in Go and writes them to the API
server. The controller manager separately runs controllerruntime.NewManager with
only Scheme and HealthProbeBindAddress and no webhook server
(controllermanager.go:46-49).

That split leaves three obstacles to running both on one Manager:

1. RBAC scope. build/admission/clusterrole.yaml grants create/update on
   mutatingwebhookconfigurations and validatingwebhookconfigurations, plus
   create on certificatesigningrequests and create/update on
   certificatesigningrequests/approval. The controller manager's ClusterRole has
   none of these, so consolidating as-is gives it cluster-wide write access to
   webhook configurations and the ability to self-approve CSRs.

2. The webhook server is not leader-gated. controller-runtime's
   DefaultServer.NeedLeaderElection() returns false, so it runs on every replica
   while only reconcilers are gated. registerValidateWebhook and
   registerMutatingWebhook (common.go:19-63) do Get, overwrite existing.Webhooks,
   then Update, with no resourceVersion conflict handling. Admission defaults to
   one replica (values.yaml replicaCount: 1), so this is latent today.

3. The CA bundle is read once at startup and written into the webhook
   configuration. controller-runtime reloads the serving certificate from its
   cert directory, but nothing updates a CABundle already written to the cluster,
   so rotating the mounted secret leaves a stale CA behind.

Webhook configurations applied declaratively with an explicit CA injection
mechanism, so a consolidated controller manager needs neither
admissionregistration write permission nor CSR approval rights.

    # permissions admission has
    grep -n "admissionregistration\|certificatesigningrequests" \
      build/admission/clusterrole.yaml

    # same query against the controller manager: no matches
    grep -n "admissionregistration\|certificatesigningrequests" \
      manifests/charts/cloudcore/templates/rbac_controllermanager.yaml

    # webhook server is not leader-gated: returns false
    grep -A2 "func (\*DefaultServer) NeedLeaderElection" \
      vendor/sigs.k8s.io/controller-runtime/pkg/webhook/server.go

If self-registration has to stay for upgrade compatibility, it needs leader
gating, conflict retry, and a CABundle refresh path.

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

Priyanshu6968:
**What type of PR is this?**
/kind bug

Hardens admission's webhook self-registration and certificate handling, which
today cannot survive a certificate rotation without a restart:

- `registerValidateWebhook` / `registerMutatingWebhook` did get -> overwrite
  `Webhooks` -> update with no resourceVersion conflict handling, and a create
  losing the race to another writer returned `AlreadyExists` straight to `Run`.
  Both now register under `retry.OnError`, treating conflict and already-exists
  as retriable so the registration re-reads and converges.
- The CA bundle was read once at startup and copied into the webhook
  configurations. It is now re-read periodically and published again when it
  changes, so a rotated CA reaches the API server before the previous one is
  gone.
- The serving certificate was loaded once by `configTLS`. Refreshing only the
  published `CABundle` would break the webhook after a CA rotation, since the
  process would keep presenting a leaf that no longer chains to it, so the
  certificate is now served through a reloader that picks up the rotated key
  pair. It keeps the last pair that loaded successfully when the files are
  unreadable or when a rotation is caught mid write.

Fixes #7220 

```release-note
Fixed the admission webhook not picking up a rotated certificate: the CA bundle
published in the webhook configurations and the certificate served by the webhook
are now reloaded instead of staying at the values read when the process started.
```

### DIFF
--- a/cloud/pkg/admissioncontroller/admission.go
+++ b/cloud/pkg/admissioncontroller/admission.go
@@ -1,12 +1,14 @@
 package admissioncontroller
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
 	"net/http"
 	"os"
+	"time"
 
 	admissionv1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
@@ -15,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -41,6 +44,11 @@ const (
 	MutatingNodeUpgradeWebhookName = "mutatingnodeupgradejob.kubeedge.io"
 
 	AutonomyLabel = "app-offline.kubeedge.io=autonomy"
+
+	// caBundleRefreshPeriod is how often the CA certificate is re-read from disk to
+	// check whether the CA bundle published in the webhook configurations is still
+	// the one the webhook is served with.
+	caBundleRefreshPeriod = time.Minute
 )
 
 var scheme = runtime.NewScheme()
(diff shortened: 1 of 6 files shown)


Priyanshu6968:
@DoisLONG sorry for randomly spamming these review requests, these PRs have been open from some time but i recently discovered that you are mentor of the LFX Project that i applied for, so i thought of getting your review on my work.
can you please look at this one.

PR #7245, opened 2026-08-14
fix: preserve servicebus timeout options

### LINKED ISSUE #7244: [Bug] Servicebus timeout defaults overwrite configured values
**What happened**:

`GetURLClient()` uses `fallthrough` while defaulting timeout options. When `HandshakeTimeout` is zero, it overwrites a caller-provided `ResponseHeaderTimeout`. When only the response timeout is zero, it uses the handshake default instead of the declared response timeout default.

Each timeout option should be defaulted independently. A configured response header timeout must remain unchanged, and an omitted response header timeout must use `DefaultURLClientOption.ResponseHeaderTimeout`.

1. Call `GetURLClient(&URLClientOption{ResponseHeaderTimeout: 5 * time.Second})`.
2. Inspect the returned transport.
3. `ResponseHeaderTimeout` becomes 30 seconds instead of 5 seconds because `HandshakeTimeout` defaults first and falls through.
4. Call `GetURLClient(&URLClientOption{HandshakeTimeout: 5 * time.Second})`.
5. The response header timeout becomes 30 seconds instead of the declared 60 second default.

The correction only replaces the fallthrough with two independent default checks. A focused table test covers both cases.

**Environment**:
- Kubernetes version (use `kubectl version`): Not required
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): master branch

  - Hardware configuration (e.g. `lscpu`): Not required
  - OS (e.g. `cat /etc/os-release`): macOS
  - Kernel (e.g. `uname -a`): Not required
  - Go version (e.g. `go version`): Go toolchain used by repository tests
  - Others: `go test ./edge/pkg/servicebus/util` passes

  </details>

  - edgecore version (e.g. `edgecore --version`): Not required
  - Hardware configuration (e.g. `lscpu`): Not required
  - OS (e.g. `cat /etc/os-release`): Not required
  - Kernel (e.g. `uname -a`): Not required
  - Go version (e.g. `go version`): Go toolchain used by repository tests
  - Others: Not required

  </details>

blackdragoon26:
**What type of PR is this?**

/kind bug

Corrects inconsistent timeout defaulting in the exported `GetURLClient()` helper. Handshake and response-header timeouts are now defaulted independently, preserving explicitly configured values and using the declared defaults when omitted.

`GetURLClient()` currently has no production caller within the repository, so this fixes a latent helper bug and protects future callers without changing the active ServiceBus runtime path.

Fixes #7244

Added a focused table test covering a configured response timeout and an independently defaulted response timeout. `go test ./edge/pkg/servicebus/util` passes on macOS.

The current ServiceBus runtime assigns its HTTP client directly in `Start()`. This PR intentionally does not wire `GetURLClient()` into that path because doing so would broaden the scope and change existing timeout behavior.

No.

```release-note
NONE
```


blackdragoon26:
Kindly review this PR
@chendave  @Shelley-BaoYue

blackdragoon26:
> Kindly review this PR @chendave @Shelley-BaoYue

Re-pinging for the review, thanks

omlahore:
I tried both new cases against master and got 30s out of each, the one configured to 5s and the one that should have been 60s. They pass on your branch.

I couldn't find anything calling GetURLClient though, apart from this package's own tests. servicebus.go just does uc.Client = htc in Start(), so I guess that option path isn't really reachable at the moment, unless I missed a caller.

blackdragoon26:
> I tried both new cases against master and got 30s out of each, the one configured to 5s and the one that should have been 60s. They pass on your branch.
> 
> I couldn't find anything calling GetURLClient though, apart from this package's own tests. servicebus.go just does uc.Client = htc in Start(), so I guess that option path isn't really reachable at the moment, unless I missed a caller.

Thanks for verifying. You’re right, GetURLClient() currently has no production caller, so this fixes a latent bug in the exported helper rather than the active ServiceBus path.

PR #7246, opened 2026-08-14
hack: replace hardcoded docker cp with runtime-agnostic helper

### LINKED ISSUE #7243: [Bug] local-up-kubeedge.sh hardcodes docker cp even when containerd is the runtime
The `hack/local-up-kubeedge.sh` script hardcodes the use of `docker cp` for certificate extraction (lines 344-345). Even when containerd is configured as the runtime via the `CONTAINER_RUNTIME` environment variable, the script fails if Docker is not installed on the system.

**Why is this needed**:
The script should use the appropriate container CLI (e.g., `crictl` or `nerdctl`) based on the configured runtime, rather than hardcoding `docker`.

Prateek-og:
The generate_streamserver_cert function in local-up-kubeedge.sh hardcodes docker cp to extract certificates from the kind control-plane container. This fails on systems using Podman as the kind provider without Docker installed. Add a kind_container_cp helper that auto-detects Docker or Podman and uses the appropriate CLI.

Fixes #7243

**What type of PR is this?**

Fixes #

```release-note

```

### DIFF
--- a/hack/local-up-kubeedge.sh
+++ b/hack/local-up-kubeedge.sh
@@ -321,6 +321,21 @@ function healthcheck {
   fi
 }
 
+# Copy files from a kind node container using whichever container CLI is available.
+# kind nodes are Docker or Podman containers; detect the appropriate CLI.
+function kind_container_cp {
+  local src="$1"
+  local dst="$2"
+  if command -v docker &>/dev/null; then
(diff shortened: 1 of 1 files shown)


Prateek-og:
@fisherxu...PTAL

PR #7255, opened 2026-08-18
admissioncontroller: reject AdmissionReview with no request

### LINKED ISSUE #7254: admissioncontroller: all webhook handlers panic on an AdmissionReview with no request field
**What happened**:

`serve` decodes the request body into an `admissionv1.AdmissionReview`
(common.go:92). A body that is valid JSON and carries `apiVersion` and `kind`
but no `request` field decodes without error and leaves `Request` nil. `serve`
then calls the hook (:96) and afterwards reads
`requestedAdmissionReview.Request.UID` (:98).

Every hook dereferences `review.Request` as its first statement:
admit_device.go:19, admit_devicemodel.go:19, admit_rule.go:24,
admit_ruleendpoint.go:16, admit_nodeupgradejob.go:43 and
mutate_offlinemigration.go:24. All of them panic on a nil `Request`, and :98
panics as well for any hook that returns without touching `Request`.

net/http recovers the panic per connection, so the process stays up, but the
handler writes no response and the connection is dropped. The API server sees
the webhook call fail. The rules, ruleendpoints and nodeupgradejobs webhooks
are registered with `FailurePolicy: Fail` (admission.go:247, :275, :303), so
while this happens those resources cannot be created, updated or deleted.

The webhook serves TLS without requiring a client certificate, so any workload
that can reach the admission Service can send this body; it does not have to
come from the API server.

`serve` rejects an AdmissionReview with no `request` and returns an
AdmissionResponse carrying the error, the same way it already does when the
body fails to decode (common.go:93-94).

Add to `cloud/pkg/admissioncontroller/common_test.go` and run
`go test -gcflags "all=-N -l" ./cloud/pkg/admissioncontroller/...`:

    func TestServeNilRequest(t *testing.T) {
        raw := `{"apiVersion":"admission.k8s.io/v1","kind":"AdmissionReview"}`
        serve(httpfake.NewResponseWriter(), &http.Request{
            Header: map[string][]string{"Content-Type": {"application/json"}},
            Body:   io.NopCloser(bytes.NewReader([]byte(raw))),
        }, admitRule)
    }

    panic: runtime error: invalid memory address or nil pointer dereference

The same panic reproduces with `admitDevice`, `admitDeviceModel`,
`admitRuleEndpoint`, `admitNodeUpgradeJob` and `mutateOfflineMigration`.

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

Priyanshu6968:
**What type of PR is this?**
/kind bug

`serve` decoded an `AdmissionReview` body that carried `apiVersion` and `kind` but no `request` without error, leaving `Request` nil. All six validating/mutating hooks dereference `review.Request` as their first statement, and `serve` itself then reads `requestedAdmissionReview.Request.UID`, so any such body panicked the handler. `net/http` recovers per connection so cloudcore stays up, but no response is written and the connection is dropped; the rules, ruleendpoints and nodeupgradejobs webhooks use `FailurePolicy: Fail`, so those resources become uncreatable/unupdatable/undeletable while it happens. The webhook does not require a client certificate, so the body need not come from the API server.
`serve` is the single entry point shared by every hook, so the review is rejected there and an `AdmissionResponse` carrying the error is returned, the same way the existing decode-failure branch does. The hooks are unchanged.

Fixes #7254 

```release-note
Fixed a panic in the cloudcore admission controller webhooks when an AdmissionReview request without a `request` field is received.
```

### DIFF
--- a/cloud/pkg/admissioncontroller/common.go
+++ b/cloud/pkg/admissioncontroller/common.go
@@ -3,6 +3,7 @@ package admissioncontroller
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 
@@ -62,6 +63,9 @@ func registerMutatingWebhook(client admissionregistrationv1client.MutatingWebhoo
 	return nil
 }
 
+// errNilAdmissionRequest is returned when a decoded AdmissionReview carries no request.
+var errNilAdmissionRequest = errors.New("the request field of the AdmissionReview is nil")
+
 // hookFunc is the type we use for all of our validators and mutators
 type hookFunc func(admissionv1.AdmissionReview) *admissionv1.AdmissionResponse
 
@@ -92,6 +96,11 @@ func serve(w http.ResponseWriter, r *http.Request, hook hookFunc) {
 	if _, _, err := deserializer.Decode(body, nil, &requestedAdmissionReview); err != nil {
 		klog.Errorf("decode failed with error: %v", err)
 		responseAdmissionReview.Response = toAdmissionResponse(err)
+	} else if requestedAdmissionReview.Request == nil {
+		// Every hook dereferences the request, so reject the review here
+		// instead of letting the hook panic on a nil pointer.
+		klog.Errorf("invalid admission review: %v", errNilAdmissionRequest)
+		responseAdmissionReview.Response = toAdmissionResponse(errNilAdmissionRequest)
 	} else {
 		responseAdmissionReview.Response = hook(requestedAdmissionReview)
 		// Return the same UID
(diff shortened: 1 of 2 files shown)


Priyanshu6968:
@DoisLONG please have a look if you get time sir. Please note that the failing check is not due to this PR, but due to an upstream issue with the mapper image build.

DoisLONG (commented):
Thanks for the PR.

The fix is small and in the right place. `serve()` is the common entry point for all admission webhook handlers, so rejecting a decoded AdmissionReview with a nil `Request` there prevents every individual hook from needing to defend against the same nil pointer case.

Without this check, a body with `apiVersion` and `kind` but no `request` can be decoded successfully, then later panic when the hook or `serve()` itself dereferences `review.Request`. Returning a failed AdmissionResponse is consistent with the existing decode-error path and avoids dropping the connection.

The added test covers the missing-request AdmissionReview case and verifies that the response is denied with the expected error message.

/lgtm

PR #7257, opened 2026-08-18
controllermanager: let the manager own the node cache

### LINKED ISSUE #7256: controllermanager: newAndStartCache races on its named return and drops the cache Start error
**What happened**:

`newAndStartCache` (controllermanager.go:76-101) builds a second `cache.Cache`
for `corev1.Node`, starts it in a bare goroutine that assigns to the named
return value `err` (:88-90), then reads that same `err` on the calling
goroutine after `WaitForCacheSync` (:91-95).

Three things follow:

1. The write at :89 and the read at :92 are unsynchronized. This is a data
   race on `err`.

2. The check is dead in practice. `Start(ctx)` blocks until `ctx` is done, so
   `err` is still nil when :92 reads it, and the goroutine's write lands long
   after `newAndStartCache` has returned. If the cache stops with an error
   while cloudcore is running, nothing observes it.

3. The cache is never handed to the manager. `NewControllerManager` passes it
   straight to `setupControllers` (:65-72) and `mgr.Add(` never appears in the
   file, so the cache sits outside the manager's runnable lifecycle: it is not
   stopped with the manager, not covered by its readiness gating, and not
   leader-election aware if leader election is ever enabled.

The Node cache is owned by the manager, and a cache that fails to start or
stops early is surfaced instead of being written into a variable nobody reads.

    # the goroutine writes the named return value, the caller reads it
    sed -n '76,101p' cloud/pkg/controllermanager/controllermanager.go

    # the cache is never registered as a manager runnable: no matches
    grep -n "mgr\.Add(" cloud/pkg/controllermanager/controllermanager.go

The separate cache also looks redundant. It caches only `corev1.Node`, and its
only consumers are `verifyNodeByNames` (node_verify.go:74) and
`verifyNodeBySelector` (node_verify.go:96), which call `Get` and `List` at
reconcile time, never during `SetupWithManager`. `mgr.GetCache()` already
provides that and is started and stopped by the manager, so the three nodetask
controllers could take it instead and `newAndStartCache` could be removed.

**Environment**:

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

  - Kernel (e.g. `uname -a`):
  - Others:

  </details>

Priyanshu6968:
**What type of PR is this?**
/kind bug

`newAndStartCache` built a second `cache.Cache` for `corev1.Node`, started it in a bare goroutine that assigned to the named return value `err`, and then read that same `err` on the calling goroutine after `WaitForCacheSync`. Three problems followed: the write and the read were unsynchronized (a data race on `err`); the check was dead in practice, because `Start(ctx)` blocks until the context is done, so `err` was still nil when it was read and a cache that stopped with an error was never observed; and the cache was never handed to the manager, so it sat outside the manager's runnable lifecycle; not stopped with the manager, not covered by its readiness gating, not leader-election aware.
The separate cache was also redundant. It cached only `corev1.Node`, and its only consumers, `verifyNodeByNames` and `verifyNodeBySelector`, `Get` and `List` at reconcile time, never during `SetupWithManager`. `mgr.GetCache()` already provides that and is started, synced and stopped by the manager, so `setupControllers` now takes the cache from the manager and `newAndStartCache` is removed. The nodegroup controller already watches `corev1.Node` through the manager cache, so this also removes a duplicate node watch.

Fixes #7256 

The package had no unit test file. The added one builds a manager with an injected static RESTMapper so the controllers can be set up without an API server, and covers both the successful setup path and the error path of `NewControllerManager`.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/controllermanager/controllermanager.go
+++ b/cloud/pkg/controllermanager/controllermanager.go
@@ -2,11 +2,9 @@ package controllermanager
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -14,8 +12,6 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/cache"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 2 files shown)


Priyanshu6968:
@DoisLONG please have a look if you get time sir. Please note that the failing check is not due to this PR, but due to an upstream issue with the mapper image build.

DoisLONG (commented):
Thanks for the PR.

The issue is clear: `newAndStartCache()` created a separate `cache.Cache`, started it outside the controller-runtime manager lifecycle, and also wrote to the named return `err` from a goroutine while the caller read it later. Removing that standalone cache avoids the data race and lets the manager own cache start, sync, stop, readiness, and lifecycle behavior.

Using `mgr.GetCache()` in `setupControllers()` also makes sense here. The node-related controllers can use the manager-owned cache instead of a duplicate Node-only cache, and the manager already handles cache startup and synchronization before controllers run.

The added tests cover the setup path with an injected RESTMapper and the error path from `NewControllerManager()`.

Main CI is still failing, but it looks unrelated to this PR. The failure is in the CRI-O mapper image build path, and that is already being handled separately by #7263. The other relevant checks are green, with ARM64 cancelled.

I do not see any remaining code blocker from my side.

/lgtm

PR #7260, opened 2026-08-21
policycontroller: send ServiceAccountAccess to every node, not only new ones

mkhon:
/kind bug

A node recorded in status.NodeList that never received the object was unreachable afterwards. syncRules() writes the node list before it sends, and then sends only to addNodes, which is derived from that same list: once a node is recorded, addNodes never contains it again. Anything happening between the two steps loses it permanently - a cloudCore restart, the spec Client.Update() error path that returns Requeue before send2Edge(), a BuildResource() or MessageLayer.Send() failure inside send2Edge() which only logs and continues, or a later deleteObjectSyncs() sweep that drops the node's ObjectSync. The node goes on running pods that authenticate against a ServiceAccount the edge never received, and the only recovery is editing status.NodeList by hand.

Send to every node instead. enqueueAckMessage() drops a message whose resourceVersion is not newer than the one recorded in that node's ObjectSync, so a node already in sync costs nothing on the wire, while a node that missed an earlier send gets an ObjectSync created and the object delivered. Per-node send failures become self-healing on the next reconcile without having to be tracked. InsertOperation becomes UpdateOperation: the edge stores both through handleMessage() -> dao.InsertOrUpdate(), which is INSERT OR REPLACE, so an update creates the row on a node that never had it, and processUpdate() ACKs the policyController source exactly as processInsert() does.

Record the node list only after the sends, in both branches, so a node is recorded only once a send has been attempted for it. This does mean the message carries the resourceVersion from before the status write, so syncController resends once from its lister on a following pass, after which the ObjectSync settles on the current revision.

Reconciles are event-driven plus controller-runtime's 10h resync, so sending to every node is not a hot loop.

TestSyncRules is updated for the two cases that go through the unchanged-spec branch: an added node now receives an update rather than an insert, and the nodes already in status.NodeList receive one too.

```release-note
NONE
```


mkhon:
I will merge to #6673

Issue #7262, opened 2026-08-22
tests/scripts/generate_mapper.sh: CRI-O mapper import tags a non-existent image name


RedZapdos123:
**What happened**:

On latest `master`, the `cri-o` branch in `tests/scripts/generate_mapper.sh` hardcodes a `podman tag` source image name that does not match the image name loaded from `modbus-mapper.tar`.

Current code:

- `docker build -f Dockerfile_nostream -t modbus-e2e-mapper:v1.0.0 .`
- `docker save -o modbus-mapper.tar modbus-e2e-mapper:v1.0.0`
- `sudo podman load -i modbus-mapper.tar && sudo podman tag localhost/v1.0.0:latest docker.io/library/modbus-e2e-mapper:v1.0.0`

In the failing `container runtime e2e test (cri-o)` job on August 22, 2026, `podman load` reports:

- `Loaded image: docker.io/library/modbus-e2e-mapper:v1.0.0`

The very next command then fails with:

- `Error: localhost/v1.0.0:latest: image not known`

This breaks the `cri-o` runtime e2e lane before the mapper tests can run.

**What you expected to happen**:

The `cri-o` mapper import path should either:

- use the actual image name returned by `podman load`, or
- skip retagging when the loaded image already has the expected `modbus-e2e-mapper:v1.0.0` name.

The test lane should not depend on a hardcoded `localhost/v1.0.0:latest` source image that is not actually present.

**How to reproduce it (as minimally and precisely as possible)**:

1. Check out latest `master` at `c2aa3803236186ca26d840eb7051200ff5683367`.
2. Inspect `tests/scripts/generate_mapper.sh` around the CRI-O branch:
   - the mapper image is built and saved as `modbus-e2e-mapper:v1.0.0`
   - the `cri-o` branch then tags `localhost/v1.0.0:latest`
3. Run the `container runtime e2e test (cri-o)` lane, or run the same script on a Linux host with Podman:

   ```bash
   CONTAINER_RUNTIME=cri-o bash -x tests/scripts/generate_mapper.sh
   ```

4. Observe that `podman load -i modbus-mapper.tar` loads `docker.io/library/modbus-e2e-mapper:v1.0.0`.
5. The follow-up `podman tag localhost/v1.0.0:latest docker.io/library/modbus-e2e-mapper:v1.0.0` fails because the source image does not exist.

**Anything else we need to know?**:

I verified two things:

- the full failure in GitHub Actions:
[...]

PR #7263, opened 2026-08-22
fix(tests): tag CRI-O mapper from load output

RedZapdos123:
## Description:

Issue #7262 reported that the `cri-o` branch in `tests/scripts/generate_mapper.sh` hardcoded `localhost/v1.0.0:latest` as the source image after `podman load`.

That breaks the mapper import path because the loaded image name comes from the archive metadata, not from that hardcoded tag. In the failing `container runtime e2e test (cri-o)` job on August 22, 2026, `podman load` returned `docker.io/library/modbus-e2e-mapper:v1.0.0`, so the follow-up `podman tag localhost/v1.0.0:latest ...` failed before the mapper tests could run.

This PR fixes the issue in the minimal correct way:

- keep the mapper image tag in one variable
- capture the actual image name returned by `podman load`
- retag that loaded image to the short `modbus-e2e-mapper:v1.0.0` name used by the e2e deployment
- fail fast if the load output does not contain a `Loaded image` line

The change is limited to `tests/scripts/generate_mapper.sh`.

## Validation:

- `bash -n tests/scripts/generate_mapper.sh`
- `docker build -q -t modbus-e2e-mapper:7262-proof-<pid> <tempdir>`
- `docker save -o <tempdir>/modbus-mapper.tar modbus-e2e-mapper:7262-proof-<pid>`
- `docker load -i <tempdir>/modbus-mapper.tar`
- focused temporary harness that executes the patched CRI-O import block against `Loaded image: docker.io/library/modbus-e2e-mapper:v1.0.0`, `Loaded image: modbus-e2e-mapper:v1.0.0`, and invalid output with no `Loaded image` line

## Checklist:

- [x] I updated only `tests/scripts/generate_mapper.sh`.
- [x] I validated the CRI-O import logic against both fully qualified and short loaded image names.

## Release note:
```release-note
NONE
```

### DIFF
--- a/tests/scripts/generate_mapper.sh
+++ b/tests/scripts/generate_mapper.sh
@@ -17,6 +17,7 @@
 curpath=$PWD
 echo $PWD
 CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"containerd"}
+mapper_image=modbus-e2e-mapper:v1.0.0
 
 # build mapper project
 cd ${curpath}/staging/src/github.com/kubeedge/mapper-framework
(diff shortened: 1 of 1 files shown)


RedZapdos123:
The CI failing test is unrelated to this PR because the `cri-o` job now passes and the remaining failure is in the `isulad` `e2e` lane with pod startup and `statefulset` timeout errors.

Shelley-BaoYue (commented):
/lgtm 
/approve 
Maybe we need cherry-pick this PR to previous release(v1.19-1.23)?

PR #7269, opened 2026-08-22
admissioncontroller: accept valid JSON content-type parameters

### LINKED ISSUE #7264: admissioncontroller: valid application/json; charset=utf-8 requests are silently dropped
**What happened**:

In `cloud/pkg/admissioncontroller/common.go`, the admission webhook handler checks `Content-Type` with an exact string compare:

- `contentType := r.Header.Get("Content-Type")`
- `if contentType != "application/json" { ... return }`

Because of that, a valid request with `Content-Type: application/json; charset=utf-8` is rejected before decoding the `AdmissionReview`, before invoking the hook, and before writing an error response body.

On current `master`, this results in the request being silently dropped by the handler.

The admission webhook should accept valid JSON content types such as `application/json; charset=utf-8`, or at minimum return a proper HTTP error / `AdmissionReview` response instead of returning an empty body.

1. Check out latest `master` at `c2aa3803236186ca26d840eb7051200ff5683367`.
2. Inspect the current content-type gate in `cloud/pkg/admissioncontroller/common.go`.
3. Run this focused WSL repro against the current handler behavior:

   ```bash
   go test -v ./cloud/pkg/admissioncontroller -run TestIssue7105ContentTypeCharsetRepro -count=1
   ```

4. Observed output on current `master`:

   ```text
   content-type="application/json" status=200 body_len=111 hook_called=true
   content-type="application/json; charset=utf-8" status=200 body_len=0 hook_called=false
   ```

5. The handler also logs:

   ```text
   contentType=application/json; charset=utf-8, expect application/json
   ```

This is a small admission hardening bug and appears directly in scope of the controller/admission modernization work discussed in `#7105`.

As of Saturday, August 22, 2026, I did not find a direct matching issue or PR for this exact content-type handling problem.

**Environment**:
- Kubernetes version (use `kubectl version`): N/A for focused handler repro
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): latest `master` at `c2aa3803236186ca26d840eb7051200ff5683367`

  - OS (e.g. `cat /etc/os-release`): WSL
  - Go version (e.g. `go version`): `go version go1.26.4 linux/amd64`
  - Others: Focused package-level repro only

  </details>

  - edgecore version (e.g. `edgecore --version`): N/A
  - Others: N/A

  </details>

RedZapdos123:
## Description:

Issue #7264 reported that valid admission webhook requests with `Content-Type: application/json; charset=utf-8` were silently dropped because `serve()` compared the header to the exact string `application/json`.

When that happened, the handler returned immediately after logging the mismatch, so the hook function was never invoked for a valid JSON request and the caller got no proper webhook response body.

This PR fixes the issue in the minimal correct way:

- parse the `Content-Type` header with `mime.ParseMediaType`
- accept `application/json` even when parameters like `charset=utf-8` are present
- return `415 Unsupported Media Type` for real invalid content types instead of silently returning
- add regression tests for exact JSON, JSON with parameters, and invalid content type paths

The change is limited to `cloud/pkg/admissioncontroller/common.go` and `cloud/pkg/admissioncontroller/common_test.go`.

Closes #7264.

## Validation:

- `go test ./cloud/pkg/admissioncontroller -count=1`
- `go test -run TestServe -count=1 ./cloud/pkg/admissioncontroller`

## Checklist:

- [x] I updated only `cloud/pkg/admissioncontroller/common.go` and `cloud/pkg/admissioncontroller/common_test.go`.
- [x] I added regression tests for valid JSON content types with parameters and invalid content type rejection.

## Release note:
```release-note
NONE
```

### DIFF
--- a/cloud/pkg/admissioncontroller/common.go
+++ b/cloud/pkg/admissioncontroller/common.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"io"
+	"mime"
 	"net/http"
 
 	admissionv1 "k8s.io/api/admission/v1"
@@ -76,8 +77,10 @@ func serve(w http.ResponseWriter, r *http.Request, hook hookFunc) {
 
 	// verify the content type is accurate
 	contentType := r.Header.Get("Content-Type")
-	if contentType != "application/json" {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	if err != nil || mediaType != "application/json" {
 		klog.Errorf("contentType=%s, expect application/json", contentType)
+		http.Error(w, "invalid Content-Type, expect application/json", http.StatusUnsupportedMediaType)
 		return
 	}
 
--- a/cloud/pkg/admissioncontroller/common_test.go
+++ b/cloud/pkg/admissioncontroller/common_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -13,8 +14,6 @@ import (
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
-
-	"github.com/kubeedge/kubeedge/cloud/test/httpfake"
 )
 
 func TestRegisterValidateWebhook(t *testing.T) {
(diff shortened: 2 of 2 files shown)


RedZapdos123:
The failing cri-o job is unrelated to this PR and is fixed by #7263.

omlahore:
I tried the charset and xml subtests against the old serve() and both fail there, and they pass on your branch. It conflicts with master now though. #7255 added a "nil admission request" subtest to TestServe on 09-10 that still uses httpfake.NewResponseWriter, so after a rebase that one needs httptest too, and w.Body.Bytes() where it unmarshals the response.

PR #7276, opened 2026-08-27
edgeapplication: make replaceEnv deterministic to preserve env var order

### LINKED ISSUE #7275: edgeapplication: non-deterministic environment variable ordering in replaceEnv triggers unnecessary pod restarts
**What happened**:

In `cloud/pkg/controllermanager/edgeapplication/overridemanager/envoverrider.go`, the `replaceEnv` function builds the updated environment variable slice by iterating over a Go map:

https://github.com/kubeedge/kubeedge/blob/master/cloud/pkg/controllermanager/edgeapplication/overridemanager/envoverrider.go#L158-L178

```go
func replaceEnv(curEnv []corev1.EnvVar, replaceValues []corev1.EnvVar) []corev1.EnvVar {
	newEnv := make([]corev1.EnvVar, 0, len(curEnv))
	currentMap := make(map[string]corev1.EnvVar)

	// Populate current map with existing environment variables
	for _, envVar := range curEnv {
		currentMap[envVar.Name] = envVar
	}

	// Replace or add new environment variables
	for _, replaceVar := range replaceValues {
		currentMap[replaceVar.Name] = replaceVar
	}

	// Convert map back to slice
	for _, envVar := range currentMap {
		newEnv = append(newEnv, envVar)
	}

	return newEnv
}
```

Because Go map iteration order is deliberately randomized by design, each reconciliation or application of `replaceEnv` produces the `[]corev1.EnvVar` slice in a non-deterministic order.

When Kubernetes compares the spec of existing workloads against the newly reconciled spec, differing slice orders in container `env` are detected as changes, causing:
1. Spurious spec updates and noisy diffs on every controller sync.
2. Unnecessary rolling updates / pod restarts of edge applications.
3. Flaky or non-deterministic test assertions when comparing patched manifests.

`replaceEnv` should preserve the original order of `curEnv` (replacing variables in-place) and append any newly added variables at the end of the slice deterministically.

Suggested fix:
```go
func replaceEnv(curEnv []corev1.EnvVar, replaceValues []corev1.EnvVar) []corev1.EnvVar {
	replaceMap := make(map[string]corev1.EnvVar, len(replaceValues))
	for _, replaceVar := range replaceValues {
		replaceMap[replaceVar.Name] = replaceVar
	}

	newEnv := make([]corev1.EnvVar, 0, len(curEnv)+len(replaceValues))
	applied := make(map[string]bool, len(replaceValues))

	// Preserve original order and replace in-place
	for _, envVar := range curEnv {
		if replacement, ok := replaceMap[envVar.Name]; ok {
			newEnv = append(newEnv, replacement)
			applied[envVar.Name] = true
		} else {
			newEnv = append(newEnv, envVar)
		}
	}

	// Append any net-new variables in the order specified in replaceValues
	for _, replaceVar := range replaceValues {
		if !applied[replaceVar.Name] {
			newEnv = append(newEnv, replaceVar)
		}
	}

	return newEnv
}
```

1. Configure an `EdgeApplication` with multiple environment variables and an `EnvOverrider` using `replace` operation.
2. Reconcile or call `replaceEnv` repeatedly with `curEnv = [A, B, C, D]`.
3. Observe that `newEnv` produces different element orders (e.g., `[B, D, A, C]`, `[C, A, D, B]`, etc.) across runs.

Preserving slice order is critical in Kubernetes resource manifests because Kubernetes treats list order differences in container env lists as object modifications.

**Environment**:
- Kubernetes version (use `kubectl version`): all
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): master

  - Component: `controllermanager` (`edgeapplication/overridemanager`)

  </details>

  - N/A (Cloud component)

  </details>

rahulkr182:
**What type of PR is this?**

/kind bug

In `cloud/pkg/controllermanager/edgeapplication/overridemanager/envoverrider.go`, `replaceEnv` previously built the returned slice by iterating over a Go map (`for _, envVar := range currentMap`).

Because Go map iteration order is non-deterministic, this randomized the order of `[]corev1.EnvVar` on every reconciliation, causing Kubernetes to detect spurious spec differences and trigger unnecessary pod restarts and rolling updates.

This PR:
1. Updates `replaceEnv()` to preserve the original ordering of existing environment variables and replace them in-place.
2. Appends any net-new variables in the deterministic order they appear in `replaceValues`.
3. Updates `TestReplaceEnv` to assert exact slice equality `assert.Equal(t, tt.expected, result)` ensuring slice ordering is verified.

Fixes #7275 

All unit tests in `controllermanager` pass including with `-race` and `-count=5`.

```release-note
NONE
```

### DIFF
--- a/cloud/pkg/controllermanager/edgeapplication/overridemanager/envoverrider.go
+++ b/cloud/pkg/controllermanager/edgeapplication/overridemanager/envoverrider.go
@@ -156,22 +156,29 @@ func overrideEnv(curEnv []corev1.EnvVar, envOverrider *v1alpha1.EnvOverrider) ([
 }
 
 func replaceEnv(curEnv []corev1.EnvVar, replaceValues []corev1.EnvVar) []corev1.EnvVar {
-	newEnv := make([]corev1.EnvVar, 0, len(curEnv))
-	currentMap := make(map[string]corev1.EnvVar)
-
-	// Populate current map with existing environment variables
-	for _, envVar := range curEnv {
-		currentMap[envVar.Name] = envVar
-	}
-
-	// Replace or add new environment variables
+	replaceMap := make(map[string]corev1.EnvVar, len(replaceValues))
 	for _, replaceVar := range replaceValues {
-		currentMap[replaceVar.Name] = replaceVar
+		replaceMap[replaceVar.Name] = replaceVar
 	}
 
-	// Convert map back to slice
-	for _, envVar := range currentMap {
-		newEnv = append(newEnv, envVar)
+	newEnv := make([]corev1.EnvVar, 0, len(curEnv)+len(replaceValues))
+	applied := make(map[string]struct{}, len(replaceValues))
+
+	// Preserve original order and replace in-place
+	for _, envVar := range curEnv {
+		if replacement, ok := replaceMap[envVar.Name]; ok {
+			newEnv = append(newEnv, replacement)
+			applied[envVar.Name] = struct{}{}
+		} else {
+			newEnv = append(newEnv, envVar)
+		}
+	}
+
+	// Append any net-new variables in deterministic order
+	for _, replaceVar := range replaceValues {
+		if _, ok := applied[replaceVar.Name]; !ok {
+			newEnv = append(newEnv, replaceVar)
+		}
 	}
 
 	return newEnv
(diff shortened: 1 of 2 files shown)


omlahore (commented):
I ran the updated test on master and it fails about half the time since the order comes from map iteration, and it passes every time with this. Both copies get added now if an override lists the same new name twice, master only kept the last one. kubelet takes the last one anyway so I don't feel strongly about it 🙂

PR #7284, opened 2026-09-03
cloudhub: fix panic on double close of the ack channel

adity1raut:
**What type of PR is this?**

/kind bug

`NodeSession.ReceiveMessageAck` closes the ack channel through three separate steps:

```go
ackChan, exist := ns.ackMessageCache.Load(parentID)
if exist {
    close(ackChan.(chan struct{}))
    ns.ackMessageCache.Delete(parentID)
}
```

`Load` → `close` → `Delete` is not atomic. Two acks carrying the same `parentID` can both pass the `Load` before either one reaches the `Delete`, and the second `close` panics with `close of closed channel`. The panic is not recovered, so it takes down the whole `cloudcore` process.

This is not a corner case. `sendMessageWithRetry` resends the *same* message — keeping its ID — up to four times while waiting for an ack, so the edge node acking each copy produces several acks with an identical `parentID` by design. Over QUIC those acks are then handled concurrently, because `pkg/viaduct/pkg/conn/quic.go` spawns one `handleMessage` goroutine per stream.

The fix is to use `sync.Map.LoadAndDelete`, which is atomic, so exactly one caller obtains the channel and closes it.

None filed; found while auditing the cloudhub session code.

**Reproduction**

The regression test added in this PR (`TestNodeSessionReceiveMessageAckConcurrently`) drives concurrent acks for a single `parentID`. Against `master` it panics:

```
panic: close of closed channel

goroutine 419 [running]:
...session.(*NodeSession).ReceiveMessageAck(0x0?, {0x20eb0ed, 0x24})
	cloud/pkg/cloudhub/session/node_session.go:130 +0x65
```

With this change it passes.

**Testing**

- `go test -count=1 ./cloud/pkg/cloudhub/...` — the `session` package passes.
- `go vet ./cloud/pkg/cloudhub/...` — clean.
- Two failures in that tree are pre-existing on an unmodified `master` and unrelated to this change: the `-race` warning in `TestNodeSessionKeepAliveCheck` (gomock/`terminateOnce`) and `TestCheckNode/node_not_found_returns_404` in `servers/httpserver/node`.

```release-note
Fixed a panic in CloudHub that could crash cloudcore when an edge node sent duplicate acknowledgements for the same message.
```


adity1raut:
The three red `container runtime e2e test` jobs here are pre-existing and unrelated to this change, which only touches `cloud/pkg/cloudhub/session`.

The same jobs fail on a clean `master` build with no PR changes — run [32947876117](https://github.com/kubeedge/kubeedge/actions/runs/32947876117) at `b414e0a2`, the commit this branch is based on:

| job | on master `b414e0a2` | on this PR |
|---|---|---|
| `container runtime e2e test (cri-o)` | failure | failure |
| `container runtime e2e test (docker)` | cancelled (fail-fast) | cancelled |
| `container runtime e2e test (isulad)` | cancelled (fail-fast) | cancelled |
| `E2e test (WebSocket, v1.30.0)` | failure | pass |

Both runs fail at the same step in `tests/scripts/generate_mapper.sh`, with `cri-o` failing first and fail-fast cancelling the other two matrix legs:

```
+ sudo podman load -i modbus-mapper.tar
Loaded image: docker.io/library/modbus-e2e-mapper:v1.0.0
+ sudo podman tag localhost/v1.0.0:latest docker.io/library/modbus-e2e-mapper:v1.0.0
```

Newer podman keeps the `repository:tag` from the archive instead of naming it `localhost/v1.0.0:latest`, so the hardcoded re-tag fails. I opened #7285 with a fix for that script — it needs a maintainer to approve its workflows before its own `cri-o` job can confirm it.

Shelley-BaoYue (commented):
Thanks for your contributions! The PR https://github.com/kubeedge/kubeedge/pull/7263 solved the CRI-O CI has been merged. You can rebase the newest master branch.

adity1raut:
@Shelley-BaoYue PTAL

Shelley-BaoYue (commented):
/lgtm 
/approve

PR #7285, opened 2026-09-03
tests: fix modbus mapper image import for the cri-o e2e job

adity1raut:
**What type of PR is this?**

/kind bug
/kind failing-test

The `container runtime e2e test (cri-o)` job is currently red on `master`, and it takes the `docker` and `isulad` legs of the matrix down with it via fail-fast, so all three container runtime jobs go red on every PR.

The failure is in `tests/scripts/generate_mapper.sh`:

```
+ sudo podman load -i modbus-mapper.tar
Loaded image: docker.io/library/modbus-e2e-mapper:v1.0.0
+ sudo podman tag localhost/v1.0.0:latest docker.io/library/modbus-e2e-mapper:v1.0.0
make: *** [Makefile:296: e2e] Error 1
```

Older podman releases named an image loaded from a docker archive `localhost/v1.0.0:latest`, so the script re-tagged it to the name the e2e manifests expect. Newer podman keeps the `repository:tag` recorded in the archive — as the `Loaded image:` line above shows, it is *already* `docker.io/library/modbus-e2e-mapper:v1.0.0`. The hardcoded re-tag then fails on an image name that no longer exists, and because it is chained with `&&` the non-zero status fails the whole `make e2e` target.

This PR re-tags only when the expected name is not present after the load, so the script works with both the old and the new podman naming behaviour.

None filed.

**Evidence this is pre-existing and not PR-specific**:

The same three jobs fail on a clean `master` build with no PR changes — run [32947876117](https://github.com/kubeedge/kubeedge/actions/runs/32947876117) at `b414e0a2`:

| job | conclusion |
|---|---|
| `container runtime e2e test (cri-o)` | failure |
| `container runtime e2e test (docker)` | cancelled (fail-fast) |
| `container runtime e2e test (isulad)` | cancelled (fail-fast) |

and that run shows the identical `image not known` error at the same step.

**Testing**

`bash -n tests/scripts/generate_mapper.sh` passes. The real verification is this PR's own `container runtime e2e test (cri-o)` job, which exercises the changed code path.

```release-note
NONE
```


adity1raut:
Close due to already having the PR

adity1raut:
@Shelley-BaoYue, I think this PR is ready to be opened.

Shelley-BaoYue:
Hi, @adity1raut cri-o e2e job has already been fixed. Still thanks for your contributions!

adity1raut:
> Hi, @adity1raut cri-o e2e job has already been fixed. Still thanks for your contributions!

https://github.com/kubeedge/kubeedge/actions/runs/34740506035/job/103679293347 

@Shelley-BaoYue, could you please check this one? This test is still failing.

Shelley-BaoYue:
> > Hi, @adity1raut cri-o e2e job has already been fixed. Still thanks for your contributions!
> 
> https://github.com/kubeedge/kubeedge/actions/runs/34740506035/job/103679293347
> 
> @Shelley-BaoYue, could you please check this one? This test is still failing.

You're right! I'm sorry I didn't notice it at first. There's a conflict you need to resolve. : )

Shelley-BaoYue (commented):
Thanks for fixing this issue. 
/lgtm 
/approve

PR #7287, opened 2026-09-05
edgecontroller: report failures to patch reserved node labels

adity1raut:
# What type of PR is this?

/kind bug

# What this PR does / why we need it

`createNode` creates the edge node and then patches back the `kubernetes.io` labels that the `noderestriction` admission plugin does not allow the kubelet to set at creation time.

The result of that patch operation is currently not propagated correctly.

```go
node, err = uc.kubeClient.CoreV1().Nodes().Create(...)
if err == nil && len(kubernetesReversedLabels) > 0 {
    patchBytes, err := json.Marshal(...)
    if err == nil {
        node, err = uc.kubeClient.CoreV1().Nodes().Patch(...)
    }
}
return node, err
````

 The short variable declaration introduces a new `err` inside the `if` block. Therefore, when `Patch` returns an error, that error is assigned to the inner `err` variable and is discarded when the block exits.

 `node` is not redeclared, so it continues to refer to the outer variable and is overwritten with the result of the failed patch operation.

 The function then returns the outer `err`, which remains `nil` because the patch block is only entered after the node `Create` operation succeeds.

 As a result, a node whose reserved labels could not be applied is incorrectly reported as a successful registration.

 `registerNode` checks `err != nil`, receives `nil`, and sends the edge node an `ObjectResp` with a successful result even though the reserved labels were not applied:

```
ObjectResp{
    Object: <the failed patch result>,
    Err:    nil,
}
```

 The node itself does exist in the cluster because the creation succeeded, but the `kubernetes.io/*` labels are silently missing.

 This PR fixes the issue by assigning the patch error to the enclosing `err` variable so that the failure is correctly propagated to the caller.

 The created node is also preserved when the patch fails, since the node has already been successfully created in the cluster at that point.

 # Which issue(s) does this PR fix?

 None filed; this issue was found while auditing the upstream edgecontroller registration path.

 # Reproduction

 The issue can be reproduced using a fake clientset whose `patch nodes` reactor returns an error.

 On `master`, `createNode` incorrectly returns a `nil` error:

```
createNode returned node=&Node{ObjectMeta:{ ... map[kubernetes.io/os:linux] ...}} err=<nil>
```

 This means the caller incorrectly considers the node registration successful even though applying the reserved labels failed.

 The regression test added in this PR, `TestCreateNodeReservedLabelPatch`, covers this behavior.

 The `patch failure is reported to the caller` subtest fails on `master` with:

```
Error:      An error is expected but got nil.
Messages:   a failed reserved label patch must not be reported as a successful registration
```

 With this change, the test correctly receives the patch error from `createNode`.

 A second subtest verifies the successful path and ensures that both reserved and non-reserved labels are preserved when the patch succeeds.

 # Testing

 The following checks were run successfully:

 - `go test ./cloud/pkg/edgecontroller/...`
- `go vet ./cloud/pkg/edgecontroller/...`
- `gofmt`

 # Does this PR introduce a user-facing change?

```
Fixed edge node registration reporting success when the kubernetes.io reserved labels could not be applied to the node.
```


Shelley-BaoYue (commented):
/lgtm 
/approve

PR #7291, opened 2026-09-11
edge/taskmanager: fix operator precedence in node upgrade reportActionStatus

adity1raut:
**What type of PR is this?**

/kind bug

In the node upgrade runner, `reportActionStatus` should only build the version extend when the response is a `*nodeUpgradeJobActionResponse` and the action is `BackUp`, `Upgrade` or `RollBack`:

```go
if resp, ok := resp.(*nodeUpgradeJobActionResponse); ok &&
	action == string(operationsv1alpha2.NodeUpgradeJobActionBackUp) ||
	action == string(operationsv1alpha2.NodeUpgradeJobActionUpgrade) ||
	action == string(operationsv1alpha2.NodeUpgradeJobActionRollBack) {
	extend = taskmsg.FormatNodeUpgradeJobExtend(resp.FromVersion, resp.ToVersion)
}
```

Because `&&` binds more tightly than `||`, Go evaluates this as `(ok && action == BackUp) || action == Upgrade || action == RollBack`.

`RunAction` reports a `*baseActionResponse` when the spec serializer, the pre-run hook or the action lookup fails. In that case the type assertion fails and the shadowed `resp` is nil. For `Upgrade` and `RollBack` the condition is still true, so `resp.FromVersion` dereferences a nil pointer and edgecore panics. One way this happens: after a failed upgrade, `ReportUpgradeStatus` runs the `RollBack` action and `upgradeDao.Save` fails in `preRun`.

This PR wraps the action checks in parentheses, so the extend is only built after the type assertion succeeds. Reports for a `*nodeUpgradeJobActionResponse` behave exactly as before.

None filed. I found this while reviewing the node upgrade status reporting path.

I added three cases to `TestNodeUpgradeJobReportActionStatus` that pass a `*baseActionResponse` for `BackUp`, `Upgrade` and `RollBack`. The test loop now also checks that `reportActionStatus` does not panic and that the status is still reported. On `master`, the `Upgrade` and `RollBack` cases fail with:

```
Panic value: runtime error: invalid memory address or nil pointer dereference
... actions.nodeUpgradeJobActionHandler.reportActionStatus(...)
    edge/pkg/taskmanager/actions/nodeupgradejob.go:328
```

Tests run on this branch:

- `go test -gcflags "all=-N -l" ./edge/pkg/taskmanager/actions/ -run TestNodeUpgradeJobReportActionStatus -v`: all 6 subtests pass
- `go test -gcflags "all=-N -l" ./edge/pkg/taskmanager/...`: pass
- `go vet ./edge/pkg/taskmanager/actions/`: no issues
- `staticcheck ./edge/pkg/taskmanager/actions/...`: no issues
- `gofmt -l` on the modified files: no changes needed

The second commit fixes the CRI-O container runtime e2e job, which currently fails on `master` too. `E2E_MAPPER_1` fails because the mapper pod never leaves Pending. On the current runner image, `podman load` names the saved mapper archive `localhost/v1.0.0:latest`. The script then tags it with the short name, which podman stores as `localhost/modbus-e2e-mapper:v1.0.0`. CRI-O resolves the deployment's short name to `docker.io/library/modbus-e2e-mapper:v1.0.0`, so it can't find the image. The script now tags the loaded image with that `docker.io/library` name. I can move this commit to a separate PR if you prefer.

```release-note
Fixed an edgecore panic when reporting the status of a node upgrade Upgrade or RollBack action that failed before it ran.
```


adity1raut:
@Shelley-BaoYue PTAL

omlahore (commented):
I ran the new cases in a linux container. Upgrade and rollback pre-run panic on master and pass with this, so I think the fix looks good 👌

PR #7294, opened 2026-09-12
edgeapplication: preserve optional environment references

Hanabi9248:
**What type of PR is this?**

/kind bug

Adding an environment variable through an EdgeApplication override rewrites the existing env array. This drops `optional` from ConfigMap and Secret key references, making an optional missing object or key required. It also drops `fieldRef` without `apiVersion` and `resourceFieldRef` without `containerName` or `divisor`, leaving an empty `valueFrom`.

Preserve the optional flag, including omitted versus explicit `false`. Retain downward API references when their optional fields are absent, using the default divisor of `1`.

Round-trip regressions add an unrelated env variable and check the existing references. They cover ConfigMap/Secret optional flags and downward API fields with and without explicit values.

The current revision (`675ac58`) passes the affected package with `-race`, `go vet`, and `go test ./cloud/pkg/controllermanager/edgeapplication/...` on Linux with Go 1.27.1. The earlier revision (`35036e4`) also passed `make verify`, `make lint`, `make test PROFILE=y`, and `make integrationtest` on Ubuntu 22.04 with Go 1.23.12 in [this validation run](https://github.com/Hanabi9248/kubeedge/actions/runs/34684818977).

```release-note
EdgeApplication environment overrides preserve optional ConfigMap and Secret key references and downward API references with omitted optional fields.
```


omlahore (commented):
I checked the new cases, the four with an explicit optional fail on master and pass here. fieldRef and resourceFieldRef lose fields the same way though, a fieldRef with just fieldPath or a resourceFieldRef without divisor turns into valueFrom: {} after adding an unrelated env and the API server rejects that. Not sure if you want that in this PR or a separate one 🙏

Hanabi9248:
Thanks for checking these cases. Included the downward API references in 675ac58 since they go through the same env rewrite. `apiVersion` and `containerName` no longer gate the conversion, and an omitted divisor uses its default of `1`.

Added six round-trip cases: the four with omitted fields fail before this follow-up and pass now. The affected package passes with `-race` on Linux, along with `go vet` and the edgeapplication package tests.

WillardHu:
Please add a sign-off to your all commits to resolve the DCO check.

https://github.com/kubeedge/kubeedge/pull/7294/checks?check_run_id=104714910414

omlahore:
Hey, thanks for adding those. I checked 675ac58 and the new downward API cases fail without it and pass with it, so I think it looks good 😄

Hanabi9248:
Added sign-offs to both commits; DCO now passes. The code is unchanged. Thanks for checking the downward API cases as well.

Hanabi9248:
Thanks for the review and LGTM last week. The downward API cases and commit sign-offs are addressed. Could an approver take another look when convenient? Please let me know if anything else is needed for approval.

PR #7301, opened 2026-09-14
synccontroller: set the GVK on the orphaned-sync delete stub

skupr:
**What type of PR is this?**

/kind bug

Force-delete a pod while an edge node is offline and the node comes back looking almost clean:
kubelet removes the pod, the v1 `meta` table is cleared, the cloud deletes the ObjectSync and
considers the node reconciled. But `meta_v2`, the table MetaServer serves from, keeps the pod
forever.

The two messages that would have done the right thing never arrive. EdgeController's watch
dispatches a final update and a proper delete, both carrying the full pod with intact TypeMeta,
but the node has no session, so `messageDispatcher` creates a pool for it that nothing drains and
both are discarded. The only message that survives to be retried is the one
`gcOrphanedObjectSync()` fabricates — the object is gone from K8s, so there is nothing real left
to send — and it sets namespace, name and UID with no GVK.

The edge then rejects it. `imitator.Event()` decodes with `unstructured.UnstructuredJSONScheme`,
which requires a Kind:

```
imitator.go:258] failed to unmarshal message content to unstructured obj:
Object 'Kind' is missing in '{"metadata":{"name":"probe","namespace":"default","uid":"..."}}'
```

It returns zero events, so `meta_v2` is neither written nor its watch hook triggered. Nothing
reports this: `Inject()` has no error return, the v1 delete that follows succeeds, metaManager
ACKs, and CloudHub deletes the ObjectSync on that ACK — which ends the retries and makes the row
permanent. `KeyFuncObj()` would reject the stub on the same grounds even if the decode passed.

Ordinary deletes never hit either guard, because `manager/common.go:31` runs `util.SetMetaType()`
on every informer event. This is specific to the offline path, and it is not pod-specific: one
force-delete also stranded the pod's `ServiceAccountAccess`, and any kind SyncController
garbage-collects can leave a row behind.

While the node is connected the row is invisible, since MetaServer serves LIST from the cloud. On
the next disconnect it serves the phantom from local storage, as `Running`, with no containers
behind it. A cloud LIST is never written back — the `InsertOrUpdateObj()` call in metaserver
`storage.go:196` is commented out — so nothing prunes it.

This PR stamps the GVK from the ObjectSync spec, which already carries both fields. Cluster-scoped
syncs get the same treatment.

Fixes #

Opened as a draft: the fix and its test are complete and verified, but I would like a maintainer's
read on two things before marking it ready.

1. Whether stamping the GVK in `gcOrphanedObjectSync()` is the right layer, versus having
   `buildEdgeControllerMessage()` do it for every message it builds.
2. Whether the same treatment is wanted for any other fabricated body I have not found.

Verified on a v1.22 cluster with **only cloudCore replaced**, leaving the edge node on its
released build — which is enough to see the effect, because the imitator's rejection is
`klog.Errorf` and its *absence* is observable at default verbosity.

| | before | after |
|---|---|---|
| delivered stub body | `{"metadata":{"name":"ghost-probe","namespace":"default","uid":"..."}}` | `map[apiVersion:v1 kind:Pod metadata:map[...]]` |
| CRD stub body | `{"metadata":{"name":"default",...}}` | `map[apiVersion:policy.kubeedge.io/v1alpha1 kind:ServiceAccountAccess ...]` |
| imitator decode failures | 2 | 0 |
| stranded `meta_v2` rows | 2 | 0 |
| MetaServer while disconnected | serves the pod as `phase=Running` | empty list, 404 on GET |

The stub still carries an empty PodSpec, so the `edged.go:394` branch that turns a stub delete
into a full re-list from the v1 table is unaffected — that is what makes kubelet drop the pod
today even though `meta_v2` keeps it.

**What this PR deliberately does not fix**: the message loss described above. The correct,
fully-formed delete is still dropped into a pool nobody drains
(`message_dispatcher.go:510`), which I confirmed is still the case after this change. That
deserves its own fix.

```release-note
Fixed objects being left in an edge node's local metaserver storage forever when they are deleted
from Kubernetes while the node is disconnected. MetaServer served such objects as still present
whenever the node fell back to local storage.
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


omlahore (commented):
I checked the new test, it fails on master and passes here. The cluster-scoped change isn't covered though, I reverted it in clusterobjectsync.go and the whole package still passes. TestGcOrphanedClusterObjectSyncFuncIntegration already stubs buildEdgeControllerMessageFunc, so I think checking the content's kind and apiVersion there would cover it 🤔

skupr:
> I checked the new test, it fails on master and passes here. The cluster-scoped change isn't covered though, I reverted it in clusterobjectsync.go and the whole package still passes. TestGcOrphanedClusterObjectSyncFuncIntegration already stubs buildEdgeControllerMessageFunc, so I think checking the content's kind and apiVersion there would cover it 🤔

Thank you!
Extended TestGcOrphanedClusterObjectSyncFuncIntegration, and it now covers both changes - reverting either change with tests kept reveals errors.

omlahore:
Thanks for extending the test. I reverted each change on its own and both fail now, including the cluster-scoped one in TestGcOrphanedClusterObjectSyncFuncIntegration, so I think it looks good.

PR #7304, opened 2026-09-15
dynamiccontroller: reject watches for uninformable resources instead of exiting

### LINKED ISSUE #7303: dynamiccontroller exits or hangs when an edge watches a resource cloudcore can't watch
**What happened**:

When an edge node watches a resource through metaserver, cloudcore's dynamiccontroller creates an informer for that resource on first use. If creating it fails, cloudcore either exits or gets stuck.

`ProcessApplication` / `processWatchApp` → `HandlerCenter.AddListener` → `ForResource` → `NewCommonResourceEventHandler`:

https://github.com/kubeedge/kubeedge/blob/68e45c5ef1ee49f9ca3b61f93e0bbe915f398b11/cloud/pkg/dynamiccontroller/application/eventhandler.go#L119-L123

```go
informerPair, err := genericinformers.GetInformersManager().GetInformerPair(gvr)
if err != nil {
    klog.Exitf("get informer for %s err: %v", gvr.String(), err)
}
```

There are two failure modes:

1. **The resource doesn't exist.** `informers.forResource` calls `mapper.KindFor(gvr)`. For a CRD that isn't installed this returns an error, and `klog.Exitf` stops cloudcore. The edge agent keeps watch applications for 5 minutes after they close and re-sends them through `watchapp/sync` when it reconnects, so the same watch can make cloudcore crash again after every restart.

2. **The resource exists, but cloudcore's ClusterRole can't list/watch it.** `addInformerPair` calls `cache.WaitForCacheSync(ifs.stopCh, ...)`, and `stopCh` only closes when cloudcore shuts down. The reflector keeps getting `forbidden`, so the cache never syncs and the call never returns. During that wait it holds:
   - `informers.lock`, which blocks every other `GetInformerPair` / `GetLister` caller
   - `handlerCenter.handlerLock`, which blocks every other `ForResource` / `DeleteListener`

   `Center.Process` runs synchronously on the single `receiveMessage` goroutine. That means metaserver list, get and watch requests from **every** edge node stall, not just the node that sent the watch.

   https://github.com/kubeedge/kubeedge/blob/68e45c5ef1ee49f9ca3b61f93e0bbe915f398b11/cloud/pkg/common/informers/informer_manager.go#L218-L252

The failed informer entry also stays in `informersByGVR`, because it is stored before the sync wait. Any later caller gets an unsynced informer back with a nil error.

A watch for a resource cloudcore can't inform should be rejected and the error returned to the edge, like any other failed Application. It should not stop cloudcore or block message handling for other nodes.

Authorization is off by default (`cloudHub.authorization.enable: false`), so `checkNodePermission` does not reject the watch first.

Crash:
1. Run cloudcore and an edge node with metaserver enabled, using the default configuration.
2. From a pod on the edge node (or on the node itself), watch a CRD that is not installed in the cluster through metaserver:
   `kubectl --server=http://127.0.0.1:10550 get foos.example.com -w`
3. cloudcore exits with `get informer for example.com/v1, Resource=foos err: no matches for ...`.
4. After cloudcore restarts and the edge reconnects, the watch is re-sent and cloudcore exits again.

Hang:
1. Install a CRD that cloudcore's ClusterRole does not grant `list`/`watch` on.
2. Watch it from the edge through metaserver as above.
3. cloudcore stays up but logs `failed to list ... is forbidden` repeatedly. Metaserver list/get/watch requests from all edge nodes stop getting responses.

Suggested fix:
- Have `NewCommonResourceEventHandler` and `ForResource` return an error instead of calling `klog.Exitf`, and reject the Application so the edge gets the error.
- Put a time limit on the cache-sync wait in `addInformerPair`, and don't hold `informers.lock` or `handlerLock` while waiting.
- If the sync fails, remove the informer entry so a later watch can retry cleanly.

Related, but not duplicates: #6632 and #5035 replace `Exitf` with `panic`, which still takes cloudcore down.

**Environment**:
- Kubernetes version (use `kubectl version`): any
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`): master (68e45c5ef1ee49f9ca3b61f93e0bbe915f398b11)

  - Others: default cloudcore config (authorization disabled)

  </details>

  - edgecore version (e.g. `edgecore --version`): master
  - Others: metaserver enabled

  </details>

adity1raut:
**What type of PR is this?**

/kind bug

An edge watch through metaserver reaches `HandlerCenter.ForResource`, which builds an informer for the resource the first time it is watched. Two cases break cloudcore (details in #7303):

- **The resource doesn't exist.** `mapper.KindFor` fails and `NewCommonResourceEventHandler` calls `klog.Exitf`, so cloudcore exits. The edge re-sends its watch applications when it reconnects, so the same watch can take cloudcore down again after every restart.
- **cloudcore's ClusterRole can't list the resource.** The informer never syncs, and `GetInformerPair` waits in `cache.WaitForCacheSync` until cloudcore stops. During that wait it holds the informer manager lock and `handlerLock`. Applications are processed on a single goroutine, so metaserver list, get and watch stall for every edge node.

This PR:

- Makes `NewCommonResourceEventHandler` and `ForResource` return errors instead of exiting. `AddListener` already turns an error into a rejected Application, so the edge now gets the error.
- Bounds the cache-sync wait in `GetInformerPair` to 5s, which is below the 10s the edge waits for an Application response, so the rejection reaches the edge.
- Holds neither the informer manager lock nor `handlerLock` while waiting.
- Caches an informer in `informersByGVR` only after it has synced. Previously the entry was stored before the wait, so after a failed sync later callers got an unsynced informer and a nil error.
- Keeps waiting at startup: `newDynamicController` retries the `nodes` and `services` handlers until they are ready, as it did before, so a slow initial sync in a large cluster doesn't stop cloudcore.

Fixes #7303

- `GetInformerPair` is also used by synccontroller through `GetLister`. There it now returns an error after 5s instead of blocking the worker. synccontroller already returns on that error and retries on its next reconcile.
- When `GetInformerPair` gives up, the informer keeps running in its shared factory, since the factories can't stop a single informer. So a later watch succeeds once the resource becomes listable, and a large resource that takes longer than 5s to sync the first time is accepted on retry. For a resource that stays forbidden, the reflector keeps logging `forbidden` with backoff.
- Processing is still serial, so each watch for an unlistable resource still blocks the Application goroutine for up to 5s. Before this PR it blocked until cloudcore stopped. Parallel processing is the existing `TODO` in `Center.Process` and is out of scope here.

Testing:

- New `cloud/pkg/common/informers/informer_manager_test.go` uses a fake dynamic client to cover three cases: a resource the RESTMapper doesn't know, a resource that syncs, and a resource whose list is forbidden. The forbidden case checks that the lock is free during the wait, that the call gives up, and that the informer is not cached. Against the old `GetInformerPair`, that test fails in both ways the issue describes: the lock stays held and the call never returns.
- `eventhandler_test.go` now covers `ForResource`/`AddListener` returning the informer error without registering a handler or listener, `ForResource` not holding `handlerLock` while getting the informer, and `NewCommonResourceEventHandler` returning an `AddEventHandler` error.
- `go test -race -gcflags "all=-N -l" ./cloud/pkg/common/informers/... ./cloud/pkg/dynamiccontroller/...` passes, and `go mod tidy -diff` is clean.

```release-note
cloudcore no longer exits or stops serving metaserver requests when an edge node watches a resource that does not exist or that cloudcore is not allowed to list. The watch is rejected and the error is returned to the edge.
```


adity1raut:
@Shelley-BaoYue  Can You Please Take a Look

Issue #4429, opened 2022-11-20
pod will be terminated when edge node reconnect


shao77622:
**What happened**:
make edge node disconnected from cloud core
(by modify cloudcore IP address with a non-exits host  in  /etc/kubeedge/config/edgecore.yaml , then systemctl restart edgecore.), after a while , reconnect.. 
when edge node connect to cloud core, the pod will be terminated, and pod  with another name will be scheduled.
**What you expected to happen**:
the pod will not be terminated, as v1.12.0 does.
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.22.12
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`):1.12.1
- <details><summary>Cloud nodes Environment:</summary>

  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>
- <details><summary>Edge nodes Environment:</summary>

  - edgecore version (e.g. `edgecore --version`):
  - Hardware configuration (e.g. `lscpu`):
  - OS (e.g. `cat /etc/os-release`):
  - Kernel (e.g. `uname -a`):
  - Go version (e.g. `go version`):
  - Others:

  </details>

wackxu:
Do you add tolerations for you deploy to make the pod not to be migrated when node is offline?

      tolerations:
        - effect: NoExecute
          key: node.kubernetes.io/unreachable
          operator: Exists

shao77622:
> 
No tolerations ,  but it's DaemonSet. 
I tested with kubeedge:v1.12.0, the pod will not be recreate.

shao77622:
-----sample.yaml-----
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: redis-3.2.9
  namespace: edge
  labels:
    app: redis
spec:
  selector:
    matchLabels:
      app: redis
  template:
    metadata:
      labels:
        app: redis
    spec:
      nodeSelector:
        xxx/redis: 3.2.9
      hostNetwork: true
      containers:
        - args:
            - '--requirepass'
            - '888888'
          image: redis:3.2.9
          name: redis
          ports:
            - containerPort: 6379
              name: redis
              protocol: TCP
          volumeMounts:
            - mountPath: /etc/localtime
[...]

PR #7306, opened 2026-09-16
metamanager: query cached service account tokens by key when offline

harshrajdebug:
**What type of PR is this?**

/kind bug

An edge node that has lost its connection to the cloud serves service account tokens out of its local database. `processQuery` picks between two lookups based on `resID`:

```go
if requireRemoteQuery(resType) && connect.IsConnected() {
    m.processRemote(message)
    return
}

if resID == "" {
    metas, err = m.metaService.QueryMeta("type", resType)
} else {
    metas, err = m.metaService.QueryMeta("key", resKey)
}
```

For service account tokens `parseResource` returns a hardcoded empty `resID`, so the by-type branch runs and the key it computed one line earlier goes unused. Two things follow from that:

- **The query returns every token cached on the node.** `QueryMeta("type", resType)` matches on the resource type, so it returns one row per cached token whichever token was asked for.
- **The caller then rejects the result.** `handleServiceAccountTokenFromMetaDB` accepts exactly one row and fails everything else with `serviceaccount length from meta db is N`.

So a node holding more than one cached token cannot read back any of them while it is offline, even when every one of those tokens is fresh and unexpired. Kubelet cannot populate the projected `kube-api-access-*` volume, `MountVolume.SetUp` fails, and the pod does not start. That is the offline autonomy symptom reported in #6414, #5958, #4441 and #4429, and the log in #6414 shows lookups for two different namespaces on the same node, so that node was holding at least two tokens.

While the node is connected none of this is reachable, because `processQuery` returns early through `processRemote`. The bug only appears once the node is disconnected, which is the case edge autonomy exists to cover.

The by-key lookup is how this worked originally. ea5598c33 added service account token caching with the key derived in a separate `getSpecialResourceKey`, and `processQuery` called `QueryMeta("key", resKey)`. 468af07f2 moved key derivation into `parseResource` and dropped both `getSpecialResourceKey` and the by-key call, and `resID` became an empty string in that move. `handleMessage` still stores tokens under the same `parseResource` key, so reading by key is what matches the write path.

This PR:

- Returns the service account name as `resID` from `parseResource`, so `processQuery` looks the token up by the key that was computed for it. This is a one line change and it restores the pre-468af07f2 behaviour.
- Adds `TestParseResourceServiceAccountTokenResID`, which pins `resID` and the key for a service account token message. It fails on master with `resID is empty, so processQuery would query cached tokens by type instead of by key` and passes with the change.
- Adds `TestParseResourceResID` over the resource shapes `parseResource` accepts, so the service account token case stays consistent with the rest.

Related to #6414, #5958, #4441 and #4429.

I have deliberately not written `Fixes`, because those issues need more than this change. Two further defects in `edge/pkg/metamanager/client/serviceaccount.go` feed the same symptom:

- `getTokenLocally` calls `DeleteMetaByKey` as soon as `requiresRefresh` returns true, so the cached row is destroyed before any replacement has been obtained. Offline the replacement never arrives. #6828 from @mkhon covers this half.
- `GetServiceAccountToken` has no fallback to a cached token when the refresh fails. Upstream kubelet separates "no cached token", "cached token has expired" and "cached token is stale but still valid" in `pkg/kubelet/token/token_manager.go`, and returns the cached token for the third. KubeEdge has no equivalent branch and no `expired()` check.

Both of those are worth fixing and neither belongs in this PR. I would rather send one defect at a time and leave those issues open until all three are addressed.

- `resID` is read only by `processQuery`. The other five `parseResource` call sites discard it, so no other behaviour changes shape.
- The non service account path already sets `resID` to the third token for a three part resource. This change makes the service account token case agree with it.
- I measured the related timing while tracing this. At the 3607 second TTL that appears in the #6414 logs, `requiresRefresh` starts demanding a refresh 49 minutes in, while the token still has 11 minutes and 7 seconds of validity. The `token expired` message that `getTokenLocally` logs at that point is inaccurate, which is part of why this has been hard to read from logs. That belongs to the two defects above and is not touched here.
- `edge/pkg/metamanager/metaserver/handlerfactory` fails on my machine, and it fails identically with this change reverted, so it looks unrelated to this PR.
- I reproduced this against the real code and a real sqlite database through `dao.Init`, with two fresh unexpired tokens cached: the by-type query returns 2 rows and `handleServiceAccountTokenFromMetaDB` rejects them. I have not reproduced the full symptom on a live cluster with a node cut off from cloudcore, so the connection to those four issues is strongly indicated and I would stop short of calling it proven.

```release-note
Fixed an issue where an edge node that had cached more than one service account token could not read any of them back while disconnected from the cloud, which prevented pods mounting a service account token from starting.
```


omlahore (commented):
I think this and #7307 don't line up once both are in. #7307 makes KeyFunc add the exp timestamp when the status has one, so rows stored from the cloud response end up under base/<exp>, but the key parseResource builds here comes from kubelet's request with no status, so QueryMeta("key", resKey) is looking for the bare base key and won't find a row. Maybe this needs the same prefix lookup newestCachedGeneration does?

harshrajdebug:
@omlahore you are right, and it is more definite than not finding a row: the two keys diverge by construction.

`parseResource` serves both directions. `handleMessage` uses it for the `ResponseOperation` write, where the body is the cloud's TokenRequest with `Status.ExpirationTimestamp` populated, so #7307's `KeyFunc` appends the suffix. `processQuery` uses it for kubelet's `QueryOperation`, where the body is spec-only, so `KeyFunc` takes its defensive bare-base branch. Merging both branches locally and printing the two keys for one token:

```
written = "default"/"default"/[]string(nil)/3607/v1.BoundObjectReference{Kind:"Pod", APIVersion:"", Name:"nginx", UID:"a3452707-5b1b-45e4-bc12-f5c3a7169127"}/1790151743000000000
read    = "default"/"default"/[]string(nil)/3607/v1.BoundObjectReference{Kind:"Pod", APIVersion:"", Name:"nginx", UID:"a3452707-5b1b-45e4-bc12-f5c3a7169127"}
```

So `QueryMeta("key", resKey)` looks for a key nothing is ever stored under, and the offline read returns zero rows.

One thing to add about the prefix lookup, because it is not sufficient on its own. `handleServiceAccountTokenFromMetaDB` still rejects anything that is not exactly one row, and #7307 deliberately keeps the previous generation alive across the refresh-propagation window. A bare prefix swap would trade a zero-row miss for a two-row miss inside exactly the window that change exists to create. The lookup has to select the newest un-expired generation the way `newestCachedGeneration` does, or that length check has to be relaxed.

I would rather not fold that into this PR while both are still unreviewed. This one is a correct fix against master on its own, and reconciling here would make it depend on the other. Whichever lands second is the one that has to carry it, and I will do that as soon as either gets a review. If you would prefer it here instead, say so and I will move it.

omlahore:
@harshrajdebug yeah that makes sense, I'd keep it out of this one too. Whichever one lands second can carry it like you said.

PR #7307, opened 2026-09-16
metamanager: fall back to the cached service account token when refresh fails

harshrajdebug:
**What type of PR is this?**

/kind bug

A node that has lost its connection to the cloud cannot start pods that mount a service account token. Three separate defects combine to cause it. This PR fixes two of them, in `edge/pkg/metamanager/client/serviceaccount.go`, and #7306 fixes the third, which lives in the metamanager read path.

**Updated 2026-09-17.** This PR now includes @mkhon's commit from #6828, carried with his authorship intact, after he offered it on that thread. It supersedes #6828, so there is one path to review.

The first commit is his. `getTokenLocally` deleted the cached row as soon as `requiresRefresh` returned true, before any replacement existed. At the 3607 second TTL kubelet requests for a projected volume, that happens 49 minutes in, while the token still has 11 minutes and 7 seconds of validity. Offline the replacement never arrives, so the deletion is unrecoverable. His commit stops the deletion and stores each refreshed token under its own per-generation key, so the previous generation survives the window during which a pod may still be presenting it.

The second commit is mine, and covers two gaps that remain after his.

`GetServiceAccountToken` had no fallback. It asked for the cached token and, on any error, went to `getTokenRemotely`. A token merely due for a refresh was therefore treated exactly like one that is unusable, and offline the remote call cannot succeed. Kubelet cannot populate the projected `kube-api-access-*` volume, `MountVolume.SetUp` fails, and the pod does not start. That is the symptom reported in #6414, #5958, #4441 and #4429. Upstream kubelet serves the cached token when a refresh fails and the cached one is still usable (`pkg/kubelet/token/token_manager.go`), and this restores that behaviour.

To do it, `newestCachedGeneration` is split out of `getTokenLocally` so the caller can see the cached token and make the decision, since only the caller knows whether a refresh can succeed. `getTokenLocally` keeps its previous signature, error strings and semantics, so the tests in the first commit still describe it accurately. Worth noting that `newestCachedGeneration` returns un-expired generations only, so anything it hands back is safe to serve. Upstream needs a third branch for an expired cached token because its cache holds a single entry that may have expired; the per-generation scheme filters those at read time instead.

`CheckTokenExist` treated presence in metaDB as proof that a token is usable. That held only while rows were deleted at 80% of TTL. Rows now outlive the refresh threshold, and an expired generation survives until the next GC sweep, so presence alone no longer means usable. It matters because it is the only check MetaServer performs when no public key is available for signature verification (`edge/pkg/metamanager/metaserver/auth/auth.go`), which is exactly the offline case. Without the expiry test an expired token would authenticate until the sweep removed it.

Related to #6414, #5958, #4441 and #4429.

I have not written `Fixes`, because those issues also need #7306, which restores the by-key lookup `processQuery` uses to read cached tokens while offline. All three defects have to be fixed before a disconnected node reliably starts a pod that mounts a service account token.

- **On #6828.** @mkhon said on that thread: "you can copy missing functionality from this PR to yours. Absolutely no problems with this as soon as this gets landed into main/master somehow". His commit is included here unmodified and under his authorship, so review of that half is review of his work.
- The tests I added drive the real metaDB through `dao.Init`, with no stubbing of the data layer, so they cover the per-generation keys end to end. `TestGenerationsCoexistAcrossRefresh` asserts that two generations coexist after a refresh and that both authenticate, which is the propagation window the first commit exists to protect.
- Reverting only my two changes and keeping the tests fails `TestGetServiceAccountTokenServesStaleTokenWhenOffline` and `TestCheckTokenExistRejectsExpired`, so they pin the behaviour.
- **Two local failures, neither caused by this PR, both worth stating plainly.** `TestGetTokenLocally` from the first commit fails on my machine. I isolated the cause: on darwin/arm64 with Go 1.26, gomonkey does not re-patch a method within a single test function, and a probe using a fresh `gomonkey.Patches` still returned the first stub. That test patches the same method several times in one function, so every case after the first sees the earlier stub. It fails the same way on the unmodified #6828 branch, so it is my platform. I have left the test alone and that check rests on CI.
- Separately, the whole `edge/pkg/metamanager/client` package hits `fatal error: runtime: name offset base pointer out of range` here, and it does so on a clean master checkout with no changes applied, so it looks like the same toolchain problem. The tests I added pass when run directly.
- While reading `CheckTokenExist` I noticed it returns false for the entire lookup if any single cached row fails to unmarshal, where skipping that row would be safer. I have left it alone as a separate concern.

```release-note
Edge nodes now keep using a cached service account token that has passed its refresh threshold but has not expired when the cloud is unreachable, so pods mounting a service account token can still start while the node is disconnected. Refreshed tokens are stored per generation so the previous token keeps working until the pod picks up the new one, and expired tokens are no longer accepted by the metaserver token check.
```


harshrajdebug:
I have reworked this to carry @mkhon's commit from #6828, after he offered it there: "you can copy missing functionality from this PR to yours. Absolutely no problems with this as soon as this gets landed into main/master somehow". His commit is included unmodified and under his authorship, so this now supersedes #6828, and there is one path for a reviewer to follow.

The PR description is updated. In short, his commit stops the cached row being deleted before a replacement exists and stores each refreshed token under a per-generation key so the previous generation survives the window in which a pod may still be presenting it. My commit adds the fallback in `GetServiceAccountToken` and the expiry test in `CheckTokenExist`, neither of which was covered by #6828.

Folding the two together changed one thing in my half. I split `newestCachedGeneration` out of his `getTokenLocally` so the caller can see the cached token and decide, and because his loop already filters expired generations, anything it returns is un-expired by construction. The three way branch I had before collapses to two. Upstream kubelet needs the third case because its cache holds a single entry that may have expired, and the per-generation scheme filters those at read time instead.

The tests I added drive the real metaDB through `dao.Init`, with no stubbing of the data layer. `TestGenerationsCoexistAcrossRefresh` asserts that two generations coexist after a refresh and that both still authenticate, which is the propagation window his commit exists to protect and the one thing my earlier version of this PR got wrong.

Two local failures that I want to state plainly, so nobody trips over them. `TestGetTokenLocally`, from his commit, fails on my machine, and it fails identically on the unmodified #6828 branch. The cause is that gomonkey does not re-patch a method within a single test function on darwin/arm64 with Go 1.26; a probe with a fresh `gomonkey.Patches` still returned the first stub. That test patches the same method several times in one function, so every case after the first reads the earlier stub. I have not touched the test, and that check rests on CI. Separately the whole package hits `fatal error: runtime: name offset base pointer out of range` here, including on a clean master checkout with nothing applied.

#7306 is still needed alongside this. It restores the by-key lookup that `processQuery` uses to read cached tokens while offline, which is a third defect in the same path and lives in a different file.

PR #7312, opened 2026-09-18
fix silent build failure when docker group is absent grep exits 1 whe…

### LINKED ISSUE #7311: build_with_container.sh fails silently (no output) when /etc/group has no docker entry
**What happened**:

On macOS, the default build fails immediately without printing any diagnostic output:
Image

Even the script's first ``echo`` is never reached, so there is no indication of what went wrong.

The issue is caused by this line in ``hack/make-rules/build_with_container.sh``:

```bash  
DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group | cut -f3 -d:)}"   
```

On macOS, `/etc/group` does not contain a docker entry because Docker Desktop does not create one. Since the script uses both `set -o errexit` and `set -o pipefail`, the failed grep causes the script to exit immediately.

As a result, the build stops before the script's echo and before Docker is ever invoked.

This affects the default build path because ``BUILD_WITH_CONTAINER`` is set to true in the Makefile.

Running:

```bash
make all WHAT=cloudcore
```

on macOS should either:

1. build successfully inside the container, or
2. fail with a clear error message explaining what is missing and how to proceed.

A silent Error 1 makes the problem difficult to diagnose.

**Steps to reproduce**:

On macOS with Docker Desktop installed and running:

```bash
git clone https://github.com/kubeedge/kubeedge
cd kubeedge
make all WHAT=cloudcore
```

The underlying failure can also be reproduced independently:

```bash
$ bash -c 'set -o errexit; set -o pipefail; \
  DOCKER_GID="${DOCKER_GID:-$(grep "^docker:" /etc/group | cut -f3 -d:)}"; \
  echo "reached echo, GID=$DOCKER_GID"'
$ echo $?
1 
```

The echo is never reached.

**Possible fix**:

One approach is to tolerate the missing docker group and only pass the -u option when a GID is available:

```bash
DOCKER_GID="${DOCKER_GID:-$(grep '^docker:' /etc/group 2>/dev/null | cut -f3 -d: || true)}"

USER_OPTS=()
if [[ -n "${DOCKER_GID}" ]]; then
  USER_OPTS=(-u "${UID}:${DOCKER_GID}")
fi 
```

This would allow Docker Desktop on `macOS` to handle bind-mount ownership without requiring a docker group entry.

If container builds are intentionally not supported on macOS, the script should instead exit with an explicit message such as:

Container builds are not supported on macOS.
Please run with BUILD_WITH_CONTAINER=false.

**Existing workarounds**:

The following workarounds currently work, but are not discoverable from the error:

```bash
make all WHAT=cloudcore BUILD_WITH_CONTAINER=false

or:

DOCKER_GID=0 make all WHAT=cloudcore 
```

**Related issue**:

This appears to be the same root cause reported in `#3881 (fix build_with_container.sh)` in May 2022. That PR was not reviewed, became stale in July 2022, and was automatically closed in August 2022. The affected line remains unchanged on master.

macOS is otherwise handled explicitly in the repository's tooling. For example, `hack/lib/lint.sh` contains `darwin-specific` handling and provides instructions for installing GNU sed on macOS.

This makes the current behavior of `build_with_container.sh` inconsistent with the rest of the tooling.

**Environment**:

- KubeEdge: master @ d062d2983
- Kubernetes: n/a — build-time issue, no cluster involved
- Hardware: Apple Silicon (arm64)
- OS: macOS 27.0
- Go: go1.23.12 darwin/arm64
- Docker: 29.8.0
- GNU Make: 3.81 (macOS default)

tarone-saloni:
**What type of PR is this?**

/kind bug

On macOS, `/etc/group` does not contain a `docker` group entry because Docker Desktop does not create one.

`build_with_container.sh` currently resolves the Docker group ID using:

```bash
grep '^docker:' /etc/group
```

When the group does not exist, `grep` exits with status 1. Since the script uses `set -o errexit` and `set -o pipefail`, this causes the script to exit immediately before printing any output or invoking Docker.

This PR makes the Docker group lookup tolerant of a missing group and only passes `-u "${UID}:${DOCKER_GID}"` when a Docker group ID is actually
available.

On systems where the `docker` group exists, the existing behavior remains unchanged.

Fixes #7311

Tested with a stubbed `docker` command using the default macOS Bash 3.2:

- When no `docker` group exists, the script continues and omits `-u`.
- When `DOCKER_GID=999` is set, `-u 501:999` is passed as before.

A string is used instead of an array because Bash 3.2 can fail on empty array expansion when `set -o nounset` is enabled.

NONE


tarone-saloni:
PTAL @fisherxu @kevin-wangzefeng

Issue #7006, opened 2026-07-13
router: Rest.Forward leaks a goroutine and can hang on timeout or client disconnect


Priyanshu6968:
**What happened**:

The cloud router forwards a message inside `Rest.Forward` by spawning a worker
goroutine and waiting on unbuffered channels:

    // cloud/pkg/router/provider/rest/rest.go:117-127
    stop := make(chan struct{})
    respch := make(chan interface{})   // unbuffered
    errch := make(chan error)          // unbuffered
    go func() {
        resp, err := target.GoToTarget(res, stop)
        if err != nil { errch <- err; return }
        respch <- resp
    }()

`Forward` then selects on respch/errch/timer/request-context. On the timeout
branch (rest.go:174-181) and the client disconnect branch (rest.go:182-189) it
returns immediately. After it returns, nothing drains `respch`/`errch` again, so
when the worker goroutine finally finishes `GoToTarget`, its `respch <- resp`
(126) or `errch <- err` (123) blocks forever on the unbuffered channel — leaking
that goroutine on every slow or disconnected request.

The timeout/disconnect branches also do `stop <- struct{}{}` (178/188) on an
unbuffered `stop` channel, but the two most common targets ignore it:
`Rest.GoToTarget` (rest.go:194, `_ chan struct{}`) and `EventBus.GoToTarget`
(eventbus.go:120, `_ chan struct{}`) never read `stop`; only
`ServiceBus.GoToTarget` does. So for rest/eventbus targets `stop <- struct{}{}`
blocks the Forward goroutine too, holding the HTTP request handler open.

**What you expected to happen**:

A slow target or a disconnected client should free the worker goroutine and the
request resources, not leak a goroutine per event. Sends to the result channels
must not be able to block forever, and the stop signal must not block when the
target does not consume it.

**How to reproduce it (as minimally and precisely as possible)**:

1. Configure a router rule with a REST source forwarding to a rest or eventbus
   target.
2. Make the target respond slower than `restTimeout` (default 60s), or
   disconnect the client mid-request.
3. The request returns (timeout/disconnect), but the worker goroutine spawned at
   rest.go:120 stays blocked on `respch`/`errch` and never exits. Under repeated
   slow/disconnected requests, goroutine count grows without bound (visible in a
   pprof goroutine profile).

**Environment**:
- Kubernetes version (use `kubectl version`):
- KubeEdge version(e.g. `cloudcore --version` and `edgecore --version`):
[...]

PR #7315, opened 2026-09-19
router: prevent stale response callback from blocking message dispatch

### LINKED ISSUE #7314: router: a stale response callback can permanently deadlock the router's message dispatch loop
**What happened**:

`listener.Process()` is the router's single message dispatch goroutine, started from router.go:52.

When a message has a `ParentID`, `HandleMessage()` invokes the registered callback synchronously. The callback registered by servicebus.GoToTarget performs a blocking send on an unbuffered channel:

```
if stop != nil {
    listener.MessageHandlerInstance.SetCallback(messageID, func(message *model.Message) {
        response = message
        stop <- struct{}{}
    })
    <-stop
    listener.MessageHandlerInstance.DelCallback(messageID)
}
```

The problem is that stop has one receiver but two possible senders: the response callback and the timeout path in `Rest.Forward`.

This creates two ways for the callback to block permanently:

1. Timeout and response race
If the timeout signal and the edge response arrive at nearly the same time, <-stop consumes one of them. If it consumes the timeout signal, the response callback is left trying to send to a channel with no receiver.

2. DelCallback window
After <-stop returns, the callback is still registered until `DelCallback()` executes. If a response arrives in that small window, `HandleMessage()` invokes the callback even though GoToTarget is no longer receiving from stop.

In both cases, the callback blocks inside `listener.Process()`. Since callback execution is synchronous, `Process()` never gets back to `beehiveContext.Receive()`, so the router stops processing all subsequent messages.

This is therefore not limited to a single request. A stale response can stall the entire router module until CloudCore is restarted.

There are also two related issues in the same code path:

response can be written by the callback while being read by GoToTarget after a timeout wakeup without synchronization. This is a data race under -race.If `GoToTarget` remains blocked, DelCallback() is never reached, so the corresponding entry remains in callbackHandlers.

**Relation to #7016**:

Making the stop signal non-blocking alone does not completely solve this.

SetCallback() happens before <-stop. If the timeout signal is sent during that window and the send is made non-blocking, the signal can simply be dropped. GoToTarget then waits forever with the callback still registered, and a later response can still block listener.Process().

The callback delivery and callback lifetime therefore need to be handled together.

**What should happen**:

A late, duplicate, or otherwise stale response should never be able to block the router's message dispatch loop.

Once the request has timed out or the receiver has gone away, a response should be safely ignored rather than blocking the dispatch goroutine. Callback registration should also be cleaned up when the waiting goroutine exits, including the timeout/abandoned-request path.

**How to reproduce**:

1. Configure a router rule with a REST source and a servicebus target.
2. Send a request whose edge response arrives at approximately the same time as restTimeout (60s by default).
3. The request returns 408, but the router stops processing subsequent messages.
4. A goroutine profile shows listener.Process blocked on a channel send inside the callback registered by servicebus.GoToTarget.

The following unit test reproduces the underlying problem without requiring a cluster:

```
func TestHandleMessageWedgesDispatchLoopOnStaleCallback(t *testing.T) {
	stop := make(chan struct{}) // unbuffered, nobody receiving

	MessageHandlerInstance.SetCallback("parent-1", func(*model.Message) {
		stop <- struct{}{} // same blocking send used by servicebus.GoToTarget
	})

	reply := model.NewMessage("parent-1")
	reply.Header.ParentID = "parent-1"

	done := make(chan struct{})
	go func() {
		_ = MessageHandlerInstance.HandleMessage(reply)
		close(done)
	}()

	select {
	case <-done:
	case <-time.After(2 * time.Second):
		t.Fatal("HandleMessage blocked on a stale callback; " +
			"Process() invokes callbacks synchronously, so this can deadlock " +
			"the router's message dispatch loop")
	}
}
```

**Environment**

* KubeEdge: master (172a4b5d3)
* Affected area: cloud/pkg/router
* Feature: Rule / RuleEndpoint
* Related: #7016

tarone-saloni:
**What type of PR is this?**

/kind bug

`listener.Process()` is the router module's single message dispatch goroutine. For any message carrying a ParentID it calls `mh.callback(message)` synchronously, which invokes the registered closure directly.

The closure registered by `ServiceBus.GoToTarget` did a blocking send on the caller's unbuffered `stop` channel:

```go
listener.MessageHandlerInstance.SetCallback(messageID, func(message *model.Message) {
    response = message
    stop <- struct{}{}
})
<-stop
listener.MessageHandlerInstance.DelCallback(messageID)
```

That channel has one receive but two senders — this callback, and the timeout branch of `Rest.Forward`. Once `<-stop` has consumed one of them, any later send blocks forever. Two orderings reach that state: a reply arriving at the same moment the timer fires, and a reply arriving in the window after `<-stop` returns but before `DelCallback` runs. In either case the send never completes, `Process()` never gets back to `beehiveContext.Receive()`, and the router stops handling messages for every rule and endpoint until cloudcore is restarted.

The same block had two smaller problems: `response` was written by the callback goroutine and read by `GoToTarget` after a timeout wakeup with nothing ordering the two, and `DelCallback` was skipped whenever `GoToTarget` returned by any path other than a normal reply, leaving the entry in `callbackHandlers`.

This PR:

- delivers the reply over a private buffered channel using a non-blocking send, so the callback can never park the dispatch    goroutine regardless of what the receiver is doing
- waits on that channel alongside `stop` in a `select`, so `stop` is back to one sender and one receiver
- moves `DelCallback` to a `defer` so the registration is always cleaned up
- replaces the `Load` + `Delete` pair in `MessageHandler.callback` with `LoadAndDelete`, making delivery atomically one-shot and closing the window described above

`response` is now assigned only by `GoToTarget` itself, which removes the race.

Adds tests for the `listener` package, which had none: dispatch does not block on an abandoned receiver, a callback fires exactly once under concurrent replies, and `DelCallback` prevents invocation.

Fixes #7314

This is scoped to the dispatch-loop deadlock and deliberately does not touch `cloud/pkg/router/provider/rest/rest.go`. The per-request goroutine leak and hang in `Rest.Forward` is a separate defect already tracked in #7006 with an open fix in #7016, and the two changes are independent.

One consequence of #7016 is worth noting here: on client disconnect `Rest.Forward` currently returns without ever signalling `stop`, because `request.Context().Done()` isonly ever closed, so the `if !ok` guard returns before the `stop <- struct{}{}` below it. `GoToTarget` therefore stays parked and its deferred `DelCallback` does not run. That path is in `rest.go` and is closed by #7016, not by this PR.

Verified with `go build ./cloud/...`, `go vet ./cloud/pkg/router/...` and `go test ./cloud/pkg/router/... -race`.

Fixed a deadlock in the cloud router where a late or duplicate response from an edge node could permanently stop the router module from processing messages.


omlahore (commented):
The GoToTarget change looks right to me. I ran the new tests against master's message.go 30 times though and they all passed, and I think the abandoned-receiver one can't fail anyway since it registers its own non-blocking callback. I'm not sure there's an easy way to hit the window before DelCallback in a test either.

tarone-saloni:
> The GoToTarget change looks right to me. I ran the new tests against master's message.go 30 times though and they all passed, and I think the abandoned-receiver one can't fail anyway since it registers its own non-blocking callback. I'm not sure there's an easy way to hit the window before DelCallback in a test either.

Yeah, good catch. You're right they also pass on master's `message.go`, so they weren't really testing anything.For the once-only test, I made it deterministic: park the first delivery in the callback, then send a second one. It fails consistently with `Load` + `Delete` and passes with `LoadAndDelete`.The abandoned-receiver test can't really fail as-is since the changed closure is inside `GoToTarget`. I can pull it into a small helper and test that, or just drop the test. Which do you prefer?

tarone-saloni:
PTAL @GsssC @zc2638

PR #7318, opened 2026-09-20
policycontroller: don't abort VisitRulesFor on a dangling roleRef

### LINKED ISSUE #7317: policycontroller: one dangling roleRef drops the rest of a service account's bindings
I tried VisitRulesFor with two ClusterRoleBindings for the same service account, one pointing at a ClusterRole that doesn't exist and one at a real one, and the access object came back with no rules at all. GetRoleReferenceRules returns an error for the missing role and the loop returns right there ([reconcile.go#L589-L593](https://github.com/kubeedge/kubeedge/blob/d062d298/cloud/pkg/policycontroller/manager/reconcile.go#L589-L593)), so every binding after it is skipped. The RoleBinding loop below does the same at [L612-L616](https://github.com/kubeedge/kubeedge/blob/d062d298/cloud/pkg/policycontroller/manager/reconcile.go#L612-L616). Upstream Kubernetes hands that error to the visitor and carries on with the rest of the bindings.

So one stale binding quietly strips a service account of access it should still have on the edge. I checked it at master d062d298 with a unit test against the fake client, and the log line was "failed to get rules for clusterrolebinding dangling".

pujitha24:
**What type of PR is this?**

/kind bug

`VisitRulesFor` in `cloud/pkg/policycontroller/manager/reconcile.go` scans a service
account's ClusterRoleBindings and RoleBindings and resolves each one's `RoleRef` into
policy rules. If any single binding's `RoleRef` points at a Role/ClusterRole that no
longer exists (e.g. the Role was deleted but the binding wasn't cleaned up), the lookup
returns a NotFound error and the code did `return` immediately — aborting the scan and
silently dropping every other still-valid binding for that service account from the
resulting `ServiceAccountAccess` object. One stale binding therefore stripped an edge
service account of access it should still have. Upstream Kubernetes' equivalent RBAC
visitor logs the error for the offending binding and continues with the rest.

Fixes #

Approach: changed `return` to `continue` in both the ClusterRoleBinding loop and the
RoleBinding loop inside `VisitRulesFor`, so a dangling `RoleRef` only skips that one
binding (still logged via the existing `klog.Errorf`) instead of aborting the whole
visit. The two `List()` error paths above each loop are untouched and still `return`,
since those represent a failure to read the dataset at all rather than one bad entry.

Validation:
- Added `TestVisitRulesForSkipsDanglingRoleRef` in `reconcile_test.go`, covering both
  loops with one resolvable and one dangling binding each. Confirmed it fails against
  the old code (`got []` for both assertions — the dangling ClusterRoleBinding drops
  the good one, and the RoleBinding loop is never even reached) and passes with the fix.
- `go build ./...` passed.
- `go test ./cloud/pkg/policycontroller/...` passed.
- `go vet ./cloud/pkg/policycontroller/...` and `gofmt -l` on the changed files: clean.
- `golangci-lint run` (repo-pinned v1.64.5, via `GOTOOLCHAIN=go1.23.12` to match this
  repo's go.mod toolchain) on `./cloud/pkg/policycontroller/...`: no issues.
- HEAD is at upstream master tip (d062d2983); master's push-triggered CI (Main CI
  WorkFlow, FOSSA, CIFuzz) is green.
- Not run: a live cluster/e2e reproduction. This is a defect provable in a unit test
  (aborting a loop on the first resolution error vs. skipping just that entry), so the
  failing-then-passing test above is the reproduction for this class of change.

```release-note
Fixed a bug in the cloud policycontroller where a single ClusterRoleBinding or
RoleBinding with a dangling RoleRef (pointing at a deleted Role/ClusterRole) caused
every other binding for that service account to be silently dropped from its computed
access, instead of just skipping the unresolvable binding.
```

Fixes #7317


DoisLONG (commented):
Thanks for the PR.

The fix is small and makes sense. A single ClusterRoleBinding or RoleBinding with a dangling RoleRef should not abort the whole `VisitRulesFor()` scan. Skipping only the bad binding while keeping the existing log allows the remaining valid bindings to still be reflected in `ServiceAccountAccess`.

I also like that the two list-error paths are left as `return`, because those represent failure to read the full binding list, while a dangling RoleRef is only one bad entry inside an otherwise readable list.

The added regression test covers both paths: one dangling and one valid ClusterRoleBinding, and one dangling and one valid RoleBinding. It verifies that the valid bindings are still visited.

/lgtm

